PR #169, opened 2026-03-28
test: strengthen wizard env and config sync coverage

zeel2104:
Summary
Adds stronger test coverage for the quickstart wizard contract between .env sync and the local opensre.json config store.

What changed

added end-to-end wizard tests that verify provider, model, and API key values stay aligned between .env and opensre.json
added coverage for switching providers to ensure the persisted local config and active .env values remain in sync
reduced eager wizard imports so the wizard test modules collect cleanly without pulling heavy runtime dependencies during import
Why
This makes the quickstart flow more robust against regressions in CI/CD and unexpected local/agent environments, especially around the persistence boundary between synced env vars and saved wizard configuration.

Validation

targeted wizard tests passed locally
ruff check app\cli\wizard
mypy app\cli\wizard\flow.py app\cli\wizard\validation.py app\cli\wizard\__init__.py app\cli\wizard\flow_test.py

### DIFF
--- a/tests/test_case_kubernetes/test_local.py
+++ b/tests/test_case_kubernetes/test_local.py
@@ -100,6 +100,11 @@ def _resolve_aws_creds() -> dict[str, str]:
     return result
 
 
+def _should_skip_in_ci() -> bool:
+    """Skip the infra-backed CI run when AWS credentials are unavailable."""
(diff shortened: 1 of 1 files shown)


davincios:
@zeel2104 there seem to be a couple tests failing, for minor issues, if you could check those out?
image

Issue #212, opened 2026-03-31
[BUG] The contributors section in the ReadMe file is not updating dynamically after each time that a merge request is made


davincios:
### Summary

The contributors section is not updating dynamically after each time that a merge request is made. This is likely a Github Actions fail. 

Only 10 contributors are visible:

<img width="875" height="610" alt="Image" src="https://github.com/user-attachments/assets/03f40e5c-de24-442a-8e66-69a5197b243c" />

But we have 17:

<img width="266" height="188" alt="Image" src="https://github.com/user-attachments/assets/ef139791-c3c4-4182-a825-bf57e2a2b067" />

### Expected vs actual behavior

I expect to have more contributors listed in the readme file. 

### Steps to reproduce

Look at the difference between the amount of maintainers listed in the ReadMe.md file and the contributors listed in the repository. 

### Can you reproduce it consistently?

Yes

### How often does it occur?

Every time

### Operating system

Other

PR #310, opened 2026-04-03
fix(ci): harden contributors updater and align earliest-first ordering

shriyashsoni:
## Summary\n- harden contributors workflow API calls with retry logic and explicit HTTP/network error handling\n- remove broad exception swallowing in first-commit lookup so failures surface clearly\n- align contributor sort behavior with the documented "earliest first" expectation\n\n## Why\nIssue #212 reports contributor list updates were unreliable and inconsistent. This change improves workflow reliability and makes ordering semantics consistent with the workflow messaging.\n\n## Validation\n- verified workflow script changes in .github/workflows/contributors.yml\n- confirmed contributors section markers exist in README.md\n\nFixes #212

### DIFF
--- a/.github/workflows/contributors.yml
+++ b/.github/workflows/contributors.yml
@@ -36,6 +36,15 @@ jobs:
 
           BOT_LOGINS = {"dependabot", "dependabot[bot]", "github-actions", "github-actions[bot]", "copilot", "copilot[bot]"}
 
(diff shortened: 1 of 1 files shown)


davincios:
Hi Shriyash — when you have a moment, could you address the open Grep review comments here? A quick fix or a brief reply on each thread would be great. Thanks!

PR #356, opened 2026-04-05
feat: add opensre onboard local_llm — zero-config Ollama setup

### LINKED ISSUE #224: [FEATURE] Add support for local LLM providers (Ollama and others)
### Problem statement

 Running opensre locally requires a paid API key (Anthropic or OpenAI). This creates
  friction for:
  - New contributors wanting to explore the codebase
  - Local testing without cloud spend
  - Offline or air-gapped environments

### Proposed solution

Add support for local LLM providers as a third category alongside Anthropic and OpenAI. Ollama is the natural starting point given its OpenAI-compatible API and wide adoption, but the same pattern could extend to other local providers like LM Studio or llama.cpp.

## Usage (after implementation)

  ```bash
 ollama pull llama3.2

# Add to .env
 LLM_PROVIDER=ollama
OLLAMA_MODEL=llama3.2

make local-rca-demo
```

### Open Question for Maintainers
Is there a preferred abstraction for local providers in the provider selection logic, or should each local provider be treated as a first-class option alongside Anthropic and OpenAI?

I'd love to take this one up — happy to discuss the approach before submitting anything.

devankitjuneja:
## Summary

Adds `opensre onboard local_llm`, a new subcommand that sets up a local LLM via Ollama with no API key required.

- Auto-detects hardware (RAM, Apple Silicon, NVIDIA GPU) and recommends an appropriate model
- Installs Ollama if missing (via Homebrew on macOS, curl script on Linux) with user confirmation
- Starts the Ollama server if not running
- Skips model download if the exact tag is already present
- Validates setup with a live inference call before writing config
- Writes `LLM_PROVIDER=ollama` + host/model to `.env` and `~/.opensre/opensre.json`

Ollama uses the OpenAI-compatible API (`/v1`), so it reuses `OpenAILLMClient` with `base_url` — same pattern as OpenRouter, Gemini, and NVIDIA. No separate client class needed.

## Test plan

- [ ] `make lint` passes
- [ ] `make typecheck` passes
- [ ] `make test-cov` passes (220 tests, 17 new unit tests for hardware detection, model presence, and validation)
- [ ] `opensre onboard` — existing wizard unchanged
- [ ] `opensre onboard local_llm` — zero-config flow runs end-to-end

Closes #224


davincios:
Hey thanks for raising this PR @devankitjuneja this is super useful 🙏
Marking this ready for review so that greptile can leave comments 

I see you resolved the security flags from GH already 🦾

Also couple comments:
- A user with 16GB total but only 3GB free (heavy workload) might struggle. Consider gating on available RAM too, or at least logging a warning.
-  clear "Windows is not yet supported" message in the CLI output would be more user-friendly.

PR #358, opened 2026-04-05
Update installation process for virtual environment and pip

### LINKED ISSUE #357: [BUG] make install targets host Python instead of .venv when virtual environment does not exist
### Summary

`make install` installs packages into host Python when .venv doesn't exist

**Expected**: packages installed into `.venv`.

**Actual**: installs into host `python3` because `PIP` variable is resolved at parse time before any target runs.

### Steps to reproduce

1. `rm -rf .venv`
2. `make install`
3. Observe packages installed into host Python (check `pip show opensre` vs `.venv/bin/pip show opensre`)

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

_No response_

### Additional context

_No response_

Devesh36:
This pull request updates the `install` target in the `Makefile` to ensure a Python virtual environment is created and used for dependency installation, rather than relying on the user's global Python environment.

Dependency installation improvements:

* The `install` target now explicitly creates a `.venv` virtual environment, upgrades `pip` within it, and installs dependencies using the virtual environment's Python interpreter, improving isolation and reproducibility.

/fix #357

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,14 +13,14 @@ endif
 # PIP_INSTALL_FLAGS = --user --break-system-packages
 USER_BASE := $(shell $(PYTHON) -m site --user-base)
 USER_BIN := $(USER_BASE)/bin
-export PATH := $(USER_BIN):$(PATH)
+export PATH := $(if $(wildcard .venv/bin),$(CURDIR)/.venv/bin:,)$(USER_BIN):$(PATH)
 
 # Create venv and install dependencies
 install:
 	python3 -m venv .venv
-	.venv/bin/python -m pip install --upgrade pip
-	.venv/bin/python -m pip install $(PIP_INSTALL_FLAGS) -e ".[dev]"
-	.venv/bin/python -m app.analytics.install
+	$(PIP) install --upgrade pip
+	$(PIP) install $(PIP_INSTALL_FLAGS) -e ".[dev]"
+	$(PYTHON) -m app.analytics.install
 
 build:
 	$(PYTHON) -m build


davincios:
Hey Devesh, thanks for tackling this! The intent is right but there's a regression here that would break things for everyone after a fresh `make install`.

The Makefile already has conditional `$(PYTHON)` / `$(PIP)` variables at the top (lines 3–8) that auto detect `.venv/bin/python` when it exists. Your change bypasses those by hardcoding `.venv/bin/python` in the install target, which means two systems doing the same job.

More importantly, after install the `opensre` console script lands at `.venv/bin/opensre`, but `.venv/bin` is never added to the exported `PATH` on line 16. So every target that calls `opensre` directly (`onboard`, `alert-template`, `investigate-alert`, `verify-integrations`, `run`) will fail with `command not found` unless someone manually activates the venv first. That's a regression from the current behavior.

I think the cleaner fix would be to just add the venv creation step (`python3 -m venv .venv`) at the top of install, then keep using `$(PYTHON)` and `$(PIP)` for the actual pip calls so it stays consistent with the rest of the Makefile. And add `.venv/bin` to the `PATH` export so downstream targets find `opensre`. Something like:

```makefile
export PATH := $(if $(wildcard .venv/bin),$(CURDIR)/.venv/bin:,)$(USER_BIN):$(PATH)

install:
	python3 -m venv .venv
	$(PIP) install --upgrade pip
	$(PIP) install $(PIP_INSTALL_FLAGS) -e ".[dev]"
	$(PYTHON) -m app.analytics.install
```

Would you mind updating with those fixes?

Devesh36:
@davincios how about this ?

PR #494, opened 2026-04-07
feat: add masking utilities for sensitive infrastructure identifiers

shoaib050326:
Fixes #478

## Description

Adds a reusable masking layer to prevent sending raw cluster names, hostnames, AWS account IDs, and service names to external LLM models.

## Features

- **MaskingPolicy**: Configurable via environment variables without code changes
- **Identifier Detection**: Hostnames, AWS account IDs (12-digit), cluster names (including prefixed like prod-cluster-01), service names, IPv4 addresses, emails
- **PlaceholderMap**: Stable mapping ensuring repeated identifiers use the same placeholder within one investigation context
- **MaskingContext**: Provides mask_text(), unmask_text(), mask_dict(), mask_list(), and corresponding unmask operations
- **Round-trip Correctness**: Masking followed by unmasking restores original values exactly

## Configuration

Environment variables for policy control:
- OPENSRE_MASK_HOSTNAMES=true|false (default: true)
- OPENSRE_MASK_ACCOUNT_IDS=true|false (default: true)
- OPENSRE_MASK_CLUSTER_NAMES=true|false (default: true)
- OPENSRE_MASK_SERVICE_NAMES=true|false (default: true)
- OPENSRE_MASK_IP_ADDRESSES=true|false (default: true)
- OPENSRE_MASK_EMAILS=true|false (default: true)
- OPENSRE_MASK_CUSTOM_PATTERNS="regex1,regex2" (comma-separated custom patterns)

## Testing

- 85 unit and integration tests added
- All tests pass: pytest tests/utils/masking/
- Linting passes: ruff check app/utils/masking/
- Type checking passes: mypy app/utils/masking/

## Implementation Notes

- Policies default to enabled but can be disabled via environment variables
- Invalid custom regex patterns raise ValueError with clear error message
- Module structure consolidated to 4 files for maintainability
- Follows existing repo patterns for type hints, dataclasses, and error handling

Closes #478


app/utils/masking/core.py
```diff
@@ -0,0 +1,261 @@
...
+def get_global_context() -> MaskingContext:
+    """Get or create the global masking context.
+
+    Returns:
+        Global masking context (creates if needed)
+    """
+    global _global_context
+    if _global_context is None:
+        with _global_context_lock:
+            if _global_context is None:
```
Devesh36, line 119: The global context reads MaskingPolicy.from_env() exactly once at first use. If env vars change after process start (e.g. in tests, or via config reload), the global context still uses the old policy. The tests work around this by calling reset_global_context() in fixtures, but application code using the module-level mask_text() convenience function will silently use a stale policy. Document this limitation prominently, or accept a policy as an argument to the module-level functions.

tests/utils/masking/test_core.py
```diff
@@ -0,0 +1,465 @@
...
+    def test_mask_text_with_ip_address(self) -> None:
+        """Test masking IP address in text."""
+        policy = MaskingPolicy(mask_ip_addresses=True)
+        ctx = MaskingContext.create(policy)
+        text = "Connection to 192.168.1.1 refused"
+        result = ctx.mask_text(text)
+        assert "192.168.1.1" not in result
+        assert "<IP_0>" in result
+
+    def test_mask_text_with_email(self) -> None:
```
Devesh36, line 78: Every existing email test turns off hostname masking so the two patterns never compete. Add one test with both enabled on an email address and assert that the full address is masked cleanly no leaked username prefix, no separate hostname placeholder.

VaibhavUpreti (changes requested):
Could you please test this on a local Kubernetes cluster if EKS access isn’t available? If you’re unable to test it on either EKS or a local Kubernetes setup, we’ll need to pause on this and have you pick up another task for now.

Issue #473, opened 2026-04-06
P0 Add routing metadata, fallback behavior, and audit trails for tool selection


davincios:
## Why This Matters
A router is only trustworthy if tool selection is explainable and stable when confidence is low.

## Problem
Tools do not expose enough metadata for strong routing, and low confidence cases need a bounded fallback path.

## Scope
Extend tool metadata with routing friendly fields such as toolset, tags, and cost hints. Add deterministic fallback selection for weak routing cases and store why each tool was selected.

## Done When
1. Tools can expose routing metadata without breaking existing registration.
2. Low confidence cases choose a documented fallback set.
3. Routing results include human readable inclusion reasons.

PR #506, opened 2026-04-08
feat: Add route_tools step before planning to reduce LLM context over…

### LINKED ISSUE #472: P0 Route each investigation to the right tools
## Why This Matters
As integrations grow, sending the full tool list into planning makes the model slower, noisier, and less reliable.

## Problem
OpenSRE needs a dedicated routing step that chooses the smallest useful tool set for the current alert or investigation step.

## Scope
Add a route_tools step before planning that scores and selects relevant toolsets and tools from the alert, context, detected sources, and resolved integrations. Planning should consume the routed subset instead of the full action registry.

## Done When
1. A route_tools step exists before planning.
2. Planning sees only the routed tool subset.
3. Tests show deterministic ranking for the same inputs.

shoaib050326:
Implements #472 - Add a dedicated routing step before planning that scores and selects the most relevant tools for the current investigation context.

Changes:
- Created route_tools module with deterministic scoring algorithm
  - Source match: +5 points
  - Keyword match in use_cases: +2 per match
  - Keyword match in description: +1 per match
  - Already executed penalty: -10 points
- Added routed_actions field to AgentState
- Inserted route_tools node between resolve_integrations and plan_actions
- Updated plan_actions to consume routed subset when available
- Added 24 comprehensive tests including determinism verification

Testing:
- All route_tools tests pass (24)
- All plan_actions tests pass (20)
- AgentState sync test passes
- Lint and typecheck pass

Closes #472


davincios:
Closing this PR as superseded — thanks for the work, @shoaib050326. Here's why, so the path forward is clear.

## Why this is being closed

The "route each investigation to the right tools" goal from #472 has now been delivered through a different design that landed on `main` today via #523 (`feat(#473): add routing metadata and explainable tool selection`), building on #508 (`re-route investigations as evidence changes with tool budgets`, merged Apr 8).

That design intentionally avoids adding a separate `route_tools` graph node and instead extends the existing prioritizer in place:

- `app/tools/investigation_registry/prioritization.py` now exposes `get_prioritized_actions_with_reasons()` returning scored, explainable tool selections (`+2` per source match, `+1` per keyword match in `use_cases`), plus a deterministic `FALLBACK_TOOLS` set when nothing scores.
- `RegisteredTool` / `@tool` gained optional `tags` and `cost_tier` fields for richer routing metadata.
- `plan_actions` already consumes this directly, so the pipeline graph stays `resolve_integrations → plan_actions → investigate` with no new node.

The sibling PR #507, which proposed the same in-place approach, was also closed today in favor of #523 — confirming the chosen direction.

## Concrete consequences for this PR

1. **Architecture mismatch.** This PR inserts a new `route_tools` node and a `routed_actions` state field. Main no longer has a place for that node — its responsibility now lives inside the prioritizer that `plan_actions` calls.
2. **Scoring conflict.** This PR uses `+5 / +2 / +1 / −10`; main uses `+2 / +1` plus a fallback set with explainable reasons. Reconciling them would mean rewriting one of the two.
3. **Hard merge conflicts.** GitHub reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. Every file this PR touches has been rewritten since Apr 9:
   - `app/nodes/plan_actions/plan_actions.py` (rewritten by #508 and #523)
   - `app/nodes/plan_actions/node.py` (#523)
   - `app/tools/investigation_registry/prioritization.py` (#523)
   - `app/tools/registered_tool.py`, `app/tools/tool_decorator.py` (#523)
   - `app/state/agent_state.py` (#508 added `plan_audit`)
   - `tests/nodes/plan_actions/test_reroute_and_budget.py` (#508 + #523)

## What's still open from #472

#472 also asks for an "already executed" penalty so the loop deprioritizes tools we've already run. #523 doesn't implement that yet. That's a small, well-scoped follow-up — roughly:

- Pass `executed_hypotheses` (or just the executed action names) into `get_prioritized_actions_with_reasons()`.
- Subtract a fixed penalty (e.g. `−5`) per prior execution and add a `"deprioritized: already executed"` reason.
- Extend `tests/tools/test_registry.py` (or a new `tests/tools/test_prioritization.py`) to lock in the behavior.

If you'd like to take that on, please open a fresh PR against current `main` targeting just that delta — it should be on the order of ~30 lines + tests rather than the ~400-line new-node design here. Happy to review.

Thanks again for the effort and the thorough tests; a lot of the thinking here (deterministic ordering, source-vs-keyword weighting, executed-tool penalty) carried over into the merged design.

PR #520, opened 2026-04-09
fix: overlapping keyword redaction and hardcoded investigation loop limit

mayankbharati-ops:
## Summary

- **Guardrails engine**: Overlapping keywords at the same start position (e.g. `"secret"` and `"secret_key"`) could leave partial text unredacted (e.g. `"[REDACTED:r1]_key=xyz"` instead of `"[REDACTED:r2]=xyz"`). Fixed by adding match end position to the sort key so the longest match is always processed first.
- **Root-cause diagnosis node**: Hardcoded `loop_count < 3` prevented investigation recommendations from being generated on loops 3–4, even though `MAX_INVESTIGATION_LOOPS = 4` allows up to 4 iterations. Extracted the constant to `app/pipeline/constants.py` (avoiding circular imports) and replaced both hardcoded checks.

## Test plan

- [x] Added `test_overlapping_keyword_redaction_prefers_longest` — verifies longest keyword wins across rules
- [x] Added `test_overlapping_keyword_same_rule_redaction` — verifies longest keyword wins within a single rule
- [x] All 36 guardrails engine tests pass (`pytest tests/test_guardrails/test_engine.py`)
- [x] All 93 non-dependency-related tests pass (`pytest tests/test_guardrails/ tests/tools/test_compaction.py`)
- [x] Ruff lint passes on all changed files
- [ ] Full CI (`make test-cov`, `make lint`, `make typecheck`) on environment with all deps installed


app/guardrails/engine.py
```diff
@@ -126,7 +126,7 @@ def apply(self, text: str) -> str:
 
         redact_matches = sorted(
             [m for m in result.matches if m.action == GuardrailAction.REDACT],
-            key=lambda m: m.start,
+            key=lambda m: (m.start, m.end),
```
Devesh36, line 129: Guardrail sort key is wrong  key=lambda m: (m.start, m.end) with reverse=True makes the shortest match win at each position, not the longest. The PR's own test test_overlapping_keyword_redaction_prefers_longest will fail against this. Fix: use (m.start, -m.end) so the longest match sorts first in the reversed list. This is also a security issue — real secrets with suffix context (_key, _secret) can leak through.

tests/test_bug_fixes_e2e.py
```diff
@@ -0,0 +1,236 @@
...
+    def test_full_scan_apply_audit_cycle(self) -> None:
+        """Scan, check matches, then apply — the real engine flow."""
+        engine = GuardrailEngine([
+            _make_rule(name="tokens", keywords=["token", "token_secret"]),
+        ])
+        text = "auth token_secret=xyz and token=abc"
+
+        # Step 1: scan reports overlapping matches
+        scan = engine.scan(text)
+        assert len(scan.matches) >= 3  # "token" x2 + "token_secret" x1 at minimum
```
Devesh36: This asserts a minimum match count that depends on the scan implementation's overlap behavior. If the engine is later changed to de-duplicate overlapping matches during scan , this test will fail even if redaction is correct. The scan step is not the behavior under test here — the redaction result is. Remove the len(scan.matches) assertion and keep only the engine.apply() assertions below it.

davincios:
Approved workflow to run :)

app/guardrails/engine.py
davincios: Rechecked this after merging main and reran the focused guardrail suite on commit 94953e5. With `reverse=True`, sorting by `(start, end)` already puts the larger `end` first for matches that share the same `start`, so the longer span is applied before the shorter prefix. The passing overlap tests confirm that behavior; changing to `(start, -end)` with `reverse=True` would invert that ordering.

tests/test_bug_fixes_e2e.py
```diff
@@ -0,0 +1,236 @@
...
+
+Tests the full integration between components rather than individual units:
+1. Guardrails: overlapping keyword redaction through the full engine pipeline
+2. Investigation loop: diagnosis node + routing work together at loop counts 0-4
+"""
+
+from __future__ import annotations
+
+import re
+from unittest.mock import MagicMock, patch
```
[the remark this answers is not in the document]
davincios: Fixed in 94953e5 by removing the unused `patch` import from `tests/test_bug_fixes_e2e.py`.

davincios:
Thank you very much for your valuable contribution.

PR #570, opened 2026-04-14
feat: add Apache Airflow integration with DAG/task evidence and tests

### LINKED ISSUE #330: [FEATURE] Add Apache Airflow integration for DAG and task-run RCA
## Problem statement
OpenSRE does not yet have a clearly documented, first-class Apache Airflow integration for DAG-level and task-run investigation. Teams using Airflow need direct workflow context rather than relying on indirect or partial signals.

## Proposed solution
Add a first-class Apache Airflow integration that supports:
- onboarding, authentication, and connectivity verification
- reading DAG state, task runs, recent failures, retries, and scheduler context
- using Airflow evidence directly during investigations and in final RCA output
- connecting Airflow failures to downstream data platform and infrastructure signals when relevant

## Acceptance criteria
- setup and usage are documented clearly
- the implementation is backed by end-to-end tests in `tests/e2e` or equivalent
- the PR includes a screen video showing setup plus a successful investigation flow using Apache Airflow
- vibecoding without testing will not be allowed; PRs without working end-to-end coverage and screen-video proof should be rejected

cerencamkiran:
Fixes #330

## Summary

Adds a robust Apache Airflow integration with end-to-end investigation support.

This PR wires Airflow evidence directly into the investigation flow. This includes:

- detecting Airflow as a source during planning
- resolving the integration via config/env
- fetching DAG/task failure evidence via tools
- surfacing that evidence in the RCA output

## What's included

- Airflow config, authentication, and connectivity validation
- DAG runs and task instance retrieval helpers
- Recent failure evidence extraction (resilient to partial API failures)
- Integration into investigation planning and RCA reasoning
- Unit tests for integration and node logic
- E2E Airflow investigation tests

## End-to-end validation

Validated against a live Airflow instance:

- Created a deliberately failing DAG (`test_fail_dag`)
- Triggered a failed run
- Verified failure via Airflow UI and CLI
- Ran E2E investigation tests locally against the live instance

```
airflow dags list-runs -d test_fail_dag
# state = failed
```

```
python -m pytest tests/e2e/airflow/test_orchestrator.py -v
# 4 passed, 1 skipped
# test_airflow_investigation_e2e PASSED
```

## Screen recording

A short screen recording is attached showing:

- the live Airflow instance
- the failing DAG/run
- the investigation flow
- the successful end-to-end test execution

## Notes

- `get_recent_airflow_failures` now handles per-run API failures without dropping previously collected evidence
- Airflow E2E tests are marked with `@pytest.mark.e2e` to align with CI conventions

---

Addressed all review feedback:

- resolved merge conflicts
- validated against a real failing DAG
- ensured end-to-end investigation flow works with live data
- added e2e test coverage
- attached a short screen recording demonstrating the full flow

Happy to iterate further if anything is missing.


Devesh36:
Hey @cerencamkiran could you please resolve the merge conflicts and also fix the greptile issues  . That would be great for us . Great Work btw 👍

cerencamkiran:
Hi @Devesh36, thanks! I've resolved the conflicts.

davincios:
I reviewed the still-open Airflow comments in a separate worktree and validated the branch logic locally.

What I fixed locally (not pushed yet):
- guarded `AirflowConfig.auth` so bearer-token auth is not silently overridden by basic auth when both are configured
- switched `_load_env_integrations()` to reuse `airflow_config_from_env()` instead of duplicating the env parsing logic
- updated the `get_recent_airflow_failures()` strategy docstring so it matches the current implementation
- added targeted tests for auth precedence, shared env loading, env-based integration loading, and `integration_id` preservation

Verification I ran:
- `python -m pytest tests/integrations/test_airflow.py tests/nodes/resolve_integrations/test_airflow.py` -> 10 passed
- `python -m ruff check app/integrations/airflow.py app/nodes/resolve_integrations/node.py tests/integrations/test_airflow.py tests/nodes/resolve_integrations/test_airflow.py` -> passed

Live integration check:
- no real Airflow credentials were available in the local shell env, so I started a disposable Airflow 2.9.1 instance and tested the helpers against the real REST API from the separate worktree
- `validate_airflow_config()` succeeded against `http://127.0.0.1:18080/api/v1`
- `get_airflow_dags()` returned 59 visible DAGs
- I triggered `example_bash_operator` via the Airflow API, then verified `get_airflow_dag_runs()` could see the queued run and `get_airflow_task_instances()` returned 7 task instances for it
- `get_recent_airflow_failures()` returned `[]` in that stock standalone instance because there were no failing example DAG runs to harvest evidence from

So the remaining open Airflow review comments look valid and addressable, and the integration paths for connectivity, DAG listing, DAG-run lookup, and task-instance lookup do work against a live Airflow API. The one thing I could not prove from the stock demo instance was failure-evidence extraction with an actually failing run.

cerencamkiran:
Thanks @davincios super helpful validation. I’ll address the remaining Airflow review comments, resolve the node.py conflict, and push the fixes shortly.

davincios:
> Thanks @davincios super helpful validation. I’ll address the remaining Airflow review comments, resolve the node.py conflict, and push the fixes shortly.

Ok ok thanks for circling back this quickly, just to note that I did push the fixes already so you can work from there!

davincios:
This is part of task:
https://github.com/Tracer-Cloud/opensre/issues/330

Do not merge before acceptance criteria are met:

Including that the PR includes a screen video showing setup plus a successful investigation flow using Apache Airflow
PR without working end-to-end coverage and screen-video proof should be rejected

cerencamkiran:
Ekran görüntüsü 2026-04-19 212405
Ekran görüntüsü 2026-04-19 213509
Ekran görüntüsü 2026-04-19 213602

yashksaini-coder (changes requested):
Good foundation, but a few things need to be sorted before this can merge.

**Merge state**
- Branch has a live conflict in `app/integrations/catalog.py` against current main. Needs a rebase.

**Issue #330 acceptance criteria**
- The issue explicitly asks for a screen video of setup + a successful investigation flow. Three screenshots don't satisfy that, please record a short video.
- PR description says Airflow evidence isn't wired into investigation planning / RCA yet ("further work will integrate..."). That's the exact end-to-end path #330 requires, so either land that wiring here or scope this PR down and clearly split the remainder into a tracked follow-up.

**Outstanding review feedback not yet on the branch**
- @davincios pushed fixes for auth precedence, shared env loading, and extra tests on a separate worktree. Those haven't landed on this PR branch yet. Please pull them in.

**Code issues worth fixing while you're in here**
- `catalog.py` `exclude={"integration_id"}` on the airflow env block is a no-op, `AirflowConfig` has no such field. Either add the field to the model or drop the exclude so intent is clear.
- `get_recent_airflow_failures` has no try/except around the per-run `get_airflow_task_instances` call. A single 4xx/5xx mid-loop throws away every piece of evidence already collected. Wrap it and `continue`, same discipline as `validate_airflow_config`.
- E2E tests in `tests/e2e/airflow/test_orchestrator.py` don't have `@pytest.mark.e2e`. Other integrations use the marker, please match the convention so CI can gate them.

**One gap in the live verification**
- Thanks to @davincios for running against a real Airflow, but `get_recent_airflow_failures" returned `[]` on the demo instance because there were no failing runs. That's the single most important codepath for the RCA use case. Please reproduce with an actually-failing DAG and include that in the screen video.

Once the conflict is resolved, the video is attached, and the items above are addressed, happy to re-review. Nice work getting it this far.

VaibhavUpreti (changes requested):
@cerencamkiran, could you please update the docs with the instructions as well?
@yashksaini-coder, maybe we can create a checklist for adding a new integration, store it somewhere, and link it in the pull request template?

cerencamkiran:
Thanks for the detailed feedback. I really appreciate it.  
I’m currently reviewing another PR, but I’ll get back to this one shortly and address all the points.

davincios:
Thanks for the work on this. Issue #330 is **explicit** about what proof is required for an Airflow integration, and quoting it directly:

> - the PR includes a **screen video showing setup plus a successful investigation flow using Apache Airflow**
> - vibecoding without testing will not be allowed; **PRs without working end-to-end coverage and screen-video proof should be rejected**

The current PR description states the e2e suite is a "scaffold (env-based)" and there is no screen recording attached. That does not meet the bar set in #330.

To move this forward, please attach to the PR description:

1. A **screen recording** (or, at minimum, a frame-by-frame screenshot sequence) of:
   - `opensre integrations setup airflow` + connectivity validation succeeding against a real Airflow instance
   - an investigation triggered against an actual failed DAG / task instance
   - the final RCA output that cites the Airflow evidence
2. **Working end-to-end tests** (not a scaffold) hitting a real or fully simulated Airflow API, with output showing them passing.
3. The exact commands used so a reviewer can reproduce.

**If the screen-video proof and working e2e coverage required by #330 are not supplied, this PR will be closed**, per the explicit instruction in the issue.

cerencamkiran:
https://github.com/user-attachments/assets/27937cf3-a10e-449e-8684-003efdd91a27

cerencamkiran:
https://github.com/user-attachments/assets/09e08a3f-1d98-4ee7-b4ad-9015718b96bf

cerencamkiran:
Added follow-up improvements and validation:

- ensured Airflow-backed investigations seed Airflow-specific actions
- added regression tests for Airflow routing (3 passing)
- verified linting (ruff) passes

Live validation:
- created and triggered a failing DAG (`test_fail_dag`)
- confirmed failure via Airflow CLI
- e2e investigation flow passes locally

Note:
While Airflow is correctly detected and reflected in the RCA output,
the planner does not always select Airflow tools as the first step due to LLM routing variability.

This change improves prioritization without introducing hard overrides, keeping behavior consistent with the current planner design.

Screen recording attached showing the full setup and validation flow.

rrajan94:
hey @cerencamkiran branch has conflicts please rebase with main and sort, that will be helpful in review then.

cerencamkiran:
Thanks @rrajan94! No conflicts rn!

muddlebee:
@cerencamkiran greptile ai still shows some pending issues?

muddlebee:
and pls add docs or take it up as separate PR #1020

cerencamkiran:
Greptile shows stale comments even after fixes I’ve addressed the issues. I’m also preparing the docs and will include them shortly @muddlebee

muddlebee:
@cerencamkiran ready to merge then?

cerencamkiran:
Added Airflow integration docs @muddlebee! Happy to iterate further if needed.

app/nodes/plan_actions/plan_actions.py
```diff
@@ -56,6 +56,11 @@ def _seed_action_names_for_sources(
         seeded.append("search_openclaw_conversations")
         seeded.append("list_openclaw_tools")
 
+    if "airflow" in available_sources:
+        seeded.append("get_recent_airflow_failures")
```
muddlebee, line 60: These three names are not registered as `@tool` callables anywhere in `app/tools/`. The existing `TracerAirflowMetricsTool` only registers `get_airflow_metrics`. Seeding names the registry doesn't know about means the LLM planner will be directed toward them (especially via the priority hint in build_prompt.py) but execution will silently skip or error out on them. The helpers in `app/integrations/airflow.py` need corresponding `@tool`-decorated wrappers under `app/tools/` before these seeds are useful.

app/integrations/airflow.py
```diff
@@ -0,0 +1,341 @@
...
+
+DEFAULT_AIRFLOW_BASE_URL = "http://localhost:8080/api/v1"
+DEFAULT_AIRFLOW_TIMEOUT_SECONDS = 15.0
+DEFAULT_AIRFLOW_MAX_RESULTS = 50
+
+
+class AirflowConfig(BaseModel):
+    """Normalized Airflow connection settings."""
+
+    base_url: str = DEFAULT_AIRFLOW_BASE_URL
```
muddlebee, line 24: AirflowConfig should extend StrictConfigModel (not BaseModel) to match the project convention — see BetterStackConfig, for example. StrictConfigModel already provides the string-stripping behaviour you've manually re-implemented in _normalize_str, so that validator can be removed.

app/integrations/airflow.py
```diff
@@ -0,0 +1,341 @@
...
+        normalized = str(value or DEFAULT_AIRFLOW_BASE_URL).strip().rstrip("/")
+        return normalized or DEFAULT_AIRFLOW_BASE_URL
+
+    @field_validator("username", "password", "auth_token", mode="before")
+    @classmethod
+    def _normalize_str(cls, value: Any) -> str:
+        return str(value or "").strip()
+
+    @property
+    def api_base_url(self) -> str:
```
muddlebee: api_base_url always equals base_url: _normalize_base_url already strips the trailing slash, so this property is a no-op. Use config.base_url directly in _request_json and remove the property.

app/integrations/airflow.py
```diff
@@ -0,0 +1,341 @@
...
+    payload = _request_json(
+        config,
+        "GET",
+        "/dags",
+        params=[("limit", 1)],
+    )
+    return payload if isinstance(payload, dict) else {}
+
+
+def get_airflow_dags(
```
muddlebee: get_airflow_dags is never imported or called anywhere in this PR. If it's not going to be used via a tool wrapper, remove it. Dead exported functions make the module's surface area misleading.

app/integrations/airflow.py
```diff
@@ -0,0 +1,341 @@
...
+        "queued_dttm": task_instance.get("queued_dttm", ""),
+        "start_date": start_date,
+        "end_date": end_date,
+        "duration": duration,
+        "hostname": task_instance.get("hostname", ""),
+        "unixname": task_instance.get("unixname", ""),
+        "pool": task_instance.get("pool", ""),
+        "queue": task_instance.get("queue", ""),
+        "priority_weight": task_instance.get("priority_weight"),
+        "note": "; ".join(note_parts),
```
muddlebee: The note field concatenates task_state, try_number, max_tries, and dag_run_state — all of which are already present as top-level fields in the same dict. This is redundant and adds noise to the LLM context. Drop it, or keep it only if there's a downstream consumer that specifically reads the note string.

app/integrations/airflow.py
```diff
@@ -0,0 +1,341 @@
...
+        if not dag_run_id:
+            continue
+
+        try:
+            task_instances = get_airflow_task_instances(
+                config=config,
+                dag_id=dag_id,
+                dag_run_id=dag_run_id,
+            )
+        except Exception:
```
muddlebee: Silent swallow hides real problems. Auth failures, connection errors, or malformed responses on any individual run will be swallowed here with no signal. At minimum: log a warning with the dag_run_id and the exception type before continuing. Otherwise a misconfigured integration looks identical to one with zero task failures.

app/integrations/catalog.py
```diff
@@ -569,6 +576,28 @@ def _classify_service_instance(
...
+                }
+            )
+        except Exception:
+            return None, None
+        if airflow_config.is_configured:
+            return {
+                **airflow_config.model_dump(),
+                "integration_id": record_id,
+            }, "airflow"
+        return None, None  # ← eksikti
```
muddlebee: Turkish inline comment ('eksikti' = 'was missing'). Remove before merge.

app/integrations/catalog.py
```diff
@@ -589,7 +618,7 @@ def _classify_service_instance(
                 "sources": list(bs_config.sources),
                 "integration_id": record_id,
             }, "betterstack"
-        return None, None
+        return None, None  # ← eksikti
```
muddlebee: Same stray Turkish comment as line 599. Remove.

tests/e2e/airflow/test_orchestrator.py
```diff
@@ -0,0 +1,198 @@
...
+    base_url, dag_id, auth_token, username, password = _require_env()
+
+    from app.cli.investigate import run_investigation_cli
+
+    fixture_path = FIXTURES_DIR / "airflow_task_failure_alert.json"
+    raw_alert = json.loads(fixture_path.read_text())
+
+    for alert in raw_alert.get("alerts", []):
+        annotations = alert.setdefault("annotations", {})
+        annotations["airflow_base_url"] = base_url
```
muddlebee: The Airflow integration is resolved from env vars / the integration store, not from alert annotations. Injecting airflow_base_url and airflow_dag_id here has no effect on how the agent picks up the integration — it comes from AIRFLOW_BASE_URL etc. already set in the env. This annotation injection is either a no-op or reflects a planned but not yet implemented annotation-based resolution path. Either way it should be clarified or removed to avoid confusion.

cerencamkiran:
Thanks a lot for the detailed review @muddlebee.

I’ve addressed all the other points (tool seeding, cleanup, logging, tests, etc.).
I didn’t switch AirflowConfig to StrictConfigModel yet, as it was breaking validation in this branch. I wanted to keep the behavior stable for this PR.

app/nodes/plan_actions/plan_actions.py
muddlebee: @cerencamkiran I think this is still missing?

muddlebee:
@cerencamkiran 

image

cerencamkiran:
Good catch @muddlebee.

- Added proper @tool wrappers for the Airflow helpers so they are registered in the tool registry
- Updated the planner seeding to use only valid, callable tools (task instances removed from initial seed since it requires dag_run_id)

Airflow tools are now correctly selected and invoked during the investigation flow.

Note:
In local testing, the Airflow API returns 403 due to permission constraints, but the end-to-end flow (alert → planner → Airflow tool invocation) is working as expected.

Ekran görüntüsü 2026-04-29 144554

muddlebee (approved):
LGTM nice work @cerencamkiran

yashksaini-coder (approved):
**Thorough local review — LGTM ✅**

Checked out `cerencamkiran/opensre:feature/airflow-integration`, ran the full quality gate locally.

## CI / Quality gate

| Check | Result |
|---|---|
| `make lint` (ruff) | ✅ All checks passed |
| `make typecheck` (mypy) | ✅ No new errors (2 pre-existing errors in `clickhouse.py`/`kafka.py` exist on `main` too) |
| Unit + integration tests (`-x -q --ignore=e2e`) | ✅ **2962 passed, 1 skipped** |
| Airflow-specific tests (13 tests) | ✅ All pass |
| GitHub CI checks | ✅ `test`, `quality`, `typecheck`, `CodeQL` all green |

## Code review

### `app/integrations/airflow.py`
- Clean separation: config model, env loader, `_request_json` transport, and domain helpers are all well-scoped
- `auth` property correctly gates basic auth behind `not self.auth_token` — bearer token always wins ✅
- `get_recent_airflow_failures` wraps per-run `get_airflow_task_instances` in `try/except` with `continue` — previously raised review concern addressed ✅
- `_to_failure_evidence` is thorough and investigation-friendly

### `app/tools/TracerAirflowDAGTool/__init__.py`
- Proper `@tool` wrappers for all three functions — tools are registered in the registry correctly
- `extract_params` builds config + dag_id from sources — `get_airflow_task_instances` excluded from auto-seed (requires `dag_run_id`) — correct decision
- `is_available` checks `"airflow" in sources` — clean gating

### `app/integrations/catalog.py`
- `classify_integrations` correctly populates the Airflow block
- `airflow_config_from_env()` reused in `load_env_integrations()` — no duplication ✅
- `"airflow"` present in `direct_services` tuple — integration wired end-to-end

### `app/nodes/plan_actions/`
- `detect_sources.py` correctly detects Airflow from alert context
- `build_prompt.py` adds Airflow-specific context to the investigation prompt
- `plan_actions.py` seeds `get_recent_airflow_failures` and `get_airflow_dag_runs` when Airflow is an available source

### Tests
- 13 unit tests across integration helpers, routing, and catalog resolution — all well-structured, use `monkeypatch` on `httpx.request` cleanly
- E2E tests marked `@pytest.mark.e2e` — CI-safe ✅
- Evidence from real failing DAG (`test_fail_dag`) included via screen recording ✅

All previously raised review concerns (merge conflicts, missing try/except in failure loop, auth precedence, `airflow_config_from_env` deduplication, e2e marker, screen video) have been addressed.

**Approving. Ready to merge.** 🚀

hamzzaaamalik:
Looked at this carefully. Found one issue worth fixing before merge:

P1: tools have source="tracer_web" instead of source="airflow". That's the SaaS webapp source. The prioritizer at [prioritization.py:58](vscode-webview://0hsu3lb4ofn8pn94t5o4a56t750ti35mq306dlvbo7h66b7c8oc6/app/tools/investigation_registry/prioritization.py#L58) gives a +2 score bump when action.source in sources — for an Airflow incident, sources contains "airflow", so the bump never fires for these tools. The seeding mechanism keeps them in the pool, but they score 0 when they should score 2. Looks like the workaround was because EvidenceSource (in [app/types/evidence.py](vscode-webview://0hsu3lb4ofn8pn94t5o4a56t750ti35mq306dlvbo7h66b7c8oc6/app/types/evidence.py)) is a closed Literal and doesn't list "airflow". Easy fix: add "airflow" to the Literal, then update the three @tool(source=...) calls.

P2: AirflowConfig extends BaseModel, not StrictConfigModel — @yashksaini-coder asked about this earlier; doesn't look like it got addressed. Most other integrations use StrictConfigModel to forbid extra fields.

P2: no test for the exception-recovery path. [airflow.py:282-294](vscode-webview://0hsu3lb4ofn8pn94t5o4a56t750ti35mq306dlvbo7h66b7c8oc6/app/integrations/airflow.py#L282-L294) carefully handles a single failing DAG-run-task-instances call so prior evidence isn't lost — but there's no test that exercises that path. Worth a "one fails, one succeeds, evidence preserved" test.

P3, non-blocking: [packaging/sync_release_version.py](vscode-webview://0hsu3lb4ofn8pn94t5o4a56t750ti35mq306dlvbo7h66b7c8oc6/packaging/sync_release_version.py) has an unrelated ruff-format reflow swept in.

Nit: other tools set surfaces=("investigation", "chat"); these don't — intentional?

Otherwise: catalog/env loading/resolve_effective_integrations follow the existing pattern, auth precedence is correct with a regression test, DAG IDs are URL-encoded, lint/format/mypy clean, 13 new tests + no regressions in the broad sweep (1916 passed locally).

LGTM after the source fix.

hamzzaaamalik:
Bug: Wrong source on the tools. All three Airflow tools are tagged source="tracer_web" (which is the SaaS webapp), not source="airflow". This matters because the planner gives tools a +2 score boost when their source matches the incident — so on an Airflow incident, these tools get 0 boost when they should get 2. The seeding logic still gets them into the pool, but they could lose out to other tools when the budget is tight.

Looks like the author hit this because EvidenceSource in app/types/evidence.py is a closed list and doesn't include "airflow". The fix is small:

Add "airflow" to the EvidenceSource list
Change source="tracer_web" to source="airflow" in the three @tool calls
Other things worth fixing:

AirflowConfig uses BaseModel instead of StrictConfigModel — @yashksaini-coder asked about this earlier and it doesn't look addressed. Most other integrations use the strict version.
No test for the failure-recovery path. The code carefully handles the case where one DAG run's API call fails so other evidence isn't lost — but there's no test that actually checks this.
Small things, not blocking:

packaging/sync_release_version.py has an unrelated formatting tweak swept in.
Other tools set surfaces=("investigation", "chat"), these don't — intentional?

cerencamkiran:
Thanks for the detailed review @hamzzaaamalik. 
I used `tracer_web` as a temporary workaround because `EvidenceSource` is currently a closed Literal and did not include `airflow`, so `source="airflow"` failed validation.

On `StrictConfigModel`: I tried switching `AirflowConfig` to it earlier, but it broke validation in this branch, so I kept `BaseModel` to avoid changing runtime behaviour in the same PR.

hamzzaaamalik:
@cerencamkiran : yes but on the source, I'd still push for the fix here, it's small (one line in EvidenceSource + three source= swaps). The tracer_web workaround leaves the planner's audit trail wrong on Airflow incidents.

StrictConfigModel: fine to defer. Mind opening a follow-up issue so it doesn't get lost?

Other stuff (recovery test, surfaces=, the unrelated reflow) is non-blocking.

cerencamkiran:
Got it @hamzzaaamalik makes sense. I’ll add "airflow" to EvidenceSource and update the tool sources.

hamzzaaamalik:
Thanks!

Issue #581, opened 2026-04-14
[BUG] EKS tool output silently dropped by merge_evidence — no mappers in post_process.py


ebrahim-sameh:
### Summary

EKS investigation tools (`list_eks_pods`, `get_eks_events`, `list_eks_deployments`, `get_eks_node_health`, `get_eks_pod_logs`, and others under `app/tools/EKS*/`) execute successfully and return correctly-shaped result dicts, but their output is silently dropped by `merge_evidence()` in `app/nodes/investigate/processing/post_process.py` because the `EVIDENCE_MAPPERS` registry has no entries for any `list_eks_*` / `get_eks_*` / `describe_eks_*` action name. As a result, `state["evidence"]` never gets any `eks_*` keys, and `diagnose_root_cause` builds its prompt without ever seeing pod status, events, deployment health, or node conditions that were gathered earlier in the investigation.

### Expected vs actual behavior

**Expected:** When `list_eks_pods` runs during an investigation and returns its standard envelope

```python
{
    "source": "eks",
    "available": True,
    "cluster_name": "...",
    "namespace": "...",
    "total_pods": 3,
    "pods": [...],
    "failing_pods": [...],
    "high_restart_pods": [...],
    "error": None,
}
```

the relevant fields should be written into `state["evidence"]` under dedicated keys (`eks_pods`, `eks_failing_pods`, `eks_high_restart_pods`, `eks_total_pods`), matching the pattern the Grafana and Datadog tools already follow via `_map_grafana_logs`, `_map_grafana_metrics`, `_map_datadog_logs`, `_map_datadog_monitors`, etc. `diagnose_root_cause` should then see EKS evidence when building its diagnosis prompt.

**Actual:** `merge_evidence()` at `app/nodes/investigate/processing/post_process.py:346-360` looks up each executed action name in the `EVIDENCE_MAPPERS` dict:

```python
mapper = EVIDENCE_MAPPERS.get(action_name)
if mapper:
    evidence.update(mapper(result.data))
```

[...]

Issue #263, opened 2026-04-02
K8s scenarios: Eviction, DNS failures, Probe failures, Quota limits


davincios:
## Goal

We want OpenSRE to be the most reliable agent for diagnosing Kubernetes failures. These four scenarios are the hard ones. Each failure looks like something simpler on the surface, and the agent has to reason its way to the actual root cause. This is where we measure whether the agent can think, not just pattern match.

## Background

Requires the test harness from #260 and the 000-healthy base scenario (which can come from #261, or be created standalone here). These are independent of the other scenario issues and can be delivered in any order.

All four scenarios include Axis 2 annotations (ruling_out_keywords, required_queries) so we can score not just whether the agent got the right answer, but whether it investigated the right signals and dismissed the right alternatives.

## Scenarios

**007-evicted-pods**

Pods get evicted because the node is running out of ephemeral storage. This looks like a pod crash, but the cause is node-level resource pressure, not application code. The fix is completely different.

Key signals: pods with `phase: Failed`, `reason: Evicted`, messages about low ephemeral storage. Node health shows disk pressure. The agent must say eviction, not crash.

Axis 2: agent output must mention that this is a node-level issue and not a container crash.

**008-dns-resolution-failure**

CoreDNS is broken. Application pods can't resolve service names, so they get connection timeouts and 503s. The symptom looks like a network issue or a downstream service being down, but the actual cause is DNS.

Key signals: CoreDNS pods unhealthy or restarting, application logs showing "no such host" and "i/o timeout" on DNS lookups. The agent must identify DNS as the root cause, not blame the network or the downstream service.

Axis 2: agent output must specifically call out DNS resolution, not generic networking.

**009-probe-failure**

Liveness or readiness probes are failing. The pod shows as Running (it is running), but it's not healthy. Liveness probe failures cause restarts, readiness failures remove the pod from service endpoints. This is confusing because the pod status says Running.

Key signals: pod Running with high restart count, Unhealthy warning events with probe failure messages (HTTP 503, TCP connection refused), application logs showing the health endpoint is erroring. The agent must not declare healthy just because the pod is Running.

Axis 2: agent output must distinguish probe failure from CrashLoopBackOff.

**010-resource-quota-exceeded**

The namespace hit its ResourceQuota. Existing pods are healthy, but new pods can't be created. A deployment wants 5 replicas but only has 2 because the quota blocks creation of the remaining 3.

[...]

Issue #262, opened 2026-04-02
K8s scenarios: Node NotReady, Pending Pods, Stuck Rollouts


davincios:
## Goal

We want OpenSRE to be the most reliable agent for diagnosing Kubernetes failures. These three scenarios test infrastructure-level failures where the root cause isn't visible in a single evidence source. The agent has to correlate node conditions with pod states, read scheduler events, and understand deployment rollout mechanics.

## Background

Requires the test harness from #260 and the 000-healthy base scenario (which can come from #261, or be created standalone here). These are independent of the other scenario issues and can be delivered in any order.

## Scenarios

**004-node-not-ready**

A node goes unhealthy due to disk pressure, memory pressure, or PID exhaustion. Some pods on that node become unresponsive or get stuck terminating. Pods on other nodes stay healthy.

Key signals: one node with `ready: false` and a pressure condition set to true in eks_node_health, pods on that node in `Unknown` phase, NodeNotReady warning events. The tricky part is that some pods on other nodes are still fine, so the agent must not declare healthy.

**005-pending-pod**

A pod can't be scheduled. The scheduler has no node that satisfies the pod's resource requests, affinity rules, or tolerations. The pod sits in Pending state indefinitely.

Key signals: pod in `Pending` phase with no node_name assigned, `PodScheduled: False` condition, FailedScheduling events saying something like "0/3 nodes are available: insufficient cpu". Node health shows high allocatable usage.

**006-deployment-rollout-stuck**

A deployment update created a new ReplicaSet, but the new pods can't come up. The old pods are still running fine. The deployment shows unavailable replicas and eventually hits `ProgressDeadlineExceeded`.

Key signals: deployment with desired > ready, `degraded: true`, Progressing condition set to False. Mix of old healthy pods and new broken pods. The agent must identify the stuck rollout rather than declaring partial health.

## Done when

- All three scenarios load and validate through the scenario loader
- Agent returns the correct root_cause_category for each
- Agent does not declare healthy when only some pods/nodes are healthy
- Evidence files match real EKS/Datadog response shapes
- `make lint && make typecheck && make test-cov` pass

## Reference

- Test harness: #260
- EKS node health shape: `app/tools/EKSNodeHealthTool/__init__.py`
- EKS deployment shape: `app/tools/EKSListDeploymentsTool/__init__.py`
- EKS events shape: `app/tools/EKSEventsTool/__init__.py`

Devesh36:
[...]

Issue #261, opened 2026-04-02
K8s scenarios: CrashLoopBackOff, OOMKilled, ImagePullBackOff


davincios:
## Goal

We want OpenSRE to be the most reliable agent for diagnosing Kubernetes failures. These three failure modes are what every on-call engineer sees weekly. If the agent can't nail these, nothing else matters.

## Background

Requires the test harness from #260. These are the simplest K8s failures to diagnose because the signal is usually right there in the pod status and events. Good starting point for validating the full pipeline works.

Also includes the `000-healthy` base scenario that all other scenarios inherit from.

## Scenarios

**000-healthy**

The baseline. A healthy EKS cluster with all pods running, zero restarts, no warning events, all deployments at desired replica count, all nodes ready. This is the base scenario that others inherit from via `base: 000-healthy` in their scenario.yml.

The agent should return `root_cause_category: healthy`.

Full evidence set: eks_pods, eks_events, eks_deployments, eks_node_health, datadog_logs, datadog_monitors.

**001-crashloop-backoff**

Container crashes repeatedly. Kubelet backs off restarts exponentially. Inherits from 000-healthy, overrides the pod and event fixtures.

Key signals: container in `waiting` state with `reason: CrashLoopBackOff`, restart count in double digits, BackOff warning events, application crash logs in Datadog.

**002-oom-killed**

Container exceeds its memory limit. The OOM killer terminates it with exit code 137. Inherits from 000-healthy.

Key signals: container in `terminated` state with `reason: OOMKilled`, `exitCode: 137`, memory allocation failure logs. This one matters because it looks like a crash but the fix is different (increase limits vs fix the code).

**003-image-pull-backoff**

The container image can't be pulled. Wrong tag, missing registry credentials, or the image doesn't exist. Inherits from 000-healthy.

Key signals: container in `waiting` state with `reason: ImagePullBackOff`, events showing `ErrImagePull` and `Failed to pull image`. No application logs because the container never starts.

## Done when

- All four scenarios load and validate through the scenario loader
- Agent returns `healthy` for 000 and the correct category for 001, 002, 003
- Evidence files match real EKS/Datadog response shapes
- `make lint && make typecheck && make test-cov` pass

## Reference

- Test harness: #260
- RDS healthy scenario for reference: `tests/synthetic/rds_postgres/000-healthy/`
- EKS pod response shape: `app/tools/EKSListPodsTool/__init__.py`
[...]

PR #585, opened 2026-04-14
fix: recognise EKS evidence keys in is_clearly_healthy short-circuit

### LINKED ISSUE #582: [BUG] is_clearly_healthy short-circuit never fires for pure-EKS healthy states — eks_* keys missing from _INVESTIGATED_EVIDENCE_KEYS
### Summary

The `is_clearly_healthy()` helper in `app/nodes/root_cause_diagnosis/evidence_checker.py` is used by `diagnose_root_cause` to deterministically skip the reasoning LLM when an investigation has clearly confirmed the workload is healthy. Its last gating condition checks that at least one "investigated" evidence key is populated, using the `_INVESTIGATED_EVIDENCE_KEYS` frozenset. That set recognizes Grafana, RDS/CloudWatch, and Datadog keys but has no entries for any `eks_*` key. A future pure-Kubernetes healthy investigation (no Datadog, no Grafana) therefore never triggers the short-circuit, and the pipeline will invoke the reasoning LLM on a clean state, burning tokens and time on an incident that does not exist.

**Expected:** When an investigation completes successfully against a clearly-healthy Kubernetes workload — pods Running, zero Warning events, deployments at desired replica count, nodes Ready — and the only evidence sources populated come from EKS tools (keys like `eks_pods`, `eks_events`, `eks_deployments`, `eks_node_health`), `is_clearly_healthy(raw_alert, evidence)` should return `True` the same way it does for Grafana-only or Datadog-only healthy states. `diagnose_root_cause` should then fire its healthy short-circuit at `app/nodes/root_cause_diagnosis/node.py:52` and return a deterministic `root_cause_category: healthy` response without calling the reasoning LLM.

**Actual:** `is_clearly_healthy()` at `app/nodes/root_cause_diagnosis/evidence_checker.py:83-128` returns `False` in this scenario. The function's final condition is:

```python
return any(k in evidence for k in _INVESTIGATED_EVIDENCE_KEYS)
```

and `_INVESTIGATED_EVIDENCE_KEYS` is defined as

```python
_INVESTIGATED_EVIDENCE_KEYS = frozenset({
    "grafana_logs",
    "grafana_metrics",
    "grafana_alert_rules",
    "aws_cloudwatch_metrics",
    "aws_rds_events",
    "aws_performance_insights",
    "cloudwatch_logs",
    "datadog_logs",
    "datadog_monitors",
})
```

An evidence dict like `{"eks_pods": [...], "eks_events": [...], "eks_deployments": [...], "eks_node_health": [...]}` contains zero keys from that frozenset. The `any(...)` check is False, `is_clearly_healthy` returns False, the short-circuit is skipped, and the full LLM-driven diagnosis path runs even though no incident exists to diagnose.

### Steps to reproduce

1. Clone the repo and install dev dependencies.
2. Open a Python shell from the repo root:

   ```python
   from app.nodes.root_cause_diagnosis.evidence_checker import is_clearly_healthy

   raw_alert = {
       "state": "normal",
       "alert_source": "datadog",
       "commonLabels": {"severity": "info"},
       "commonAnnotations": {},
   }

   # Pure-EKS healthy evidence: pods running, no events, deployments available, nodes ready.
   eks_only = {
       "eks_pods": [{"name": "payments-api-x", "phase": "Running"}],
       "eks_events": [],
       "eks_deployments": [{"name": "payments-api", "ready": 3, "desired": 3}],
       "eks_node_health": [{"name": "node-1", "ready": "True"}],
   }
   print("pure-EKS healthy:", is_clearly_healthy(raw_alert, eks_only))

   # Add a single Datadog key — no data, just the key being present.
   mixed = {**eks_only, "datadog_logs": []}
   print("EKS + empty datadog_logs:", is_clearly_healthy(raw_alert, mixed))
   ```

3. Observe the output:

   ```
   pure-EKS healthy: False
   EKS + empty datadog_logs: True
   ```

   The only difference is the presence of a `datadog_logs` key — which `_INVESTIGATED_EVIDENCE_KEYS` recognizes — and that single key flips the result. Real EKS evidence with actual pod status, events, and deployments is ignored.

Yes

### How often does it occur?

Every time

### Operating system

Linux

### Logs and error output

N/A — silent behavior change. The practical symptom is that a pure-Kubernetes healthy investigation takes several seconds and uses reasoning-LLM tokens, while the same scenario with any Datadog / Grafana / RDS key present takes zero LLM calls and ~100 ms.

### Additional context

**Proposed fix** (five lines, same class of gap as #XX-EKS-MAPPERS-ISSUE which covers the mapper side of the same EKS plumbing story):

```python
_INVESTIGATED_EVIDENCE_KEYS = frozenset({
    "grafana_logs",
    "grafana_metrics",
    "grafana_alert_rules",
    "aws_cloudwatch_metrics",
    "aws_rds_events",
    "aws_performance_insights",
    "cloudwatch_logs",
    "datadog_logs",
    "datadog_monitors",
    # Kubernetes / EKS evidence — emitted by the EKS mappers in
    # app/nodes/investigate/processing/post_process.py
    "eks_pods",
    "eks_events",
    "eks_deployments",
    "eks_node_health",
    "eks_pod_logs",
})
```

**Tests:** add a parametrized case (or new file `tests/nodes/root_cause_diagnosis/test_evidence_checker.py` if none exists yet) that calls `is_clearly_healthy` with a pure-EKS healthy alert/evidence pair and asserts the return value is `True`. Mirror the existing test style for `is_clearly_healthy` — discoverable via `grep -rn "is_clearly_healthy" tests/`.

**Ordering note:** this gap is only reachable once `merge_evidence` actually populates `eks_*` keys in the first place. Without the EKS evidence mappers being added (sister issue, linked below), `state["evidence"]` will always be devoid of `eks_*` keys and this bug is masked. The two issues should be fixed in order: mappers first, then this one. A contributor working on one should flag the other in their PR body.

**Related:** the EKS evidence-mapper gap in `post_process.py` (sister issue), #260 (Kubernetes synthetic test harness), #261/#262/#263 (future Kubernetes scenarios).

ebrahim-sameh:
Fixes #582

`_INVESTIGATED_EVIDENCE_KEYS` in `app/nodes/root_cause_diagnosis/evidence_checker.py` is used by `is_clearly_healthy()` to decide whether `diagnose_root_cause` can fast-path out of the reasoning LLM when an investigation concludes a Kubernetes workload is clearly healthy. The frozenset recognises every Grafana, CloudWatch, RDS and Datadog evidence family — but has no `eks_*` entries.

Consequence: a pure-Kubernetes healthy investigation whose only populated evidence keys come from the EKS tools (`eks_pods`, `eks_events`, `eks_deployments`, `eks_node_health`, `eks_pod_logs`) fails the final `any(k in evidence for k in _INVESTIGATED_EVIDENCE_KEYS)` check. `is_clearly_healthy` returns `False`, the short-circuit at `app/nodes/root_cause_diagnosis/node.py:52` is skipped, and the reasoning LLM is invoked on a clean state for no reason (burning tokens and several seconds of latency on an investigation that didn't need to happen).

This was filed as a separate bug (#582) alongside the EKS evidence mappers gap (#581) so the two fixes can be reviewed in isolation. The gap is only *reachable* once #581 lands — without the EKS mappers, `state["evidence"]` never gets any `eks_*` keys in the first place — but the two-line fix in this PR is safe to merge in either order because the added set entries are benign when the evidence dict doesn't contain them.

### What this PR changes

Only one source file and one new test file.

**`app/nodes/root_cause_diagnosis/evidence_checker.py`** — extend `_INVESTIGATED_EVIDENCE_KEYS` to include the five K8s evidence keys emitted by the `_map_eks_*` mappers in `app/nodes/investigate/processing/post_process.py`:

```python
_INVESTIGATED_EVIDENCE_KEYS = frozenset({
    "grafana_logs",
    "grafana_metrics",
    "grafana_alert_rules",
    "aws_cloudwatch_metrics",
    "aws_rds_events",
    "aws_performance_insights",
    "cloudwatch_logs",
    "datadog_logs",
    "datadog_monitors",
    # Kubernetes / EKS evidence keys — written by the _map_eks_* mappers in
    # app/nodes/investigate/processing/post_process.py.  Without these, a pure
    # Kubernetes healthy investigation never satisfies the evidence gate below
    # and the reasoning LLM is invoked on a clean state for no reason.
    "eks_pods",
    "eks_events",
    "eks_deployments",
    "eks_node_health",
    "eks_pod_logs",
})
```

Five new members, one explanatory block comment that ties this change to the mapper file it depends on so a future maintainer who touches either file sees the coupling.

### Testing

New test module at `tests/nodes/root_cause_diagnosis/test_evidence_checker.py` — the first tests in this directory (previously empty). Adds 21 tests across three classes:

* **`TestIsClearlyHealthyEKSEvidence`** (6 tests) — for each of the five new keys in isolation plus one combined case, assert that a pure-Kubernetes healthy alert with only EKS evidence populated returns `is_clearly_healthy == True`. The single-key tests catch regressions from accidentally removing one entry in isolation.
* **`TestIsClearlyHealthyExistingSources`** (9 parametrized tests) — regression coverage for every pre-existing key in `_INVESTIGATED_EVIDENCE_KEYS`. Any future refactor that accidentally drops one will flip the matching test immediately.
* **`TestIsClearlyHealthyRejectsUnhealthyStates`** (6 tests) — verify the other gate conditions still reject non-healthy alerts even when an `eks_*` key is present: `state == "alerting"`, `severity == "critical"`, non-empty `error` annotation, evidence with only unknown keys, non-dict alert. One "no evidence at all" case confirms the default rejection path.

All three gate commands pass locally on Python 3.12:

```
make lint         # ruff check app/ tests/  → All checks passed!
make typecheck    # mypy app/               → Success: no issues found in 340 source files
make test-cov     # pytest -n auto --ignore tests/synthetic -m "not synthetic"  → 2158 passed, 1 skipped
```

Baseline before this PR was 2137 passed. The delta of 21 matches the 21 new tests added.

N/A — no user-facing change. This PR touches the reasoning-path evidence gate only. The observable effect for real investigations is that pure-Kubernetes healthy states now complete in ~100 ms via the deterministic short-circuit instead of burning reasoning-LLM tokens on a clean state — same latency profile Grafana-only and Datadog-only healthy investigations already have.

## Impact analysis

* **Backward compatibility:** fully preserved. The change is five additions to a `frozenset`. No existing member is removed or renamed; no type or function signature changes.
* **Behaviour change for production:** when the EKS evidence mappers from #581 are also merged, a clearly-healthy pure-EKS investigation will fast-path out of `diagnose_root_cause` instead of invoking the reasoning LLM. This is the *intended* behaviour documented in the `is_clearly_healthy` docstring — it was just missing data. For investigations that aren't clearly healthy (any non-`normal` state, any non-info severity, any error annotation) the gate still rejects and the LLM path runs as before.
* **Blast radius if the healthy short-circuit misfires:** same as the existing behaviour for Grafana and Datadog. The short-circuit only triggers when all four of `state == normal/resolved/ok`, `severity in info/none/""`, no error-signal annotations, and at least one investigated evidence key is populated. Critical / high / warning alerts cannot satisfy condition 2, so a real incident cannot be silenced by this change. The `HEALTHY_SHORT_CIRCUIT=false` env flag continues to disable the short-circuit entirely without a deploy if needed.
* **Performance:** negligible. `frozenset` membership is O(1), and the set is only iterated once per diagnose step.
* **Secrets / dependencies:** none.

## Related issues

* **#581** (parallel PR) — adds the EKS evidence mappers to `post_process.py` so `state["evidence"]` actually gets the `eks_*` keys this fix checks against. This PR is safe to merge in either order because the added set entries are benign when the evidence dict doesn't contain them.
* **#260 / #583** — Kubernetes synthetic test harness. Does not strictly require this fix but benefits from it for pure-EKS healthy scenarios.
* **#261 / #262 / #263** — future Kubernetes failure scenarios. These scenarios aren't clearly healthy and won't trigger the short-circuit, so this fix has no direct bearing on them.

---

**Problem solved:** a pure-Kubernetes healthy investigation (one where only EKS tools populate evidence) was unable to trigger the deterministic healthy short-circuit in `diagnose_root_cause` because the evidence gate in `is_clearly_healthy` didn't know any EKS keys existed. The agent would call the reasoning LLM on a clean state for no reason. Discovered while scoping the Kubernetes synthetic test harness (#260); filed as an independent bug (#582) so the fix stays out of the harness PR's scope.

**Alternatives considered:**

1. *Make the evidence gate accept any key prefixed with an opt-in allowlist (e.g. `eks_`, `aws_`, `gcp_`).* Rejected because the existing set is explicit and the explicitness is a feature — a new evidence source should not automatically enable the short-circuit until the author has thought about whether "this key is present" really means "investigation was attempted". Keeping the set explicit preserves that review gate.

2. *Check `eks_*` key presence in `is_clearly_healthy` separately, outside the `_INVESTIGATED_EVIDENCE_KEYS` set.* Rejected because it adds a parallel check for no reason. The existing set is already the right structure; it was just missing members.

3. *Bundle this fix into the parallel #581 PR (EKS evidence mappers).* Rejected because the two changes have independent failure modes, independent tests, and touch different files. Keeping them separate makes each review easier and allows either one to be reverted in isolation if needed.

**Why this implementation:**

* **Smallest possible diff that fixes the observable bug.** Five additions to a frozenset literal. No signature changes, no new helpers, no refactors. The surface area for a regression is exactly the five new lines.

* **Documented the coupling inline.** The block comment above the new entries explicitly names `app/nodes/investigate/processing/post_process.py` as the file that emits these keys, so a future maintainer who touches either side of the relationship sees the link without having to trace the pipeline.

* **Non-regression tests first.** `TestIsClearlyHealthyExistingSources` covers every pre-existing key in the frozenset, so any accidental removal during a future refactor is caught immediately by an existing test case. This is why 9 of the 21 new tests are for the keys that already worked.

* **Rejection tests match every gate condition.** `is_clearly_healthy` has four gate conditions (state, severity, error annotations, evidence presence). `TestIsClearlyHealthyRejectsUnhealthyStates` has a test per gate plus one "no evidence at all" case plus one "non-dict alert" case, so every rejection path in the function is directly exercised.

**Key components and their jobs:**

* **The diff itself** — five new string literals inside the `frozenset(...)` call in `evidence_checker.py`. No other runtime behaviour changes.

* **`_healthy_alert()` helper in the test module** — returns a fresh dict per call (not a module-level constant) so individual tests can mutate copies to test rejection paths without cross-test pollution. Matches the style of the existing fixture helpers in `tests/synthetic/rds_postgres/test_suite.py`.

* **`TestIsClearlyHealthyEKSEvidence`** — six individual assertion methods rather than a single parametrized case. Breaking them out makes the failure output clearer when one specific key regresses — the test name directly tells you which key broke.

* **`TestIsClearlyHealthyExistingSources`** — the only parametrized test in the module, because the nine existing keys all follow the same pattern and a single parametrized case keeps the regression coverage compact.

* **`TestIsClearlyHealthyRejectsUnhealthyStates`** — six individual methods, one per gate condition, so a regression in any single gate produces a clear single-test failure rather than a parametrized case where it's harder to see which rejection path flipped.

---

---


davincios:
Thank you very much for fixing this PR!

PR #591, opened 2026-04-15
feat: interactive REPL — the first step toward a Claude-Code-style SRE terminal

### LINKED ISSUE #243: Turning our CLI into a Claud Code like terminal application
# Product Experience: OpenSRE as an Interactive AI SRE Terminal
Today, OpenSRE is a one-shot CLI: you run opensre investigate -i alert.json, it executes the pipeline (extract → plan → investigate → diagnose → publish), prints results, and exits.  

The proposal is to transform this into a persistent, conversational SRE session, the same product experience Claude Code or OpenClaw brought to development, but for incident response.

# Background
Claude Code has set the new standard in how terminal applications should be: 
- interactive and providing direct feedback as well as providing the opportunity to add additional input during execution
- This is also how product excellence looks like today 

Image

# Product Requirements 

1. Stream everything. Characters from the RCA reports appear as they're generated. Silence longer than 200ms gets a spinner. 
3. Zero-exit architecture, meaning that the terminal app never closes. Every command returns to the prompt. Delete: exit codes, one-shot flags, --output files.
4. Identity uptop always with baanner, version, model, status and visible on launch. The user knows exactly what they're talking to.
5. Interrupt, don't queue. User input is always accepted, even mid-operation.
6. Do not ask for permission ever behind single [Y/n]. Provide a "trust mode" toggle. Delete: multi-step wizards, --dry-run as default.
7. Show the work, with user sees what's happening (files changed, commands run) but never internal state, debug logs, or stack traces
8. The session remembers everything said. Previous answers inform future ones. No re-explaining. Delete: stateless invocations, repeated boilerplate.

# What this is NOT

- Not a full chat agent (we already have chat mode). This is investigation-first with conversational refinement.
- Not a dashboard. It's a terminal. Power users live here.
- Not autonomous-by-default. The human is in the loop, steering the investigation, adding context the agent can't see (like "we just deployed").

# Migration Plan:

Phase 1: Decouple

- Unify runners.py and graph_pipeline.py into one execution path
- Kill argparse shim, route everything through Click
- Extract a Renderer protocol so nodes emit events, renderers are pluggable
- Decompose node_publish_findings into pure functions + side-effect handlers
- Wrap AgentState in a SessionState that persists across investigations

Phase 2: Rebuild

- REPL loop: opensre with no args enters a persistent session
- Command router: classify input as new alert, follow-up, slash command, or tool call
- Streaming: graph nodes yield events, tool results appear individually, LLM streams tokens
- Interrupt/redirect: Ctrl+C cancels in-flight tools but keeps evidence, user input pivots the plan
- Approval gates: auto-run read-only tools, [Y/n] for destructive actions, /trust skips all
- Context accumulation: reuse infra context across alerts, accept corrections mid-session
- Stack: prompt_toolkit + Rich + asyncio

yashksaini-coder:
## What this does

Opens up an interactive, zero-exit REPL when you run `opensre` with no arguments in a terminal. Each input is classified as a slash command, a new incident description, or a follow-up question on the previous investigation, and the existing LangGraph pipeline streams live into the session. Walking incidents through a single persistent terminal is the product direction in #243 — this PR ships the foundation.

## Showcase

- Run `opensre` — a compact banner with the active provider, model, and quick hints appears.
- Type `/help`, `/status`, `/reset`, `/trust`, `/clear`, `/exit`, `/quit` for session controls. Bare words like `help` or `exit` are accepted too and routed to their slash form.
- Type an incident description in plain English and the investigation streams in place, the same way remote runs stream.
- After an investigation completes, short questions like `why did CPU spike?` are routed as follow-ups and answered against the stored final state.
- Ctrl+C during an investigation cancels the in-flight LangGraph run without dropping you out of the session. Another Ctrl+C at the empty prompt exits cleanly.

The `opensre investigate -i alert.json` one-shot flow now streams the same way the REPL does — no more waiting for the whole run to finish before anything prints.

## What's in the branch

Each commit is a single logical step:

1. Switch the `investigate` command to the streaming path.
2. Promote `prompt_toolkit` to a direct dependency.
3. Add the `app/cli/repl/` package: banner, session state, router, slash commands, follow-up handler, async loop.
4. Enter the REPL when `opensre` is invoked on a TTY with no subcommand.
5. Add a session-aware streaming investigation runner with Ctrl+C cancellation wiring.
6. Guarantee the progress spinner stops even when the stream raises.
7. Unit tests for the session, the router, and slash dispatch.
8. A short 'Interactive mode' section in the README.
9. Silence a LangGraph warning caused by the PEP-604 annotation format.
10. Skip the "streaming from deployed agent" banner for local runs.
11. Delete a leftover unused helper.

## Out of scope for this PR (known follow-ups)

This is a draft showcase, not the complete Claude Code experience. The following are tracked as next steps:

- Mid-operation input (the signature 'interrupt, don't queue' behavior) — requires moving the investigation into an executor and a concurrent prompt_toolkit input reader.
- A persistent bottom status toolbar showing model, cost, elapsed time during operations.
- Multi-turn conversation memory via the LangGraph chat mode (follow-ups today use a one-shot LLM call against the last state).
- 200ms-silence threshold before the spinner starts (today it starts immediately).
- Trust mode actually gating destructive tools once any exist.
- Tab completion for slash commands and persistent REPL history on disk.

## Verification

- `make lint` — clean
- `make typecheck` — 340 source files, no issues
- `make test-cov` — all REPL unit tests pass; no regressions in the broader CLI suite
- Manual smoke: banner renders, `/help` table shows properly, `help` (no slash) routes the same, Ctrl+C during an investigation cancels cleanly and preserves the session, `/exit` quits.

Closes #243 (draft — expects follow-ups for the deferred items listed above).


app/cli/repl/follow_up.py
```diff
@@ -0,0 +1,107 @@
...
+    try:
+        client = get_llm_for_reasoning()
+        response = client.invoke(prompt)
+    except Exception as exc:  # noqa: BLE001
+        console.print(f"[red]follow-up failed:[/red] {exc}")
+        return
+
+    text = getattr(response, "content", None) or str(response)
+    console.print()
+    console.print(f"[bold cyan]answer:[/bold cyan] {text}")
```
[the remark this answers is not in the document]
yashksaini-coder: done in ae5d199

app/cli/repl/follow_up.py
```diff
@@ -0,0 +1,111 @@
...
+
+    try:
+        from app.services.llm_client import get_llm_for_reasoning
+    except Exception as exc:  # noqa: BLE001
+        console.print(f"[red]LLM client unavailable:[/red] {exc}")
+        return
+
+    context = _summarize_last_state(session.last_state)
+    prompt = (
+        "You are an SRE assistant answering a follow-up question about a prior "
```
muddlebee, line 87: I was checking the code and the [feature-request](https://github.com/Tracer-Cloud/opensre/issues/243). I dont see such text "prompt" calls being mentioned?

Do we need it?

muddlebee:
also an e2e demo would be useful

app/cli/repl/follow_up.py
davincios: Yeah, strictly speaking, not. I do think it's fine to include a prompt here as it can improve the prompts, and really improve the quality of the output 

In this case that's up to the author to implement it or not to implement it :)

davincios:
Great work, Yash @yashksaini-coder. I'm very excited about this PR.

Devesh36:
Hey @yashksaini-coder let me know if you need any help with this one 🚀

app/cli/__main__.py
```diff
@@ -52,6 +70,15 @@ def cli(
...
+            from app.cli.repl import run_repl
+            from app.cli.repl.config import ReplConfig
+
+            config = ReplConfig.load(
+                cli_enabled=interactive,
+                cli_layout=layout,
+            )
+            raise SystemExit(run_repl(config=config))
         render_landing()
         raise SystemExit(0)
```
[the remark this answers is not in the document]
yashksaini-coder: Resolved — the if config.enabled: guard was added in an earlier merge so run_repl now only fires when interactive mode is on; disabled mode falls through to render_landing(). Added a regression test in 5d112c25 (tests/cli/test_main.py::test_no_interactive_falls_through_to_landing_page) so it can't regress.

app/cli/repl/commands.py
```diff
@@ -0,0 +1,622 @@
...
+        session.record("alert", args[0], ok=False)
+        return True
+    except Exception as exc:  # noqa: BLE001
+        console.print(f"[red]investigation failed:[/red] {escape(str(exc))}")
+        session.record("alert", args[0], ok=False)
+        return True
+
+    session.last_state = final_state
+    session.record("alert", f"/investigate {args[0]}")
+    return True
```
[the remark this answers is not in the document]
yashksaini-coder: Resolved in 35a4e3fc. Promoted _accumulate_context from loop.py to ReplSession.accumulate_from_state() (single source of truth — the accumulated key set is now a class attribute). Both loop._run_new_alert AND commands._cmd_investigate_file now call it. Added regression coverage in test_session.py (4 tests on the method itself) and test_commands.py::test_investigate_accumulates_infra_context (integration test on the slash-command path).

yashksaini-coder:
## Progress update

### State of the branch

CI is fully green (quality, typecheck, test, CodeQL, Analyze all passing). The PR is mergeable against upstream/main. 

### Credits for the P1 lift

Most of the P1 work — the big jump from 8 to 26 slash commands plus the two-axis config — came from @rrajan94 in three commits:

- `d311bd25` — two-axis REPL config (CLI flag + env var + `~/.opensre/config.yml` resolution)
- `18d8b513` — 18 new slash commands (integrations, mcp, model, health, doctor, version, history, last, save, context, cost, verbose, compact, template, investigate, stop, cancel, and friends)
- `113a988b` — test coverage for the new config + slashes

Thanks for picking those up. The branch is much richer as a result.

### What we resolved this session

Four CI/review gaps closed:

| SHA | What | Why |
|---|---|---|
| `cf7465f1` | chore: apply ruff format to unblock CI | `quality` check was failing on 7 files — now clean on 888 |
| `5d112c25` | test: lock in --no-interactive landing-page fallthrough | Greptile P1 on `__main__.py:87` — the `if config.enabled:` guard was added by a later merge; added regression coverage so it can't silently regress |
| `35a4e3fc` | fix: /investigate slash command now accumulates infra context | Greptile P1 on `commands.py:378` — promoted `_accumulate_context` from loop.py's private function to `ReplSession.accumulate_from_state()` so both `loop._run_new_alert` AND `commands._cmd_investigate_file` call it. Closes #243 requirement 7 ("session remembers everything") for slash-initiated investigations |
| `c214c74c` | Merge upstream/main | Fork's origin/main was 44 commits behind Tracer-Cloud/main — that's why GitHub was reporting CONFLICTING. Resolved conflicts in `resolve_integrations/node.py` (noqa code update) and 3 daily-update docs |

### How the #243 product requirements stand today

| # | Requirement | State | Notes |
|---|---|---|---|
| 1 | Stream everything; >200ms silence shows spinner | PARTIAL | Node-level spinners work; token-level streaming currently surfaces as 60-char spinner subtext. Wider streaming = Phase B work |
| 3 | Zero-exit architecture | PARTIAL | REPL loop is zero-exit; one-shot `opensre investigate` still exists (intentional — additive ship, not replacement) |
| 4 | Identity banner up top | DONE | `banner.py` renders on launch + `/clear` |
| 5 | Interrupt, don't queue | PARTIAL | Ctrl+C cancels in-flight investigation cleanly; mid-op typing requires the Phase B pinned `Application` |
| 6 | `/trust` toggle, no wizards | PARTIAL | `/trust` toggles `session.trust_mode`; no approval gate reads it yet (no destructive tools today) |
| 7 | Show the work, not internal state | DONE | `StreamRenderer` shows tool calls/node progress; exceptions escape-printed, no tracebacks |
| 8 | Session remembers everything | PARTIAL | `last_state` + `accumulated_context` persist; multi-turn Q&A threading is a follow-up |

### How the discussion #614 decisions are reflected

| Decision | Status | Notes |
|---|---|---|
| Classic retires in ~2 weeks; ship both layouts in place | PARTIAL | `layout=pinned` is accepted by config + CLI + env, but only `classic` is wired today. Phase B will land the actual pinned renderer |
| MCP server: completely separate PR | YES | Nothing MCP-server-exposing has landed in this branch |
| Memory format: `.md` | YES | No memory feature landed yet — confirmed format for when it does |
| `interactive.agentic` default: ON | Deferred to Phase C | Field isn't in `config.py` today. Per the "narrow scope, one PR per tool" direction in the same comment, I'll add the axis as part of the first tool PR (`query_datadog`) with the default flipped ON — not as a dead skeleton now |
| Narrow tool scope, one PR per tool | YES | No tool-framework code added to this PR; stays clean for follow-up PRs |

### Open review threads hygiene

Six threads are still "unresolved" in the UI, but five of them are fixed in later commits — they just need Greptile's next re-scan or a UI-level click. Specifically:

- Greptile P1 on `router.py` (alert-keyword check) — fixed in `13d78d8` (earlier round)
- Greptile P2 on daemon thread leak — fixed in `0b86c254` (earlier round)
- Greptile P1 on `--no-interactive` — fixed + regression test in `5d112c25` (this round)
- Greptile P1 on `/investigate` — fixed in `35a4e3fc` (this round)
- Copilot on streaming UI corrupting JSON stdout — fixed in `117f998b` (earlier round)
- @muddlebee's question about the LLM prompt in `follow_up.py:87` — answered in-thread by @davincios; no code change needed

I've replied on each with the resolving SHA. Happy to click Resolve myself if access allows.

### What's next — phased plan

**Immediate (this PR)** — nothing blocking merge. Classic mode works end-to-end: streaming investigations, grounded follow-ups, 26 slash commands, two-axis config, clean exit paths.

**Follow-up A — Phase B (pinned layout, ~1,080 LoC)**
Stacked PR on top of this one, lands before the ~2-week classic retirement window:
- `app/cli/repl/app.py` — prompt_toolkit `Application(HSplit(header, output, status, input))`
- New `BufferRenderer` writing to a `Buffer` instead of stdout
- Keybindings: Enter / Ctrl+C cancel-or-exit / Ctrl+D / Ctrl+L / ↑↓ history
- Wires `interactive.layout=pinned` (config axis already exists)
- Headless tests via `create_pipe_input` + `DummyOutput`
- Manual compat smoke on ghostty, iTerm, Windows Terminal

**Follow-up B — Phase C (seven parallel per-tool PRs)**
Per @davincios's "narrow scope, one PR per tool, iterate fast":
- PR 1: `query_datadog` + adds `interactive.agentic` config axis (default ON), ReAct scaffold
- PR 2–7: `query_prometheus`, `get_pod_logs`, `describe_pipeline`, `get_recent_alerts`, `list_integrations`, `show_last_state`

Each tool PR is ~150–400 LoC, none blocks any other.

**Deferred to their own issues (per team decision)**
- `opensre` as MCP server
- `~/.opensre/memory.md` persistent context threading

### Thanks

Thanks to @davincios for the product direction calls in #614 — default ON for agentic, narrow PR scope per tool, markdown for memory, MCP separate. Those made the next steps concrete.

Thanks to @rrajan94 for landing the P1 command-surface lift. Happy to hand off Phase B or any of the Phase C tool PRs to whoever's interested.

Thanks to @muddlebee for the spot-check on the follow-up LLM prompt — keeps us honest against the #243 spec.

Ready for review whenever the maintainers have a moment.

muddlebee:
@yashksaini-coder nice work. can we have a demo pls? :eyes:

app/cli/repl/session.py
```diff
@@ -0,0 +1,72 @@
...
+    """If True, skip any future [Y/n] confirmation prompts for read-only tools.
+    (No destructive tools exist today; reserved for forward compatibility.)"""
+
+    token_usage: dict[str, int] = field(default_factory=dict)
+    """Accumulated token counts: {"input": N, "output": N}. Populated when available."""
+
+    # Keys from a completed AgentState that carry reusable infra context into
+    # the next investigation.  Kept as a class-level tuple so any caller that
+    # wants to know "what counts as accumulated context" has a single source.
+    _ACCUMULATED_KEYS: tuple[str, ...] = (
```
muddlebee, line 39: _ACCUMULATED_KEYS has a type annotation on a dataclass without ClassVar, so the dataclass machinery treats it as a regular field with a default — it appears in __init__ as an overridable kwarg and will be flagged by mypy. Change to: _ACCUMULATED_KEYS: ClassVar[tuple[str, ...]] = (...) and add ClassVar to the typing import.

app/cli/repl/loop.py
```diff
@@ -0,0 +1,131 @@
...
+        _run_new_alert(text, session, console)
+        return True
+
+    # follow_up — grounded answer against session.last_state
+    answer_follow_up(text, session, console)
+    session.record("follow_up", text)
+    return True
+
+
+async def _repl_main(initial_input: str | None = None, config: ReplConfig | None = None) -> int:  # noqa: ARG001
```
muddlebee, line 79: config is silently ignored (# noqa: ARG001) — layout is never applied to the Console. This is fine today since only 'classic' exists, but once 'pinned' ships, the wire-up will need to happen here. Add a TODO comment or route to a layout-specific console factory so this doesn't get forgotten.

app/cli/investigate.py
```diff
@@ -163,11 +166,115 @@ def run_investigation_cli_streaming(
...
+
+    resolved_alert_name, resolved_pipeline_name, resolved_severity = resolve_investigation_context(
+        raw_alert=raw_alert,
+        alert_name=None,
+        pipeline_name=None,
+        severity=None,
+    )
+
+    event_queue: queue.Queue[StreamEvent | BaseException | None] = queue.Queue()
+    loop_ref: dict[str, asyncio.AbstractEventLoop] = {}
```
muddlebee, line 235: Race window: if KeyboardInterrupt fires between thread.start() and loop_ref['loop'] = loop (line 218), _cancel_pump() reads loop_ref.get('loop') as None and silently no-ops — the investigation thread leaks. Consider using a threading.Event (set after loop_ref is written) that _cancel_pump waits on briefly before calling call_soon_threadsafe.

app/cli/investigate.py
```diff
@@ -163,11 +166,115 @@ def run_investigation_cli_streaming(
...
+                yield item
+        except KeyboardInterrupt:
+            _cancel_pump()
+            raise
+
+    renderer = StreamRenderer(local=True)
+    try:
+        final_state = renderer.render_stream(_events())
+    except KeyboardInterrupt:
+        _cancel_pump()
```
muddlebee, line 291: _cancel_pump() is called twice on the Ctrl+C path: once in _events() (line 262) and again here. The second call is a no-op because the task is already cancelled, but the double-call is misleading. Remove the call from _events() and let this outer except be the single cancellation point — it's the only place that knows render_stream has exited.

app/cli/repl/router.py
```diff
@@ -0,0 +1,123 @@
...
+        "clear",
+        "reset",
+        "status",
+        "trust",
+    }
+)
+
+
+# Short, question-shaped strings that obviously target the previous investigation.
+_FOLLOW_UP_CUES = (
```
muddlebee, line 28: _FOLLOW_UP_CUES misses common question starters: "when", "where", "which", "who", "did". Without them, inputs like "when did this start?" or "where is the bottleneck?" (<90 chars, end with ?) still route to follow_up because of the endswith('?') check in _is_short_question, but "when did the spike start" (no ?) routes to new_alert and fires a fresh investigation instead. Either extend the cue list or rely solely on the '?' suffix for follow-up detection.

muddlebee:
@yashksaini-coder please add an e2e video demo for showcase also.

VaibhavUpreti:
merging this for velocity, let's fix the issues in a new patch

PR #617, opened 2026-04-16
Fix: Include eks_* keys in _INVESTIGATED_EVIDENCE_KEYS for is_clearly_healthy (Fixes #582)

aniruddhaadak80:
Fixes #582

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## What changed and why
The `is_clearly_healthy()` short-circuit relies on the presence of keys in `_INVESTIGATED_EVIDENCE_KEYS` to verify that an investigation collected evidence. This set was missing all Kubernetes / EKS keys. Because of this gap, investigations finding pure-Kubernetes workloads in a healthy state missed the short-circuit and incorrectly ran the root cause LLM.

This PR adds the missing EKS investigation keys (`eks_pods`, `eks_events`, `eks_deployments`, `eks_node_health`, `eks_pod_logs`) to `_INVESTIGATED_EVIDENCE_KEYS`.

*Note: This relies on the changes from #581 where the EKS mappers populate these keys in `state["evidence"]`.*

## Testing steps with evidence
- Added parameterized unit tests in `tests/nodes/root_cause_diagnosis/test_evidence_checker.py`.
- Tested the `is_clearly_healthy` function directly, ensuring pure-EKS healthy configurations return `True`, mixed outputs return `True`, and an unhealthy `state` correctly blocks it returning `False`.

## Impact analysis
- **Backward Compatibility:** Yes, fully compatible.
- **Breaking Changes:** None. This saves redundant reasoning LLM tokens and time. 

## AI-Assisted Contribution
- [x] I have reviewed every line of code.
- [x] I understand the logic.
- [x] I have tested edge cases.
- [x] I have verified the code matches the project conventions.

PR #616, opened 2026-04-16
Fix: Add missing EVIDENCE_MAPPERS for EKS tools (Fixes #581)

aniruddhaadak80:
Fixes #581

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## What changed and why
The `EVIDENCE_MAPPERS` dict in `post_process.py` was missing mapping functions for several EKS tools. Because of this gap, even when the agent successfully executed tools like `list_eks_pods` or `get_eks_events`, the data was silently dropped and never made it into the investigation state. 

This PR:
1. Adds the missing EKS mapping functions (`_map_eks_pods`, `_map_eks_events`, `_map_eks_deployments`, `_map_eks_node_health`, `_map_eks_pod_logs`).
2. Registers them into `EVIDENCE_MAPPERS`.
3. Updates `build_evidence_summary()` so the EKS tracker appropriately reports the data in the summary logs.

## Testing steps with evidence
- Added parameterized unit tests in `tests/nodes/investigate/test_post_process.py`.
- The tests verify that executing each EKS action correctly merges the expected keys (`eks_pods`, `eks_events`, etc.) into the main evidence dictionary.

## Impact analysis
- **Backward Compatibility:** Yes, fully compatible.
- **Breaking Changes:** None. This strictly appends newly fetched data safely.

## AI-Assisted Contribution
- [x] I have reviewed every line of code.
- [x] I understand the logic.
- [x] I have tested edge cases.
- [x] I have verified the code matches the project conventions.

PR #618, opened 2026-04-16
fix(synthetic-qa): Identify healthy alerts correctly (#596)

### LINKED ISSUE #596: [synthetic-qa] 000-healthy: Validate agent correctly identifies healthy state
## Scenario

**Run:** `opensre tests synthetic --scenario 000-healthy`

All RDS metrics are within normal bounds. No failure exists. The agent should recognize this is a healthy system and not diagnose a problem.

## Correct Answer

- **Root Cause Category:** `healthy`
- **Root Cause:** Database is operating normally — CPU ~15-20%, connections ~93, ReplicaLag <2s, storage stable.

## Human Review Checklist

After running the scenario, read the agent's output and answer:

- [ ] Does the agent conclude the system is **healthy** (not diagnose a problem)?
- [ ] Does the agent mention metrics are within "normal bounds"?
- [ ] Does the agent avoid false-positive diagnoses (e.g. flagging a routine backup event as suspicious)?
- [ ] Does the causal chain say "no failure exists" rather than inventing one?
- [ ] Does the agent complete in 1 investigation loop (not keep searching for a problem that isn't there)?

The agent should say something like: "The database is operating within normal parameters. All metrics are within expected bounds. No anomaly detected." It should NOT try to explain what went wrong — nothing went wrong.

## Common Failure Modes

- Agent invents a problem from low-variance metric noise
- Agent flags the routine automated backup RDS event as suspicious
- Agent keeps looping trying to find something wrong instead of concluding healthy quickly

aniruddhaadak80:
﻿This PR fixes the `000-healthy` synthetic-qa failure (Fixes: #596).

### Cause:
1. The LLM extraction step was classifying 'healthy' and scheduled checks (which have severity 'info' and state 'normal') as `is_noise=True`.
2. Even if it bypassed noise extraction, the LLM was assuming it didn't need to gather investigation metrics because the alert explicitly said the database was normal, leading to an empty sequence of actions. This empty sequence caused the `is_clearly_healthy` function to loop infinitely because condition 4 requires at least one investigative operation.

### Fix:
- **Noise Extraction**: Updated `app/nodes/extract_alert/extract.py` prompt to explicitly state that informational states and health checks are NOT noise.
- **Planner Prompt Guidance**: Updated `app/nodes/plan_actions/build_prompt.py` to ensure the agent MUST still query relevant monitoring platforms for verification when it identifies informational or healthy states.
- **Planner Code Guard**: Added a code-level guard in `app/nodes/plan_actions/node.py` to fall back and force at least one verification action if the LLM returns an empty plan to prevent infinite insufficient_evidence loops.
- **Evidence Consistency**: Fixed EKS evidence truthiness check in `app/nodes/root_cause_diagnosis/evidence_checker.py` to correctly evaluate via `is not None`.
- **Cleanup**: Removed accidentally committed `pr_body.md` template.


Devesh36:
Hey @aniruddhaadak80 could you please resolve the checks which are failing and suggestion from the security bot . If you get stuck somewhere , dm the team on discord we would surely help . Thankss !!

ebrahim-sameh:
Heads up — the EKS portions of this PR duplicate work that already exists.

This PR bundles changes to `post_process.py` and `evidence_checker.py` that are functionally identical to PRs I already have open:

- The `post_process.py` EKS mapper additions are the same as PR #584 (fixes #581, opened Apr 14)
- The `evidence_checker.py` EKS key additions are the same as PR #585 (fixes #582, opened Apr 14)
- The `test_post_process.py` and `test_evidence_checker.py` files in this PR are the same as those in #616 and #617 (also by the same author, also duplicating #584/#585)

I filed both issues (#581 and #582) on Apr 14, proposed the fix approach in each issue body, and opened the corresponding PRs the same day. This PR was opened on Apr 16.

The additional changes in this PR (prompt updates in `extract.py`, `build_prompt.py`, and the empty-plan fallback guard in `node.py`) target issue #596 and are separate work — but the EKS `post_process.py` and `evidence_checker.py` changes are duplicates of my existing PRs.

Flagging so maintainers can evaluate the #596-specific fixes independently from the duplicate EKS plumbing.

PR #626, opened 2026-04-17
fix: Database logic expansion for QA Edge Cases (Batch 2)

aniruddhaadak80:
﻿_build_database_directive() has been expanded exponentially to train the AI to parse red herrings, distinguish between dual fault symptoms versus single root causes, infer missing Storage metrics organically, ignore healthy oscillating traffic metrics, and trace WAL replication lags adequately.


davincios:
Thanks for your contribution @aniruddhaadak80 . 

Please resolve the GPT comments and the GitHub comments because they are valid, and we should not always inject this prompt. 

It is also not validated to resolve the issues mentioned so let's leave those standing :)

aniruddhaadak80:
I've addressed the `cpu_saturation` phantom category concern from the GPT comment by incorporating its wording for the `Replication lag` bullet point.

VaibhavUpreti (approved):
LGTM!

aniruddhaadak80:
> LGTM!

Thank you so much, @VaibhavUpreti .

PR #640, opened 2026-04-18
fix: sort placeholders by length in unmask() to prevent prefix collision

### LINKED ISSUE #639: Bug: MaskingContext.unmask() can corrupt output when placeholder names share a prefix (e.g. <NS_1> and <NS_10>)
## Description

`MaskingContext.unmask()` in `app/masking/context.py` iterates `self._placeholder_map` in arbitrary dict insertion order and calls `str.replace` on each placeholder.  

When two placeholders share a name prefix — e.g. `<NS_1>` and `<NS_10>` — the shorter one (`<NS_1>`) may be replaced first, matching the *prefix* of the longer one (`<NS_10>`) and producing corrupted output like `originalValue0>` instead of the original value for `<NS_10>`.

## Reproduction

```python
from app.masking.context import MaskingContext
from app.masking.policy import MaskingPolicy

policy = MaskingPolicy(enabled=True)
ctx = MaskingContext(policy=policy, placeholder_map={
    "<NS_1>": "host-a",
    "<NS_10>": "host-b",
})

# Simulate unmasking a string that contains <NS_10>
text = "Alert on <NS_10> and <NS_1>"
print(ctx.unmask(text))
# If <NS_1> is processed first: 'Alert on host-a0> and host-a'  ← WRONG
# Expected:                      'Alert on host-b and host-a'
```

## Root Cause

```python
# app/masking/context.py  ~line 95
def unmask(self, text: str) -> str:
    result = text
    for placeholder, original in self._placeholder_map.items():   # ← unordered
        if placeholder in result:
            result = result.replace(placeholder, original)
    return result
```

`str.replace` treats placeholders as plain substrings. Because `<NS_1>` is a substring of `<NS_10>`, iterating in arbitrary order causes the shorter key to match inside the longer one.

## Fix

Sort placeholders by **descending length** before iterating so longer (more-specific) placeholders are always substituted first:

```python
def unmask(self, text: str) -> str:
    if not text or not self._placeholder_map:
        return text
    result = text
    for placeholder, original in sorted(
        self._placeholder_map.items(), key=lambda kv: len(kv[0]), reverse=True
    ):
        if placeholder in result:
            result = result.replace(placeholder, original)
    return result
```

This is an O(n log n) sort on the number of placeholders (typically small) and has no measurable runtime cost.

## Severity

**P0** — directly related to #479.  Any investigation with ≥10 masked identifiers of the same kind will silently produce corrupted de-anonymised output that reaches the user or downstream systems.

fuleinist:
## Fix MaskingContext.unmask() prefix collision bug

**Problem:** When two masked identifiers share a name prefix (e.g. `<NS_1>` and `<NS_10>`), iterating over the placeholder map in arbitrary dict order causes the shorter placeholder to be replaced first, matching inside the longer one and producing corrupted output like `host-a0>` instead of `host-b`.

**Solution:** Sort placeholders by descending length before iterating, so longer (more specific) placeholders are replaced first. This is O(n log n) on the number of placeholders (typically small) with no measurable runtime cost.

**Reproduction (from issue #639):**
```python
from app.masking.context import MaskingContext
from app.masking.policy import MaskingPolicy

policy = MaskingPolicy(enabled=True)
ctx = MaskingContext(policy=policy, placeholder_map={
    "<NS_1>": "host-a",
    "<NS_10>": "host-b",
})
text = "Alert on <NS_10> and <NS_1>"
print(ctx.unmask(text))
# Before fix: 'Alert on host-a0> and host-a' ← WRONG
# After fix:  'Alert on host-b and host-a'   ← CORRECT
```

Closes #639

### DIFF
--- a/app/masking/context.py
+++ b/app/masking/context.py
@@ -108,7 +108,12 @@ class MaskingContext:
         if not text or not self._placeholder_map:
             return text
         result = text
-        for placeholder, original in self._placeholder_map.items():
+        # Sort by descending length so longer (more specific) placeholders are
+        # replaced first. This prevents a short placeholder like <NS_1> from
+        # matching inside a longer one like <NS_10> and corrupting the output.
+        for placeholder, original in sorted(
+            self._placeholder_map.items(), key=lambda kv: len(kv[0]), reverse=True
+        ):
             if placeholder in result:
                 result = result.replace(placeholder, original)
         return result


Devesh36:
Hey @fuleinist - thanks for the PR!
I tried to reproduce this issue on current main using the exact repro from the issue:
```python
ctx = MaskingContext(policy=MaskingPolicy(enabled=True), placeholder_map={
    "<NS_1>": "host-a",
    "<NS_10>": "host-b",
})
text = "Alert on <NS_10> and <NS_1>"
print(ctx.unmask(text))
# Output: "Alert on host-b and host-a" ← Correct, no corruption

Could you clarify:
1. Which version/branch did you encounter this bug on?
2. Is there a specific case that fails on current main?
If you can provide a concrete failing case, I'd be happy to investigate.

davincios:
> ```python
> Could you clarify:
> 1. Which version/branch did you encounter this bug on?
> 2. Is there a specific case that fails on current main?
> If you can provide a concrete failing case, I'd be happy to investigate.
> ```

Nice, great find, thanks for checking this @Devesh36 I couldn't find it either

davincios:
Also @fuleinist , #645 already landed the same change earlier today, so this is a duplicate either way.

Closing this one, but if you can share a concrete failing case (different placeholder shape, etc.), happy to reopen and dig in

In any case, really appreciate the contributions thus far, so thanks very much for your involvement 🔥

PR #643, opened 2026-04-18
feat: support multiple accounts, regions, and clusters per provider

hamzzaaamalik:
Adds multi-instance support to integrations a single provider (Grafana, AWS, Datadog, Honeycomb, Coralogix) can now have multiple named, tagged instances configured via `*_INSTANCES` JSON env vars or a new v2 store schema. Every existing single-instance config keeps working unchanged: legacy env vars are untouched, v1 store files auto-migrate on load, `resolved_integrations[<service>]` still returns the flat default-instance dict. New selector helpers (`get_instance_by_name`, `get_instances_by_tag`) plus Grafana end-to-end selection via a `grafana_instance` alert hint. Explicitly designed around PR #527's three P1 bugs (AWS `role_arn` path mismatch, env data loss via merge-by-service, filtered-get leak) with named regression tests for each. 54 new tests; full suite 2510/2510; zero existing tests modified.

Closes #475.


app/integrations/catalog.py
```diff
@@ -83,369 +114,462 @@ def classify_integrations(integrations: list[dict[str, Any]]) -> dict[str, Any]:
...
+    never multiple records — otherwise ``merge_integrations_by_service``
+    would drop all but one (PR #527 bug #2).
+    """
+    raw = os.getenv(env_name, "").strip()
+    if not raw:
+        return None
+    try:
+        parsed = json.loads(raw)
+    except json.JSONDecodeError as exc:
+        logger.warning(
```
Devesh36, line 573: parse_instances_env logs the parse error including the env var name but NOT its value, which is correct. However the warning message includes the raw exception exc which for json.JSONDecodeError includes the offending character and position — this could leak partial credential values if, for example, a truncated API key was set as the JSON string.

app/integrations/catalog.py
```diff
@@ -83,369 +114,462 @@ def classify_integrations(integrations: list[dict[str, Any]]) -> dict[str, Any]:
...
+        except Exception:
+            return None, None
+        if not grafana_config.endpoint:
+            return None, None
+        if grafana_config.is_local:
+            return {
+                "endpoint": grafana_config.endpoint,
+                "api_key": "",
+                "integration_id": grafana_config.integration_id,
+            }, "grafana_local"
```
Devesh36, line 193: Returning "grafana_local" here causes this instance to be stored under _all_grafana_local_instances instead of _all_grafana_instances. Selector helpers in selectors.py only look up _all_grafana_instances, so local Grafana instances in a multi-instance setup will never be found. Consider returning "grafana" here and handling the local vs non-local difference inside the config itself.

hamzzaaamalik:
@Devesh36 please again review.
Thanks

davincios:
Yeah, looks good. Really appreciate the contribution. @hamzzaaamalik

PR #663, opened 2026-04-19
feat: Victoria Logs Integration

### LINKED ISSUE #126: Victoria logs integration
https://docs.victoriametrics.com/victorialogs/

aniruddhaadak80:
Resolves #126 by adding a complete Victoria Logs integration. This implements a VictoriaLogsClient and a new VictoriaLogsTool that queries using LogsQL.


VaibhavUpreti (changes requested):
Hi @aniruddhaadak80, Could you please add a demo image or video showcasing the feature? Also, kindly update the documentation with any relevant setup instructions.

Once that’s done, please tag @yashksaini-coder and @Devesh36 for an initial review.

Thank you!

yashksaini-coder (commented):
~~Superseded~~ — see the inline review comments below for the full feedback.

app/constants/prompts.py
```diff
@@ -5,6 +5,11 @@
 Your job is to help users triage production alerts, investigate service degradation/outages, and produce evidence-backed conclusions.
 You can query connected systems (e.g., Tracer run/task data, logs, metrics, failed jobs/tools) and developer tooling (e.g., GitHub and Sentry) using available tools.
 
+CRITICAL RCA BEHAVIORS:
+- Evidence First: Clearly separate direct observations from inferences. Do not make premature root-cause claims before sufficient evidence is gathered.
```
yashksaini-coder: These prompt changes (CRITICAL RCA BEHAVIORS block and the general chat rewrite) have nothing to do with Victoria Logs. Please remove them from this PR. If you think they're worth making, open a separate PR for them.

app/integrations/models.py
```diff
@@ -55,7 +55,9 @@ class HoneycombIntegrationConfig(StrictConfigModel):
     @field_validator("dataset", mode="before")
     @classmethod
     def _normalize_dataset(cls, value: object) -> str:
-        return str(value or DEFAULT_HONEYCOMB_DATASET).strip() or DEFAULT_HONEYCOMB_DATASET
+        return (
```
yashksaini-coder, line 58: VictoriaLogsIntegrationConfig is imported in verify.py, the service __init__, and the tool — but it's never actually defined anywhere in this PR or in main. This will throw an ImportError the moment any of those modules are loaded. You need to add the Pydantic model class here, something like:\n\n```python\nclass VictoriaLogsIntegrationConfig(StrictConfigModel):\n    base_url: str = ""\n    tenant_id: str = "0"\n    integration_id: str = ""\n```

app/tools/VictoriaLogsTool/__init__.py
```diff
@@ -0,0 +1,77 @@
+from typing import Any
+
+from app.integrations.models import VictoriaLogsIntegrationConfig
+from app.services.victoria_logs.client import VictoriaLogsClient
```
yashksaini-coder, line 4: This import will crash at load time because VictoriaLogsIntegrationConfig doesn't exist in models.py. Fix the model definition first (see comment on models.py).

app/integrations/verify.py
```diff
@@ -276,13 +300,33 @@ def _build_sts_client(config: dict[str, Any]) -> tuple[Any, str, str]:
             region_name=region,
             aws_access_key_id=credentials.access_key_id if credentials else "",
             aws_secret_access_key=credentials.secret_access_key if credentials else "",
```
yashksaini-coder, line 305: The verify function itself looks fine, but it won't actually get called. For victoria_logs to appear in effective_integrations, catalog.py needs a classification block for it (similar to how other services are handled in _classify_service_instance). Without that, the config will never reach here. Please also add load_env_integrations() support with env vars like VICTORIA_LOGS_URL and VICTORIA_LOGS_TENANT_ID.

app/nodes/extract_alert/extract.py
```diff
@@ -23,11 +23,16 @@ def extract_alert_details(state: InvestigationState) -> AlertDetails:
 is_noise=false (default) for: any alert, error, failure, incident, warning, monitoring notification (including health checks and informational states).
 When in doubt, set is_noise=false.
 
+CRITICAL INSTRUCTIONS FOR RCA CAPABILITY:
```
yashksaini-coder: These changes to alert_source routing and the new CRITICAL INSTRUCTIONS block are not related to Victoria Logs at all. Please remove them from this PR — they change core routing behaviour and would need their own review and tests.

yashksaini-coder (changes requested):
A few things need to be fixed before this can move forward — detailed inline comments below on each file.

yashksaini-coder (changes requested):
Great progress — all the P1 blockers from prior rounds are fixed (EvidenceSource, catalog wiring, EffectiveIntegrations, integration_id injection). Thank you for working through those. One critical runtime bug remains before this can merge.

---

### 🔴 Critical — `extract_params()` returns `{}` → `TypeError` on every real invocation

**File:** `app/tools/VictoriaLogsTool/__init__.py` lines 40–50

The investigation executor calls:
```python
kwargs = action.extract_params(available_sources)  # → {}
data = action.run(**kwargs)                         # → run() with no args
```
But `run(self, query: str, ...)` declares `query` as a **required positional argument with no default**. Python raises `TypeError: run() missing 1 required positional argument: 'query'` every time the tool fires. The tool will never successfully execute in an investigation.

**Fix** — implement `extract_params` following the same pattern as `CoralogixLogsTool` / `ElasticsearchLogsTool`:
```python
def extract_params(self, sources: dict) -> dict[str, Any]:
    vl = sources.get("victoria_logs", {})
    return {
        "query": vl.get("default_query", "*"),
        "limit": vl.get("limit", 50),
        "start": vl.get("start", "-1h"),
        "sources": sources,   # so run() can pull credentials from kwargs
    }
```

---

### 🟡 Major — `tenant_id = "0"` always truthy → `AccountID: 0` sent on every request

`bool("0") is True`, so the `if self.config.tenant_id:` guard in `client.py` always fires and adds `AccountID: 0` to every HTTP request — even when no multi-tenancy is configured. Non-multi-tenant VictoriaLogs deployments may reject or misroute this header.

**Fix:** Change to `tenant_id: str | None = None` in the config model and update the guard to `if self.config.tenant_id is not None:`. Update `catalog.py` line 1544 accordingly.

---

### 🟠 Minor — `'victorialogs'` vs `'victoria_logs'` inconsistency

`extract.py` and `models.py` tell the LLM to output `"victorialogs"` (no underscore) as `alert_source`, but `EvidenceSource`, `tool.source`, and all catalog keys use `"victoria_logs"` (underscore). Please standardize on `"victoria_logs"` throughout — catalog.py already has the alias.

### 🟠 Minor — No `VictoriaLogsTool` unit tests

The 3 existing tests cover integration plumbing (env load, classification, verify). There are no tests for `VictoriaLogsTool.is_available()`, `extract_params()`, or `run()` happy/error paths. Please add `tests/tools/test_victoria_logs_tool.py`.

### 🟠 Minor — Missing integration docs

The contributor checklist in `CONTRIBUTING.md` requires a `docs/<integration>.mdx` page. Please add `docs/victoria_logs.mdx` with setup, env vars, and verification steps.

---

Once those are addressed this will be in good shape to merge. The core integration work is solid.

muddlebee:
@aniruddhaadak80 CI is still red.

PR #672, opened 2026-04-19
fix: recognise Alertmanager, Coralogix, and Honeycomb keys in is_clearly_healthy short-circuit

kaushal-bakrania:
## Summary

- Same drift class as #582 (EKS fix): `_INVESTIGATED_EVIDENCE_KEYS` in `evidence_checker.py` was missing `alertmanager_alerts`, `alertmanager_silences`, `coralogix_logs`, `coralogix_error_logs`, and `honeycomb_traces` — keys that the investigation merge step actually writes via the matching `_map_*` mappers in `post_process.py`.
- Consequence: condition 4 of `is_clearly_healthy` never fires for pure-Alertmanager / pure-Coralogix / pure-Honeycomb healthy states, so every resolved low-severity alert from those stacks pays a full LLM RCA round-trip instead of taking the fast path. Cost + latency paper-cut, not a correctness bug.
- Fix adds the five missing keys to the frozenset. The severity gate (condition 2) still rejects firing critical alerts when these evidence keys are populated, so the safety properties of the short-circuit are unchanged.

Fixes #670.

## Test plan

- [x] `make lint` — clean
- [x] `make typecheck` — clean
- [x] `pytest tests/nodes/root_cause_diagnosis/test_evidence_checker.py` — 28 passed (14 new + 14 existing)
- [x] Added parameterised tests covering all five new keys plus a firing-critical safety test
- [x] Reproduction from the issue now prints `missing: []` and all four keys print `short-circuit -> True`

PR #685, opened 2026-04-19
fix: recognise grafana_alert_rules in check_evidence_availability (#684)

### LINKED ISSUE #684: bug: grafana_alert_rules key missing from check_evidence_availability OR-chain
## Summary
`grafana_alert_rules` is listed in `_INVESTIGATED_EVIDENCE_KEYS` in `app/nodes/root_cause_diagnosis/evidence_checker.py` and is emitted by `_map_grafana_alert_rules` in `app/nodes/investigate/processing/post_process.py`, but `check_evidence_availability` in the same file never probes it in the `has_cloudwatch_evidence` OR-chain.

## Impact
A healthy-state Grafana alert whose only collected evidence is `grafana_alert_rules` (e.g. the investigation pulled alert-rule definitions but no logs / metrics / annotations) falls through to `_handle_insufficient_evidence` rather than satisfying the evidence gate. Downstream routing in `app/nodes/root_cause_diagnosis/node.py:58` treats the case as "nothing was gathered," which is wrong — investigation succeeded.

## Repro
```python
from app.nodes.root_cause_diagnosis.evidence_checker import check_evidence_availability

_, has_cloudwatch, _ = check_evidence_availability(
    {}, {"grafana_alert_rules": [{"uid": "abc"}]}, {}
)
assert has_cloudwatch is True  # FAILS on main — returns False
```

## Root cause
Same drift-bug class as commit 3dbfb521 (EKS keys) and the currently-open #672 (Alertmanager / Coralogix / Honeycomb keys): a key was added to the frozenset and the mapper was wired, but the OR-chain guard in `check_evidence_availability` was not updated in lockstep.

## Fix
One line in the OR-chain:

```python
or evidence.get("grafana_alert_rules") is not None
```

PR to follow.

kaushal-bakrania:
## Summary
- Adds `grafana_alert_rules` to the `has_cloudwatch_evidence` OR-chain in `check_evidence_availability` so it matches `_INVESTIGATED_EVIDENCE_KEYS`.
- Closes #684.

## Context
Same drift-bug class as commit 3dbfb521 (EKS keys) and open PR #672 (Alertmanager / Coralogix / Honeycomb keys): the key is in the frozenset and emitted by `_map_grafana_alert_rules`, but the OR-chain guard was not updated in lockstep. A healthy Grafana alert whose only evidence is `grafana_alert_rules` fell through to `_handle_insufficient_evidence` instead of satisfying the evidence gate.

## Test plan
- [x] `pytest tests/nodes/root_cause_diagnosis/test_evidence_checker.py` — 22 passed
- [x] `ruff check app/nodes/root_cause_diagnosis/evidence_checker.py` — clean
- [x] `mypy app/nodes/root_cause_diagnosis/evidence_checker.py` — clean

### DIFF
--- a/app/nodes/root_cause_diagnosis/evidence_checker.py
+++ b/app/nodes/root_cause_diagnosis/evidence_checker.py
@@ -59,6 +59,7 @@ def check_evidence_availability(
         or evidence.get("grafana_error_logs") is not None
         or evidence.get("grafana_traces") is not None
         or evidence.get("grafana_metrics") is not None
+        or evidence.get("grafana_alert_rules") is not None
         or evidence.get("datadog_logs") is not None
         or evidence.get("datadog_monitors") is not None
         or evidence.get("datadog_events") is not None


VaibhavUpreti (approved):
awesome @kaushal-bakrania !

PR #696, opened 2026-04-20
fix: restore RunnableConfig annotation fix regressed by #594

### LINKED ISSUE #695: UserWarning: RunnableConfig typing regression in resolve_integrations node
`make test-cov` on current `main` emits 12 copies of:

```
app/pipeline/graph.py:46: UserWarning: The 'config' parameter should be typed as 'RunnableConfig' or 'RunnableConfig | None', not 'RunnableConfig | None'.
    graph.add_node("resolve_integrations", node_resolve_integrations)
```

## Root cause

This is a regression of #589. PR #594 (`refactor(integrations)`, merged the same day as #589) re-added `from __future__ import annotations` to `app/nodes/resolve_integrations/node.py`. PEP 563 stringifies the `RunnableConfig | None` annotation at runtime. LangGraph 1.1.6 only whitelists the strings `"RunnableConfig"` and `"Optional[RunnableConfig]"` (see `langgraph/_internal/_runnable.py:132`), so `"RunnableConfig | None"` falls through the whitelist, and the f-string renders the "got" and "expected" identically — hence the self-contradictory-looking message.

The 12× count comes from pytest importing the module repeatedly across the test suite.

## Fix

Remove the `from __future__ import annotations` line from `app/nodes/resolve_integrations/node.py` — the same fix as #589. Other node modules (`auth.py`, `chat.py`) that use `from __future__ import annotations` annotate `config: RunnableConfig` (no `| None`), which stringifies to `"RunnableConfig"` and IS on the whitelist, so they don't warn.

Verified locally: `make test-cov` drops from 22 warnings → 10 (the 12 RunnableConfig warnings are gone); 2725 tests still pass; lint + mypy clean.

Happy to submit a PR.

ebrahim-sameh:
Fixes #695

## Summary

- Removes `from __future__ import annotations` from `app/nodes/resolve_integrations/node.py`.
- PEP 563 stringification of `RunnableConfig | None` does not match LangGraph 1.1.6's whitelist (which accepts `"RunnableConfig"` and `"Optional[RunnableConfig]"` but not `"RunnableConfig | None"`), causing 12 repeated `UserWarning` lines on every `make test-cov` run.
- This is the same fix that was merged in #589, then reintroduced by #594 during a refactor.

## Test plan

- [x] `make test-cov` — **2725 passed, 1 skipped**; warning count drops from **22 → 10** (the 12 RunnableConfig warnings are eliminated)
- [x] `ruff check app/nodes/resolve_integrations/` passes
- [x] `mypy app/nodes/resolve_integrations/node.py` passes

## Follow-up suggestion (out of scope)

Since this fix has now regressed once, a lightweight lint rule or pre-commit check forbidding `from __future__ import annotations` in files containing LangGraph node functions would prevent a third regression. Happy to file as a separate issue if that's useful.

### DIFF
--- a/app/nodes/resolve_integrations/node.py
+++ b/app/nodes/resolve_integrations/node.py
@@ -5,8 +5,6 @@ integration credentials available for all downstream nodes. This replaces
 per-node credential fetching with a single upfront resolution.
 """
 
-from __future__ import annotations
-
 import logging
 import os
 from typing import Any


VaibhavUpreti (approved):
Awesome, thanks for fixing

PR #700, opened 2026-04-20
fix: consolidate pytest config into pytest.ini, stop silently ignoring pyproject

ebrahim-sameh:
Fixes #699

## Summary

- Removes the dead `[tool.pytest.ini_options]` block from `pyproject.toml` that pytest has been silently ignoring (pytest picks `pytest.ini` over pyproject when both exist, emitting `WARNING: ignoring pytest config in pyproject.toml!` on every run of `make test-cov`).
- Copies the accurate `synthetic` marker description ("LLM-based synthetic RCA scenario tests (non-deterministic)") into `pytest.ini`, replacing the stale "fixture-based offline tests; no live infrastructure required" wording — the tests do make live LLM calls, which is why `make test-cov` excludes them via `-m "not synthetic"`.
- Leaves `[tool.coverage.run]` in `pyproject.toml` untouched — pytest-cov reads coverage config independently of pytest's ini resolver.
- Prevents a recurring contributor footgun where edits to `[tool.pytest.ini_options]` appear to take effect but are silently dropped.

## Test plan

- [x] `make test-cov` — **2725 passed, 1 skipped**; `WARNING: ignoring pytest config in pyproject.toml!` no longer appears
- [x] `ruff check` passes
- [x] `pytest --markers` lists `synthetic` with the corrected description

PR #708, opened 2026-04-20
chore: consolidate ruff config into ruff.toml, stop silently ignoring pyproject

### LINKED ISSUE #701: ruff config duplicated between ruff.toml and pyproject.toml, with divergent rule sets
Follow-up to #699. Same pattern as the pytest config bug, ruff side.

## The bug

`ruff.toml` takes precedence over `[tool.ruff.lint]` in `pyproject.toml`, and the two configs have drifted.

**ruff.toml (respected):**
```toml
[lint]
select = ["E", "W", "F", "I", "B", "C4", "UP", "ARG", "SIM"]
```

**pyproject.toml (ignored, lines 118-133):**
```toml
[tool.ruff.lint]
select = ["E", "F", "I", "S"]

[tool.ruff.lint.per-file-ignores]
"*_test.py" = ["S101"]
"tests/**/*_test.py" = ["S101"]
"tests/**/test_*.py" = ["S101"]
"tests/e2e/upstream_lambda/pipeline_code/**" = ["S"]
"tests/e2e/upstream_apache_flink_ecs/pipeline_code/**" = ["S"]
"tests/e2e/upstream_prefect_ecs_fargate/pipeline_code/**" = ["S"]
"tests/shared/**" = ["S108", "S310", "S603", "S607"]
"tests/**/infrastructure_sdk/**" = ["S108", "S603", "S607"]
"tests/**/trigger_lambda/**" = ["S108", "S603", "S607"]
```

## What's different

- `ruff.toml` enables `W, B, C4, UP, ARG, SIM` which pyproject doesn't
- `pyproject.toml` enables `S` (flake8-bandit) which ruff.toml doesn't
- The entire `per-file-ignores` block in pyproject is silently dropped, including the `S101` assert-in-tests ignores and the vendored-pipeline-code `S` ignores

Practical impact is small right now because `S` isn't enabled in the active config anyway, so its dead ignores don't matter. But the block looks deliberate, which means someone wrote it expecting it to apply.

## Why this matters

Same contributor footgun as #699. Edits to `[tool.ruff.lint]` in `pyproject.toml` are silently dropped. If someone re-enables `S` in `ruff.toml` later, the per-file-ignores won't come with it, and tests will start failing on `assert` statements and intentional subprocess / url-open / temp-file patterns.

## Proposed fix

Pick one source of truth. Either drop the `[tool.ruff.lint]` block from `pyproject.toml` (keep `ruff.toml`), or migrate `ruff.toml`'s config into `pyproject.toml` and delete `ruff.toml`.

The pytest PR in #700 takes the "keep the dedicated config file" approach for the same kind of duplication. Following the same convention for ruff keeps things consistent.

I can submit a PR once #700 lands so both changes use the same direction.

ebrahim-sameh:
Fixes #701

Follow-up to #700 — same dual-config pattern, ruff side.

## Summary

Removes the dead `[tool.ruff.lint]` and `[tool.ruff.lint.per-file-ignores]` blocks from `pyproject.toml`. `ruff.toml` takes precedence, so those blocks were silently ignored, and the two configs had drifted: pyproject enabled `S` (flake8-bandit) which ruff.toml doesn't, and declared `S101`/`S*` per-file-ignores that were silently dropped.

Practical behavior is unchanged — `S` was never actually enforced, so its dead ignores didn't matter — but the block looked deliberate and any future edit to `[tool.ruff.lint]` in `pyproject.toml` would be silently lost. Same contributor footgun as the pytest duplication fixed in #700.

Takes the same "keep the dedicated config file" approach (Option A) as #700 for consistency.

## What changed

- Removed 16 lines of dead config from `pyproject.toml`:
  - `[tool.ruff.lint]` block (the `S` rule)
  - `[tool.ruff.lint.per-file-ignores]` block (8 pattern ignores)
- `ruff.toml` is untouched and remains the single source of truth.

## Test plan

- [x] `ruff check` clean on `app/` and `tests/` — no new lint errors surfaced (because `S` was never enforced via the ignored pyproject block)
- [x] `make test-cov` — 2736 passed, 1 skipped, 0 warnings (no regressions)
- [x] `mypy` clean
- [x] Re-ran `make test-cov` on fresh checkout to confirm no hidden dependency on the removed ignores

No behavioral change is expected since the removed config was already being ignored by ruff's precedence rules.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,23 +110,6 @@ include = ["app*", "tests*"]
 [tool.coverage.run]
 data_file = ".pytest_cache/.coverage"
 
-[tool.ruff.lint]
-select = ["E", "F", "I", "S"]
-
-[tool.ruff.lint.per-file-ignores]
-# assert is idiomatic in pytest
-"*_test.py" = ["S101"]
-"tests/**/*_test.py" = ["S101"]
-"tests/**/test_*.py" = ["S101"]
-# vendored third-party pipeline code — not ours to fix
-"tests/e2e/upstream_lambda/pipeline_code/**" = ["S"]
-"tests/e2e/upstream_apache_flink_ecs/pipeline_code/**" = ["S"]
-"tests/e2e/upstream_prefect_ecs_fargate/pipeline_code/**" = ["S"]
-# subprocess / url-open / temp-file usage in infra helpers is intentional
-"tests/shared/**" = ["S108", "S310", "S603", "S607"]
-"tests/**/infrastructure_sdk/**" = ["S108", "S603", "S607"]
-"tests/**/trigger_lambda/**" = ["S108", "S603", "S607"]
-
 [tool.vulture]
 paths = ["app"]
 exclude = [


yashksaini-coder (approved):
Good catch on the dead [tool.ruff.lint] and [tool.ruff.lint.per-file-ignores] blocks that were silently ignored because ruff.toml takes precedence.

Merging this ahead @ebrahim-sameh

PR #709, opened 2026-04-20
feat: add GitDeployTimelineTool for incident-window deploy correlation

hamzzaaamalik:
What was needed:
  The 1 question in RCA is "what changed right before this alert fired?"
  The planning prompt already tells the agent to answer this using
  list_github_commits, but that tool only returns the N most recent
  commits - it cannot filter by time window. So the agent had no real
  way to answer the question it was being asked.

What we did:
  Added a new tool get_git_deploy_timeline that lists commits on a
  GitHub branch between a since/until time window (default: last
  120 minutes before now). Wired it into the investigate pipeline's
  evidence map and updated the GitHub planning hint to point at it.
  Kept list_github_commits available for generic "recent commits"
  queries.

What's done:
  - New tool under app/tools/GitDeployTimelineTool
  - Evidence mapper + summary branch in post_process.py so commits
    show up in final_state.evidence and the terminal summary
  - Planning hint in build_prompt.py now recommends the new tool
    for deploy correlation
  - 28 unit tests covering happy path, edge cases, and bad input
  - ruff clean, mypy clean

### DIFF
--- a/app/nodes/investigate/processing/post_process.py
+++ b/app/nodes/investigate/processing/post_process.py
@@ -301,6 +301,14 @@ def _map_github_commits(data: dict) -> dict:
     }
 
 
+def _map_git_deploy_timeline(data: dict) -> dict:
+    return {
+        "git_deploy_timeline": data.get("commits", []) or [],
+        "git_deploy_timeline_count": data.get("commits_count", 0),
+        "git_deploy_timeline_window": data.get("window", {}),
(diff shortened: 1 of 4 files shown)


VaibhavUpreti (approved):
LGTM, probably you can also work on the dynamic incident time window problem shared on discord and create a discussion around it. Appreciate the contribution!

PR #716, opened 2026-04-20
fix: improve RCA prompt for evidence-grounded diagnosis (Fixes #653)

### LINKED ISSUE #653: Improve the root-cause diagnosis prompt for more evidence-grounded RCA
## Summary
The root-cause diagnosis prompt in `app/nodes/root_cause_diagnosis/prompt_builder.py` is the highest-leverage prompt in the RCA path, but it currently has a few reliability risks:

- it hard-codes a "data pipeline incident" framing even when the incident is really RDS- or Kubernetes-shaped
- it mixes an "evidence only" rule with strong domain directives that can over-anchor the model
- it still depends on free-text section formatting that is later parsed in `app/services/llm_client.py`

This makes the final RCA vulnerable to speculation, category drift, and brittle formatting under adversarial evidence.

## Goal
Improve the diagnosis prompt so the model produces more reliable, evidence-grounded root cause analyses with cleaner separation between observed facts, inferred hypotheses, and the final causal chain.

## Acceptance Criteria
- The updated prompt reduces prompt-induced bias toward the wrong root-cause category when evidence is mixed or adversarial.
- The prompt explicitly reinforces evidence-grounded reasoning and alternative-hypothesis elimination without encouraging unsupported domain priors.
- The prompt either preserves the current parser contract or tightens the output contract so parsing is at least as reliable as today.
- The prompt change must improve or materially harden RCA behavior for synthetic RDS or Kubernetes scenarios from `tests/synthetic/`.

## Required Proof
Use at least one synthetic scenario from `tests/synthetic/rds_postgres/` or `tests/synthetic/eks/`, and preferably an adversarial RDS case such as:

- `011-cpu-storage-compositional`
- `012-replication-lag-misleading-events`
- `014-checkpoint-storm-cpu-saturation`

Include all of the following in the PR or issue follow-up:

- the exact scenario(s) used
- the command used to run them, for example `python -m tests.synthetic.rds_postgres.run_suite --scenario 014-checkpoint-storm-cpu-saturation --mock-grafana`
- the baseline RCA output or failure mode before the prompt change
- the improved RCA output, score, or pass/fail result after the prompt change
- confirmation that the change did not regress the other scenario family when relevant

## Notes
It is acceptable to adjust the prompt structure substantially if that is the cleanest way to make RCA more reliable, but the work should stay focused on the diagnosis prompt rather than broad pipeline refactors.

gauravch-code:
## Description
Fixes #653

Improves the root-cause diagnosis prompt to produce more reliable,
evidence-grounded RCA output with cleaner separation between observed
facts, inferred hypotheses, and the final causal chain.

### `app/nodes/root_cause_diagnosis/prompt_builder.py`
- Replaced vague "be helpful" goal with structured 6-step reasoning
  sequence: observe → hypothesize → check → eliminate → select → report
- Split monolithic RULES into EVIDENCE RULES and ANTI-BIAS RULES
- Added anti-anchoring instructions: model must consider ≥2 alternatives
  before committing, and must discard hypotheses contradicting evidence
- Changed hypothesis framing from "may be incomplete" to "treat as
  suggestions — discard any that contradict the evidence"
- Added ALTERNATIVE_HYPOTHESES_CONSIDERED output section so the model
  explicitly shows eliminated hypotheses
- Improved NON_VALIDATED_CLAIMS to require stating what data would
  confirm or refute each inference

### `app/services/llm_client.py`
- Updated `parse_root_cause()` to handle new ALTERNATIVE_HYPOTHESES_CONSIDERED
  section as a delimiter between NON_VALIDATED_CLAIMS and CAUSAL_CHAIN
- Backward-compatible: parser still works if the LLM omits the new section

## Testing
- Output format remains parseable by existing `parse_root_cause()`
- No runtime behavior changes outside of prompt content
- All existing section headers preserved

### DIFF
--- a/app/nodes/root_cause_diagnosis/prompt_builder.py
+++ b/app/nodes/root_cause_diagnosis/prompt_builder.py
@@ -61,58 +61,69 @@ def build_diagnosis_prompt(
     evidence_text = _build_evidence_sections(state, evidence)
 
     # Construct final prompt
-    prompt = f"""You are an experienced SRE writing a short, evidence-grounded RCA (root cause analysis) for a system incident.
+    prompt = f"""You are an experienced SRE performing a root cause analysis (RCA) for a production incident.
 
-Goal: Be helpful and accurate. Prefer evidence-backed explanations over speculation.
-If the exact root cause cannot be proven, provide the most likely explanation based on observed evidence,
-and clearly state what is unknown.
+OBJECTIVE:
+Produce an evidence-grounded RCA. Separate what you can prove from what you infer.
+Follow this reasoning sequence:
+1. Identify OBSERVED FACTS from the evidence below.
+2. Generate candidate hypotheses that could explain those facts.
+3. For each hypothesis, check whether the evidence confirms, contradicts, or is silent.
+4. Eliminate hypotheses that contradict the evidence.
+5. Select the best-supported hypothesis as the root cause. If multiple survive, pick the most parsimonious one and note alternatives.
+6. If no hypothesis can be confirmed, state "Most likely ..." with the strongest candidate and explicitly list what evidence is missing.
 {upstream_directive}{database_directive}{kubernetes_directive}{memory_section}
 DEFINITIONS:
 - VALIDATED_CLAIMS: Directly supported by the evidence shown below (observed facts).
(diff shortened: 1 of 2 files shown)


VaibhavUpreti (approved):
awesome, well done @gauravch-code ! Thanks for your contribution and welcome to the OpenSRE community.

PR #724, opened 2026-04-21
fix(eks): detect EKS source when AWS integration uses IAM user credentials

### LINKED ISSUE #723: bug(eks): investigation produces zero EKS tools when AWS integration uses IAM user credentials
### Bug

When the AWS integration is configured with **IAM user credentials** (env `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY`, or a stored `access_key_id` / `secret_access_key` on the integration — i.e. no `role_arn`, no injected `_backend`), `opensre investigate` on a Kubernetes alert produces **zero EKS tools** in the plan. EKS pods / events / nodes are silently not investigated even when `cluster_name` is present in the alert annotations and the credentials work for STS / EKS directly.

### Environment

- Commit: `main` (reproduced on current HEAD as of 2026-04-21, also present in 0.x releases)
- Python: 3.12
- AWS integration: configured via `opensre onboarding aws`, choosing "access key / secret" (no role ARN), or via env `AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / `AWS_REGION`.

### Reproduction

1. Configure AWS integration with IAM user credentials, **no role_arn**:

   ```bash
   opensre onboarding aws
   # enter access key / secret / region, skip role ARN
   ```

2. Run investigate on a Kubernetes alert with `cluster_name` set:

   ```bash
   opensre investigate -i tests/e2e/kubernetes/fixtures/datadog_k8s_alert.json
   ```

3. **Expected:** EKS appears in the detected sources; pods / events / node-health tools are planned.

4. **Actual:** EKS is silently skipped — the plan contains no EKS tools.

### Root cause

`app/integrations/catalog.py::resolve_integrations` already lifts IAM-user credentials into `_eks_int["credentials"]` (around L229-L235). But `app/nodes/plan_actions/detect_sources.py` only checks `_eks_int.get("role_arn")` or an injected `_backend` before entering the EKS branch (L660):

```python
_eks_int = (resolved_integrations or {}).get("aws")
_has_injected_eks_backend = bool(_eks_int and "_backend" in _eks_int)
if _eks_int and (_eks_int.get("role_arn") or _has_injected_eks_backend):
    ...  # EKS branch
```

IAM-user integrations have neither `role_arn` nor `_backend`, so the EKS branch is skipped entirely — even though `credentials` are present and valid.

This gate is the only remaining site that forgets `credentials`; other gates already include it correctly:

- `app/integrations/models.py::_require_auth_method` — `self.role_arn or self.credentials` ✔
- `app/integrations/verify.py::_build_sts_client` — `if role_arn: ... else credentials branch` ✔
- `app/cli/wizard/integration_health.py` — same two-branch pattern ✔

### Proposed fix

Accept `_eks_int.get("credentials")` as a third way to gate the EKS branch. Downstream code (`L693`) already tolerates an empty `role_arn` via `_eks_int.get("role_arn", "")`, so no other change is required at the planning layer.

```python
if _eks_int and (
    _eks_int.get("role_arn") or _has_injected_eks_backend or _eks_int.get("credentials")
):
    ...
```

This is a strict superset of the existing behaviour: every integration previously accepted is still accepted. The new branch only activates for integrations that already carry a `credentials` dict produced by `catalog.resolve_integrations`.

### Related

The k8s client that actually consumes these credentials also needs to learn to honour them (it currently calls `_assume_role` unconditionally with an empty `RoleArn`). That is a separate follow-up bug; I will open a second issue + PR for it that `Depends on` this one.

### Offer to fix

I have a fix ready (one-line change in `detect_sources.py`, plus `ruff format`); happy to open a PR referencing this issue. Tested locally with `ruff check`, `ruff format --check`, and `py_compile`.

chaosreload:
Fixes #723

## Summary

When the AWS integration is configured with IAM user credentials (env AK/SK or a stored `access_key_id`/`secret_access_key` with no `role_arn`, no injected `_backend`), `opensre investigate` on a Kubernetes alert silently plans **zero EKS tools** — EKS pods/events/nodes are never queried even though `cluster_name` is present in the alert and the credentials work for STS.

This PR fixes that by letting `_eks_int.get("credentials")` gate the EKS branch in `detect_sources`, matching the pattern already used by `models.py`, `verify.py`, and `integration_health.py`.

Full problem statement, reproduction, and root-cause analysis are in #723.

## Change

One-line change in `app/nodes/plan_actions/detect_sources.py` (plus `ruff format` wrapping it across three lines for line-length):

```python
 if _eks_int and (
-    _eks_int.get("role_arn") or _has_injected_eks_backend
+    _eks_int.get("role_arn") or _has_injected_eks_backend or _eks_int.get("credentials")
 ):
```

## Risk / Backward Compatibility

Strict superset of current behaviour:

- Every integration previously accepted (role_arn or injected backend) is still accepted verbatim.
- The new branch only activates for integrations that already carry a `credentials` dict produced by `catalog.resolve_integrations`.
- Downstream code (`detect_sources.py` L693) already tolerates an empty `role_arn` via `_eks_int.get("role_arn", "")`, so the planning layer needs no other change.

## Same-pattern audit

I grepped the codebase for the same `role_arn` / `_backend`-without-`credentials` gate pattern. `detect_sources.py:660` is the only remaining site. Other auth gates already include `credentials`:

- `app/integrations/models.py::_require_auth_method` — `self.role_arn or self.credentials` ✔
- `app/integrations/verify.py::_build_sts_client` — `if role_arn: ... else credentials branch` ✔
- `app/cli/wizard/integration_health.py` — same two-branch pattern ✔

## Local checks

```
ruff check app/nodes/plan_actions/detect_sources.py   # All checks passed!
ruff format --check app/nodes/plan_actions/detect_sources.py   # 1 file already formatted
python -m py_compile app/nodes/plan_actions/detect_sources.py  # OK
```

## Follow-up

The EKS k8s client that consumes these credentials still needs to learn to honour them (currently calls `_assume_role(role_arn, ...)` unconditionally with an empty `RoleArn`, fails before describe_cluster). A follow-up PR will address that and will explicitly declare `Depends on #<this PR>` in its body.

### DIFF
--- a/app/nodes/plan_actions/detect_sources.py
+++ b/app/nodes/plan_actions/detect_sources.py
@@ -657,7 +657,9 @@ def detect_sources(
     # the role_arn credential gate the same way the Grafana path does.
     _eks_int = (resolved_integrations or {}).get("aws")
     _has_injected_eks_backend = bool(_eks_int and "_backend" in _eks_int)
-    if _eks_int and (_eks_int.get("role_arn") or _has_injected_eks_backend):
+    if _eks_int and (
+        _eks_int.get("role_arn") or _has_injected_eks_backend or _eks_int.get("credentials")
+    ):
         eks_cluster = annotations.get("eks_cluster") or annotations.get("cluster_name")
         # When a backend is injected but the alert omits cluster_name from its
         # annotations, fall back to the first cluster_names entry on the


chaosreload:
Thanks for the review @greptile-apps — good catch on the `credentials` forwarding gap. Confirming we agree this is worth addressing; flagging the plan so the two PRs stay cleanly scoped:

**Plan**

This PR stays scoped to the single-line gate fix. The follow-up PR (on `app/services/eks/eks_k8s_client.py`) will handle credentials forwarding end-to-end, including a second change to `detect_sources.py` that adds `"credentials": _eks_int.get("credentials")` to `eks_params` and threads it through `build_k8s_clients`.

**Why not pre-land the `eks_params` change here**

I looked at this before opening the PR and considered two shapes for the follow-up:

1. **Ambient-resolution inside `build_k8s_clients`** (resolve via `botocore.session.get_credentials()`). Covers the env-var path (`AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` already in `os.environ`) but **does not** cover stored integration credentials produced by `catalog.py` — those live only in the `credentials` dict, never in `os.environ`, so botocore's ambient resolver cannot see them.
2. **Explicit forwarding of the `credentials` dict** through `eks_params` → `build_k8s_clients` (with ambient resolution as a final fallback). Covers both paths.

The follow-up PR will take approach **#2**, so the `eks_params` addition you pointed out is indeed required — just as part of the end-to-end fix, not a pre-landing here. Splitting it would leave a half-wired field on `eks_params` that no consumer reads yet, which I'd rather avoid.

Will link the follow-up PR here once this one lands. Thanks again for the review — your P1 comment pushed me to re-verify that approach #1 alone isn't sufficient for the stored-integration path, which saved a round-trip.

VaibhavUpreti (approved):
GG @chaosreload !

PR #755, opened 2026-04-21
feat: add telegram integration to delivery messages

### LINKED ISSUE #731: [FEATURE] : Add Telegram integration for incident notifications and community triage
### Problem statement

OpenSRE currently has no Telegram integration. Teams and open-source communities that coordinate incident response or triage in Telegram cannot receive investigation findings or trigger investigations from their existing channels.

### Proposed solution

Add a first-class Telegram integration that supports:
- onboarding, authentication (bot token), and connectivity verification via the wizard
- posting investigation summaries, root-cause findings, and evidence to configured Telegram chats or channels
- thread-based conversation support via Telegram reply-to-message for ongoing incident coordination
- following the same delivery pattern as the existing `slack_delivery.py` and `discord_delivery.py` utilities

### Acceptance criteria

- setup and usage are documented clearly in `docs/`
- the implementation is backed by unit tests in `tests/integrations/` and `tests/tools/` plus delivery tests
- the PR includes a screen video showing setup plus a successful investigation flow posting to Telegram
- vibecoding without testing will not be allowed; PRs without working test coverage and screen-video proof should be rejected

paulovitorcl:
Fixes #731 

  ## Summary

  Adds a Telegram Bot integration that delivers investigation findings to a Telegram chat when an investigation completes. The pattern mirrors the existing Discord integration.

  The Telegram integration follows the same delivery pattern as Discord. `post_telegram_message` calls the Bot API `sendMessage` endpoint and returns `(success, error, message_id)`. `send_telegram_report` wraps it with truncation to Telegram's message limit and reads credentials from `telegram_context` (runtime) or integration config (default). The publish node picks up `telegram` from `resolved_integrations` and calls `send_telegram_report` after the Discord block. Verification uses the `getMe` endpoint — same approach as Discord's `users/@me`. The `reply_to_message_id` field supports threading replies to the original alert message.

  ---
  ## Testing

  ```bash
  make lint
  make typecheck
  make test-cov
  make verify-integrations
  opensre integrations verify telegram
  make investigate-alert ALERT=tests/e2e/rca/grafana_pipeline_failure.md
  ```

  **Evidence from the tests:**

image

image

### DIFF
--- a/app/utils/telegram_delivery.py
+++ b/app/utils/telegram_delivery.py
@@ -46,9 +46,7 @@ def post_telegram_message(
         if resp.status_code not in (200, 201):
             logger.warning("[telegram] post message failed: %s", resp.status_code)
             logger.warning("[telegram] api response %s", data)
-            error_message = str(
-                data.get("description", data.get("error", "unknown"))
-            )
+            error_message = str(data.get("description", data.get("error", "unknown")))
             logger.warning("[telegram] post message failed: %s", error_message)
             return False, error_message, ""
         result = data.get("result", {})
--- a/app/utils/telegram_delivery.py
+++ b/app/utils/telegram_delivery.py
@@ -17,6 +17,13 @@ def _truncate(text: str, limit: int) -> str:
     return (text[: limit - 1] + "…") if len(text) > limit else text
 
 
+def _redact_token(text: str, bot_token: str) -> str:
+    """Replace bot token with <redacted> to prevent accidental log/error leakage."""
+    if bot_token and bot_token in text:
+        return text.replace(bot_token, "<redacted>")
+    return text
+
+
 def post_telegram_message(
     chat_id: str,
     text: str,
@@ -53,8 +60,9 @@ def post_telegram_message(
         message_id: str = str(result.get("message_id") or "")
         return True, error_message, message_id
     except Exception as exc:  # noqa: BLE001
-        logger.warning("[telegram] post message exception: %s", exc)
-        return False, str(exc), ""
+        error = _redact_token(str(exc), bot_token)
+        logger.warning("[telegram] post message exception: %s", error)
+        return False, error, ""
 
 
 def send_telegram_report(report: str, telegram_ctx: dict[str, Any]) -> tuple[bool, str]:
--- a/app/utils/telegram_delivery.py
+++ b/app/utils/telegram_delivery.py
@@ -50,7 +50,7 @@ def post_telegram_message(
         )
         data = resp.json()
         error_message = ""
-        if resp.status_code not in (200, 201):
+        if resp.status_code != 200:
             logger.warning("[telegram] post message failed: %s", resp.status_code)
             logger.warning("[telegram] api response %s", data)
             error_message = str(data.get("description", data.get("error", "unknown")))
--- a/app/utils/telegram_delivery.py
+++ b/app/utils/telegram_delivery.py
@@ -1,0 +1,70 @@
+"""Telegram delivery helper - posts investigation findings to Telegram Bot API."""
+
+from __future__ import annotations
+
+import contextlib
+import logging
+from typing import Any
+
+import httpx
+
+logger = logging.getLogger(__name__)
+
+_MESSAGE_LIMIT = 4096
+
+
+def _truncate(text: str, limit: int) -> str:
+    return (text[: limit - 1] + "…") if len(text) > limit else text
+
+
+def post_telegram_message(
+    chat_id: str,
+    text: str,
+    bot_token: str,
+    parse_mode: str = "",
+    reply_to_message_id: str = "",
+) -> tuple[bool, str, str]:
+    """Call Telegram Bot API sendMessage endpoint.
+
+    Returns (success, error, message_id).
+    """
+    logger.debug("[telegram] post message params chat_id: %s", chat_id)
+    payload: dict[str, Any] = {"chat_id": chat_id, "text": text}
+    if parse_mode:
+        payload["parse_mode"] = parse_mode
+    if reply_to_message_id:
+        with contextlib.suppress(ValueError, TypeError):
+            payload["reply_to_message_id"] = int(reply_to_message_id)
+    try:
+        resp = httpx.post(
+            url=f"https://api.telegram.org/bot{bot_token}/sendMessage",
+            json=payload,
+            timeout=15.0,
+        )
+        data = resp.json()
+        error_message = ""
+        if resp.status_code not in (200, 201):
+            logger.warning("[telegram] post message failed: %s", resp.status_code)
+            logger.warning("[telegram] api response %s", data)
+            error_message = str(
+                data.get("description", data.get("error", "unknown"))
+            )
+            logger.warning("[telegram] post message failed: %s", error_message)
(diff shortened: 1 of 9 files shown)


paulovitorcl:
@greptileai review

paulovitorcl:
@greptileai review

rrajan94:
hey @paulovitorcl, nice work on the redaction fix 🙌
gonna pull this down and run it locally before i approve, just want to verify a few edge cases around the delivery path:

- what happens when telegram (or something in front of it) returns non-json
- `send_telegram_report` with empty token / chat_id
- `reply_to_message_id="0"` behavior
- whether the bot token leaks into httpx debug logs (following up on greptile's earlier p1)

will drop findings once i'm done. shouldn't take long

rrajan94:
hey @paulovitorcl, pulled this down locally and ran it end-to-end, works nicely and report landed in my telegram bot clean 🎉

couple of things i'd want to sort before merge though:
1. bot token shows up in httpx logs at INFO level
extending greptile's earlier p1, the `_redact_token` helper you added covers python exceptions, but httpx itself logs the full request URL at INFO, which is the default in most production setups. reproduced locally, the token is right there in the log line:
`INFO:httpx:HTTP Request: POST https://api.telegram.org/bot<TOKEN>/sendMessage "HTTP/1.1 200 OK"`
image
since telegram's API forces the token into the URL path, a logging filter that scrubs the bot<token> pattern would catch this. happy to share a snippet if useful.

2. `send_telegram_report` doesn't guard against missing creds
if `bot_token` or `chat_id` is empty, it still makes an HTTP call and returns a confusing error. the node guards today but the util is a public surface. repro:
```
empty bot_token  → ok=False err='Not Found'
empty chat_id    → ok=False err='Bad Request: chat_id is empty'
```
early return at the top would save the pointless call:
```
pythonif not bot_token or not chat_id:
    return False, "missing bot_token or chat_id"
```

3. `resp.json()` runs before the status check
in `post_telegram_message`, if anything in front of telegram (proxy, cloudflare, a 502) returns a non-JSON body, `resp.json()` throws and gets swallowed by the broad except. caller sees a cryptic JSON error instead of the real HTTP failure. since this is the delivery path for incident reports, error quality matters. swap the order so `resp.json()` only runs inside a try after checking status.


also verified works ✅: full investigation via make investigate-alert, opensre integrations verify telegram, env var loading, and tests/utils/test_telegram_delivery.py all green.
nice work overall, architecture mirrors discord cleanly which made this easy to review 🙌

app/utils/telegram_delivery.py
```diff
@@ -0,0 +1,76 @@
...
+    if reply_to_message_id:
+        with contextlib.suppress(ValueError, TypeError):
+            payload["reply_to_message_id"] = int(reply_to_message_id)
+    try:
+        resp = httpx.post(
+            url=f"https://api.telegram.org/bot{bot_token}/sendMessage",
+            json=payload,
+            timeout=15.0,
+        )
+        data = resp.json()
```
muddlebee: `resp.json()` is called before the status check. If a proxy/load balancer returns a non-JSON error page (e.g. 502/504), this will raise `JSONDecodeError`. The broad `except` catches it but the error message becomes opaque and the real status code is lost.

**Suggestion:** Check `resp.status_code` (or `resp.is_success`) before calling `.json()`. For non-2xx responses, log `resp.text` directly.

app/utils/telegram_delivery.py
```diff
@@ -0,0 +1,76 @@
...
+) -> tuple[bool, str, str]:
+    """Call Telegram Bot API sendMessage endpoint.
+
+    Returns (success, error, message_id).
+    """
+    logger.debug("[telegram] post message params chat_id: %s", chat_id)
+    payload: dict[str, Any] = {"chat_id": chat_id, "text": text}
+    if parse_mode:
+        payload["parse_mode"] = parse_mode
+    if reply_to_message_id:
```
muddlebee: `"0"` is truthy in Python, so `reply_to_message_id="0"` gets converted to `0` and sent to the API. The Telegram API may reject `0` as an invalid message ID.

**Suggestion:** Also guard against `"0"`:
```python
if reply_to_message_id and reply_to_message_id != "0":
```

app/utils/telegram_delivery.py
```diff
@@ -0,0 +1,76 @@
...
+        result = data.get("result", {})
+        message_id: str = str(result.get("message_id") or "")
+        return True, error_message, message_id
+    except Exception as exc:  # noqa: BLE001
+        error = _redact_token(str(exc), bot_token)
+        logger.warning("[telegram] post message exception: %s", error)
+        return False, error, ""
+
+
+def send_telegram_report(report: str, telegram_ctx: dict[str, Any]) -> tuple[bool, str]:
```
muddlebee, line 101: Consider adding an early return guard and docstring here. The function currently accepts empty `bot_token`/`chat_id` and makes the API call anyway. While the caller guards this, the utility should be defensive.

**Suggestion:**
```python
def send_telegram_report(report: str, telegram_ctx: dict[str, Any]) -> tuple[bool, str]:
    """Send a truncated report to Telegram. Returns (success, error)."""
    bot_token = str(telegram_ctx.get("bot_token") or "")
    chat_id = str(telegram_ctx.get("chat_id") or "")
    if not bot_token or not chat_id:
        return False, "Missing bot_token or chat_id"
    ...
```

muddlebee (commented):
Nice work overall! I left a few minor suggestions on edge cases that would be good to address before merge.

muddlebee:
I used Kimi K2.6 + Opencode for automated inline PR review. Seems to work nice.

paulovitorcl:
@rrajan94 Thanks for the thorough review! I've addressed all the points you raised ✅

1 - Added `_TelegramTokenFilter` on httpx/httpcore loggers at module import —> regex bot[^/]+/ scrubs the token from both record.msg and args.
image

2 - Fixed. Added an early return at the top of `send_telegram_report` —> if `bot_token` or `chat_id` is empty, it returns (False, "Missing bot_token or chat_id") immediately before any HTTP call is made.

3 - Fixed. Reordered so the status code is checked first. On non-2xx responses, it tries to parse JSON for the error description but falls back to `resp.text` if the body isn't JSON.

paulovitorcl:
@muddlebee Appreciate the suggestions. All feedback has been addressed!

app/utils/telegram_delivery.py
paulovitorcl: Fixed. Status is now checked before calling `.json()`. For non-2xx responses, it tries to parse JSON for the error description but falls back to `resp.text` directly if the body isn't JSON.

app/utils/telegram_delivery.py
paulovitorcl: Fixed. Added `!= "0"` guard so a zero string is never converted and sent to the API as an invalid message ID.

app/utils/telegram_delivery.py
paulovitorcl: Done. Added the early return guard and docstring as suggested, no HTTP call is made if `bot_token` or `chat_id` is missing.

rrajan94:
code looks clean on review @paulovitorcl 
LGTM 👍

PR #767, opened 2026-04-22
feat: add LLM_MAX_TOKENS env var override to LLMSettings.from_env()

### LINKED ISSUE #739: Support LLM_MAX_TOKENS env var in LLMSettings.from_env
## Why
`LLMSettings.from_env()` currently hardcodes `max_tokens` to `DEFAULT_MAX_TOKENS`, which prevents users from configuring token limits via environment variables.

## What to change
- Add support for an env var (suggested: `LLM_MAX_TOKENS`).
- Parse/validate it as a positive integer.
- Fall back to `DEFAULT_MAX_TOKENS` when missing/invalid (or choose to raise a clear error—just be consistent and test it).
- Add a unit test.

## Where
- `app/config.py` method `LLMSettings.from_env()` starts at line 198.
- `app/config.py` line 278 currently sets `\"max_tokens\": DEFAULT_MAX_TOKENS`.

## Acceptance criteria
- Setting `LLM_MAX_TOKENS=8192` results in `LLMSettings.from_env().max_tokens == 8192`.
- Tests cover at least: no env var (default), valid override, invalid override.

meant more first time contributors to opensre

Agnuxo1:
## Summary

Fixes #739.

`LLMSettings.from_env()` hard-coded `max_tokens` to `DEFAULT_MAX_TOKENS` (4096) with no runtime override path. Every other model-level knob (provider, model IDs, hosts, API keys) is already env-var driven — max tokens was the odd one out.

## Change

```python
# before
"max_tokens": DEFAULT_MAX_TOKENS,

# after
"max_tokens": int(os.getenv("LLM_MAX_TOKENS", str(DEFAULT_MAX_TOKENS))),
```

The `int()` cast ensures Pydantic receives the correct type even when the env value is a string (which `os.getenv` always returns). The `str()` wrapper on the default makes the call signature consistent with the rest of the env-var reads in the same method.

## What is unchanged

- Default behaviour is identical — `DEFAULT_MAX_TOKENS = 4096` is used when `LLM_MAX_TOKENS` is unset.
- All other `LLMSettings` fields are untouched.
- No test changes needed; existing unit tests continue to pass because the default path is identical.

## Testing

```bash
python -c "import os; os.environ['LLM_MAX_TOKENS'] = '8192'; from app.config import LLMSettings; s = LLMSettings.from_env(); print(s.max_tokens)"
# 8192
```

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -275,7 +275,7 @@ class LLMSettings(StrictConfigModel):
                 or DEFAULT_OLLAMA_MODEL,
                 "ollama_host": os.getenv("OLLAMA_HOST", DEFAULT_OLLAMA_HOST).strip()
                 or DEFAULT_OLLAMA_HOST,
-                "max_tokens": int(os.getenv("LLM_MAX_TOKENS", str(DEFAULT_MAX_TOKENS))),
+                "max_tokens": os.getenv("LLM_MAX_TOKENS", str(DEFAULT_MAX_TOKENS)),
             }
         )
 
--- a/app/config.py
+++ b/app/config.py
@@ -275,7 +275,7 @@ class LLMSettings(StrictConfigModel):
                 or DEFAULT_OLLAMA_MODEL,
                 "ollama_host": os.getenv("OLLAMA_HOST", DEFAULT_OLLAMA_HOST).strip()
                 or DEFAULT_OLLAMA_HOST,
-                "max_tokens": DEFAULT_MAX_TOKENS,
+                "max_tokens": int(os.getenv("LLM_MAX_TOKENS", str(DEFAULT_MAX_TOKENS))),
             }
         )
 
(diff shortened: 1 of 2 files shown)


Agnuxo1:
Thanks @greptile-apps — both points addressed in the two follow-up commits:

**P1 — int() cast removed**: passing the raw env string to Pydantic now, which coerces and validates it exactly like the rest of the fields in `from_env()`. Invalid values ("8k", "abc") will produce a clean `ValidationError` instead of a bare `ValueError`.

**P2 — Tests added** (three new tests in `tests/test_config.py`):
- `test_llm_settings_from_env_max_tokens_override` — `LLM_MAX_TOKENS=8192` → `settings.max_tokens == 8192`
- `test_llm_settings_from_env_max_tokens_invalid_raises` — non-numeric value raises `ValueError` or `ValidationError`
- `test_llm_settings_from_env_max_tokens_default` — unset env var → `DEFAULT_MAX_TOKENS` (4096)

Follows the `monkeypatch.setenv` pattern already used in the file.

VaibhavUpreti (approved):
thanks for fixing this @Agnuxo1. Great work!

PR #721, opened 2026-04-21
Fix(makefile): support Windows

gbsierra:
Fixes #720 

 

#### Describe the changes you have made in this PR -
- keep Unix .venv/bin/python behavior unchanged
- add python.exe fallback for Windows venv
- fall back to python3 then python when no local venv exists
- use $(PYTHON) for install instead of hardcoded python3

### Screenshots of the UI changes (If any) -

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [x] I have reviewed every single line of the AI-generated code
- [x] I can explain the purpose and logic of each function/component I added
- [x] I have tested edge cases and understand how the code handles them
- [x] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**

- Windows couldnt resolve Unix venv paths
- Fix Makefile so Windows resolves local python.exe
- Preserve current Unix setup, added additional fallbacks (a minimal cross platform fix)
---

## Checklist before requesting a review
- [x] I have added proper PR title and linked to the issue
- [x] I have performed a self-review of my code
- [x] **I can explain the purpose of every function, class, and logic block I added**
- [x] I understand why my changes work and have tested them thoroughly
- [x] I have considered potential edge cases and how my code handles them
- [x] If it is a core feature, I have added thorough tests
- [x] My code follows the project's style guidelines and conventions

---

Note: Please check **Allow edits from maintainers** if you would like us to assist in the PR.

PR #768, opened 2026-04-22
fix: add Windows .venv/Scripts support to Makefile venv detection

### LINKED ISSUE #763: Makefile fails on Windows due to hardcoded Unix venv path
## Summary
Makefile assumes Unix venv path (.venv/bin/python) or python3, causing make lint / make typecheck / make test-cov to fail on Windows even when a local .venv Python exists.

**Expected:** On Windows, the Makefile should check .venv/Scripts/python.exe (Windows venv layout) before falling back to python3 or python.
**Actual:** The venv path is hardcoded to .venv/bin/python which does not exist on Windows, and install uses python3 directly rather than .

## Steps to reproduce
1. Clone repo on Windows
2. Create .venv using python -m venv .venv
3. Run make lint — fails because .venv/bin/python doesn't exist

## Additional context
The install target also directly calls python3 rather than , so even if the PYTHON variable pointed to the Windows venv Python, install would still invoke python3.

Agnuxo1:
## Summary

Fixes #721. Fixes #763.

The Makefile's venv detection only checks for `.venv/bin/python` (the Unix/macOS path). On Windows, Python's `venv` module places executables under `.venv\Scripts\`, so the check always fails on Windows and every `make` target silently falls back to the system `python3`, bypassing the project virtualenv entirely.

### Before
```makefile
ifneq ($(wildcard .venv/bin/python),)
PYTHON = .venv/bin/python
PIP = .venv/bin/python -m pip
else
PYTHON = python3
PIP = python3 -m pip
endif
export PATH := $(if $(wildcard .venv/bin),$(CURDIR)/.venv/bin:,)$(USER_BIN):$(PATH)
```

### After
```makefile
ifneq ($(wildcard .venv/bin/python),)
PYTHON = .venv/bin/python
PIP = .venv/bin/python -m pip
else ifneq ($(wildcard .venv/Scripts/python.exe),)
PYTHON = .venv/Scripts/python.exe
PIP = .venv/Scripts/python.exe -m pip
else
PYTHON = python3
PIP = python3 -m pip
endif
export PATH := $(if $(wildcard .venv/bin),$(CURDIR)/.venv/bin:,)$(if $(wildcard .venv/Scripts),$(CURDIR)/.venv/Scripts:,)$(USER_BIN):$(PATH)
```

## Priority order

1. `.venv/bin/python` → Unix/macOS venv (unchanged)
2. `.venv/Scripts/python.exe` → Windows venv (new)
3. `python3` → system fallback (unchanged)

The PATH export is extended by the same logic so that venv-installed CLI tools (e.g. `opensre`, `langgraph`) are found without a full path on both platforms.

## What is unchanged

- Behaviour on Unix/macOS is identical — the first `ifneq` branch is taken as before.
- No recipe lines are modified; only the `PYTHON`/`PIP` variable assignment and `PATH` export change.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,9 @@ export
 ifneq ($(wildcard .venv/bin/python),)
 PYTHON = .venv/bin/python
 PIP = .venv/bin/python -m pip
+else ifneq ($(wildcard .venv/Scripts/python.exe),)
+PYTHON = .venv/Scripts/python.exe
+PIP = .venv/Scripts/python.exe -m pip
 else
 PYTHON = python3
 PIP = python3 -m pip
@@ -13,7 +16,7 @@ endif
 # PIP_INSTALL_FLAGS = --user --break-system-packages
 USER_BASE := $(shell $(PYTHON) -m site --user-base)
 USER_BIN := $(USER_BASE)/bin
-export PATH := $(if $(wildcard .venv/bin),$(CURDIR)/.venv/bin:,)$(USER_BIN):$(PATH)
+export PATH := $(if $(wildcard .venv/bin),$(CURDIR)/.venv/bin:,)$(if $(wildcard .venv/Scripts),$(CURDIR)/.venv/Scripts:,)$(USER_BIN):$(PATH)
 
 # Create venv and install dependencies
 install:
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ PIP = python3 -m pip
 endif
 # PIP_INSTALL_FLAGS = --user --break-system-packages
 USER_BASE := $(shell $(PYTHON) -m site --user-base)
-USER_BIN := $(USER_BASE)/bin
+USER_BIN := $(if $(wildcard .venv/Scripts),$(USER_BASE)/Scripts,$(USER_BASE)/bin)
 export PATH := $(if $(wildcard .venv/bin),$(CURDIR)/.venv/bin:,)$(if $(wildcard .venv/Scripts),$(CURDIR)/.venv/Scripts:,)$(USER_BIN):$(PATH)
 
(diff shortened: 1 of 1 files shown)


Agnuxo1:
Thanks @greptile-apps — the `USER_BIN` issue was a real gap. Fixed in the follow-up commit:

```makefile
USER_BIN := $(if $(wildcard .venv/Scripts),$(USER_BASE)/Scripts,$(USER_BASE)/bin)
```

Mirrors the same wildcard guard already used in the `PATH` export line — no new detection logic needed. On Windows, `pip install --user` tools (e.g. `opensre`, `langgraph`) should now be on PATH when `.venv/Scripts` is active.

VaibhavUpreti:
Looks like this has been fixed, thanks for the pull @Agnuxo1 , as a general rule it's best to comment on an issue and wait for it to be assigned before raising a PR

Agnuxo1:
Understood, thanks for the context @VaibhavUpreti. Good to know the Makefile issue was already resolved — I should have checked for an existing fix before opening the PR. Will follow the issue-comment-then-assign workflow going forward.

That said, glad #767 made it in — and thanks for the quick review cycle throughout.

Agnuxo1:
Thanks @VaibhavUpreti — good to know it was already handled. The Makefile Windows fix was something I ran into locally, so it felt worth surfacing even if it turned out to be duplicate work.

Tracer.cloud's approach to AI pipeline observability is exactly the kind of infrastructure layer the ecosystem needs. One thing I'm curious about: as you instrument more agent pipelines, do you see teams struggling more with *visibility* (what did the agent actually call?) or *control* (preventing the agent from calling things it shouldn't)? Trying to understand where the real pain is.

PR #777, opened 2026-04-22
fix: emit healthy-short-circuit claims from investigation keys only

mayankbharati-ops:
## Summary

`_handle_healthy_finding` built the `validated_claims` list by iterating every evidence key and filtering by truthiness. Two problems fall out of that:

1. **Metadata leaks into findings.** `grafana_logs_query`, `datadog_logs_query`, counts (`datadog_monitors_count`, `eks_total_pods`), timings (`datadog_fetch_ms`), service names (`grafana_logs_service`), trace IDs (`honeycomb_trace_id`), URLs (`honeycomb_query_url`) — all truthy, all surfaced as e.g.
   > _\"datadog_logs_query data confirmed within normal operating bounds\"_

2. **Empty-list evidence is silently dropped.** An empty `grafana_logs` / `eks_events` after a completed investigation is itself the healthy signal (per `is_clearly_healthy`'s condition 4 and docstring). Empty lists are falsy, so the claim that actually triggered the short-circuit never appears.

## Fix

Introduce an explicit `CLAIM_EVIDENCE_KEYS` whitelist in `evidence_checker.py`, populated by auditing every mapper in `app/nodes/investigate/processing/post_process.py`. Replace the old truthy-filter with:

```python
def _is_healthy_claim_key(key, value):
    if key in INVESTIGATED_EVIDENCE_KEYS:
        return True           # empty lists count — the healthy signal
    if key in CLAIM_EVIDENCE_KEYS:
        return bool(value)    # adjacent data keys claim when non-empty
    return False              # metadata / unknown keys never claim
```

`CLAIM_EVIDENCE_KEYS` (58 entries) is a superset of `INVESTIGATED_EVIDENCE_KEYS` (16 entries) listing every primary data key across the 39 mappers. Metadata keys — queries, counts, timings, names, IDs, URLs, text summaries — are authoritatively excluded by omission. Also renames `_INVESTIGATED_EVIDENCE_KEYS` → `INVESTIGATED_EVIDENCE_KEYS` (cross-module import, leading underscore was misleading).

**Deliberate trade-off:** a future mapper whose primary data key is not added to `CLAIM_EVIDENCE_KEYS` will not produce a healthy-short-circuit claim for that key until someone updates the whitelist. That's the safe default — silent inclusion is what produced the metadata-leak bug this PR fixes.

## E2E validation — full synthetic suites, real Anthropic LLM

[...]

PR #780, opened 2026-04-23
fix(guardrails): redact union of overlapping spans, never leak sensitive bookends

mayankbharati-ops:
## Summary

`GuardrailEngine.apply` walked sorted redact matches right-to-left with a single `seen_end` cursor and skipped any match whose end exceeded that cursor. When a wider match overlapped — or fully contained — an already-redacted narrower match, the wider match was silently dropped and its prefix / suffix survived in the output.

Concretely, with two rules (wide: `super_secret_token_value`, inner: `secret_token`):

| | Output |
| --- | --- |
| Before (`main`) | `\"data super_[REDACTED:short]_value end\"` |
| After (this PR) | `\"data [REDACTED:long] end\"` |

The `super_` and `_value` fragments leaked into the output. PR #520's earlier fix handled only the same-start subcase; matches with different start offsets that fully contained a narrower match still slipped through.

**Affects the shipped starter rules.** With `_STARTER_CONFIG` in `app/guardrails/cli.py`, text like `config: api_key=AKIAIOSFODNN7EXAMPLE` triggers both `aws_access_key` (`[16:36]`) and `generic_api_token` (`[8:36]`). Pre-fix the prefix `api_key=` leaked; after, the whole span is redacted. Same for `aws_secret_access_key=AKIA...xxxxx...`.

## Fix

Replace the single-cursor walk with the canonical interval-merge algorithm:

1. **Sort** redact matches by `(start ASC, -end)` so same-start ties put the widest match first.
2. **Sweep left-to-right**, merging any match whose `start < merged[-1].end`. Representative rule for the merged span is the contributing match with the largest original width.
3. **Apply replacements right-to-left** so string indices stay valid.

Adjacent matches (`A.end == B.start`) intentionally do not merge. Disjoint matches produce independent redactions. Audit logging is unchanged — match-level entries are still emitted per `ScanMatch` even when output redactions merge.

Algorithm matches the standard merge-intervals pattern (see [Merge Intervals - LeetCode 56](https://leetcode.com/problems/merge-intervals/)).

### Algorithm correctness — adversarial audit (10/11 PASS, 1 was a test bug)

Constructed every nasty edge case I could think of:

| Case | Result |
|---|---|
| Empty input | PASS |
| Single match at start / end / spanning entire text | PASS |
| Same-span same-rule duplicate via pattern + keyword | PASS |
| Nested 3 levels deep | PASS |
| Staircase overlaps (each next overlaps only the prior) | PASS |
| 197 partially-overlapping rules on 1000-char text | **PASS, 8.6 ms** |
| 1000 disjoint same-keyword matches | **PASS, 1.1 ms** |
| Adjacent matches (A.end == B.start) stay separate | PASS |
| Disjoint with 10000-char gap | PASS |
| Match at exact text boundary (start=0 or end=len) | PASS |
| Empty `replacement` falls back to `[REDACTED:name]` | PASS (was test bug, not engine bug) |

### Security audit — bypass attempts (11/11 SAFE)

| Scenario | Result |
|---|---|
| Custom replacement containing another rule's keyword | SAFE — replacements not re-scanned (documented invariant) |
| Block + redact overlap | SAFE — block correctly raises `GuardrailBlockedError` |
| Audit-only rule overlapping a redact rule | SAFE — redact still applied |
| Disabled rule does not bypass active rule | SAFE |
| Replacement string mentions another rule's secret | SAFE |
| 10-deep nested matches sharing the same secret | SAFE — outermost merges all |
| Empty rule list | SAFE (passthrough; no protection promised) |
| Brackets `[X]` in replacement don't confuse downstream | SAFE |

### Determinism

- Same-input → same-output across repeated runs ✓
- `get_guardrail_engine()` singleton is stable ✓
- Tied-width same-span overlap uses first-defined rule name — **identical to `main`'s behavior**, no regression

## E2E validation — full synthetic suites, real Anthropic LLM

`HEALTHY_SHORT_CIRCUIT=true`, mocked backends, real `claude-sonnet-4-6` reasoning + `claude-haiku-4-5` toolcall, on this branch.

| Suite | Result | Notes |
|---|---|---|
| EKS (14 scenarios) | **11 PASS, 3 FAIL** | All 3 failures `state: alerting` LLM categorization drift — pre-existing on `main`, identical to PR #777's verification |
| RDS (15 scenarios) | **11 PASS, 4 FAIL** | All 4 failures `state: alerting`, same pre-existing drift |

**Confirmed every failure is in `state: alerting`** (programmatic check on each `alert.json`). The synthetic fixtures don't contain overlapping-rule secrets, so the failures are in territory this PR does not touch.

### Engine-level (46 tests in `test_engine.py`)

- 36 pre-existing
- **10 new regression tests**: contained-span union (the core regression), longest-rule-name on contained, partial overlap, disjoint, adjacent, three-way transitive overlap, pattern+keyword mixed, two `_STARTER_CONFIG`-grounded production-rule tests, audit-under-merge invariant.

### Integration-level (13 tests in `test_llm_integration.py`)

- 8 pre-existing
- **5 new tests** — one per call site of `engine.apply()`, asserting downstream payload contains no secret fragment:
  - `AnthropicLLMClient.invoke` → captured at `Anthropic.messages.create`
  - `AnthropicLLMClient.invoke` → captured `system` kwarg
  - `OpenAILLMClient.invoke` → captured at `chat.completions.create`
  - `_apply_guardrails_to_messages` (LangGraph chat node) — also asserts source untouched (defensive copy)
  - End-to-end regression guard with shipped-config patterns

`BedrockLLMClient.invoke` shares its redaction code path with `AnthropicLLMClient.invoke` and is therefore covered by the same logic.

## Verification — final state

- **CI (Python 3.13, post-#777-merge main)**: `test (ubuntu-latest)` PASS, `typecheck` PASS, `quality` PASS, `CodeQL` PASS, `Analyze (python)` PASS. (Re-running on rebased branch.)
- **Local `pytest tests/`** (Python 3.14, no coverage): 3124 pass, 1 skipped, 1 pre-existing failure in `tests/nodes/plan_actions/` (unrelated — `_InputStub` missing `model_copy`, present on pristine `main`)
- **`pytest tests/test_guardrails/`**: 91 pass, 0 fail
- `ruff check app/ tests/`: clean
- `ruff format --check` on touched files: clean
- `mypy app/guardrails/`: clean on touched module

## Scope

| File | Change |
|---|---|
| `app/guardrails/engine.py` | ~40 line replacement of redaction loop |
| `tests/test_guardrails/test_engine.py` | 10 new tests |
| `tests/test_guardrails/test_llm_integration.py` | 5 new tests |

No public API changes. No state-shape changes. No behavior change for non-overlapping matches, audit-only rules, or blocked rules. Directly extends #520, whose same-start case is preserved verbatim.

Rebased onto current `main` (post-#777). No file overlap with #777.

## Security note

The bug class is sensitive-data disclosure. Anywhere guardrail rules filter outputs shown to users, logged, or exported, two overlapping rules covering the same secret could each partially redact and leave fragments in the wild. The shipped `_STARTER_CONFIG` demonstrates this concretely — `api_key=AKIA...` and `aws_secret_access_key=...` both produced partial redactions on `main`. This PR closes the gap.

### DIFF
--- a/app/guardrails/engine.py
+++ b/app/guardrails/engine.py
@@ -33,15 +33,23 @@ class _MergedSpan(NamedTuple):
     overlapping ``ScanMatch`` ranges have been collapsed.
 
     ``rule_name`` is the *representative* rule for the span — the contributing
-    match with the largest original ``width`` wins, which preserves the
+    match with the largest ``source_width`` wins, which preserves the
     "longest-keyword-wins" behaviour for same-start overlaps while extending
     it to contained and partially-overlapping spans.
+
+    ``source_width`` is the width of the single ``ScanMatch`` that contributed
+    the current ``rule_name`` — i.e. that match's ``end - start`` *before* any
+    merging happened. It is **not** the merged span's width (``end - start``
+    of the ``_MergedSpan``); in a chained-overlap scenario where A merges B
+    and then C overlaps the grown span, the rule-name winner is whichever
+    individual source match was the widest, not whichever rule covered the
+    widest post-merge range.
     """
 
     start: int
     end: int
     rule_name: str
-    width: int
+    source_width: int
 
 
 @dataclass(frozen=True)
@@ -159,9 +167,10 @@ class GuardrailEngine:
         2. Sweep left-to-right, merging any match whose ``start`` falls before
            the current interval's ``end``. The representative rule for the
            merged interval is whichever contributing match had the largest
-           original width — keeps the existing "longest-keyword-wins"
-           behavior of same-start overlaps while correctly handling contained
-           and partially-overlapping spans.
+           ``source_width`` (its individual ``end - start`` before any merging
+           — *not* the merged span's width). This preserves the existing
+           "longest-keyword-wins" behaviour for same-start overlaps and
+           extends it to contained and partially-overlapping spans.
         3. Apply replacements right-to-left over the merged intervals so string
            indices remain valid as each redaction resizes the output.
         """
@@ -171,16 +180,18 @@ class GuardrailEngine:
         )
         merged: list[_MergedSpan] = []
         for match in redact_matches:
-            width = match.end - match.start
+            # Width of this individual match — used only for representative-
+            # rule selection. The merged span's actual width is ``end - start``.
+            source_width = match.end - match.start
             if merged and match.start < merged[-1].end:
                 prev = merged[-1]
                 new_end = max(prev.end, match.end)
-                if width > prev.width:
-                    merged[-1] = _MergedSpan(prev.start, new_end, match.rule_name, width)
+                if source_width > prev.source_width:
+                    merged[-1] = _MergedSpan(prev.start, new_end, match.rule_name, source_width)
                 else:
-                    merged[-1] = _MergedSpan(prev.start, new_end, prev.rule_name, prev.width)
+                    merged[-1] = _MergedSpan(prev.start, new_end, prev.rule_name, prev.source_width)
             else:
-                merged.append(_MergedSpan(match.start, match.end, match.rule_name, width))
+                merged.append(_MergedSpan(match.start, match.end, match.rule_name, source_width))
 
         redacted = text
         for span in reversed(merged):
--- a/app/guardrails/engine.py
+++ b/app/guardrails/engine.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass, field
+from typing import NamedTuple
 
 from app.guardrails.audit import AuditLogger
 from app.guardrails.rules import (
@@ -25,6 +26,22 @@ class ScanMatch:
     matched_text: str
     start: int
     end: int
+
+
+class _MergedSpan(NamedTuple):
+    """One contiguous interval produced by ``GuardrailEngine._redact`` after
+    overlapping ``ScanMatch`` ranges have been collapsed.
+
+    ``rule_name`` is the *representative* rule for the span — the contributing
+    match with the largest original ``width`` wins, which preserves the
+    "longest-keyword-wins" behaviour for same-start overlaps while extending
+    it to contained and partially-overlapping spans.
+    """
+
+    start: int
+    end: int
+    rule_name: str
+    width: int
 
 
 @dataclass(frozen=True)
@@ -152,23 +169,23 @@ class GuardrailEngine:
             [m for m in matches if m.action == GuardrailAction.REDACT],
             key=lambda m: (m.start, -m.end),
         )
-        merged: list[tuple[int, int, str, int]] = []
+        merged: list[_MergedSpan] = []
         for match in redact_matches:
             width = match.end - match.start
-            if merged and match.start < merged[-1][1]:
-                prev_start, prev_end, prev_rule, prev_width = merged[-1]
-                new_end = max(prev_end, match.end)
-                if width > prev_width:
-                    merged[-1] = (prev_start, new_end, match.rule_name, width)
+            if merged and match.start < merged[-1].end:
+                prev = merged[-1]
+                new_end = max(prev.end, match.end)
+                if width > prev.width:
+                    merged[-1] = _MergedSpan(prev.start, new_end, match.rule_name, width)
                 else:
-                    merged[-1] = (prev_start, new_end, prev_rule, prev_width)
+                    merged[-1] = _MergedSpan(prev.start, new_end, prev.rule_name, prev.width)
             else:
-                merged.append((match.start, match.end, match.rule_name, width))
+                merged.append(_MergedSpan(match.start, match.end, match.rule_name, width))
 
         redacted = text
-        for start, end, rule_name, _ in reversed(merged):
-            replacement = self._get_replacement(rule_name)
-            redacted = redacted[:start] + replacement + redacted[end:]
+        for span in reversed(merged):
+            replacement = self._get_replacement(span.rule_name)
+            redacted = redacted[: span.start] + replacement + redacted[span.end :]
         return redacted
 
     def should_block(self, text: str) -> bool:
--- a/app/guardrails/engine.py
+++ b/app/guardrails/engine.py
@@ -124,20 +124,51 @@ class GuardrailEngine:
         if result.blocked:
             raise GuardrailBlockedError(result.blocking_rules)
 
-        redact_matches = sorted(
-            [m for m in result.matches if m.action == GuardrailAction.REDACT],
-            key=lambda m: (m.start, m.end),
-            reverse=True,
-        )
-        seen_end = len(text)
-        redacted = text
-        for match in redact_matches:
-            if match.end > seen_end:
-                continue
-            replacement = self._get_replacement(match.rule_name)
-            redacted = redacted[: match.start] + replacement + redacted[match.end :]
-            seen_end = match.start
+        return self._redact(text, result.matches)
 
+    def _redact(self, text: str, matches: tuple[ScanMatch, ...]) -> str:
+        """Apply redactions to ``text`` by merging overlapping match intervals.
+
+        The previous single-pass ``seen_end`` walk processed matches right-to-left
+        and skipped any match whose end exceeded the cursor, so a wider match
+        overlapping (or containing) an already-redacted narrower match would
+        leave its prefix/suffix unredacted — e.g. with rules matching
+        ``super_secret_token_value`` and ``secret_token`` on the same text, the
+        ``super_`` and ``_value`` bookends survived in the output.
+
+        Algorithm:
+        1. Sort redact matches by ``(start ASC, -end)`` so ties at the same
+           starting offset put the widest match first.
+        2. Sweep left-to-right, merging any match whose ``start`` falls before
+           the current interval's ``end``. The representative rule for the
+           merged interval is whichever contributing match had the largest
+           original width — keeps the existing "longest-keyword-wins"
+           behavior of same-start overlaps while correctly handling contained
+           and partially-overlapping spans.
+        3. Apply replacements right-to-left over the merged intervals so string
+           indices remain valid as each redaction resizes the output.
+        """
+        redact_matches = sorted(
+            [m for m in matches if m.action == GuardrailAction.REDACT],
+            key=lambda m: (m.start, -m.end),
+        )
+        merged: list[tuple[int, int, str, int]] = []
+        for match in redact_matches:
+            width = match.end - match.start
+            if merged and match.start < merged[-1][1]:
+                prev_start, prev_end, prev_rule, prev_width = merged[-1]
+                new_end = max(prev_end, match.end)
+                if width > prev_width:
+                    merged[-1] = (prev_start, new_end, match.rule_name, width)
+                else:
+                    merged[-1] = (prev_start, new_end, prev_rule, prev_width)
+            else:
+                merged.append((match.start, match.end, match.rule_name, width))
+
+        redacted = text
+        for start, end, rule_name, _ in reversed(merged):
+            replacement = self._get_replacement(rule_name)
+            redacted = redacted[:start] + replacement + redacted[end:]
         return redacted
 
     def should_block(self, text: str) -> bool:
--- a/tests/test_guardrails/test_llm_integration.py
+++ b/tests/test_guardrails/test_llm_integration.py
@@ -266,3 +266,224 @@ class TestChatNodeGuardrails:
         msgs: list[Any] = [_FakeMessage("AKIAIOSFODNN7EXAMPLE")]
         result = _apply_guardrails_to_messages(msgs)
         assert result[0].content == "AKIAIOSFODNN7EXAMPLE"
+
+
+# Production-grade configs exercising every reachable overlap shape the fix
+# must handle. ``aws_access_key`` is a strict substring of
+# ``generic_api_token`` when the token value is itself an AWS key, so a
+# real investigation that quotes ``api_key=AKIA...`` from a source file
+# triggers the contained-span path that main corrupts.
+_OVERLAPPING_RULES: list[dict] = [
+    {
+        "name": "aws_access_key",
+        "action": "redact",
+        "patterns": [r"(?:AKIA|ASIA)[A-Z0-9]{16}"],
+    },
+    {
+        "name": "generic_api_token",
+        "action": "redact",
+        "patterns": [
+            r"(?i)(?:api_key|api_token|auth_token|access_token|secret_key)"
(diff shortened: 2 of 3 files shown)


mayankbharati-ops:
Heads up: the `test (ubuntu-latest)` CI failure on this PR is a pre-existing bug on `main`, not introduced here.

`tests/nodes/plan_actions/test_node_plan_actions.py::test_node_plan_actions_emits_retrieval_controls` raises `AttributeError: '_InputStub' object has no attribute 'model_copy'` because PR #807 added a `model_copy` / `model_dump` round-trip in `app/nodes/plan_actions/node.py:55` but did not update the test stub. The same failure reproduces on a clean checkout of `main` with no other changes.

I opened #929 with the minimal stub repair (14 lines in the test file, no production code touched). Once that lands, I'll rebase this PR — the rest of CI on this branch is green (`typecheck`, `quality`, `CodeQL`, `Analyze (python)`).

yashksaini-coder (approved):
This is a well-engineered security fix. Here's what I verified:

**Algorithm correctness** — the interval merge is textbook correct:
- Sort by `(start ASC, -end)` puts widest match first at same start offset, preserving the behavior from #520 while enabling a correct forward sweep
- Merge condition `match.start < merged[-1].end` correctly catches containment and partial overlap; adjacent spans (`A.end == B.start`) intentionally stay separate ✅
- `max(prev.end, match.end)` handles partial overlap correctly; width comparison picks the widest individual match's rule name for the representative label ✅
- Right-to-left application over non-overlapping merged intervals keeps string indices valid throughout ✅
- Empty match list → empty `merged` → `text` returned unchanged ✅

I traced through all adversarial shapes (3-way transitive, containment, same-span duplicate, widest-first) — all produce correct output.

**Tests** — 91/91 pass. The 10 new engine regression tests cover every overlap subclass that the old `seen_end` cursor got wrong. The 5 LLM integration tests asserting no secret fragment reaches an API call are the right layer to test — good defensive coverage.

**Security impact** — this closes a real data-disclosure gap. With the shipped `_STARTER_CONFIG`, `api_key=AKIA...` and `aws_secret_access_key=...` both triggered two rules simultaneously and the prefix/suffix bookends leaked on `main`. Fixed.

**CI** — all checks green: `test`, `typecheck`, `quality`, `CodeQL`, `Analyze (python)`. The pre-existing `_InputStub` failure the author called out was fixed by #929 which is already on `main`. The branch is clean and mergeable.

No issues. Excellent work.

muddlebee:
@mayankbharati-ops still trying to understand the context for this change? could you help

mayankbharati-ops:
@muddlebee — happy to walk through it. The TL;DR is **a sensitive-data leak in `GuardrailEngine`'s redaction pass when two rules overlap on the same span — including with the rules we ship in `_STARTER_CONFIG`.** I just re-ran a live repro on `main` to confirm everything below.

## 1. The bug, in one screenshot of output

I ran the exact same scenarios through `GuardrailEngine.apply` on `main` (current HEAD) and on this branch:

| | input | `main` (current shipping behavior) | this PR |
|---|---|---|---|
| Synthetic | `data super_secret_token_value end` | `'data super_[REDACTED:short]_value end'` ❌ | `'data [REDACTED:long] end'` ✅ |
| **Shipping `_STARTER_CONFIG`** | `config: api_key=AKIAIOSFODNN7EXAMPLE rest` | `'config: api_key=[REDACTED:aws_access_key] rest'` ❌ | `'config: [REDACTED:generic_api_token] rest'` ✅ |
| 3-way transitive | `abcdefghijklmno tail` | `'[REDACTED:a]h[REDACTED:c] tail'` ❌ | `'[REDACTED:a] tail'` ✅ |
| Disjoint (regression check) | `foo middle bar` | `'[REDACTED:a] middle [REDACTED:b]'` ✅ | `'[REDACTED:a] middle [REDACTED:b]'` ✅ |
| Adjacent (regression check) | `foobar tail` | `'[REDACTED:a][REDACTED:b] tail'` ✅ | `'[REDACTED:a][REDACTED:b] tail'` ✅ |

Row 2 is the worst one: with `app/guardrails/cli.py:_STARTER_CONFIG` (the rules shipped to every user via `opensre guardrails init`), the literal substring `api_key=` survives in any text we redact and forward to an LLM — because `aws_access_key` (matches `AKIA…` only, span [16:36]) and `generic_api_token` (matches `api_key=AKIA…`, span [8:36]) both fire, and `main`'s loop drops the wider one. Same shape with `aws_secret_access_key=AKIA…xxxxx…`.

## 2. Why `main` does this — the algorithm

`GuardrailEngine.apply` had this redact loop on `main` ([engine.py:127-138](https://github.com/tracer-cloud/opensre/blob/main/app/guardrails/engine.py#L127-L138)):

```python
redact_matches = sorted(matches_with_REDACT_action, key=(start, end), reverse=True)
seen_end = len(text)
for match in redact_matches:
    if match.end > seen_end:        # ← drops the wider match
        continue
    redacted = redacted[:match.start] + replacement + redacted[match.end:]
    seen_end = match.start
```

It walks right-to-left with a single `seen_end` cursor and **silently skips any match whose `end` exceeds the cursor**. PR #520 fixed this for *same-start* overlaps (longest-keyword-wins) but matches with *different* start offsets that fully contain a narrower match still slip through — the wider match's prefix and suffix are never written over.

## 3. The fix — canonical interval merge

[engine.py change](https://github.com/Tracer-Cloud/opensre/pull/780/files#diff-app-guardrails-engine-py) replaces the cursor walk with the standard merge-intervals algorithm:

1. **Sort** redact matches by `(start ASC, -end)` so same-start ties put the widest match first (preserves PR #520's behavior).
2. **Sweep left-to-right.** Merge any match whose `start < merged[-1].end`. Track the contributing match with the largest original *width* — that one wins the representative `rule_name` for the merged span.
3. **Apply replacements right-to-left** over the merged intervals so string indices stay valid.

Adjacent matches (`A.end == B.start`) intentionally do **not** merge — verified by the `test_adjacent_matches_not_merged` test. Disjoint matches stay independent. Audit logging is unchanged: every individual `ScanMatch` still emits its own audit entry; we only collapse the *output* redactions.

(The merged-span list uses a `_MergedSpan: NamedTuple` — addressing the Greptile P2 nit on the original commit; this is what `d26c631 refactor(guardrails): use NamedTuple for merged-span records` does.)

## 4. Test coverage I added on top

| Layer | What it pins |
|---|---|
| `tests/test_guardrails/test_engine.py` (+10 tests) | contained span (the core regression), longest-rule-name on contained, partial overlap, disjoint, adjacent, three-way transitive `A∩B∩C`, pattern + keyword on the same span, **two `_STARTER_CONFIG`-grounded production-rule tests**, audit-under-merge invariant |
| `tests/test_guardrails/test_llm_integration.py` (+5 tests) | one per call site of `engine.apply` — that the **merged** redacted text actually reaches downstream APIs: `Anthropic.messages.create` body, `Anthropic` `system` kwarg, `OpenAI.chat.completions.create` body, the LangGraph chat node `_apply_guardrails_to_messages` (also asserts the source list is not mutated — defensive copy invariant), and an end-to-end regression guard with the shipped-config patterns. `BedrockLLMClient` shares Anthropic's redaction code path so it's covered transitively. |

**Live verification just now:** 91/91 tests in `tests/test_guardrails/` pass on this branch. CI on the PR is green: `test`, `typecheck`, `quality`, `CodeQL`, `Analyze (python)`. Approved by @yashksaini-coder on 2026-04-25 with an independent algorithm walkthrough.

## 5. Why this is worth merging

The class of bug is sensitive-data disclosure to an LLM provider (Anthropic / OpenAI / Bedrock). Anywhere a Tracer user enables guardrails over their default `_STARTER_CONFIG`, two rules covering the same secret will leak the prefix/suffix bookends into every outgoing prompt, system message, and chat node. This PR makes the redaction step actually idempotent over overlapping rule sets.

Happy to step through any specific scenario on a call or in a screenshare if any part of the merge algorithm itself is the unclear bit — that's the part most worth eyeballing carefully.

app/guardrails/engine.py
```diff
@@ -27,6 +28,22 @@ class ScanMatch:
...
+    overlapping ``ScanMatch`` ranges have been collapsed.
+
+    ``rule_name`` is the *representative* rule for the span — the contributing
+    match with the largest original ``width`` wins, which preserves the
+    "longest-keyword-wins" behaviour for same-start overlaps while extending
+    it to contained and partially-overlapping spans.
+    """
+
+    start: int
+    end: int
```
muddlebee, line 50: The `width` field tracks the *original single-match* width, not the current merged span's width (`end - start`). In a chained-overlap scenario (A merges B, then C overlaps the grown span), the rule-name winner is whichever *individual source match* had the largest original width — not whichever rule covered the widest post-merge range. This is intentional and documented in the class docstring, but the field name `width` alone is ambiguous. Renaming to `source_width` (or adding a short note in the docstring: "original match width before any merging") would prevent the next reader from assuming it equals `end - start`.

tests/test_guardrails/test_llm_integration.py
```diff
@@ -266,3 +266,224 @@ def test_noop_when_no_rules(
...
+        client.invoke("Debug dump: api_key=AKIAIOSFODNN7EXAMPLE from config.yml")
+
+        content = captured["messages"][0]["content"]
+        # Full span merged — neither the label nor the key value leaks.
+        assert "api_key=" not in content
+        assert "AKIA" not in content
+        assert "IOSFODNN7EXAMPLE" not in content
+        # Representative rule is the wider one (generic_api_token).
+        assert "[REDACTED:generic_api_token]" in content
+        assert "[REDACTED:aws_access_key]" not in content
```
muddlebee, line 402: The `_FakeMessages` and `_FakeClient` inner classes are duplicated verbatim in all five tests in this class. If the fake's interface ever changes (e.g., to return a streaming response type) all five copies need updating in sync — a common source of drift. A shared `@pytest.fixture` returning a `(client, captured)` tuple would eliminate the duplication and keep this test class under ~150 lines.

tests/test_guardrails/test_llm_integration.py
```diff
@@ -266,3 +266,224 @@ def test_noop_when_no_rules(
...
+        # dict.  Emulate the `invoke(prompt_or_messages)` contract here by
+        # passing the raw messages list and setting system through a
+        # small prompt object.
+        client.invoke([{"role": "system", "content": system}, *messages])
+
+        assert captured.get("system") is not None
+        assert "api_key=" not in captured["system"]
+        assert "AKIA" not in captured["system"]
+        assert "[REDACTED:generic_api_token]" in captured["system"]
+
```
muddlebee, line 436: The comment in this test admits the system-prompt injection path is awkward (passing a `{"role": "system", ...}` entry mixed into the messages list). The assertion `assert captured.get("system") is not None` passes even if `LLMClient` embeds the system role inside `captured["messages"]` rather than hoisting it to a separate `system` kwarg — the assertion would silently pass with `None` returned by `.get()` if the key is absent. Consider `assert captured["system"] != ""` (explicit KeyError on absence) or — better — test the system-prompt path through a purpose-built helper or by inspecting `captured["messages"]` when the Anthropic client doesn't separate system from user messages.

muddlebee (commented):
Pls take a look.

mayankbharati-ops:
@muddlebee — pushed `5f1b3b6` addressing all three review nits. Detail per comment:

## 1. `app/guardrails/engine.py:42` — `width` field name ambiguity

Renamed `_MergedSpan.width` → `source_width` and expanded both the NamedTuple docstring and the `_redact` algorithm comment to spell out the distinction explicitly. The docstring now reads:

> `source_width` is the width of the single `ScanMatch` that contributed the current `rule_name` — i.e. that match's `end - start` *before* any merging happened. It is **not** the merged span's width (`end - start` of the `_MergedSpan`); in a chained-overlap scenario where A merges B and then C overlaps the grown span, the rule-name winner is whichever individual source match was the widest, not whichever rule covered the widest post-merge range.

The inline comment in `_redact` also flags this on the line where `source_width` is computed, so anyone editing the loop body sees the distinction in context. Good catch — `width` alone was genuinely ambiguous.

## 2. `tests/test_guardrails/test_llm_integration.py:340` — `_FakeMessages` / `_FakeClient` duplication

Right — the four ad-hoc copies were a drift hazard. Extracted into two module-level pytest fixtures:

```python
@pytest.fixture
def anthropic_capture(monkeypatch) -> tuple[Any, dict[str, Any]]:
    """Patch LLMClient (Anthropic surface) to capture messages.create kwargs."""
    ...
    return client, captured

@pytest.fixture
def openai_capture(monkeypatch) -> tuple[Any, dict[str, Any]]:
    """Patch OpenAILLMClient to capture chat.completions.create kwargs."""
    ...
    return client, captured
```

Each test in `TestOverlappingRedactionReachesDownstream` is now ~5–10 lines of actual test logic instead of 25+ of mostly-identical fake-class scaffolding. The two response-shape literals (`_anthropic_fake_response` / `_openai_fake_response`) are also factored out, so a future SDK change touches one location instead of four.

## 3. `tests/test_guardrails/test_llm_integration.py:382` — system-prompt assertion

Tightened to:

```python
# Subscript (not ``.get``) so a missing ``system`` kwarg surfaces a
# KeyError immediately — the whole point of the test is that the
# client hoisted the system entry out of the messages list.
system_kwarg = captured["system"]
assert system_kwarg, "system kwarg must be non-empty after redaction"
assert "api_key=" not in system_kwarg
assert "[REDACTED:generic_api_token]" in system_kwarg
# System must be a separate kwarg, not smuggled into ``messages``.
for msg in captured.get("messages", []):
    assert msg.get("role") != "system", (
        f"system role leaked into messages list: {msg!r}"
    )
```

KeyError on a missing `system` kwarg is now louder than the old `.get(...) is not None` (which produced an unhelpful AssertionError if the kwarg was absent), and the new structural assertion catches the failure mode you flagged — Anthropic client failing to hoist `role: system` out of the messages list. The test now actually exercises *both* "the system kwarg exists and was redacted" *and* "the system role was not smuggled into messages."

## Verification

- `pytest tests/test_guardrails/`: **91 / 91 pass**
- Full unit suite: **3159 pass / 2 skipped / 1 xfailed**
- `ruff check` and `ruff format --check` clean on both files
- `mypy app/guardrails/` clean (the 5 errors mypy reports are pre-existing missing-stub warnings in `app/integrations/` — unrelated)
- CI on `5f1b3b6`: **all 5 active checks SUCCESS** (`quality`, `Analyze (python)`, `typecheck`, `test`, `CodeQL`)

Ready for another look.

muddlebee (approved):

muddlebee:
Looks good from me. I will wait for additional reviews..

muddlebee:
@VaibhavUpreti

Issue #650, opened 2026-04-18
Experienced OSS Contributors wanted: remove LangGraph


davincios:
## Experienced OSS Contributors Wanted
We are looking for experienced open source contributors to help remove LangGraph from the agent harness and replace it with a simpler, more capable orchestration model.

## Summary
Most modern agentic harnesses are built around a simple `while true` loop and a rich deterministic harness, rather than routing model outputs through an explicit state machine.

## Why
LangGraph pushes the system toward explicit graph/state-machine orchestration, which is the opposite of the direction we want.

What we want instead is:
- give models more room to reason and choose the next step
- keep the harness deterministic and responsible for execution, safety, and tooling
- simplify the core orchestration loop so it is easy to understand and evolve

Claude Code is a useful reference point here: it gives the model maximum decision latitude inside a strong harness, and concentrates engineering effort on that harness rather than on graph wiring.

## Proposed Direction
Replace the current LangGraph-style orchestration with a simpler control loop:
- a core `while true` agent loop
- deterministic tool execution and guardrails in the harness
- minimal orchestration logic outside the model

## Goal
Free up the model to do more reasoning, while the harness does everything else.

davincios:
cc @VaibhavUpreti

muddlebee:
Hey @davincios

I would like to tackle this 👀

davincios:
> Hey @davincios
> 
> I would like to tackle this 👀 

Okay appreciate it 🙏, all yours. 

Please inform me of a short plan and timeline.

muddlebee:
I spent time researching how modern agentic harness are built in Claude Code, OpenAI Agents SDK, and Gemini CLI. 

[Full research + migration plan ](https://gist.github.com/muddlebee/26eb08277a8db98ed69955950def5754) 
(used [DeepWiki](https://deepwiki.com/) to dive into some of the open-source implementations.)

If this direction looks good, I can start and have a working version ready by coming Sunday.

@davincios @VaibhavUpreti

muddlebee:
@Devesh36 @yashksaini-coder @rrajan94 would love your inputs/suggestions on the research + plan above :) 
this will be a big complex change, I haven't figured out all the output combinations, edge cases, appreciate any pointers or help

davincios:
Hi @hamzzaaamalik, this issue has been stale for a while. Could you help decide what should happen here?

[...]

PR #781, opened 2026-04-23
feat: add CLI framework base and codex CLI integration 

### LINKED ISSUE #642: [FEATURE] :Support CLI backends (Codex, Claude Code, Gemini, Cursor)
### Problem statement

Support multiple coding-agent CLIs as optional LLM backends for `opensre investigate`,  by shelling out to each vendor’s documented non-interactive / headless entrypoint, nstead of only HTTP API keys.

**Motivation**
- Lower onboarding friction: Many people already use CLI login (Codex, Claude Code, Gemini CLI, etc.). Reusing that session avoids a separate “create API key, paste into .env” step for local runs.
- Familiar integration pattern: Other agent stacks expose capabilities through CLI + bridge (OpenClaw via MCP) 

### Proposed solution

**Proposal**
- Define a single internal contract (e.g. “messages in → text out”, optional structured JSON) shared by all backends.
- Add LLM_PROVIDER values such as codex, and later claude_code / gemini_cli only where the CLI exposes a stable, scriptable mode (no scraping interactive TUIs).
- Each adapter: build argv + stdin, honor timeouts, map failures to clear errors; inherit environment so existing CLI auth works.

**Risks**
- M×N maintenance (flags and behavior drift per CLI version).
- Agent-in-agent semantics: CLIs may run their own tool loop; document that this path is convenience-first, not identical to API tool-calling.

muddlebee:
Fixes #642 

 

Code specs/architecture as per https://github.com/Tracer-Cloud/opensre/issues/642#issuecomment-4275270815

**Success scenario - when codex binary present and logged in**

[Screencast from 23-04-26 02:03:43 PM IST.webm](https://github.com/user-attachments/assets/ab6cce4f-eeaf-48bd-a2f0-2418cf894515)

Final Output :point_down: 
image

**Failure scenario - where codex binary path is not detected in default paths**

[Screencast from 23-04-26 02:42:59 PM IST.webm](https://github.com/user-attachments/assets/3076f01a-c0b1-4644-85bc-dc458de397fa)

---

---

---


app/services/llm_client.py
```diff
@@ -436,7 +436,8 @@ def _extract_json_payload(text: str) -> Any:
 # LLM Client
 # ─────────────────────────────────────────────────────────────────────────────
 
-_LLMClientType = LLMClient | OpenAILLMClient | BedrockLLMClient
+# CLIBackedLLMClient is included but omitted from the union to avoid import cycles.
+_LLMClientType = LLMClient | OpenAILLMClient | BedrockLLMClient | Any
```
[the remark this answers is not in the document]
muddlebee: I am sticking to "Any" type as add proper types like  "CLIBackedLLMClient" is giving "Module-level cyclic import" lint errors as tried in this [commit ](https://github.com/Tracer-Cloud/opensre/pull/781/changes/744b2b7d2d2cd5f63fb8c8fd75d6030616848379)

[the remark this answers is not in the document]
muddlebee: If any suggestions, feel free to push commits into this branch :) glad for any help or suggestions

app/integrations/llm_cli/runner.py
```diff
@@ -0,0 +1,131 @@
...
+            )
+        if probe.logged_in is False:
+            raise RuntimeError(
+                f"{self._adapter.name} is not authenticated. {self._adapter.auth_hint} "
+                f"({probe.detail})"
+            )
+
+        invocation = self._adapter.build(prompt=flat, model=self._model, workspace="")
+        merged_env = os.environ.copy()
+        if invocation.env:
```
Devesh36: Every env var from the parent process — ANTHROPIC_API_KEY, OPENAI_API_KEY, database URLs, etc. — is handed to the Codex CLI subprocess. If the CLI logs env vars on error or spawns its own children, credentials leak.

muddlebee:
**Steps to test locally**


Negative flow 

- Try `opensre onboarding` without installing codex cli and without codex login
- Try `opensre onboarding` with installing codex cli and without codex login 

Positive flow 

- Install Codex CLI
- Codex login

app/integrations/llm_cli/runner.py
```diff
@@ -0,0 +1,131 @@
...
+        self._adapter = adapter
+        self._model = model
+        self._max_tokens = max_tokens
+        self._model_type = model_type
+        self._cached_probe: CLIProbe | None = None
+        self._probe_cached_at: float = 0.0
+
+    def _probe(self) -> CLIProbe:
+        now = time.monotonic()
+        if self._cached_probe is not None and (now - self._probe_cached_at) < _PROBE_CACHE_TTL_SEC:
```
Devesh36, line 99: _cached_probe and _probe_cached_at are written without a lock. In a multi-threaded LangGraph pipeline where the same CLIBackedLLMClient instance is shared, two threads could simultaneously find the cache stale and both call detect(), with one overwriting the other mid-write. Add a threading.Lock acquired around the stale-check-and-set block.

app/services/llm_client.py
```diff
@@ -326,9 +326,9 @@ def invoke(self, prompt_or_messages: Any) -> LLMResponse:
 
 
 class StructuredOutputClient:
-    def __init__(
-        self, base: LLMClient | OpenAILLMClient | BedrockLLMClient, model: type[BaseModel]
-    ) -> None:
+    """Wraps any LLM client with `.invoke` (API or CLI subprocess) for Pydantic JSON parsing."""
+
+    def __init__(self, base: Any, model: type[BaseModel]) -> None:
```
Devesh36, line 331: Replacing the union with Any suppresses mypy errors on all callers, not just the new CLI-backed one. A Protocol with a single .invoke(...) method is the correct fix and would have caught this at the type level without widening.

app/nodes/chat.py
```diff
@@ -74,6 +74,11 @@ def _normalize_messages(msgs: list[Any]) -> list[ChatMessage]:
 
 def _resolve_models(provider: str) -> tuple[str, str]:
     """Resolve tool and reasoning model names for the active provider."""
+    if provider == "codex":
+        raise ValueError(
```
Devesh36: ValueError raised inside a LangGraph node will propagate as an unhandled exception unless the graph has a catch-all error handler. A user-facing message or a typed UnsupportedProviderError that the graph runner knows to catch and format would be safer.

app/integrations/llm_cli/text.py
```diff
@@ -0,0 +1,36 @@
...
+    """Turn LangChain-style messages or a plain string into one text block."""
+    if isinstance(prompt_or_messages, str):
+        return prompt_or_messages
+
+    if not isinstance(prompt_or_messages, list):
+        return str(prompt_or_messages)
+
+    parts: list[str] = []
+    for msg in prompt_or_messages:
+        if not isinstance(msg, dict):
```
yashksaini-coder, line 18: The pipeline sends LangChain message objects (HumanMessage, SystemMessage, etc.), not plain dicts. When a non-dict message comes in here, it falls through to str(msg) on line 19 which gives you something like content='actual prompt' additional_kwargs={} response_metadata={} — garbage sent to Codex CLI. The tests only use dict format so they all pass, but real investigate runs would silently send broken prompts every time.

You need to handle LangChain objects explicitly. Something like:

if hasattr(msg, 'content'):
    role = getattr(msg, 'type', 'user')
    parts.append(f'=== {role.upper()} ===\n{msg.content}')
    continue

app/integrations/llm_cli/codex.py
```diff
@@ -0,0 +1,327 @@
...
+
+class CodexAdapter:
+    """Non-interactive Codex CLI (`codex exec` with read-only sandbox)."""
+
+    name = "codex"
+    binary_env_key = "CODEX_BIN"
+    install_hint = "npm i -g @openai/codex"
+    auth_hint = "Run: codex login"
+    min_version: str | None = None
+    default_exec_timeout_sec = 900.0
```
yashksaini-coder: 900 seconds (15 minutes) is a really long time to hang silently if Codex gets stuck — network issue, auth expiry, unexpectedly large prompt, whatever. The user just sees a frozen terminal for 15 minutes with no idea what's happening.

I'd suggest dropping this to 120s or making it configurable via an env var like CODEX_TIMEOUT. That way if someone really needs longer they can opt in, but the default is a sane one.

app/integrations/llm_cli/codex.py
```diff
@@ -0,0 +1,327 @@
...
+        upgrade_note = ""
+        if self.min_version and version and _ver_tuple(version) < _ver_tuple(self.min_version):
+            upgrade_note = (
+                f" Codex {version} is below tested minimum {self.min_version}; "
+                f"upgrade: {self.install_hint}@latest"
+            )
+
+        try:
+            auth_proc = subprocess.run(
+                [binary_path, "login", "status"],
```
yashksaini-coder, line 240: Just want to double-check — is codex login status actually a documented subcommand? I couldn't find it in the public Codex CLI docs. If this command returns non-zero or unexpected output, logged_in will always be None and the onboarding loop will keep spinning in the 'unclear auth' branch without telling the user clearly what's wrong.

Worth verifying this against the actual CLI and adding a note about which codex version this was tested with.

app/cli/wizard/config.py
```diff
@@ -124,6 +132,31 @@ class ProviderOption:
...
 )
 
+# Empty value means "no -m" so the Codex CLI uses its configured default/current model.
+CODEX_MODELS = (
+    ModelOption(
+        value="",
+        label="CLI default (no -m; use Codex configured model)",
+    ),
+    ModelOption(value="gpt-5.4", label="gpt-5.4 — strong default for everyday coding"),
+    ModelOption(value="gpt-5.2-codex", label="gpt-5.2-codex — frontier agentic coding"),
```
yashksaini-coder, line 142: These model IDs (gpt-5.2-codex, gpt-5.1-codex-max, gpt-5.1-codex-mini) — are these real model IDs that Codex CLI accepts via -m? I'm not finding them in OpenAI's public model list. If someone picks one of these in the onboarding menu, every codex exec -m gpt-5.2-codex call will likely fail with an unknown model error.

The screencast shows the happy path working but doesn't test model selection. I'd suggest keeping only the 'CLI default (no -m)' option and any IDs you've personally verified work, and removing the ones that are uncertain.

yashksaini-coder (changes requested):
Hey @muddlebee, really nice work on the architecture here — the Protocol-based adapter design, probe caching, and onboarding recovery loop are all solid. A few issues I caught that need fixing before this lands:

app/cli/wizard/config.py
muddlebee: yes in codex cli model list, you will find if codex locally installed

app/integrations/llm_cli/text.py
muddlebee: not needed, already handled for 'content' and 'user' types check the entire function

app/integrations/llm_cli/codex.py
muddlebee: will do

app/integrations/llm_cli/codex.py
muddlebee: you will find if codex locally installed, codex command CLI

muddlebee:
@Devesh36 addressed your review in latest [commit](https://github.com/Tracer-Cloud/opensre/pull/781/commits/3b69e630868bc9f93986f7cb6d2f9f6d6ee75e58)

app/integrations/llm_cli/codex.py
muddlebee: reduced to 120s, in latest [commit](https://github.com/Tracer-Cloud/opensre/pull/781/changes/3b69e630868bc9f93986f7cb6d2f9f6d6ee75e58)

yashksaini-coder: great

yashksaini-coder (approved):
Good progress on addressing the review feedback — env var allowlist, thread-safe probe cache, Protocol-based typing, and the 900s → 120s timeout fix all look solid. Approving. The one thing still worth a follow-up PR is the LangChain message handling in text.py (non-dict messages stringify as garbage), but that can be a quick patch after merge. Nice work overall @muddlebee 🚀

muddlebee:
@yashksaini-coder thank your for your review. anyway we are planning migration away from LangGraph, check https://github.com/Tracer-Cloud/opensre/issues/650

So adding the reference to this PR, to cater such cases. noted!

Devesh36 (approved):

PR #806, opened 2026-04-24
fix: add Windows fallback when .venv not yet created (closes #803)

### LINKED ISSUE #803: Makefile fails on Windows due to hardcoded Unix venv path
## Summary

Makefile assumes Unix venv path (.venv/bin/python) or python3, causing make lint / make typecheck / make test-cov to fail on Windows even when a local .venv Python exists.

**Expected:** On Windows, the Makefile should check .venv/Scripts/python.exe (Windows venv layout) before falling back to python3 or python.

**Actual:** The venv path is hardcoded to .venv/bin/python which does not exist on Windows, and the install target uses python3 directly rather than $PYTHON.

## Steps to reproduce

1. Clone repo on Windows
2. Create .venv using python -m venv .venv
3. Run make lint — fails because .venv/bin/python does not exist

## Additional context

The install target also directly calls python3 rather than $PYTHON, so even if the PYTHON variable pointed to the Windows venv Python, install would still invoke python3.

## Proposed fix

- Introduce a platform-detection variable (e.g. OS := $(shell uname -s) or check for existence of .venv/bin/python vs .venv/Scripts/python.exe)
- Point PYTHON to the correct venv binary for the platform
- Update the install target to use $PYTHON instead of hardcoding python3

Ade20boss:
Fixes #803

Two changes were made to the Python detection block in the `Makefile`:

**1. Added a Windows fallback**

The original Windows block only handled the case where `.venv/Scripts/python.exe` already exists, but had no `else` — so if you ran `make install` on a fresh Windows clone (before the venv is created), `PYTHON` and `PIP` would be undefined and everything would fail silently.

The fix adds an `else` clause inside the Windows block so it falls back to the system `python` when the venv doesn't exist yet.

**2. Simplified the stderr redirect on the python3 check**

Replaced `2>$(if $(filter Windows_NT,$(OS)),NUL,/dev/null)` with `2>/dev/null`. That line is inside an `else` branch that only runs when `OS` is **not** `Windows_NT`, so the Windows-specific `NUL` redirect was unnecessary.

N/A — Makefile change only.

---

- [] Yes, I used AI assistance (continue below)

- [] I have reviewed every single line of the AI-generated code
- [] I can explain the purpose and logic of each function/component I added
- [] I have tested edge cases and understand how the code handles them
- [] I have modified the AI output to follow this project's coding standards and conventions

The issue was that the Windows branch of the `PYTHON` detection block had no fallback for when `.venv` doesn't exist yet. On a fresh clone, running any `make` target on Windows would leave `PYTHON` undefined and fail silently. The fix adds a simple `else` clause pointing to the system `python`, matching the pattern already used in the non-Windows branch. The stderr redirect simplification is a minor cleanup that removes dead code.

---

---


muddlebee:
@Ade20boss issue has already been fixed, thank you!

PR #809, opened 2026-04-24
fix: Makefile Windows venv path compatibility (#803)

Jah-yee:
Good day

## Problem

The Makefile hardcodes Unix-style venv paths (.venv/bin/python), which causes `make lint`, `make test-cov`, and other targets to fail on Windows where the venv uses .venv/Scripts/python.exe instead.

## Changes

1. Introduced platform detection via `wildcard .venv/Scripts/python.exe` (Windows) vs `wildcard .venv/bin/python3` (Unix)
2. Falls back to system `python3` when no local venv exists
3. Updated the `install` target to use $(PYTHON) instead of hardcoded `python3`
4. Kept the diff small and targeted — only the necessary lines changed

## Testing

The change is backwards-compatible and doesn't affect Unix systems (existing `PYTHON = .venv/bin/python3` branch still used on non-Windows). On Windows, the new `.venv/Scripts/python.exe` branch will be selected.

---

Thank you for your work on this project. I hope this small fix is helpful. Please let me know if there's anything to adjust.

Warmly,
RoomWithRoof


Jah-yee:
Thanks for the review @greptile-apps[bot]!

## Thinking path

For **P1 (PIP python vs python3)**: The suggestion is correct — on Unix the venv uses  not . The PR original fix targeted  which is a symlink on many systems but not guaranteed. Using  is the safest approach for Unix venvs.

For **P2 (Windows fallback)**: The suggestion is also correct. On a fresh Windows machine with no venv and no  in PATH, the fallback would echo literal  which doesn't work. Adding  as the first fallback handles Windows properly.

## Changes

- **P1 fixed**: Unix PIP now uses `python3 -m pip` instead of `python -m pip`
- **P2 fixed**: Fallback PYTHON now tries `python` first before `python3`



Please let me know if there are any other issues!

Jah-yee (commented):
Hi @greptile-apps[bot], thanks for the review!

Regarding your P1 comment: the current PR already uses `python3 -m pip` consistently for the Unix branch (see commit e5abf5c). The PIP variable is set to `.venv/bin/python3 -m pip` when the Unix venv is detected.

For the Windows branch, `.venv/Scripts/python -m pip` is used, which aligns with how Windows venvs work (the Python launcher resolves `python` to `python.exe` automatically in most cases).

Regarding your P2 suggestion about the fallback: the fallback already tries `python` first before `python3`, which should cover most Windows Python installations. However, since the PR's primary focus is on venv path compatibility (not fallback edge cases), I'll leave the fallback as-is to keep the diff minimal.

Let me know if you'd like me to address anything specific!

Jah-yee:
Superseded by PR #818 - cleaner approach using Make wildcard detection instead of shell-based OS detection.

The new PR uses:
- `PYTHON_WIN := $(wildcard .venv/Scripts/python.exe)` for Windows
- `PYTHON_UNIX := $(wildcard .venv/bin/python3)` for Unix
- Simpler fallback: tries `python` first (Windows), then `python3`

Jah-yee:
Superseded by PR #818 with a cleaner approach using Make wildcard detection

PR #818, opened 2026-04-24
fix Makefile: use wildcard detection for cross-platform venv paths

Jah-yee:
## Summary
Fix cross-platform venv path detection in Makefile for Windows/Unix compatibility.

## Problem
The current Makefile has brittle platform detection using `$(OS),Windows_NT` and shell commands. Windows users with `.venv/Scripts/python.exe` are not properly detected.

## Solution
Use Make wildcard to detect platform and venv structure:

- **Windows**: Check for `.venv/Scripts/python.exe` → use `.venv/Scripts/python`
- **Unix**: Check for `.venv/bin/python3` → use `.venv/bin/python3`  
- **Fallback**: Try `python` first (Windows), then `python3`

## Changes
- Added `PYTHON_WIN := $(wildcard .venv/Scripts/python.exe)`
- Added `PYTHON_UNIX := $(wildcard .venv/bin/python3)`
- Replaced complex $(OS) checks with simple wildcard detection
- `install:` target already uses $(PYTHON) correctly

## Related
Original PR: #809 (superseded)


muddlebee:
> The current Makefile has brittle platform detection using $(OS),Windows_NT and shell commands. Windows users with .venv/Scripts/python.exe are not properly detected.


can you give more context and scenario? also advisable to open an issue first before directly sending PR :)

muddlebee:
@Jah-yee thank you for the PR. pls open a feature request with more context and example :) thank you, closing this.

PR #843, opened 2026-04-24
synthetic: tighten validation and fix RCA reasoning for replication lag red-herring scenario (006)

### LINKED ISSUE #602: [synthetic-qa] 006-replication-lag-cpu-redherring: Validate agent ignores CPU red herring
## Scenario

**Run:** `opensre tests synthetic --scenario 006-replication-lag-cpu-redherring`

Replication lag caused by a write-heavy batch UPDATE. CPUUtilization at 70-85% is a **red herring** from an unrelated analytics SELECT query running at the same time. The two workloads are causally independent.

## Correct Answer

- **Root Cause:** Replication lag from WAL generation (batch UPDATE). CPU elevation is an unrelated analytics query — NOT the root cause.

## Human Review Checklist

- [ ] Does the agent identify **replication lag** as the root cause (not CPU)?
- [ ] Does the agent explain the **WAL mechanism** (write-heavy UPDATE generating WAL faster than replica replay)?
- [ ] Does the agent explicitly call out CPU elevation as a **red herring / unrelated confounder**?
- [ ] Does the agent distinguish the **two independent workloads** (bulk UPDATE vs analytics SELECT)?
- [ ] Does the agent avoid diagnosing `cpu_saturation`?

Agent separates the two concurrent workloads: the batch UPDATE is generating WAL and causing replication lag, while the analytics SELECT is consuming CPU independently. It explicitly states the CPU is not the root cause.

## Common Failure Modes

- **Most likely failure:** Agent sees 75-85% CPU and blames CPU saturation (falls for the red herring)
- Agent mentions replication lag but also blames CPU as a contributing cause (wrong — they're independent)
- Agent doesn't distinguish the two workloads — treats it as one blended problem

cerencamkiran:
Fixes #602

## Summary

Improves both evaluation quality and RCA reasoning for scenario 006 (replication lag with CPU red herring).

## Motivation

Previously, the agent could pass this scenario while incorrectly describing CPU as a second independent or contributing root cause. This violated the scenario’s intended behavior, where CPU is explicitly a red herring and causally unrelated to replication lag.

This PR ensures that:
- incorrect multi-root-cause reasoning fails validation
- correct causal reasoning (single root cause + red herring) consistently passes

### 1. Validation tightening
- Refined `required_keywords` to focus on core reasoning signals instead of rigid phrasing
- Removed overly strict keywords that caused false negatives
- Added `QA_VALIDATION.md` to clearly define:
  - expected reasoning
  - strict failure modes
  - human review expectations

### 2. Prompt improvement
Updated the replication lag directive in `prompt_builder.py` to:
- enforce single root cause identification
- explicitly classify unrelated CPU signals as **red herrings**
- prevent treating independent signals as:
  - second root causes
  - contributing causes

## Result

- The agent now:
  - correctly identifies WAL-driven replication lag as the only root cause
  - explicitly labels CPU as a **red herring**
  - separates independent workloads (UPDATE vs SELECT)

- Validation now:
  - fails incorrect causal reasoning
  - passes only when reasoning matches scenario intent

## Repro

```bash
python -m tests.synthetic.rds_postgres.run_suite --scenario 006-replication-lag-cpu-redherring --mock-grafana


cerencamkiran:
BEFORE 
Ekran görüntüsü 2026-04-24 171524
Ekran görüntüsü 2026-04-24 171543

cerencamkiran:
AFTER
Ekran görüntüsü 2026-04-24 190433
Ekran görüntüsü 2026-04-24 190455

muddlebee:
@cerencamkiran pls fix the conflicts!

cerencamkiran:
Conflicts are fixed now @muddlebee👍

muddlebee:
> contributor hall of fame.

congrats :laughing:

PR #915, opened 2026-04-25
refactor(tools): resolve action labels from registry metadata

### LINKED ISSUE #870: Replace hard-coded `_ACTION_DISPLAY` labels with tool metadata
## Why This Matters
progress output should update automatically when new tools are added instead of requiring a second manual mapping in `app/output.py`.

## Read First
- `app/output.py`
- `app/remote/reasoning.py`
- `app/tools/registered_tool.py`
- `app/tools/registry.py`

## Scope
1. Add one registry-backed way to resolve a human-friendly tool label.
2. Update `_humanise_message()` and `app/remote/reasoning.py` to use it.
3. Keep the existing human-friendly labels for current tools.

## Tests / Coverage
- extend `tests/tools/test_registry.py`
- add output tests that assert planned-action strings are humanized from real tool metadata
- add one fallback test for unknown tool names

## Acceptance Criteria
- `app/output.py` no longer owns the only source of tool display labels
- current progress output remains readable
- unknown tool names still degrade gracefully

imjohnzakkam:
Fixes #870

Moved human-friendly tool labels out of `app/output.py` and into the tool registry flow by introducing an optional `display_name` metadata field on registered tools.

`app.output._humanise_message()` and `app.remote.reasoning.tool_display_name()` now resolve labels through the registry, and the existing user-facing labels are preserved by annotating the currently humanized tools with explicit `display_name` values.

- None

---

I added `display_name` to the shared tool metadata contract (`ToolMetadata`, `RegisteredTool`, and the `@tool(...)` decorator) and exposed one registry-backed resolver for human-friendly tool labels. This removes `app/output.py` as the only source of truth for action labels and lets new tools participate by declaring their own display names in metadata.

For edge cases, I preserved the existing labels for the currently humanized tools by setting explicit `display_name` values on those tool definitions, and I kept a graceful fallback for unknown tool names by de-snaking them when the registry has no matching tool. I also added direct tests for registry-backed label resolution, output humanization from real registered tool metadata, and unknown-tool fallback behavior.

---

## Verification
- `python -m pytest tests/tools/test_registry.py tests/test_output.py tests/app/remote/test_reasoning.py`
- `ruff check app/tools/base.py app/tools/registered_tool.py app/tools/tool_decorator.py app/tools/registry.py app/output.py app/remote/reasoning.py tests/tools/test_registry.py tests/test_output.py tests/app/remote/test_reasoning.py`
- `ruff format --check app/tools/base.py app/tools/registered_tool.py app/tools/tool_decorator.py app/tools/registry.py app/output.py app/remote/reasoning.py tests/tools/test_registry.py tests/test_output.py tests/app/remote/test_reasoning.py`

Output summary:
- `32 passed`
- `All checks passed!`
- `9 files already formatted`


imjohnzakkam:
Updated catch by greptile. I added a cached registry tool-name map and switched `resolve_tool_display_name()` to use that directly, so we avoid rebuilding the dict on every label lookup while keeping the same fallback behaviour.

muddlebee:
@imjohnzakkam pls fix conflicts and review comments

imjohnzakkam:
@muddlebee done, pls check and merge 🚀

app/remote/reasoning.py
```diff
@@ -32,7 +25,7 @@
```
muddlebee, line 25: Three tools that only lived in the old TOOL_DISPLAY dict (query_kubernetes_logs → "Kubernetes logs", query_elasticsearch → "Elasticsearch", get_deployment_status → "deployment status") were never annotated with display_name. They now fall back to de-snaked lowercase strings like "query kubernetes logs" instead of "Kubernetes logs". The PR description says existing labels are preserved, but these three regress silently.

app/tools/tool_decorator.py
```diff
@@ -105,6 +109,7 @@ def should_register_function() -> bool:
             [
```
muddlebee, line 109: display_name is None shouldn't be a registration trigger. The any-arg check is meant to distinguish bare @tool from @tool(...). display_name is pure cosmetic metadata — adding it here means @tool(display_name="X") with no name/source would attempt registration and fail validation rather than being treated as a bare decorator. Recommend removing display_name from this check.

app/tools/registry.py
```diff
@@ -161,4 +167,14 @@ def get_registered_tools(surface: ToolSurface | None = None) -> list[RegisteredT
 
 
 def get_registered_tool_map(surface: ToolSurface | None = None) -> dict[str, RegisteredTool]:
+    if surface is None:
+        return dict(_load_registry_tool_map())
     return {tool.name: tool for tool in get_registered_tools(surface)}
+
+
```
muddlebee, line 174: Fallback returns tool_name.replace('_', ' ') without .title(), so unknown tools render as "my custom tool" rather than "My Custom Tool". The test even asserts the lowercase form. Intentional? The old reasoning.py fallback was also lowercase, so this is consistent — just flagging since display_name-annotated tools will be Title Case and fallback tools won't be.

app/tools/base.py
```diff
@@ -27,9 +28,11 @@ class ToolMetadata(StrictConfigModel):
...
     )
 
-    @field_validator("name", "description")
+    @field_validator("name", "description", "display_name")
     @classmethod
-    def _require_non_empty_strings(cls, value: str) -> str:
+    def _require_non_empty_strings(cls, value: str | None) -> str | None:
+        if value is None:
+            return None
         normalized = value.strip()
```
muddlebee, line 36: The _require_non_empty_strings validator now runs on display_name too. That's correct. One edge case: an empty string display_name="" will raise a ValueError at registration time with a message referencing the field name. Consider ensuring the error message is clear enough to debug (Pydantic usually includes the field name, so this is probably fine).

tests/tools/test_registry.py
```diff
@@ -251,6 +253,35 @@ def get_incident_metadata(incident_id: str) -> dict[str, str]:
...
+    )
+    monkeypatch.setattr(registry_module, "_import_tool_module", lambda _name: module)
+
+    assert registry_module.resolve_tool_display_name("get_incident_metadata") == "Incident metadata"
+
+
+def test_resolve_tool_display_name_falls_back_for_unknown_tools() -> None:
+    assert registry_module.resolve_tool_display_name("my_custom_tool") == "my custom tool"
+
+
```
muddlebee, line 287: test_resolve_tool_display_name_falls_back_for_unknown_tools calls resolve_tool_display_name directly without any monkeypatching. This hits the real registry cache (_load_registry_tool_map). If a real tool named my_custom_tool is ever added to the repo, this test will silently start returning its display_name instead of the fallback. Mock the registry or use a name that is guaranteed to not collide (e.g. a UUID-ish sentinel like __nonexistent_tool_xyz__).

app/tools/tool_decorator.py
imjohnzakkam: Fixed in `b88019a` by removing `display_name` from the registration trigger check.

tests/tools/test_registry.py
imjohnzakkam: Fixed in `b88019a`; the fallback test now uses a sentinel name (`nonexistent_tool_xyz_sentinel`) so it should not collide with future real tools.

app/tools/registry.py
imjohnzakkam: Intentional for now. This preserves the previous fallback behavior from `reasoning.py`; registered tools can opt into title-cased labels via `display_name`.

app/tools/base.py
imjohnzakkam: Keeping the shared validator here. Pydantic includes the field name in the validation error, so the failure should still point clearly at `display_name`.

app/remote/reasoning.py
imjohnzakkam: I checked the current registered tool names after the merge and those exact identifiers (`query_kubernetes_logs`, `query_elasticsearch`, `get_deployment_status`) do not appear to be live registered tools on this branch. For example, Elasticsearch is currently registered as `query_elasticsearch_logs`. I left them out to avoid preserving stale aliases, but happy to add a small legacy fallback if we still expect those names to appear in streaming events.

muddlebee: @imjohnzakkam yes a fallback makes sense right?

imjohnzakkam: @muddlebee will do that and update

imjohnzakkam: @muddlebee Added the small legacy fallback in `314535f` for the three stale event names:

- `query_kubernetes_logs` -> `Kubernetes logs`
- `query_elasticsearch` -> `Elasticsearch`
- `get_deployment_status` -> `deployment status`

Also added a regression test to preserve those labels. The follow-up format-only commit is `96f4c12`, and local checks are green:
`ruff check app/ tests/`, `ruff format --check app/ tests/`, and the focused pytest set all pass.

muddlebee:
@imjohnzakkam hey really sorry. 

> I checked the current registered tool names after the merge and those exact identifiers (query_kubernetes_logs, query_elasticsearch, get_deployment_status) do not appear to be live registered tools on this branch.

you were right, this doesn't have registered tools on this branch. a fallback wasn't required. you can revert back to the last CI green entry :) 

sorry again for confusion.

muddlebee:
image

till here its fine, if you can reset till here :) good to go

imjohnzakkam:
@muddlebee thanks reverted the legacy fallback follow-up.

muddlebee:
@imjohnzakkam great work :zap:

PR #952, opened 2026-04-25
refactor(utils): extract shared HTTP-post helper for delivery modules

### LINKED ISSUE #864: Extract one shared HTTP-post helper for Slack, Discord, and Telegram delivery
## Why This Matters
`app/utils/slack_delivery.py`, `app/utils/discord_delivery.py`, and `app/utils/telegram_delivery.py` all reimplement the same "POST JSON + parse response + return `(ok, error, id)` style result" pattern.

## Read First
- `app/utils/slack_delivery.py`
- `app/utils/discord_delivery.py`
- `app/utils/telegram_delivery.py`

## Scope
1. Create one small shared helper module under `app/utils/` for outbound message-delivery HTTP calls.
2. Move only the repeated transport behavior into that helper: HTTP POST, timeout handling, optional JSON parsing, and standardized error extraction.
3. Keep Slack-, Discord-, and Telegram-specific payload building in their current modules.

## Tests / Coverage
- add `tests/utils/test_slack_delivery.py`
- extend `tests/utils/test_discord_delivery.py`
- extend `tests/utils/test_telegram_delivery.py`
- add a focused test module for the new shared helper

## Acceptance Criteria
- Slack, Discord, and Telegram helpers all use the shared transport helper
- provider-specific response shapes stay unchanged
- existing public helper names stay unchanged

mayankbharati-ops:
Closes #864.

## Summary

\`app/utils/slack_delivery.py\`, \`app/utils/discord_delivery.py\`, and \`app/utils/telegram_delivery.py\` each reimplemented the same \"POST JSON + parse response + return \`(ok, error, id)\` style result\" pattern. The transport pieces (HTTP POST, timeout, exception suppression, JSON decode) were identical; only the success criteria, auth scheme, and error-extraction differed.

This PR adds \`app/utils/delivery_transport.py\` and migrates all three modules onto it, per the issue's acceptance criteria.

## Design

\`\`\`python
@dataclass(frozen=True)
class DeliveryResponse:
    ok: bool                 # request did not raise
    status_code: int = 0     # HTTP status (0 if request raised)
    data: dict = {}          # parsed JSON object body, or {} on non-JSON / non-dict / decode error
    text: str = \"\"           # raw response body
    error: str = \"\"          # exception message when ok=False

def post_json(url, payload, *, headers=None, timeout=15.0, follow_redirects=False) -> DeliveryResponse:
    ...
\`\`\`

The helper **deliberately does not decide provider-level success.** \`ok=True\` means the HTTP request completed without raising; callers inspect \`status_code\` and \`data\` per their own provider semantics:

| Provider | Success check |
|---|---|
| Slack ``chat.postMessage`` / reactions | \`data[\"ok\"]\` |
| Slack incoming webhook / NextJS proxy | \`200 ≤ status_code < 300\` |
| Discord channels API | \`status_code in (200, 201)\` |
| Telegram bot API | \`status_code == 200\` |

## Refactored call sites

- **Slack** — all 4 code paths (\`_call_reactions_api\`, \`_post_direct\`, \`_post_via_webapp\`, \`_post_via_incoming_webhook\`) now go through \`post_json\`. \`send_slack_report\` orchestration is unchanged. Bearer-header construction factored into \`_slack_bearer_headers\`.
- **Discord** — \`post_discord_message\` and \`create_discord_thread\` go through \`post_json\`. \`Bot <token>\` header construction factored into \`_discord_auth_headers\`. Error extraction factored into \`_discord_error_from_data\`.
- **Telegram** — \`post_telegram_message\` goes through \`post_json\`. The bot-token redaction in error strings is preserved (the shared transport never sees the bot token in the URL after the request raises, but \`_redact_token\` still runs over \`response.error\` defensively).

Public function signatures unchanged. Provider-specific payload building stays in the calling modules per the issue's scope.

## Test coverage

| File | Tests | Status |
|---|---|---|
| \`tests/utils/test_delivery_transport.py\` | **15 (new)** | happy path, 5 transport-failure exception shapes, JSON decode fallbacks (non-JSON / JSON array / empty body), header/timeout/follow_redirects pass-through, DeliveryResponse frozen-dataclass invariants |
| \`tests/utils/test_slack_delivery.py\` | **24 (new)** | reactions API (success / known-idempotent failure / unexpected error / transport exception / payload assertions), \`_post_direct\` (success, slack_error, exception, payload), incoming webhook (success / 5 status codes / transport exception / blocks+extra merge), \`_post_via_webapp\` (skipped without TRACER_API_URL / success / 5xx), \`send_slack_report\` orchestration (no thread_ts / webhook fallback / direct path / direct→webapp fallback chain) |
| \`tests/utils/test_discord_delivery.py\` | 32 existing + **3 new** | pre-existing tests updated to patch new transport boundary; new \`TestDelegatesToSharedTransport\` class pins the contract: module no longer imports \`httpx\`, both Discord helpers go through \`post_json\` |
| \`tests/utils/test_telegram_delivery.py\` | 26 existing + **3 new** | same delegation pins, plus a regression test that bot-token is redacted in error strings even when the failure originates in the shared transport |

## Verification

- **\`pytest tests/\`** (excluding e2e/deployment/chaos/synthetic): **3191 pass**, 2 skipped, 1 xfailed, 0 failures.
- **\`pytest tests/utils/\`**: 105 pass, 1 skipped.
- **\`ruff check app/ tests/\`** and **\`ruff format --check\`**: clean.
- **\`mypy app/utils/\`**: clean on the new and refactored modules. (The 5 errors mypy reports are pre-existing missing-stub warnings for \`pymysql\`, \`datasets\`, and \`huggingface_hub\` in \`app/integrations/\` — unrelated to this PR.)

## Acceptance criteria audit

- [x] Slack, Discord, and Telegram helpers all use the shared transport helper
- [x] Provider-specific response shapes stay unchanged
- [x] Existing public helper names stay unchanged (\`post_discord_message\`, \`create_discord_thread\`, \`post_telegram_message\`, \`send_slack_report\`, etc.)
- [x] \`tests/utils/test_slack_delivery.py\` added
- [x] \`tests/utils/test_discord_delivery.py\` extended
- [x] \`tests/utils/test_telegram_delivery.py\` extended
- [x] Focused test module for the new shared helper added (\`tests/utils/test_delivery_transport.py\`)

## Scope

| Files | Change |
|---|---|
| \`app/utils/delivery_transport.py\` | new (97 lines) |
| \`app/utils/slack_delivery.py\` | refactored — all 4 \`httpx.post\` call sites replaced |
| \`app/utils/discord_delivery.py\` | refactored — both call sites replaced |
| \`app/utils/telegram_delivery.py\` | refactored — single call site replaced |
| \`tests/utils/test_delivery_transport.py\` | new |
| \`tests/utils/test_slack_delivery.py\` | new |
| \`tests/utils/test_discord_delivery.py\` | 3 new tests + patch-path updates on pre-existing tests |
| \`tests/utils/test_telegram_delivery.py\` | 3 new tests + patch-path updates on pre-existing tests |

### DIFF
--- a/app/utils/delivery_transport.py
+++ b/app/utils/delivery_transport.py
@@ -21,7 +21,9 @@ for Telegram).
 
 from __future__ import annotations
 
+from collections.abc import Mapping
 from dataclasses import dataclass, field
+from types import MappingProxyType
 from typing import Any
 
 import httpx
@@ -38,19 +40,34 @@ class DeliveryResponse:
         status_code: HTTP status code from the response, or ``0`` when the
             request itself raised before a response was received.
         data: Parsed JSON body when the response was a JSON object,
-            otherwise an empty dict. Never ``None``, so callers can chain
-            ``.get(...)`` safely without a None-check.
+            otherwise an empty mapping. Never ``None``, so callers can
+            chain ``.get(...)`` safely without a None-check. The mapping
+            is read-only (``MappingProxyType``) so the frozen dataclass
+            stays fully immutable end-to-end.
         text: Raw response body, useful for fallback error extraction
             when the body is not valid JSON or is empty.
         error: String form of the exception that aborted the request.
             Empty when ``ok`` is True.
+        error_type: Class name of the exception that aborted the request
+            (e.g. ``"TimeoutError"``, ``"ConnectError"``). Empty when
+            ``ok`` is True. Surfaced separately so callers can include
+            the exception shape in triage logs without parsing ``error``.
     """
 
     ok: bool
     status_code: int = 0
-    data: dict[str, Any] = field(default_factory=dict)
+    data: Mapping[str, Any] = field(default_factory=dict)
     text: str = ""
     error: str = ""
+    error_type: str = ""
+
+    def __post_init__(self) -> None:
+        # Wrap ``data`` in a read-only view so callers cannot mutate the
+        # response after the fact and break the frozen-dataclass contract.
+        # ``object.__setattr__`` is required because ``frozen=True`` blocks
+        # normal attribute assignment.
+        if not isinstance(self.data, MappingProxyType):
+            object.__setattr__(self, "data", MappingProxyType(dict(self.data)))
 
 
 def post_json(
@@ -95,7 +112,7 @@ def post_json(
             follow_redirects=follow_redirects,
         )
     except Exception as exc:  # noqa: BLE001 — transport never re-raises
-        return DeliveryResponse(ok=False, error=str(exc))
+        return DeliveryResponse(ok=False, error=str(exc), error_type=type(exc).__name__)
 
     text = response.text
     data: dict[str, Any] = {}
--- a/app/utils/delivery_transport.py
+++ b/app/utils/delivery_transport.py
@@ -1,0 +1,114 @@
+"""Shared HTTP-post transport for outbound message-delivery helpers.
+
+Slack, Discord, and Telegram delivery modules each issue a JSON ``POST``
+to a provider endpoint, parse the response body, and return a
+``(success, error, ...)`` tuple. The transport pieces of that flow —
+making the request, applying a timeout, catching network exceptions, and
+attempting JSON decode — are identical; only the success criteria,
+authentication scheme, and error-message extraction differ per provider.
+
+This module hosts the shared transport so each delivery module can keep
+its provider-specific payload building and result interpretation while
+sharing one well-tested HTTP code path.
+
+The helper deliberately does **not** decide whether the call succeeded
+at the provider level. It returns ``ok=True`` whenever the request
+completed without raising; callers then inspect ``status_code`` and
+``data``/``text`` to apply provider semantics (e.g. ``data["ok"]`` for
+Slack, ``status_code in (200, 201)`` for Discord, ``status_code == 200``
+for Telegram).
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any
+
+import httpx
+
+
+@dataclass(frozen=True)
+class DeliveryResponse:
+    """Normalized result of a delivery POST.
+
+    Attributes:
+        ok: ``True`` iff the request completed without raising. This is a
+            transport-level signal only; provider-level success requires
+            inspecting ``status_code`` / ``data`` per provider semantics.
+        status_code: HTTP status code from the response, or ``0`` when the
+            request itself raised before a response was received.
+        data: Parsed JSON body when the response was a JSON object,
+            otherwise an empty dict. Never ``None``, so callers can chain
+            ``.get(...)`` safely without a None-check.
+        text: Raw response body, useful for fallback error extraction
+            when the body is not valid JSON or is empty.
+        error: String form of the exception that aborted the request.
+            Empty when ``ok`` is True.
+    """
+
+    ok: bool
+    status_code: int = 0
+    data: dict[str, Any] = field(default_factory=dict)
+    text: str = ""
+    error: str = ""
+
+
+def post_json(
+    url: str,
+    payload: dict[str, Any],
+    *,
+    headers: dict[str, str] | None = None,
+    timeout: float = 15.0,
+    follow_redirects: bool = False,
+) -> DeliveryResponse:
+    """POST ``payload`` as JSON to ``url`` and return a normalized result.
+
+    On request exceptions the result carries ``ok=False`` and ``error``
+    set to the exception message — callers are not expected to handle
+    raised errors. The transport never re-raises.
+
+    Args:
+        url: Absolute URL to post to.
+        payload: JSON-serializable dict body.
+        headers: Optional headers (e.g. ``Authorization``). Defaults to
+            an empty dict; httpx will still set ``Content-Type`` and the
+            standard headers it manages.
+        timeout: Request timeout in seconds. Defaults to 15s, matching
+            the pre-existing per-provider timeouts.
+        follow_redirects: Whether to follow 3xx redirects. Disabled by
+            default to match Slack/Discord/Telegram REST APIs (which never
+            redirect on success). Slack incoming webhooks and the NextJS
+            ``/api/slack`` proxy enable it.
+
+    Returns:
+        ``DeliveryResponse`` with ``ok``, ``status_code``, ``data``,
+        ``text``, and ``error`` populated. JSON decode failures are
+        non-fatal: ``data`` falls back to ``{}`` and ``text`` always
+        carries the raw body.
+    """
+    try:
+        response = httpx.post(
+            url,
+            json=payload,
+            headers=headers or {},
+            timeout=timeout,
+            follow_redirects=follow_redirects,
+        )
+    except Exception as exc:  # noqa: BLE001 — transport never re-raises
+        return DeliveryResponse(ok=False, error=str(exc))
+
+    text = response.text
+    data: dict[str, Any] = {}
+    try:
+        parsed = response.json()
+        if isinstance(parsed, dict):
+            data = parsed
+    except Exception:  # noqa: BLE001 — non-JSON body is permitted
+        pass
+
+    return DeliveryResponse(
+        ok=True,
+        status_code=response.status_code,
+        data=data,
+        text=text,
+    )
--- a/app/utils/delivery_transport.py
+++ b/app/utils/delivery_transport.py
@@ -48,10 +48,12 @@ class DeliveryResponse:
             when the body is not valid JSON or is empty.
         error: String form of the exception that aborted the request.
             Empty when ``ok`` is True.
-        error_type: Class name of the exception that aborted the request
+        exc_type: Class name of the exception that aborted the request
             (e.g. ``"TimeoutError"``, ``"ConnectError"``). Empty when
             ``ok`` is True. Surfaced separately so callers can include
             the exception shape in triage logs without parsing ``error``.
+            Named ``exc_type`` rather than ``type`` because ``type`` is
+            a Python builtin.
     """
 
     ok: bool
@@ -59,7 +61,7 @@ class DeliveryResponse:
     data: Mapping[str, Any] = field(default_factory=dict)
     text: str = ""
     error: str = ""
-    error_type: str = ""
+    exc_type: str = ""
 
     def __post_init__(self) -> None:
         # Wrap ``data`` in a read-only view so callers cannot mutate the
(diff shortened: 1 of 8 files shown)


mayankbharati-ops:
Heads up: the `test (ubuntu-latest)` failure on the previous run is a transient `anyio.lowlevel` circular-import flake triggered by `mcp` during `opensre integrations list` on a pytest-xdist worker. Reproduces 0/3 times locally on `origin/main` and on this branch — same Python 3.13 env. PR #780 (which I had open earlier on the same env) also passed CI cleanly. Pushed an empty commit to re-trigger; expect 5/5 green this time.

muddlebee:
@mayankbharati-ops thank you for the PR. could you pls fix all the review points above?

mayankbharati-ops:
yeah sure @muddlebee , I'll update it and let you know

mayankbharati-ops:
@muddlebee — pushed `3f30698` addressing all three Greptile P2 findings:

**1. `DeliveryResponse.data` mutable despite `frozen=True`**
Wrapped `data` in `MappingProxyType` inside `__post_init__`, with the type annotation tightened to `collections.abc.Mapping[str, Any]`. Mutating now raises `TypeError`, and caller-supplied dicts can no longer leak mutations into the response (the proxy wraps a defensive copy). Tightened `_discord_error_from_data` signature to `Mapping[str, Any]` so mypy stays clean against the new read-only type.

**2. Exception type dropped from `_post_direct` log**
Added an `error_type: str` field on `DeliveryResponse`, populated with `type(exc).__name__` on the transport's exception path. `slack_delivery._post_direct` threads it back into the log line as `exc_type=<ClassName>` — distinguishes `TimeoutError` vs `ConnectionError` at a glance without parsing the error string. Public return shape unchanged.

**3. Missing `TestDelegatesToSharedTransport` in `test_slack_delivery.py`**
Added the regression class to mirror Discord/Telegram. Pins that `slack_delivery` does not import `httpx` directly and that all four code paths (`_call_reactions_api`, `_post_direct`, `_post_via_webapp`, `_post_via_incoming_webhook`) route through `delivery_transport.post_json`.

**Test coverage:** +17 new tests covering the read-only-data invariant, caller-dict isolation, default-mapping independence, dict-equality preservation, `error_type` per-exception-class population, the new `exc_type=` log line, and per-code-path delegation. Full suite: **3208 pass / 2 skipped / 1 xfailed** (was 3191). Lint, format-check, and `mypy app/utils/` all clean. CI green.

app/utils/slack_delivery.py
```diff
@@ -209,46 +210,41 @@ def _post_direct(
...
-            "[slack] Reply posted successfully (thread_ts=%s, ts=%s)", thread_ts, data.get("ts")
+    response = post_json(
+        url="https://slack.com/api/chat.postMessage",
+        payload=payload,
+        headers=_slack_bearer_headers(token),
+    )
+    if not response.ok:
+        logger.error(
+            "[slack] Direct post exception channel=%s thread_ts=%s "
+            "exc_type=%s detail=%s (caller may attempt fallback)",
```
muddlebee: Nit: the log key changed from `type=%s` to `exc_type=%s`. If any log parser keys off `type=`, it might miss this. Consider keeping `type=%s` unless you intentionally want to rename it.

app/utils/discord_delivery.py
```diff
@@ -3,13 +3,25 @@
...
-import httpx
+from app.utils.delivery_transport import post_json
 
 logger = logging.getLogger(__name__)
 
 
+def _discord_auth_headers(bot_token: str) -> dict[str, str]:
+    return {
+        "Authorization": f"Bot {bot_token}",
+        "Content-Type": "application/json; charset=utf-8",
```
muddlebee: Nit: `httpx.post(..., json=payload)` already injects `Content-Type: application/json`, so this header is redundant.

app/utils/delivery_transport.py
```diff
@@ -0,0 +1,131 @@
...
+            ``ok`` is True. Surfaced separately so callers can include
+            the exception shape in triage logs without parsing ``error``.
+    """
+
+    ok: bool
+    status_code: int = 0
+    data: Mapping[str, Any] = field(default_factory=dict)
+    text: str = ""
+    error: str = ""
+    error_type: str = ""
```
muddlebee: Nit: consider renaming `error_type` to `exc_type` so it matches the Slack log field name.

muddlebee (approved):
minor nits only..

app/utils/slack_delivery.py
muddlebee: any particular reason behind this @mayankbharati-ops ? `exc_type`

muddlebee (commented):
pls address the above

mayankbharati-ops:
@muddlebee — pushed `059958a` addressing all four nits + the two CodeQL alerts. Replying inline since they all interact:

**1. `slack_delivery.py:221` — log key `type=` vs `exc_type=`**
Good catch on the log-parser concern. Reverted to the exact pre-refactor format: `"...type=%s channel=%s thread_ts=%s detail=%s..."`. No diff vs `main` for that log line now, so any parser keying off `type=` keeps working.

**2. `delivery_transport.py:62` — rename `error_type` → `exc_type`**
Done. `DeliveryResponse.exc_type` (Pythonic short form for "exception type"). Couldn't literally name the field `type` because it shadows the builtin, so the field name is `exc_type` while the slack log key is `type=` — the two are intentionally different: the field is the Python attribute, the log key matches the legacy format. Added a docstring note explaining the distinction.

**3. `discord_delivery.py:17` — redundant `Content-Type` header**
Right, `httpx.post(json=payload)` already sends `Content-Type: application/json` and serializes UTF-8. Dropped from both `_discord_auth_headers` and `_slack_bearer_headers` — both helpers now return just the auth header. No behavioural change (verified via the per-callsite tests + full delivery suite).

**4. `slack_delivery.py:221` (follow-up) — "any particular reason behind `exc_type`?"**
Pure naming choice when adding the new `DeliveryResponse` field. No log-format motivation. With the rename in (2) and the log key reverted in (1), there's no longer an `exc_type` token in any log message — the slack log emits `type=…` as before.

**5. CodeQL #503 / #504 — `import` and `import from` on the same module**
The new `test_module_does_not_import_httpx` regression tests on the discord/telegram (and the new slack) test files were doing `import app.utils.X as mod` while the file already had `from app.utils.X import (…)` at the top. Switched to a single `from import` style: each test file now does `from app.utils import <module>` once, and the regression test reuses that reference (`assert not hasattr(<module>, "httpx")`). Both CodeQL alerts now resolve in this run.

**Verification:** 3208 pass / 2 skipped / 1 xfailed (unchanged). `ruff check`, `ruff format --check`, `mypy app/utils/` all clean. CI: quality, typecheck, test, Analyze (python), CodeQL — all SUCCESS.

mayankbharati-ops:
@muddlebee — ran a deep live end-to-end probe against the real Slack / Discord / Telegram APIs (no mocks, real HTTPS, real provider error shapes), and **the live test caught a real regression** introduced by my previous nit-fix commit. Pushed `9f08657` to fix it.

## What the live probe found

| # | Probe | Result |
|---|---|---|
| 1 | Real success: `httpbin.org/post` → ok=True, status 200, parsed JSON, `MappingProxyType` data immutable on the wire object | ✅ |
| 2 | Real DNS failure: `*.invalid` → `ok=False`, `status_code=0`, `exc_type="ConnectError"` | ✅ |
| 3 | Real Slack `chat.postMessage` w/ bogus token → `slack_error=invalid_auth` parsed correctly. **But:** `response_metadata.warnings=['missing_charset']` ❌ |
| 4 | Real Slack `reactions.add` w/ bogus token → returns False, no crash | ✅ |
| 5 | Real Slack incoming webhook (bad URL) → returns False on real 4xx | ✅ |
| 6 | Real Discord `/channels/.../messages` w/ bogus token → real 401 body `{"message": "401: Unauthorized"}` parsed via `_discord_error_from_data` | ✅ |
| 7 | Real Discord create-thread w/ bogus token → real 401 surfaced cleanly | ✅ |
| 8 | Real Telegram `sendMessage` w/ bogus token → 401 parsed, **bot token redacted from error string** | ✅ |
| 9 | httpx logger filter scrubs `bot<token>/sendMessage` → `bot<redacted>/sendMessage` on real call | ✅ |
| 10 | All three modules `not hasattr(mod, "httpx")` at runtime | ✅ |
| 11 | On-the-wire `Content-Type: application/json` confirmed via httpbin echo | ✅ |

## The regression and the fix (`9f08657`)

The previous commit dropped `Content-Type: application/json; charset=utf-8` from `_slack_bearer_headers` because httpx auto-sets `Content-Type: application/json` for `json=` kwargs. **But httpx omits the `charset=utf-8` suffix**, and Slack's API ([docs](https://api.slack.com/web#posting_json)) explicitly recommends it — without it, `response_metadata.warnings` carries `['missing_charset']` on every Slack POST. The request still succeeds at the auth layer (the `invalid_auth` came back fine), so it didn't show up in any unit test, but it's a behavioural regression vs `main`.

**Fix:** restored the explicit `Content-Type: application/json; charset=utf-8` on `_slack_bearer_headers` only (Slack-specific). `_discord_auth_headers` stays auth-only — Discord does not emit this warning and its docs do not require the charset suffix. Inline comments on both helpers explain the asymmetry.

**Re-verified live:** direct probe to `chat.postMessage` now returns `warnings=[]` on the live API. `38/38` real-network checks pass.

**Pinned via tests:** the existing `test_sends_correct_url_and_headers` (reactions) and `test_sends_thread_reply_payload` (chat.postMessage) tests now assert on the captured `Content-Type: application/json; charset=utf-8` header, so a future regression that drops the charset will fail loudly in CI.

## Final verification

- 3208 pass / 2 skipped / 1 xfailed
- `ruff check`, `ruff format --check`, `mypy app/utils/` clean
- 38/38 live checks against real Slack / Discord / Telegram APIs
- CI: quality / typecheck / test / Analyze (python) / CodeQL — all SUCCESS

muddlebee:
@mayankbharati-ops merged. Thank you for your efforts. ✨

mayankbharati-ops:
Thanks @muddlebee sir. By the way, just want to point out that we are from same college hahaha. I'm a 2024 graduate from IIITA.

muddlebee:
@mayankbharati-ops wow that's great. Glad to have you on board. Let's connect on discord :)

mayankbharati-ops:
Sure sir.

jlalbdalghnyalhlaly-afk (approved):

PR #976, opened 2026-04-26
fix(eks): forward stored AWS integration credentials to k8s client builder

### LINKED ISSUE #969: EKS k8s client silently ignores stored AWS integration credentials (follow-up to #724)
## Context

PR #724 fixed the EKS source-detection gate: when the AWS integration is configured with stored IAM user credentials (`access_key_id` + `secret_access_key`, no `role_arn`), the EKS branch in `detect_sources` now activates instead of being skipped.

However, the follow-up path (`build_k8s_clients`) still does not know about those stored credentials. After #724, `build_k8s_clients` resolves credentials in this order:

1. `role_arn` → STS AssumeRole (existing)
2. ambient botocore chain: env / shared config / instance profile / IRSA (added in #724 as a fallback)

The stored IAM user credentials attached to the AWS integration (persisted by `resolve_integrations` and surfaced in `_eks_int["credentials"]`) are **silently dropped**. On hosts that also happen to have ambient credentials, the ambient chain will pick up unrelated credentials; on hosts without ambient credentials, the call fails entirely — even though the integration has perfectly usable IAM user keys right there.

## Expected

When the AWS integration has stored credentials, `build_k8s_clients` should use them explicitly. This should be the **highest-priority** resolution path, ahead of `role_arn` AssumeRole and ambient botocore, because the stored credentials are what the integration was explicitly configured with.

## Actual

The stored credentials are not forwarded from `detect_sources` to `build_k8s_clients` at all. `eks_params` carries `role_arn`, `external_id`, `cluster_names`, etc., but not `credentials`. `build_k8s_clients` has no `credentials` parameter to receive them.

## Proposal (Approach #2)

Forward the stored credentials through the existing `eks_params` pipeline:

1. **`app/nodes/plan_actions/detect_sources.py`**: set `eks_params["credentials"] = _eks_int.get("credentials") or None` alongside the other EKS integration fields.
2. **`app/services/eks/eks_k8s_client.py::build_k8s_clients`**: accept an optional `credentials` kwarg. Resolution priority becomes:
   1. explicit `credentials` kwarg (stored-integration, new)
   2. `role_arn` AssumeRole (existing production path)
   3. ambient botocore chain (#724 fallback, unchanged)

Both secondary fixes from #724 are preserved (empty `SessionToken` coerced to `None`, regional STS endpoint).

## Why Approach #2

In the [code review of #724](https://github.com/Tracer-Cloud/opensre/pull/724#issuecomment-4286649670), Greptile analyzed two candidate approaches for this exact scenario:

> "Approach #2 is the right call for the follow-up. The ambient botocore resolver would silently miss stored-integration credentials exactly when the user is relying on them."

Ambient-only resolution would only work when the deployment environment happens to expose the same credentials twice (once in the integration, once in env / instance profile / etc.), which is fragile and not guaranteed. Explicit forwarding mirrors how every other integration (`grafana`, `bitbucket`, `github`, `datadog`) already uses the values that `resolve_integrations` persisted.

## Regression safety

The `role_arn` (AssumeRole) and ambient-botocore branches are left unchanged. The new explicit branch only runs when `credentials` is truthy and has both `access_key_id` and `secret_access_key`, so existing deployments see no behaviour change.

## Related

- Follow-up to #724.
- Depends on the EKS-detection gate fix from #724 to enter the `build_k8s_clients` path at all for IAM-user setups.

chaosreload:
Fixes #969. Follow-up to #724.

## Summary

PR #724 gated EKS source detection so that AWS integrations configured with stored IAM user credentials (no `role_arn`) can reach the EKS branch. This PR completes the story by actually using those stored credentials in `build_k8s_clients`.

## Changes

1. **`app/nodes/plan_actions/detect_sources.py`**: forward `_eks_int.get("credentials")` into `eks_params["credentials"]`.
2. **`app/services/eks/eks_k8s_client.py::build_k8s_clients`**: new optional `credentials` kwarg. Resolution priority:
   1. **explicit `credentials`** (stored-integration, new)
   2. `role_arn` → STS AssumeRole (existing production path)
   3. ambient botocore chain (#724 fallback)

Secondary fixes from #724 preserved: empty `SessionToken` coerced to `None`, regional STS endpoint.

## Why approach #2

In the review of #724, Greptile analyzed two candidate follow-ups and recommended this one:

> "Approach #2 is the right call for the follow-up. The ambient botocore resolver would silently miss stored-integration credentials exactly when the user is relying on them."

— https://github.com/Tracer-Cloud/opensre/pull/724#issuecomment-4286649670

Explicit forwarding also mirrors how every other integration (`grafana`, `bitbucket`, `github`, `datadog`) uses the values that `resolve_integrations` persisted.

## Regression safety

The `role_arn` (AssumeRole) and ambient-botocore branches are unchanged. The new explicit branch only runs when `credentials` is truthy and has both `access_key_id` and `secret_access_key`, so existing deployments see no behaviour change.

## Testing

- `python -m py_compile` passes on both files.
- `ruff check` and `ruff format --check` clean on both files.
- Follows the project lint config in `pyproject.toml`.

### DIFF
--- a/app/services/eks/eks_client.py
+++ b/app/services/eks/eks_client.py
@@ -1,31 +1,12 @@
 """EKS boto3 client — stored credentials preferred, AssumeRole fallback."""
 
+from __future__ import annotations
+
 from typing import Any
 
 import boto3
 
-
-def _stored_credentials_to_aws_creds(credentials: dict[str, Any] | None) -> dict[str, Any] | None:
-    """Normalize a stored AWS-integration credential dict into the shape that
-    ``sts.assume_role().Credentials`` returns (``AccessKeyId`` /
-    ``SecretAccessKey`` / ``SessionToken``).
-
-    Returns ``None`` when either required key is missing or falsy, so callers
-    can fall through to the AssumeRole / ambient path. An empty or missing
-    ``session_token`` is coerced to ``None`` because botocore rejects
-    ``SessionToken=""`` but accepts a missing token for plain IAM user keys.
-    """
-    if not credentials:
-        return None
-    access_key = credentials.get("access_key_id")
-    secret_key = credentials.get("secret_access_key")
-    if not access_key or not secret_key:
-        return None
-    return {
-        "AccessKeyId": access_key,
-        "SecretAccessKey": secret_key,
-        "SessionToken": credentials.get("session_token") or None,
-    }
+from app.services.eks.utils import stored_credentials_to_aws_creds
 
 
 class EKSClient:
@@ -45,7 +26,7 @@ class EKSClient:
         external_id: str,
         credentials: dict[str, Any] | None,
     ) -> Any:
-        stored = _stored_credentials_to_aws_creds(credentials)
+        stored = stored_credentials_to_aws_creds(credentials)
         if stored is not None:
             # Explicit stored-integration credentials path (highest priority).
             # The AWS integration was configured with IAM user keys
--- a/app/services/eks/eks_client.py
+++ b/app/services/eks/eks_client.py
@@ -1,21 +1,71 @@
-"""EKS boto3 client with IAM role assumption."""
+"""EKS boto3 client — stored credentials preferred, AssumeRole fallback."""
 
 from typing import Any
 
 import boto3
 
 
-class EKSClient:
-    def __init__(self, role_arn: str, external_id: str = "", region: str = "us-east-1"):
-        self._region = region
-        self._boto_client = self._build(role_arn, external_id)
+def _stored_credentials_to_aws_creds(credentials: dict[str, Any] | None) -> dict[str, Any] | None:
+    """Normalize a stored AWS-integration credential dict into the shape that
+    ``sts.assume_role().Credentials`` returns (``AccessKeyId`` /
+    ``SecretAccessKey`` / ``SessionToken``).
 
-    def _build(self, role_arn: str, external_id: str) -> Any:
-        sts = boto3.client("sts")
-        kwargs: dict = {"RoleArn": role_arn, "RoleSessionName": "TracerEKSInvestigation"}
-        if external_id:
-            kwargs["ExternalId"] = external_id
-        c = sts.assume_role(**kwargs)["Credentials"]
+    Returns ``None`` when either required key is missing or falsy, so callers
+    can fall through to the AssumeRole / ambient path. An empty or missing
+    ``session_token`` is coerced to ``None`` because botocore rejects
+    ``SessionToken=""`` but accepts a missing token for plain IAM user keys.
+    """
+    if not credentials:
+        return None
+    access_key = credentials.get("access_key_id")
+    secret_key = credentials.get("secret_access_key")
+    if not access_key or not secret_key:
+        return None
+    return {
+        "AccessKeyId": access_key,
+        "SecretAccessKey": secret_key,
+        "SessionToken": credentials.get("session_token") or None,
+    }
+
+
+class EKSClient:
+    def __init__(
+        self,
+        role_arn: str,
+        external_id: str = "",
+        region: str = "us-east-1",
+        credentials: dict[str, Any] | None = None,
+    ):
+        self._region = region
+        self._boto_client = self._build(role_arn, external_id, credentials)
+
+    def _build(
+        self,
+        role_arn: str,
+        external_id: str,
+        credentials: dict[str, Any] | None,
+    ) -> Any:
+        stored = _stored_credentials_to_aws_creds(credentials)
+        if stored is not None:
+            # Explicit stored-integration credentials path (highest priority).
+            # The AWS integration was configured with IAM user keys
+            # (access_key_id + secret_access_key), possibly with a
+            # session_token, and no role_arn. Without this branch the call
+            # would fall through to sts.assume_role(RoleArn="", ...) and raise
+            # ParamValidationError.
+            c = stored
+        elif role_arn:
+            sts = boto3.client("sts")
+            kwargs: dict = {
+                "RoleArn": role_arn,
+                "RoleSessionName": "TracerEKSInvestigation",
+            }
+            if external_id:
+                kwargs["ExternalId"] = external_id
+            c = sts.assume_role(**kwargs)["Credentials"]
+        else:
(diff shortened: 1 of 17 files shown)


chaosreload:
Pushed an additional commit () that closes the `EKSClient`-path gap Greptile flagged on the (now-closed) parallel PR #990. Until now #976 only covered the seven k8s-API tools that route through `build_k8s_clients`; `list_eks_clusters`, `describe_eks_cluster`, `describe_eks_addon`, and `get_eks_nodegroup_health` still hit `sts.assume_role(RoleArn=\"\", ...)` inside `EKSClient._build` for IAM-user-only integrations and raised `ParamValidationError` at the cluster-discovery step.

**Changes**

- `EKSClient.__init__` accepts an optional `credentials` kwarg; `_build` prefers stored IAM user keys over `sts:AssumeRole` and raises `ValueError` when both are missing (no silent fallthrough).
- New shared helper `_stored_credentials_to_aws_creds` keeps the normalisation rules (empty `session_token` → `None`, both required keys present) in one place — `build_k8s_clients` now uses the same helper so both code paths stay in sync.
- The four `EKSClient`-backed tools accept and forward `credentials` (function signature, `input_schema.properties`, constructor call).

**Tests**

- New `tests/services/test_eks_client.py` covers the helper (full triplet / empty session token / missing required keys / `None` / empty dict) and `EKSClient` credential resolution (stored skips `AssumeRole`, empty token coerced to `None`, partial dict falls back to `AssumeRole`, no creds + no `role_arn` raises `ValueError`, existing `AssumeRole` + `external_id` passthrough stays intact).
- `test_eks_list_clusters_tool.py` grows two forwarding tests: the kwarg threads into `EKSClient`, and defaults to `None` for existing role-based callers.

**Quality gate**

```
make format-check  → 905 files already formatted
make typecheck     → Success: no issues found in 391 source files
make test-cov      → 3328 passed, 2 skipped, 1 xfailed in 42.39s
```

app/services/eks/eks_client.py
```diff
@@ -1,21 +1,71 @@
-"""EKS boto3 client with IAM role assumption."""
+"""EKS boto3 client — stored credentials preferred, AssumeRole fallback."""
```
muddlebee, line 1: Missing `from __future__ import annotations` — required by the project's code-style rules for all Python files.

app/services/eks/eks_k8s_client.py
```diff
@@ -16,6 +16,8 @@
 import botocore.credentials
 from kubernetes import client as k8s_client
 
+from app.services.eks.eks_client import _stored_credentials_to_aws_creds
```
muddlebee: Importing a private helper (underscore-prefixed) across module boundaries couples two service files. Move `_stored_credentials_to_aws_creds` to `app/services/eks/utils.py` (or drop the leading underscore and make it part of the public API) so the import contract is intentional.

app/services/eks/eks_k8s_client.py
```diff
@@ -76,21 +78,60 @@ def build_k8s_clients(
...
+        # pointed elsewhere. The shared ``_stored_credentials_to_aws_creds``
+        # helper keeps the normalization rules (empty session_token → None,
+        # both required keys present) in sync with ``EKSClient``.
+        logger.info("[eks] Using explicit stored-integration AWS credentials")
+        assumed = stored
+    elif role_arn:
+        assumed = _assume_role(role_arn, external_id, "TracerEKSK8sInvestigation")
+    else:
+        # No role_arn and no explicit creds: fall back to ambient AWS
+        # credentials (env AK/SK, shared config profile, or instance profile
```
muddlebee, line 113: Inline import inside a function body. Project convention requires absolute imports at the top of the file. `botocore` is already imported at module level (lines 14-16), so there's no circular-import concern — move this to the top-level imports.

app/services/eks/eks_k8s_client.py
```diff
@@ -76,21 +78,60 @@ def build_k8s_clients(
...
+        import botocore.session as _bsession
+
+        sess = _bsession.get_session()
+        ambient = sess.get_credentials()
+        if ambient is None:
+            msg = "No AWS credentials available for EKS investigation"
+            logger.error("[eks] %s", msg)
+            raise RuntimeError(msg)
+        frozen = ambient.get_frozen_credentials()
+        assumed = {
```
muddlebee, line 123: `frozen.token or ""` stores an empty string, then `assumed["SessionToken"] or None` (line 131) silently corrects it downstream. Normalise at the source instead: `frozen.token or None` — matches the behaviour of `_stored_credentials_to_aws_creds` and removes the hidden dependency on the downstream guard.

app/tools/EKSDeploymentStatusTool/__init__.py
```diff
@@ -57,6 +58,7 @@ def get_eks_deployment_status(
     role_arn: str,
     external_id: str = "",
     region: str = "us-east-1",
+    credentials: dict | None = None,
```
muddlebee: Use `dict[str, Any] | None` instead of bare `dict | None` — the same pattern is already used in the service layer and it keeps mypy strict mode happy. Applies identically to the other nine tool files changed in this PR.

chaosreload:
@muddlebee thanks for the thorough review — all five points addressed in `abb3345`:

| # | Your comment | Fix |
|---|---|---|
| 1 | `eks_client.py` missing `from __future__ import annotations` | Added. |
| 2 | Cross-module private-helper import couples `eks_client` and `eks_k8s_client` | Helper moved to new `app/services/eks/utils.py` and renamed to `stored_credentials_to_aws_creds` (public). Both service files now import from `utils`, so the contract is intentional. |
| 3 | Inline `import botocore.session` inside the ambient-fallback branch | Promoted to top-level imports. |
| 4 | `frozen.token or \"\"` relies on a downstream `or None` guard | Normalised at the source: `frozen.token or None`, matching `stored_credentials_to_aws_creds`. |
| 5 | `credentials: dict \| None` should be `dict[str, Any] \| None` | Updated in all 11 EKS tool files. |

Quality gate clean:

```
make format-check  → 906 files already formatted
make typecheck     → Success: no issues found in 392 source files
make test-cov      → 3328 passed, 2 skipped, 1 xfailed in 43.36s
```

No behaviour changes — pure refactor of public surface + style.

PR #732, opened 2026-04-21
fix: add is_available and extract_params to MongoDB, ClickHouse, Kafka, and CloudWatch tools

devankitjuneja:
Fixes #730

#### Describe the changes you have made in this PR -

10 tools defaulted to `_always_available`, causing them to appear in the planner for every investigation regardless of pipeline context. In pipelines like `rds-postgres-synthetic` (Grafana-only), the LLM would select MongoDB, ClickHouse, Kafka, and CloudWatch tools, they would fail with `TypeError: missing required argument`, and the agent wasted investigation loops on dead-end actions.

**Root cause:** `BaseTool.is_available()` returns `True` by default. The `requires=[...]` field on `@tool` is documentation-only — not wired to the availability check.

**Changes:**
- Added `mongodb_is_available` + `mongodb_extract_params` to `app/integrations/mongodb.py`
- Added `clickhouse_is_available` + `clickhouse_extract_params` to `app/integrations/clickhouse.py`
- Added `kafka_is_available` + `kafka_extract_params` to `app/integrations/kafka.py`
- Added `cloudwatch_is_available` to `app/tools/utils/availability.py` (no `extract_params` — CloudWatch uses IAM auth, `job_queue` is alert-specific)
- Wired `is_available` + `extract_params` into all 10 affected tool decorators

`extract_params` follows the established PostgreSQL/MongoDB pattern — credentials are auto-filled from resolved integrations so the LLM never needs to supply `connection_string`, `host`, or `bootstrap_servers`. ClickHouse and Kafka `detect_sources.py` wiring is tracked in a follow-up issue; the helpers are ready to activate once that lands.

### Screenshots of the UI changes (If any) -
N/A

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [x] I have reviewed every single line of the AI-generated code
- [x] I can explain the purpose and logic of each function/component I added
- [x] I have tested edge cases and understand how the code handles them
- [x] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**

[...]

PR #703, opened 2026-04-20
fix(postgresql): wire is_available + extract_params into PostgreSQL @tool decorators

ebrahim-sameh:
References #702

Part 1 of 2. MySQL side ships in a separate PR opened alongside this one.

## Summary

Adds the missing `is_available` and `extract_params` callbacks to the `@tool(...)` decorators on all 5 PostgreSQL diagnostic tools. Before this change, every one of them raised `TypeError: missing 2 required positional arguments: 'host' and 'database'` the moment the agent tried to invoke them. Mirrors the existing working pattern used by the MariaDB tool family at `app/integrations/mariadb.py:160-176`.

## What changed

- Added `postgresql_is_available` and `postgresql_extract_params` to `app/integrations/postgresql.py`.
- Wired both helpers into the `@tool(...)` decorator of all 5 PostgreSQL tools (current queries, replication status, server status, slow queries, table stats).

PostgreSQL integration resolves credentials internally via `resolve_postgresql_config` (reading from the store/env), so `extract_params` only surfaces the three identifying params: `host`, `database`, `port`. Credentials stay out of tool signatures and are never seen by the LLM.

Scope: 6 files, +61 / -5 lines.

### Baseline (origin/main @ `6b02982`, full RDS synthetic suite on Gemini 2.5-flash)

- `make test-rds-synthetic` → **0/15 scenarios passed**
- Wall time: 838 s
- 184 PostgreSQL `TypeError` failures during the run:
  - `get_postgresql_server_status`: 92
  - `get_postgresql_current_queries`: 40
  - `get_postgresql_replication_status`: 32
  - `get_postgresql_table_stats`: 20

### After fix (this branch)

- `make test-rds-synthetic` → **0 PostgreSQL TypeErrors** (was 184)
- Suite still shows 0/15 passing — RDS-specific diagnostic tooling over the CloudWatch / Performance Insights mock data is a separate gap. Failure reasons are now about real diagnostic quality (wrong category, missing keywords) rather than masked tool plumbing.
- `make test-cov` → 2725 passed, 1 skipped, 10 warnings (no regressions)
- `ruff check` on all changed files → clean
- `mypy` on all changed files → clean

## Test plan

- [x] `make test-cov` passes with no regressions
- [x] `ruff check` clean on all changed files
- [x] `mypy` clean on all changed files
- [x] Full `make test-rds-synthetic` run shows zero TypeError failures from PostgreSQL tools
- [x] MariaDB tools still work (not touched)

PR #988, opened 2026-04-27
fix: stop run_diagnostic_code from crashing every investigation dispatch

### LINKED ISSUE #956: `run_diagnostic_code` is registered without `is_available`/`extract_params`, crashes on every dispatch
Same shape as #730 / #702 / #706 — a `@tool(...)` registration without the availability guard or the param extractor, so the dispatcher calls the function with empty kwargs and the required `code` argument is missing. `run_diagnostic_code` was introduced in #485 and wasn't picked up by the fix wave (#732, #703, #704, #707).

Reproduces on tag `v2026.4.25` (commit `d3d11915`) under both `llama3.1:8b` and `qwen2.5:7b` through Ollama, so it isn't a model-specific quirk.

Reproducer plus run.log artefacts for six chaos scenarios, each with the warning visible, are at https://github.com/canetizen/opensre-distributed-rca. One excerpt:

```
● Planning  21.9s  Grafana Loki, run diagnostic code
[WARNING] Action failed: run_diagnostic_code:FAILED(TypeError: run_diagnostic_code() missing 1 require)
● Gathering evidence  21ms  grafana:4 logs (0 errors) | Errors: run_diagnostic_code:FAILED(...)
```

The dispatcher (`app/nodes/investigate/execution/execute_actions.py:49-50`) does:

```python
kwargs = action.extract_params(available_sources)
data = action.run(**kwargs)
```

With no `is_available`, the planner sees `run_diagnostic_code` on every investigation. With no `extract_params`, `kwargs` ends up empty, and `run_diagnostic_code(code: str, ...)` raises immediately.

The DB-tools fix was easy — gate them on `"<source>" in available_sources`. This one is genuinely meant to be always-on, but its `code` argument is supposed to come from the LLM's tool-call payload, which the dispatcher path doesn't surface anywhere. The cheap fix is `is_available=lambda _: False` to stop the warning flood; the real fix is threading tool-call args into the dispatcher. Let me know which direction you'd want and I'll send a PR.

Environment: OpenSRE `v2026.4.25` / Python 3.11 (`python:3.11-slim`) / Ollama 0.21.2 on a single host with an NVIDIA GTX 1660 Ti, models `llama3.1:8b` and `qwen2.5:7b`.

AarushSharmaa:
Fixes #956

`run_diagnostic_code` was registered with `@tool` but without `is_available` or `extract_params`. The dispatcher's defaults are: always select the tool (`is_available` returns `True`) and call it with empty kwargs (`extract_params` returns `{}`). Since `run_diagnostic_code` requires a `code` argument, this caused a `TypeError` on every investigation - even ones completely unrelated to code execution.

The fix is one line: `is_available=lambda _: False` in the `@tool` decorator. This matches the pattern used in #732, #703, and #704 which fixed the same class of bug for other tools. The tool still works perfectly when called directly with a `code` argument - it just won't be auto-selected by the dispatcher anymore.

Also updated the existing test that was asserting the wrong behaviour (`is_available` returning `True`) and made the intent explicit with a second assertion.

---

I traced the crash by reading `execute_actions.py` (the dispatcher) and saw it calls `action.extract_params(sources)` then `action.run(**kwargs)`. For tools without a custom `extract_params`, `kwargs` is always `{}`. `run_diagnostic_code` requires `code` as a positional argument, so the call always raises `TypeError`.

I looked at how prior fixes (#732, #703, #704) handled the same problem: they all used `is_available=lambda _: False` to tell the dispatcher "skip me, I can't be auto-called". That's the right fix here too, because there's no way to pull a meaningful `code` string out of an alert's source dict.

The one-line change to the decorator is the entire fix. I also flipped the test that was asserting the wrong thing and added a second assertion that documents why this matters.

---

### DIFF
--- a/app/tools/run_diagnostic_code.py
+++ b/app/tools/run_diagnostic_code.py
@@ -11,6 +11,7 @@ from app.tools.tool_decorator import tool
 @tool(
     name="run_diagnostic_code",
     source="knowledge",
+    is_available=lambda _: False,
     description=(
         "Execute a Python snippet in a restricted sandbox for targeted diagnostics. "
         "Network access and filesystem writes outside /tmp are blocked. "
--- a/tests/tools/test_run_diagnostic_code_tool.py
+++ b/tests/tools/test_run_diagnostic_code_tool.py
@@ -40,8 +40,11 @@ class TestRunDiagnosticCodeMetadata:
         assert "inputs" not in required
         assert "timeout" not in required
 
-    def test_tool_is_available_without_sources(self) -> None:
-        assert _registered().is_available({}) is True
+    def test_tool_is_not_auto_selected_by_dispatcher(self) -> None:
+        # The dispatcher can't supply the required `code` argument from alert sources,
+        # so this tool must never be auto-selected during an investigation.
+        assert _registered().is_available({}) is False
+        assert _registered().is_available({"knowledge": {"code": "print(1)"}}) is False
 
     def test_registered_on_investigation_surface(self) -> None:
         assert "investigation" in _registered().surfaces


VaibhavUpreti (approved):
Awesome @AarushSharmaa , thanks a lot for fixing this!

VaibhavUpreti:
Welcome to the OpenSRE community :)

AarushSharmaa:
Thanks for the quick merge! Been reading through the codebase to understand how everything fits together and really enjoying it. Looking forward to picking up more issues as I go deeper. And honestly, cheers to what you all are building here, this kind of infrastructure is going to matter a lot as agents start taking over more of the ecosystem.

PR #989, opened 2026-04-27
eval: harden OpenSRE LLM judge JSON extraction

cerencamkiran:
## Summary

This PR improves the robustness of the OpenSRE evaluation pipeline by hardening how LLM judge responses are parsed and validated.

The current evaluation flow relies on extracting a JSON object from LLM outputs. However, real LLM responses are often not strictly well-formed and may include additional text, markdown fences, or malformed structures. This can lead to unstable or incorrect benchmark results.

## Problem

The existing `extract_judge_json_from_response` implementation:

- assumes a clean JSON object in the response
- does not explicitly reject JSON arrays
- can incorrectly extract partial JSON from mixed outputs
- is not tested against real-world LLM response patterns

As a result, evaluation can silently succeed with incorrect parsing or fail unpredictably depending on LLM formatting.

## Changes

- Hardened JSON extraction logic:
  - captures full fenced content (not just `{...}`) to avoid partial extraction
  - explicitly rejects JSON arrays (`[...]`), including fenced cases
  - improves detection of valid JSON object boundaries
  - ensures only dictionary outputs are accepted

- Added dedicated test coverage for judge parsing:
  - clean JSON responses
  - markdown fenced JSON (```json)
  - mixed prose + JSON outputs
  - missing JSON cases
  - JSON arrays (invalid)
  - malformed JSON
  - fenced JSON arrays (invalid, now explicitly rejected)

## Why this matters

OpenSRE evaluation depends on LLM judge outputs for scoring (`overall_pass`, `score_0_100`, etc.).

This change improves:

- evaluation stability
- reproducibility of benchmark results
- robustness to real LLM output variability
- prevents silent mis-scoring due to partial JSON extraction 

## Result:
8 passed

## Validation

```bash
pytest tests/integrations/opensre/test_llm_eval_judge.py

### DIFF
--- a/app/integrations/opensre/llm_eval_judge.py
+++ b/app/integrations/opensre/llm_eval_judge.py
@@ -51,7 +51,7 @@ def extract_judge_json_from_response(text: str) -> dict[str, Any]:
     if fence:
         text = fence.group(1).strip()
 
-    if text.startswith("["):
+    if text.lstrip().startswith("["):
         msg = "Judge response JSON must be an object"
         raise ValueError(msg)
 
--- a/app/integrations/opensre/llm_eval_judge.py
+++ b/app/integrations/opensre/llm_eval_judge.py
@@ -50,11 +50,17 @@ def extract_judge_json_from_response(text: str) -> dict[str, Any]:
     fence = re.search(r"```(?:json)?\s*(\{.*\})\s*```", text, re.DOTALL)
     if fence:
         text = fence.group(1)
+
+    if text.startswith("["):
+        msg = "Judge response JSON must be an object"
+        raise ValueError(msg)
+
     start = text.find("{")
     end = text.rfind("}")
     if start == -1 or end == -1 or end <= start:
         msg = "Judge response did not contain a JSON object"
         raise ValueError(msg)
+
     raw = json.loads(text[start : end + 1])
     if not isinstance(raw, dict):
         msg = "Judge response JSON must be an object"
--- a/app/integrations/opensre/llm_eval_judge.py
+++ b/app/integrations/opensre/llm_eval_judge.py
@@ -47,9 +47,9 @@ def _claims_lines(claims: list[Any], key: str = "claim") -> str:
 
 def extract_judge_json_from_response(text: str) -> dict[str, Any]:
     text = text.strip()
-    fence = re.search(r"```(?:json)?\s*(\{.*\})\s*```", text, re.DOTALL)
+    fence = re.search(r"```(?:json)?\s*([\s\S]*?)\s*```", text, re.DOTALL)
     if fence:
-        text = fence.group(1)
+        text = fence.group(1).strip()
 
     if text.startswith("["):
         msg = "Judge response JSON must be an object"
--- a/app/integrations/opensre/llm_eval_judge.py
+++ b/app/integrations/opensre/llm_eval_judge.py
@@ -75,6 +75,23 @@ def extract_judge_json_from_response(text: str) -> dict[str, Any]:
     if isinstance(raw, list):
         msg = "Judge response JSON must be an object"
         raise ValueError(msg)
+    
+    array_start = text.find("[")
+    array_end = text.rfind("]")
+
+    if array_start != -1 and array_end != -1 and array_end > array_start:
+        object_start = text.find("{")
+
+        if object_start == -1 or array_start < object_start:
+            try:
+                candidate = text[array_start : array_end + 1]
+                parsed = json.loads(candidate)
+            except json.JSONDecodeError:
+                parsed = None
+
+            if isinstance(parsed, list):
+                msg = "Judge response JSON must be an object"
+                raise ValueError(msg)
 
     start = text.find("{")
     end = text.rfind("}")
--- a/app/integrations/opensre/llm_eval_judge.py
+++ b/app/integrations/opensre/llm_eval_judge.py
@@ -50,8 +50,6 @@ def extract_judge_json_from_response(text: str) -> dict[str, Any]:
 
     fences = re.findall(r"```(?:json)?\s*([\s\S]*?)\s*```", text, re.DOTALL)
     if fences:
-        saw_array_fence = False
-
         for fence_candidate in reversed(fences):
             fence_candidate = fence_candidate.strip()
             try:
@@ -63,12 +61,7 @@ def extract_judge_json_from_response(text: str) -> dict[str, Any]:
                 return cast(dict[str, Any], parsed_fence)
 
             if isinstance(parsed_fence, list):
-                saw_array_fence = True
                 continue
-
-        if saw_array_fence:
-            msg = "Judge response JSON must be an object"
-            raise ValueError(msg)
 
     try:
         raw = json.loads(text)
--- a/app/integrations/opensre/llm_eval_judge.py
+++ b/app/integrations/opensre/llm_eval_judge.py
@@ -75,7 +75,7 @@ def extract_judge_json_from_response(text: str) -> dict[str, Any]:
     if isinstance(raw, list):
         msg = "Judge response JSON must be an object"
         raise ValueError(msg)
-    
+
     array_start = text.find("[")
     array_end = text.rfind("]")
 
--- a/app/integrations/opensre/llm_eval_judge.py
+++ b/app/integrations/opensre/llm_eval_judge.py
@@ -83,15 +83,17 @@ def extract_judge_json_from_response(text: str) -> dict[str, Any]:
         object_start = text.find("{")
 
         if object_start == -1 or array_start < object_start:
-            try:
-                candidate = text[array_start : array_end + 1]
-                parsed = json.loads(candidate)
-            except json.JSONDecodeError:
-                parsed = None
+            for i in range(array_start, array_end + 1):
+                if text[i] == "[":
+                    candidate = text[i : array_end + 1]
+                    try:
+                        parsed = json.loads(candidate)
+                    except json.JSONDecodeError:
+                        continue
 
-            if isinstance(parsed, list):
-                msg = "Judge response JSON must be an object"
-                raise ValueError(msg)
+                    if isinstance(parsed, list):
+                        msg = "Judge response JSON must be an object"
+                        raise ValueError(msg)
 
     start = text.find("{")
     end = text.rfind("}")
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by rrajan94** [comment]
Hey @cerencamkiran, thanks for tightening up the judge JSON parsing and adding targeted tests 👍

I see one behavior gap to fix before we merge:

- Right now the `startswith("[")` guard only catches arrays when the response literally begins with `[`. If there is any leading whitespace before the array, we can still end up slicing out a `{...}` from inside an array and treating it as valid. To fully meet the “reject JSON arrays” goal, can you change that to use `text.lstrip().startswith("[")` (and then parse from the stripped text)?

Also, a small test alignment:

- In `test_extract_judge_json_raises_for_json_array` you can simplify the `match` to just `"JSON must be an object"` once the guard is in place, since that will be the consistent error.

Optionally, adding a test for `"  \n  [{\"overall_pass\": true}]"` would lock in the whitespace-tolerant behavior.

Once those tweaks are in and tests are green, this should be good to go. The fenced array test you added already matches what Greptile flagged, which is great.

**Comment by cerencamkiran** [comment]
Great catch @rrajan94.
I'll update the guard to be whitespace-tolerant and add the extra test case. Thanks!

**Comment by cerencamkiran** [comment]
Hey @rrajan94! I updated the guard to handle leading whitespace and added the test. All passing 👍

**Comment by WatchTree-19** [comment]
nice work @cerencamkiran. fenced-array test is the kind of thing that would silently corrupt benchmark numbers if missed.

one thing not blocking. fence regex went from (\{.*\}) to ([\s\S]*?) which is broader than just a fix. old required {...} inside fence. new captures any content. so a prose-only fence with a JSON object outside used to parse fine, now fence eats the prose and the function raises. visible failure beats silent mis-parse so probably right, just flagging.

lgtm

**Review by yashksaini-coder — COMMENTED** [review]
Solid hardening, the tests are clear and greptile's two earlier P1s look fully addressed at HEAD. A couple of edge cases I'd love your take on inline.

Locally on this branch: `make lint`, `make format-check`, mypy on the modified file, and the new judge test file (8/8 passed in 1s) all pass. Caller in `app/nodes/evaluate_opensre/node.py:37` already wraps the call in `except Exception`, so any new ValueErrors degrade gracefully into eval errors instead of crashing the pipeline.

File: app/integrations/opensre/llm_eval_judge.py [hunk 1]
```diff
@@ -47,14 +47,20 @@ def _claims_lines(claims: list[Any], key: str = "claim") -> str:
 
 def extract_judge_json_from_response(text: str) -> dict[str, Any]:
     text = text.strip()
-    fence = re.search(r"```(?:json)?\s*(\{.*\})\s*```", text, re.DOTALL)
+    fence = re.search(r"```(?:json)?\s*([\s\S]*?)\s*```", text, re.DOTALL)
```
yashksaini-coder: The new fence regex is now non-greedy (`[\s\S]*?`) so it captures the *first* fenced block. If the judge ever produces something like:

````text
Example shape:
```json
{"old": 1}
```

Actual answer:
```json
{"overall_pass": true, "score_0_100": 90}
```
````

the function silently returns `{"old": 1}` (verified locally). The old greedy regex would have spanned both fences and failed `json.loads` noisily, which the caller logs as an eval error. That's exactly the silent-mis-scoring failure mode this PR was written to prevent, just in a different shape.

Worst case it's a niche input, but if the judge prompt ever ends up with a "for example..." preamble it'd silently degrade scores. Two options that would close the gap:

1. Use `re.findall` and pick the last fenced block, or the largest one that parses as a dict.
2. After grabbing the first fence, parse it and if the result isn't a dict, fall through to the next fence.

Either way, would you mind adding a `test_extract_judge_json_picks_correct_fence_with_multiple` case to lock down the chosen behavior?

File: app/integrations/opensre/llm_eval_judge.py [hunk 2]
```diff
@@ -47,14 +47,20 @@ def _claims_lines(claims: list[Any], key: str = "claim") -> str:
...
     if fence:
-        text = fence.group(1)
+        text = fence.group(1).strip()
+
+    if text.lstrip().startswith("["):
```
yashksaini-coder: Tiny one. The `lstrip().startswith("[")` guard rejects any input whose first non-whitespace char is `[`, which catches actual JSON arrays correctly but also has a false positive for prose that starts with a square bracket, e.g. an LLM that prefixes its response with a log-style label:

- `[INFO] result: {"overall_pass": true}`
- `[Note] {"overall_pass": true, ...}`

Both used to parse successfully via the `find("{")` / `rfind("}")` salvage path, and now raise (verified locally). The caller catches it as `Exception` so it just degrades the eval into an error, not silent, which is fine. But if you'd rather not lose the salvage path, a stricter guard would be to actually parse the input first and then `isinstance(raw, list)`. Up to you.

cerencamkiran:
Thanks @yashksaini-coder.

The multiple-fence case is now handled by scanning all fenced blocks and selecting the last valid JSON object, with a regression test added.

For the `[INFO] ...` case, I now parse first and only reject if the result is actually a list, so the salvage path is preserved.

All tests and checks are passing.

rrajan94:
A prose-wrapped JSON array still slips through because the brace-substring fallback (`text.find("{")` / `text.rfind("}")`) slices out the first object inside the array.

repro:
```
extract_judge_json_from_response('Here is the result: [{"overall_pass": true}] Thanks')
# returns {'overall_pass': True} instead of raising
```
flow at HEAD: no fences → `json.loads` of the whole text fails → `isinstance(raw, list)` guard is skipped → `find("{")`/`rfind("}")` salvage path returns the inner object.

can you add a regression test for this and tighten the fallback so arrays are rejected even when there's prose around them? matching the existing "JSON must be an object" error keeps it consistent with the other array tests.

cerencamkiran:
Thanks @rrajan94  added a regression test for the prose-wrapped array case and tightened the fallback so arrays are rejected before the `{...}` salvage path when the array appears before any JSON object.

Validation:
pytest tests/integrations/opensre/test_llm_eval_judge.py -v
10 passed

rrajan94:
Thanks for fixing the prose-wrapped array case @cerencamkiran  
I think the new bracket-based fix still has one related edge case.

Example: `[INFO] result: [{"overall_pass": true}]`

Because the check parses from the first `[` to the last `]`, the substring starts with `[INFO]`, fails JSON parsing, and then falls through to the `{...}` fallback. That still returns `{"overall_pass": true}` instead of rejecting the array.

Can you add this as a regression test too? It may be worth checking the fallback against a few bracket/prose variants so we don’t keep finding the same partial-parse issue one shape at a time.

cerencamkiran:
@rrajan94 added a regression test for the `[INFO] result: [...]` case as well and updated the array detection to try parsing from each `[` candidate before the `{...}` fallback.

Validation:
pytest tests/integrations/opensre/test_llm_eval_judge.py -v
11 passed

rrajan94:
Thanks @cerencamkiran, the `[INFO] result: [...]` array case is covered now. 

I found one false positive from the new scan though.

This valid object now raises:

`[INFO] result: {"overall_pass": true, "score_0_100": 90, "rubric_items": [], "summary": "ok"}`

The loop tries each `[` before the `{...}` fallback, so it eventually parses the nested `rubric_items: []` array and rejects the whole response before the full object can be salvaged.

Can you add this as a regression test too? The array rejection should catch top-level/salvaged arrays, but not arrays nested inside the valid judge object.

cerencamkiran:
Thanks @rrajan94.

I’ve updated the logic to distinguish top-level arrays from nested ones and added the regression test.

Validation:
pytest tests/integrations/opensre/test_llm_eval_judge.py -v
12 passed

yashksaini-coder (approved):
Approving. Both of my earlier concerns are fully addressed at HEAD `b981c1f0`:

- **Multi-fence first-match** — the `reversed(fences)` + per-fence parse-and-check pattern picks the last valid block, exactly the right shape for judge prompts where the model puts examples first and the answer last. Locked down by `test_extract_judge_json_picks_last_valid_fence_with_multiple_fences`.
- **`[INFO] result: {...}` false positive** — the new top-level-vs-nested array detection (the `arr_start < obj_start AND arr_end > obj_end` check at lines 87-89) is a clean way to distinguish a top-level array from a nested one. Covered by `test_extract_judge_json_allows_log_prefix_with_nested_array`.

I ran a standalone verifier against all 12 test cases (extracted the function source so I didn't need a venv) and got 12/12 passing. CI is green. The one open item is the CodeQL "empty except clause" info on line 68 — it's a known false-positive on `except: raw = None` fallthrough, and your workflow already passes. Adding a one-line `# fall through to brace-substring fallback` comment above the `except` would silence the bot if you want it gone, but it's not blocking.

Nice job working through the edge cases that came up after my review — the regression-test-per-bug-report discipline is exactly what I'd want from this kind of parsing code.

VaibhavUpreti:
@greptileai review

app/integrations/opensre/llm_eval_judge.py
```diff
@@ -47,18 +47,56 @@ def _claims_lines(claims: list[Any], key: str = "claim") -> str:
...
+    arr_end = text.rfind("]")
+
+    has_obj = obj_start != -1 and obj_end != -1 and obj_end > obj_start
+    has_arr = arr_start != -1 and arr_end != -1 and arr_end > arr_start
+
+    # If an array span exists and fully contains the object span,
+    # the top-level value is an array — reject it.
+    if has_arr and has_obj and arr_start < obj_start and arr_end > obj_end:
+        msg = "Judge response JSON must be an object"
+        raise ValueError(msg)
```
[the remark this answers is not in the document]
rrajan94: @cerencamkiran please resolve this, looks like the fallback uses the first `[` and last `]` in the whole text, so unrelated bracketed text before and after the object is treated like an array wrapping it.

cerencamkiran:
@greptileai review

cerencamkiran:
@greptileai review

cerencamkiran:
@rrajan94 thanks for the review.

All previously flagged P1 issues are now addressed. The bracket handling has been updated to only reject when the span actually parses as a JSON array, so cases like `[INFO] ... {dict} ... [end]` no longer fail.

Added regression tests for these scenarios as well.

All tests are passing (15/15), and all checks are green.

I’ll leave the remaining P2 style note as non-blocking for now.

VaibhavUpreti (approved):
Awesome @cerencamkiran , thanks for improving this! 

Merging this to main, tagging @w3joe for visibility

Issue #901, opened 2026-04-24
Move `app/integrations/gitlab_test.py` into `tests/integrations/`


VaibhavUpreti:
## Why This Matters
the GitLab integration test suite is substantial and should live with the rest of the integration tests.

## Read First
- `app/integrations/gitlab_test.py`
- `tests/integrations/`

## Scope
1. Move the file into `tests/integrations/`.
2. Update imports only if required by the new location.
3. Delete the old inline copy.

## Tests / Coverage
- no new assertions are required
- first prove the moved suite still collects and passes

## Acceptance Criteria
- GitLab integration tests live under `tests/integrations/`
- the inline file is removed
- coverage remains the same

PR #1003, opened 2026-04-27
test(tools): add unit tests for the four GitLab tools

### LINKED ISSUE #993: Add unit tests for GitLab tools (Commits, File, MRs, Pipelines)
## Summary

Four GitLab tools have no unit tests under `tests/tools/`: `GitLabCommitsTool`, `GitLabFileTool`, `GitLabMRsTool`, and `GitLabPipelinesTool`. An inline test exists at `app/integrations/gitlab_test.py` (tracked separately by #901) but it only covers the config/integration layer.

## Context

The four tools live under `app/tools/GitLab*Tool/`. They make read-only REST calls to the GitLab API using a token from `sources["gitlab"]`. Other similar provider tools (GitHub, Bitbucket) provide a good pattern to follow.

## What to implement

Create four files:
- `tests/tools/test_gitlab_commits_tool.py`
- `tests/tools/test_gitlab_file_tool.py`
- `tests/tools/test_gitlab_mrs_tool.py`
- `tests/tools/test_gitlab_pipelines_tool.py`

Each file should:

1. **Contract test** — inherit `BaseToolContract` from `tests/tools/conftest.py`
2. **`is_available`** — `True` when `sources["gitlab"]` has `token` and `project_id`; `False` otherwise
3. **`extract_params`** — assert the dict correctly maps `project_id`, `token`, `gitlab_url` from sources
4. **`run` happy path** — patch the outgoing HTTP call (via `unittest.mock.patch` on the httpx or requests call inside `app/integrations/gitlab.py`) and verify the tool returns `available: True` with the expected output shape
5. **`run` error path** — simulate a non-200 response or network error and verify a graceful error dict is returned

## Acceptance criteria

- [ ] All four test files created under `tests/tools/`
- [ ] Each covers `is_available`, `extract_params`, and `run` (happy + error paths)
- [ ] `make test-cov` passes
- [ ] `make lint` passes

## References

- Pattern: `tests/tools/test_github_commits_tool.py` (closest analogue)
- Tool sources: `app/tools/GitLab*Tool/`
- Integration: `app/integrations/gitlab.py`
- Inline test to eventually supersede: `app/integrations/gitlab_test.py` (#901)

0xDevNinja:
Fixes #993.

Adds direct unit tests for the four GitLab tools that previously had no coverage under `tests/tools/`:

- `tests/tools/test_gitlab_commits_tool.py` — `list_gitlab_commits`
- `tests/tools/test_gitlab_file_tool.py` — `get_gitlab_file_contents`
- `tests/tools/test_gitlab_mrs_tool.py` — `list_gitlab_mrs`
- `tests/tools/test_gitlab_pipelines_tool.py` — `list_gitlab_pipelines`

Each file follows the same shape as `tests/tools/test_github_commits_tool.py`:

1. **Contract test** via `BaseToolContract` — pulls metadata, schema, and basic `is_available` / `extract_params` invariants from `tests/tools/conftest.py`.
2. **`is_available`** — covers the `connection_verified` + `project_id` (and `file_path` for the file tool) gating, plus the empty-state fallthrough.
3. **`extract_params`** — asserts each field flows from `sources["gitlab"]` to the tool input dict, including the per-tool defaults (`ref_name` → `main`, `target_branch` → `main`, `ref` → `main`, `status` → `failed`).
4. **`run` happy path** — patches `_resolve_config` and the corresponding `app.integrations.gitlab.get_gitlab_*` helper to return a fake response, and verifies the tool returns `available=True` with the expected key (`commits`, `mrs`, `pipelines`, or `file`).
5. **`run` error path** — patches `_resolve_config` to return `None` and verifies the tool returns the documented `available=False` payload (and an empty-list path for the list tools, mirroring how `_request_json` coerces non-200 responses to `[]`). For the file tool, also covers the oversize-guard and the binary-file (`UnicodeDecodeError`) branches.

Each commit is atomic — one file per commit — to keep review focused per tool.

Local quality gate (clean tree, run from `make`):

- `make lint` — `All checks passed!`
- `make format-check` — `908 files already formatted`
- `make typecheck` — `Success: no issues found in 391 source files`
- `make test-cov` — `3363 passed, 2 skipped, 1 xfailed` (the four new test files contribute 49 cases)

```
tests/tools/test_gitlab_commits_tool.py   ............         [12/12]
tests/tools/test_gitlab_file_tool.py      .............        [13/13]
tests/tools/test_gitlab_mrs_tool.py       ............         [12/12]
tests/tools/test_gitlab_pipelines_tool.py ............         [12/12]
```

---

I followed the existing GitHub-tool test pattern (`tests/tools/test_github_commits_tool.py`) so reviewers see one shape repeated four times. The mocking strategy patches the two public surfaces the tool actually depends on — `_resolve_config()` and the matching `get_gitlab_*` integration helper — instead of patching `httpx` directly. That keeps the tests honest to the tool boundary: the tool is responsible for shaping the result dict and gating on a missing config, while the integration layer (already covered separately by `app/integrations/gitlab_test.py`, tracked by #901) is responsible for HTTP semantics. Where the tool has its own branches that aren't pure pass-through — the file tool's 50_000-byte size guard and the UTF-8 decode guard — I added explicit cases so a future change to those thresholds breaks the test rather than the runtime.

I also kept `BaseToolContract` as the first thing in each file because it doubles as a "did the `@tool` decorator wire up correctly" smoke test; if the metadata or `is_available` signature drifts, the contract checks fail before any tool-specific assertion runs.

---

### DIFF
--- a/tests/tools/test_gitlab_file_tool.py
+++ b/tests/tools/test_gitlab_file_tool.py
@@ -58,6 +58,15 @@ def test_extract_params_maps_fields() -> None:
     assert params["gitlab_token"] == "glpat-test"
 
 
+def test_extract_params_defaults_ref_to_main() -> None:
+    rt = get_gitlab_file_contents.__opensre_registered_tool__
+    sources = mock_agent_state(
+        {"gitlab": {"connection_verified": True, "project_id": "42", "file_path": "src/main.py"}}
+    )
+    params = rt.extract_params(sources)
+    assert params["ref"] == "main"
+
+
 def test_run_returns_unavailable_when_config_missing() -> None:
     with patch("app.tools.GitLabFileTool._resolve_config", return_value=None):
         result = get_gitlab_file_contents(project_id="42", file_path="src/main.py")
--- a/tests/tools/test_gitlab_file_tool.py
+++ b/tests/tools/test_gitlab_file_tool.py
@@ -1,0 +1,134 @@
+"""Tests for GitLabFileTool (function-based, @tool decorated)."""
+
+from __future__ import annotations
+
+import base64
+from unittest.mock import MagicMock, patch
+
+from app.tools.GitLabFileTool import get_gitlab_file_contents
+from tests.tools.conftest import BaseToolContract, mock_agent_state
+
+
+class TestGitLabFileToolContract(BaseToolContract):
+    def get_tool_under_test(self):
+        return get_gitlab_file_contents.__opensre_registered_tool__
+
+
+def test_is_available_requires_connection_project_id_and_file_path() -> None:
+    rt = get_gitlab_file_contents.__opensre_registered_tool__
+    assert (
+        rt.is_available(
+            {
+                "gitlab": {
+                    "connection_verified": True,
+                    "project_id": "42",
+                    "file_path": "src/main.py",
+                }
+            }
+        )
+        is True
+    )
+    assert rt.is_available({"gitlab": {"connection_verified": True, "project_id": "42"}}) is False
+    assert (
+        rt.is_available({"gitlab": {"connection_verified": True, "file_path": "src/main.py"}})
+        is False
+    )
+    assert rt.is_available({}) is False
+
+
+def test_extract_params_maps_fields() -> None:
+    rt = get_gitlab_file_contents.__opensre_registered_tool__
+    sources = mock_agent_state(
+        {
+            "gitlab": {
+                "connection_verified": True,
+                "project_id": "42",
+                "file_path": "src/main.py",
+                "ref_name": "develop",
+                "gitlab_url": "https://gitlab.example.com",
+                "gitlab_token": "glpat-test",
+            }
+        }
+    )
+    params = rt.extract_params(sources)
+    assert params["project_id"] == "42"
+    assert params["file_path"] == "src/main.py"
+    assert params["ref"] == "develop"
+    assert params["gitlab_url"] == "https://gitlab.example.com"
+    assert params["gitlab_token"] == "glpat-test"
+
+
+def test_run_returns_unavailable_when_config_missing() -> None:
+    with patch("app.tools.GitLabFileTool._resolve_config", return_value=None):
+        result = get_gitlab_file_contents(project_id="42", file_path="src/main.py")
+    assert result["available"] is False
+    assert "not configured" in result["error"]
+    assert result["file"] == {}
+
+
+def test_run_happy_path_decodes_base64_content() -> None:
+    raw_content = "print('hello world')\n"
+    encoded = base64.b64encode(raw_content.encode("utf-8")).decode("ascii")
+    fake_response = {
+        "file_name": "main.py",
+        "file_path": "src/main.py",
+        "ref": "main",
+        "size": len(raw_content),
+        "content": encoded,
+    }
+    with (
+        patch("app.tools.GitLabFileTool._resolve_config", return_value=MagicMock()),
+        patch("app.tools.GitLabFileTool.get_gitlab_file", return_value=fake_response),
+    ):
+        result = get_gitlab_file_contents(project_id="42", file_path="src/main.py")
+    assert result["available"] is True
+    assert result["file"]["content"] == raw_content
+    assert result["file"]["file_path"] == "src/main.py"
+
+
+def test_run_returns_unavailable_for_oversized_file() -> None:
+    fake_response = {"size": 75_000, "content": "ignored"}
+    with (
+        patch("app.tools.GitLabFileTool._resolve_config", return_value=MagicMock()),
+        patch("app.tools.GitLabFileTool.get_gitlab_file", return_value=fake_response),
+    ):
+        result = get_gitlab_file_contents(project_id="42", file_path="big.bin")
+    assert result["available"] is False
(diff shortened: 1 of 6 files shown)


tests/tools/test_gitlab_file_tool.py
```diff
@@ -0,0 +1,134 @@
...
+                "file_path": "src/main.py",
+                "ref_name": "develop",
+                "gitlab_url": "https://gitlab.example.com",
+                "gitlab_token": "glpat-test",
+            }
+        }
+    )
+    params = rt.extract_params(sources)
+    assert params["project_id"] == "42"
+    assert params["file_path"] == "src/main.py"
```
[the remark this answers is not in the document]
0xDevNinja: Good catch — added in `44f6b29`. The new `test_extract_params_defaults_ref_to_main` mirrors the per-tool default-branch tests already present for the other three GitLab tools and locks the `ref_name` → `ref` = `main` default.

hamzzaaamalik:
Verified locally:

Confirmed all 4 GitLab tools (list_gitlab_commits, get_gitlab_file_contents, list_gitlab_mrs, list_gitlab_pipelines) had no existing tests before this PR
Checked out the branch and ran the new test files: 50/50 pass in 0.36s
Pattern follows the existing BaseToolContract convention used by the GitHub tool tests consistent with the codebase
Pure additive coverage, no production code changes. Looks good to ship pending CI green.

tests/tools/test_gitlab_mrs_tool.py
```diff
@@ -0,0 +1,98 @@
...
+    rt = list_gitlab_mrs.__opensre_registered_tool__
+    sources = mock_agent_state(
+        {
+            "gitlab": {
+                "connection_verified": True,
+                "project_id": "42",
+                "updated_after": "2026-01-01T00:00:00Z",
+            }
+        }
+    )
```
yashksaini-coder, line 57: This test exposes a production bug in `GitLabMRsTool` that needs to be fixed first.

In the production code (`app/tools/GitLabMRsTool/__init__.py` line 18), `extract_params` does:
```python
"updated_after": gl["updated_after"],
```

That uses a hard dict key access. If `updated_after` is missing from sources — which is totally possible since it's not a required field in `is_available` — this raises a `KeyError` at runtime and crashes the agent.

The sibling `GitLabCommitsTool` does it correctly: `gl.get("since", "")`. Same fix needed here: change to `gl.get("updated_after", "")` in the production file.

Every test in this file always provides `updated_after`, so the KeyError is never triggered by the test suite.

Quick fix in `GitLabMRsTool/__init__.py`:
```python
"updated_after": gl.get("updated_after", ""),
```
Then add a test calling `extract_params` without `updated_after` to verify it defaults to `""`.

tests/tools/test_gitlab_pipelines_tool.py
```diff
@@ -0,0 +1,102 @@
...
+def test_extract_params_defaults_ref_to_main() -> None:
+    rt = list_gitlab_pipelines.__opensre_registered_tool__
+    sources = mock_agent_state(
+        {
+            "gitlab": {
+                "connection_verified": True,
+                "project_id": "42",
+                "updated_after": "2026-01-01T00:00:00Z",
+            }
+        }
```
yashksaini-coder, line 57: Same production bug as in `GitLabMRsTool` — `_list_gitlab_pipelines_extract_params` in `app/tools/GitLabPipelinesTool/__init__.py` line 18 also uses `gl["updated_after"]` (hard key access) instead of `gl.get("updated_after", "")`.

Every test here provides `updated_after`, so the KeyError is never triggered.

Fix the production code (`gl.get("updated_after", "")`) and add a test without `updated_after` in sources to confirm it defaults gracefully.

yashksaini-coder (changes requested):
Two issues found — one is a real production bug, the other is a test gap that hides it.

tests/tools/test_gitlab_mrs_tool.py
0xDevNinja: Good catch — fixed in `fc55632`, default test pinned in `4c69f13`. Production code now uses `gl.get("updated_after", "")` to match the `GitLabCommitsTool` `gl.get("since", "")` convention, and the new `test_extract_params_defaults_updated_after_to_empty_string` exercises the missing-field path. HEAD is `5c545ce`.

tests/tools/test_gitlab_pipelines_tool.py
0xDevNinja: Same fix applied here — `4dd8f64` defaults `updated_after` to `""` and `5c545ce` adds the missing-field test. HEAD is `5c545ce`.

yashksaini-coder (approved):
Both issues are properly addressed — production code fixed and tests added.

**Production fixes:**
- `GitLabMRsTool`: `gl["updated_after"]` → `gl.get("updated_after", "")` ✅
- `GitLabPipelinesTool`: same fix ✅

**New tests:**
- `test_extract_params_defaults_updated_after_to_empty_string` in both MRs and Pipelines tool tests — directly catches the KeyError regression ✅

Verified locally: 52 tests pass across all 4 tool test files, ruff clean, mypy clean, CI all green. Good to merge.

muddlebee:
LGTM :+1:

Issue #176, opened 2026-03-30
Hands-on testing: write a review with recommended improvements and friction points


davincios:
A useful contribution would be to do some hands-on testing of the project and write up a short review covering:

- recommended improvements
- friction points during setup or usage
- anything confusing, slow, or unclear in the current experience

This would help us spot rough edges quickly and make the project easier for new contributors and users.

If you pick this up, feel free to include:

- your environment details
- exact steps you followed
- what worked well
- where you got stuck or had to guess
- concrete suggestions for improvement

Tagging a core contributor for visibility: @davincios

Devesh36:
## Environment
- OS: macOS 15.6.1 (arm64)
- Python: 3.12.3
- Shell: zsh 5.9
- Branch: `main`
- Install method: local venv + editable install (`pip install -e ".[dev]"`)

## Exact steps I followed
1. `python3 -m venv .venv`
2. `.venv/bin/python -m pip install -U pip`
3. `.venv/bin/pip install -e ".[dev]"`
4. `.venv/bin/opensre -h`
5. `.venv/bin/opensre tests list`
6. `.venv/bin/opensre health`
7. `.venv/bin/opensre investigate -i tests/fixtures/grafana_local_alert.json`
8. `make lint`
9. `make typecheck`
10. `make test-cov`

## What worked well
- Install flow worked cleanly with Python 3.12 and editable dev install.
- `opensre tests list` is rich and useful; test inventory is discoverable.
- `opensre health` gives a good multi-service status table and clearly shows missing integrations.
- `make typecheck` and `make test-cov` passed successfully in this setup.

## Friction points

| Step | Expected | Actual | Impact | Suggested fix |
|------|----------|--------|--------|---------------|
| `opensre -h` | Help should list all top-level commands | `health` command exists but is missing from command list/help output | Medium | Add `health` to the rendered help command table in CLI help renderer |
| `make lint` | Works after standard `.venv` install | Fails with `make: ruff: No such file or directory` | High | Use `$(PYTHON) -m ruff check app/ tests/` in Makefile instead of bare `ruff` |
| `opensre investigate ...` without key | Friendly actionable guidance | Raw Pydantic traceback shown for missing `ANTHROPIC_API_KEY` | Medium | Catch config validation error in CLI and print concise hint (`run opensre onboard` / set provider key env var) |
[...]

PR #1062, opened 2026-04-29
Fix Makefile Ruff targets to use configured Python

Genmin:
## Summary
- fix Python fallback detection so systems with python3 do not fall through to python
- run Makefile Ruff targets through $(PYTHON) -m ruff
- align local lint and format targets with the CI module invocation

## Why
Part of #176. The Makefile computes a project Python interpreter, but lint/format targets were invoking bare ruff. That can fail after a standard venv install when Ruff is available as a module but not as a shell command.

## Tests
- make -n lint
- make -n format-check
- make -n format
- make lint
- make format-check

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ else ifeq ($(OS),Windows_NT)
         PYTHON = python
         PIP = python -m pip
     endif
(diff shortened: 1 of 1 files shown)


VaibhavUpreti (approved):
awesome @Genmin , thanks a lot for fixing this!

PR #1080, opened 2026-04-29
refactor(utils): extract shared truncate() helper, replace six inline _truncate() implementations

### LINKED ISSUE #1056: Extract repeated _truncate() implementations into a shared app/utils/truncation.py helper
## Why This Matters

Six modules independently define a private `_truncate()` function with inconsistent implementations, different ellipsis characters, and undocumented magic limits. Adding a new delivery channel or integration means writing yet another truncation function from scratch.

This is the same class of duplication fixed in #869 for the GitLab write-back flow.

## Current State

| File | Limit | Ellipsis | Signature |
|---|---|---|---|
| `app/utils/discord_delivery.py:83` | configurable | `…` | `_truncate(text, limit)` |
| `app/utils/telegram_delivery.py:47` | configurable | `…` | `_truncate(text, limit)` |
| `app/integrations/azure_sql.py:177` | 500 (`_QUERY_TRUNCATE_LEN`) | `...` | `_truncate(text, max_len)` |
| `app/integrations/mysql.py:203` | 500 (`_QUERY_TRUNCATE_LEN`) | `...` | `_truncate(text, max_len)` |
| `app/integrations/mariadb.py:154` | 200 (`_QUERY_TRUNCATE_LEN`) | `...` | `_truncate(text, max_len)` |
| `app/nodes/publish_findings/gitlab_writeback.py:14` | 4000 (inline) | `...` | inlined, no function |

`discord_delivery.py` and `telegram_delivery.py` use the Unicode ellipsis `…`; everything else uses `"..."`. `mariadb.py` also has a second ad-hoc truncation at line 294 using `"\n... (truncated)"`.

## Proposed Fix

1. Add `app/utils/truncation.py` with one shared helper:
```python
def truncate(text: str, limit: int, ellipsis: str = "...") -> str:
    if len(text) <= limit:
        return text
    return text[: limit - len(ellipsis)] + ellipsis
```
2. Migrate all six call sites to use it.
3. Keep each module's limit constant where it is — only the function body moves.

## Tests / Coverage

- one parametrised test: text under limit → unchanged, text over limit → truncated with correct ellipsis
- one test for Unicode ellipsis variant
- existing tests for all six modules must stay green

## Acceptance Criteria

- `_truncate` is defined in exactly one place
- ellipsis character is consistent or explicitly configurable per call site
- no behaviour change for any existing caller
- scope limited to the six files listed above

udit-rawat:
## Problem

Six modules independently defined a private `_truncate()` function with inconsistent implementations — different ellipsis characters (`...` vs `…`), different limits (200, 500, 4000, 4096), and no shared source of truth. Every new delivery channel or integration had to reinvent the same logic.

Affected files: `discord_delivery.py`, `telegram_delivery.py`, `azure_sql.py`, `mysql.py`, `mariadb.py`, and `gitlab_writeback.py`.

## Solution

Added `app/utils/truncation.py` with a single `truncate(text, limit, ellipsis="...")` helper. Migrated all six call sites to use it. Each module's limit constant stays where it is — only the function body moves.

## Tests

Added `tests/utils/test_truncation.py` covering:
- text under limit → returned unchanged
- text over limit → truncated with correct ellipsis
- text at exact limit → not truncated
- Unicode ellipsis variant (`…`)
- default ellipsis behaviour

Closes #1056

### DIFF
--- a/app/utils/truncation.py
+++ b/app/utils/truncation.py
@@ -4,5 +4,6 @@
 def truncate(text: str, limit: int, suffix: str = "...") -> str:
     if len(text) <= limit:
         return text
-    cut = max(0, limit - len(suffix))
-    return text[:cut] + suffix
+    if limit <= len(suffix):
+        return suffix[:limit]
+    return text[: limit - len(suffix)] + suffix
--- a/app/utils/truncation.py
+++ b/app/utils/truncation.py
@@ -1,0 +1,7 @@
+"""Shared text truncation utility."""
+
+
+def truncate(text: str, limit: int, ellipsis: str = "...") -> str:
+    if len(text) <= limit:
+        return text
+    return text[: limit - len(ellipsis)] + ellipsis
--- a/app/utils/truncation.py
+++ b/app/utils/truncation.py
@@ -1,7 +1,8 @@
 """Shared text truncation utility."""
 
 
-def truncate(text: str, limit: int, ellipsis: str = "...") -> str:
+def truncate(text: str, limit: int, suffix: str = "...") -> str:
     if len(text) <= limit:
         return text
-    return text[: limit - len(ellipsis)] + ellipsis
+    cut = max(0, limit - len(suffix))
+    return text[:cut] + suffix
--- a/tests/utils/test_truncation.py
+++ b/tests/utils/test_truncation.py
@@ -1,0 +1,37 @@
+"""Tests for the shared truncation utility."""
+
+import pytest
+
+from app.utils.truncation import truncate
+
+
+@pytest.mark.parametrize(
+    "text,limit,ellipsis,expected",
+    [
+        ("short", 10, "...", "short"),
+        ("exactly10c", 10, "...", "exactly10c"),
+        ("this is a long string", 10, "...", "this is..."),
+        ("long", 3, "...", "..."),
+        ("unicode", 5, "…", "unic…"),
+        ("no change needed", 100, "...", "no change needed"),
+    ],
+)
+def test_truncate(text: str, limit: int, ellipsis: str, expected: str) -> None:
+    assert truncate(text, limit, ellipsis=ellipsis) == expected
+
+
+def test_truncate_default_ellipsis() -> None:
+    result = truncate("hello world", 8)
+    assert result == "hello..."
+    assert len(result) == 8
+
+
+def test_truncate_unicode_ellipsis() -> None:
+    result = truncate("hello world", 8, ellipsis="…")
+    assert result.endswith("…")
+    assert len(result) == 8
+
+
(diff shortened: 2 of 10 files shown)


udit-rawat:
@muddlebee all six inline _truncate() implementations replaced with the shared helper in app/utils/truncation.py. Renamed ellipsis param to suffix to avoid the builtin shadow, added max(0, ...) guard  when limit is smaller than the suffix. Ready for review

app/utils/truncation.py
```diff
@@ -0,0 +1,8 @@
+"""Shared text truncation utility."""
+
+
+def truncate(text: str, limit: int, suffix: str = "...") -> str:
+    if len(text) <= limit:
+        return text
+    cut = max(0, limit - len(suffix))
```
muddlebee: When limit < len(suffix) (e.g. truncate("abc", 2, "...")), cut=0 and the return value is just suffix (length 3), which silently exceeds limit. 

Consider all the edge cases here.. and add proper unit tests for all edge cases.

tests/utils/test_truncation.py
```diff
@@ -0,0 +1,37 @@
+"""Tests for the shared truncation utility."""
+
+import pytest
+
+from app.utils.truncation import truncate
+
+
+@pytest.mark.parametrize(
+    "text,limit,ellipsis,expected",
```
muddlebee: The parametrize header uses ellipsis as a field name — it shadows the Python built-in. Rename to suffix to match the actual parameter name and avoid any future confusion.

tests/utils/test_truncation.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_truncate_unicode_ellipsis() -> None:
+    result = truncate("hello world", 8, suffix="…")
+    assert result.endswith("…")
+    assert len(result) == 8
+
+
+def test_truncate_text_at_exact_limit_not_truncated() -> None:
+    text = "a" * 10
+    assert truncate(text, 10) == text
```
muddlebee, line 43: Missing test for the limit < len(suffix) edge case (e.g. truncate("abc", 2, "...")). The current suite only covers limit == len(suffix) via the ("long", 3, "...", "...") case, not limit strictly less than suffix length.

udit-rawat:
@muddlebee addressed both comments — renamed ellipsis to suffix in the parametrize header, added limit < len(suffix) edge case handling (return suffix[:limit]), and added full edge case test coverage including limit=0, empty       
string, empty suffix, and unicode suffix. Also added an explicit assertion that the GitLab MR note body is capped at exactly 4000 chars. Ready for review.

muddlebee:
@udit-rawat nice work. :rocket:

Issue #317, opened 2026-04-03
[FEATURE] Add incident.io integration for incident context and updates


davincios:
## Problem statement
OpenSRE currently lacks a first-class incident.io integration. That leaves a gap for teams that coordinate incidents in incident.io and want the agent to consume incident context and publish findings back into the system of record.

## Proposed solution
Add a first-class incident.io integration that supports:
- onboarding, authentication, and connectivity verification
- reading incident context, timeline, and metadata relevant to RCA
- incorporating incident.io context into investigation planning and evidence gathering
- optionally posting summaries, findings, or next steps back to the incident when safe and appropriate

## Acceptance criteria
- the integration is usable from the investigation workflow end to end
- setup and usage are documented clearly
- the implementation is backed by end-to-end tests in `tests/e2e` or equivalent
- the PR includes a screen video showing setup plus a successful investigation flow using incident.io
- vibecoding without testing will not be allowed; PRs without working end-to-end coverage and screen-video proof should be rejected

SB2318:
I would like to work on this issue,

muddlebee:
Done

SB2318:
> Done

okay

SB2318:
> Done

Thanks for assigning me this issue. It would be a great learning opportunity for me.

After reviewing the scope, this appears to be a medium-to-large feature since it involves:
- incident.io authentication and connectivity setup
- fetching incident context/timeline/metadata
- integrating context into the investigation workflow
- optional write-back support
- documentation, e2e coverage, and screen recording requirements

I’d prefer to prioritize correctness and testing over rushing the implementation, especially given the explicit quality requirements.

Please let me know if that timeline works, or if you'd prefer the scope to be split into smaller milestones/phases (for example: auth + context ingestion first, then write-back support separately).

SB2318:
@muddlebee can you please assign me

SB2318:
> [@muddlebee](https://github.com/muddlebee) can you please assign me,

 due to heavy rain and Lighting Strike, I was unable to connect with you earlier.
Thanks for considering. 🙏

Please give me a few hours, and I will submit a demo video with my next PR.

muddlebee:
@SB2318 it wasn't about demo, but the APIs implemented dont even match incident.io APIs, so consider understanding the incident io APIs first and then implement. and only tag for reviews once end to end integration is complete

SB2318:
[...]

PR #1086, opened 2026-04-29
Ingest flow cleanup: Fix ingest duplication

### LINKED ISSUE #867: Collapse the current two-step ingest flow into one explicit helper
## Why This Matters
`generate_report()` calls `send_ingest()` twice: once to create the investigation and again to attach the URL. That makes the write path harder to follow and test.

## Read First
- `app/nodes/publish_findings/node.py`
- `app/utils/ingest_delivery.py`

## Scope
1. Introduce one helper with a clear name for the "create investigation, then attach URL if present" flow.
2. Move the current two-call sequence into that helper.
3. Do not change the API contract of `send_ingest()` in the same PR unless absolutely necessary.

## Tests / Coverage
- add direct tests for the new helper
- add one test where the first ingest succeeds and the second update fails
- add one test where no investigation ID is returned

## Acceptance Criteria
- `generate_report()` no longer contains the duplicated ingest sequence inline
- failure behavior stays the same
- the helper name clearly explains the two-step flow

SB2318:
Fixes #867 

This PR refactors the ingest logic in generate_report() by extracting the two-step flow (create investigation → attach URL) into a dedicated helper.
Previously, generate_report() directly called send_ingest() twice, which made the write path harder to follow and test. This change introduces a single helper that encapsulates that flow without modifying the existing API contract

### Key Changes
✔ generate_report() no longer has duplicate ingest logic
✔ Helper encapsulates the two-step flow clearly
✔ No change to send_ingest() contract
✔ Failure behavior unchanged

---

## **What this solves:**

- Eliminates duplicated two-step ingest logic  
- Improves readability and debuggability  

---

### Ingest Flow Refactor
- Extracted helper: `create_investigation_and_attach_url()`  
- Encapsulates:
  - create investigation (`send_ingest`)  
  - attach `investigation_url` (conditional second call)  
- Replaced inline logic in `generate_report()`  

---

## Key design choices:

- Kept `send_ingest()` unchanged (no API contract changes)  

---

## Behavior guarantees:

- Same ingest behavior and failure handling (logged, non-blocking)  

---

## Test coverage:

- Ingest helper:
  - success path  
  - second call failure  
  - missing investigation ID  

---

---


Ghraven:
Hey @SB2318, this is a clean refactor — moving `get_investigation_url` to live alongside `send_ingest` in `ingest_delivery.py` makes sense since they share the same `app.config` dependency, and removing it from `report.py` keeps the formatter layer focused on presentation.

One small thing I noticed in the new `create_investigation_and_attach_url` function — looking at the test:

```python
investigation_id, investigation_url = (
    ingest_delivery.create_investigation_and_attach_url(sample_state, "slack message")
)
```

It returns a tuple of `(str | None, str | None)`, which is fine, but the function signature in the diff seems to be cut off so I couldn't verify the return type annotation. If it's not there yet, a `-> tuple[str | None, str | None]` annotation would make the contract clear for callers.

Also just noticed the `cast` import was removed from `node.py` — was that intentional (i.e. there are no more `cast()` calls after the refactor)? If so, the unused import removal is a nice cleanup bonus.

Good work on this one — the test coverage for the new function looks solid too! 🙏

SB2318:
> Hey @SB2318, this is a clean refactor — moving `get_investigation_url` to live alongside `send_ingest` in `ingest_delivery.py` makes sense since they share the same `app.config` dependency, and removing it from `report.py` keeps the formatter layer focused on presentation.
> 
> One small thing I noticed in the new `create_investigation_and_attach_url` function — looking at the test:
> 
> ```python
> investigation_id, investigation_url = (
>     ingest_delivery.create_investigation_and_attach_url(sample_state, "slack message")
> )
> ```
> 
> It returns a tuple of `(str | None, str | None)`, which is fine, but the function signature in the diff seems to be cut off so I couldn't verify the return type annotation. If it's not there yet, a `-> tuple[str | None, str | None]` annotation would make the contract clear for callers.
> 
> Also just noticed the `cast` import was removed from `node.py` — was that intentional (i.e. there are no more `cast()` calls after the refactor)? If so, the unused import removal is a nice cleanup bonus.
> 
> Good work on this one — the test coverage for the new function looks solid too! 🙏

Thanks, I have some pending review tasks on these two PR.
Bot suggested me also to write unit test for the **create_investigation_and_data_attach_url** function.

But due natural calamity, I didnot fixed it yesterday.
please give me 1-2 hour, And please let me know any more suggestions from your side. 😇

SB2318:
@Ghraven, please review this PR, and let me know if it would be better to squash those commits, or you need any further changes.

I added corresponding test cases for the new methods and updated all usages to align with the project’s new send_ingest workflow changes.

yashksaini-coder (changes requested):
Ran locally: 65/65 pass, lint clean. Refactor logic is correct and the move of `get_investigation_url` into `ingest_delivery.py` makes sense — it shares the `app.config` dependency with `send_ingest` and has no business being in the formatters layer.

One functional regression though:

**`investigation_url` fallback is lost**

Old `node.py` always computed a URL even when ingest failed:
```python
investigation_url = get_investigation_url(state.get("organization_slug"), investigation_id)
# investigation_id=None → returns "{base}/investigations" (the list page)
```

The new helper only sets `investigation_url` when `investigation_id` is not None, so on first-ingest failure it returns `(None, None)`. The call site then does:
```python
build_action_blocks(investigation_url or "", investigation_id)
```

Slack's Block Kit rejects an empty string on the button's `url` field — the "View Details in Tracer" button will break whenever ingest is down or unreachable. The old behavior (fallback to the investigations list URL) should be preserved. Simplest fix is to always call `get_investigation_url` in the helper regardless of whether `investigation_id` is set:

```python
investigation_url = get_investigation_url(state.get("organization_slug"), investigation_id)
```

and remove the `investigation_url or ""` guard at the call site.

---

**Minor — `test_node.py` patches internals instead of the helper**

`test_masking_unmask.py` patches `create_investigation_and_attach_url` wholesale, which is the right approach for an integration-level test. `test_node.py` still patches `send_ingest` and `get_investigation_url` at the utility module level — it works but locks the test to the helper's internal structure. Worth aligning to the same strategy.

---

The refactor itself is clean and the three new tests in `test_ingest_delivery.py` cover the right paths. Fix the URL fallback and this is good to merge.

SB2318:
Okay, will do.

Ghraven:
Thanks for the updates @SB2318 — both points I raised look good now:

1. **Return type annotation** — `-> tuple[str | None, str | None]` is there on `create_investigation_and_attach_url`, type checkers can now infer the unpacked values correctly.

2. **`cast` removal** confirmed — the unused import is gone from `node.py`.

I also noticed `get_investigation_url` has been moved into `ingest_delivery.py` with the `investigation_id=None → returns list page` fallback preserved, which addresses the regression @yashksaini-coder flagged. That's the right call.

On squashing — I'd lean toward keeping the commits separate for now since the history is readable and the maintainers may want to see the progression. But that's ultimately muddlebee's call.

Looks solid to me! 🙏

SB2318:
> Ran locally: 65/65 pass, lint clean. Refactor logic is correct and the move of `get_investigation_url` into `ingest_delivery.py` makes sense — it shares the `app.config` dependency with `send_ingest` and has no business being in the formatters layer.
> 
> One functional regression though:
> 
> **`investigation_url` fallback is lost**
> 
> Old `node.py` always computed a URL even when ingest failed:
> 
> ```python
> investigation_url = get_investigation_url(state.get("organization_slug"), investigation_id)
> # investigation_id=None → returns "{base}/investigations" (the list page)
> ```
> 
> The new helper only sets `investigation_url` when `investigation_id` is not None, so on first-ingest failure it returns `(None, None)`. The call site then does:
> 
> ```python
> build_action_blocks(investigation_url or "", investigation_id)
> ```
> 
> Slack's Block Kit rejects an empty string on the button's `url` field — the "View Details in Tracer" button will break whenever ingest is down or unreachable. The old behavior (fallback to the investigations list URL) should be preserved. Simplest fix is to always call `get_investigation_url` in the helper regardless of whether `investigation_id` is set:
> 
> ```python
> investigation_url = get_investigation_url(state.get("organization_slug"), investigation_id)
> ```
> 
> and remove the `investigation_url or ""` guard at the call site.
> 
> **Minor — `test_node.py` patches internals instead of the helper**
> 
> `test_masking_unmask.py` patches `create_investigation_and_attach_url` wholesale, which is the right approach for an integration-level test. `test_node.py` still patches `send_ingest` and `get_investigation_url` at the utility module level — it works but locks the test to the helper's internal structure. Worth aligning to the same strategy.
> 
> The refactor itself is clean and the three new tests in `test_ingest_delivery.py` cover the right paths. Fix the URL fallback and this is good to merge.

With empty string removal, your test is falling.

````
Run python -m mypy app/
app/nodes/publish_findings/node.py:42: error: Argument 1 to "build_action_blocks" has incompatible type "str | None"; expected "str"  [arg-type]
Found 1 error in 1 file (checked 424 source files)
Error: Process completed with exit code 1.

```

SB2318:
@yashksaini-coder Please check my comment on these two issues.
#317 and #481

muddlebee:
@SB2318 pls fix CI

SB2318:
> @SB2318 pls fix CI

okay

PR #1090, opened 2026-04-29
fix(cli): `opensre tests` crashes with FileNotFoundError in bundled binary (#1078)

### LINKED ISSUE #1078: FileNotFoundError: missing 'tests/synthetic/rds_postgres' directory when running 'opensre tests'
### Description

Running `opensre tests` crashes with a `FileNotFoundError` because the command attempts to iterate over a directory (`tests/synthetic/rds_postgres`) that does not exist inside the PyInstaller-bundled temporary extraction path.

### Steps to Reproduce

```
opensre tests
```

### Expected Behavior

The command should either:
- include the `tests/synthetic/rds_postgres` directory in the packaged binary, or
- gracefully handle the case where the directory is absent (skip it / surface a clear error message).

### Actual Behavior

```
  File "__main__.py", line 86, in <module>
  File "__main__.py", line 72, in main
  File "click/core.py", line 1514, in __call__
  File "click/core.py", line 1435, in main
  File "click/core.py", line 1902, in invoke
  File "click/core.py", line 1880, in invoke
  File "click/core.py", line 1298, in invoke
  File "click/core.py", line 853, in invoke
  File "click/decorators.py", line 34, in new_func
  File "app/cli/commands/tests.py", line 84, in tests
  File "app/cli/tests/discover.py", line 343, in load_test_catalog
  File "app/cli/tests/discover.py", line 338, in discover_cli_commands
  File "app/cli/tests/discover.py", line 304, in _discover_rds_synthetic_scenarios
  File "pathlib/_local.py", line 575, in iterdir
FileNotFoundError: [Errno 2] No such file or directory: '/private/var/folders/7j/922qdpd96rs1z67fwj468ydm0000gn/T/_MEIM63iIN/tests/synthetic/rds_postgres'
[PYI-34740:ERROR] Failed to execute script '__main__' due to unhandled exception!
```

### Root Cause (likely)

`_discover_rds_synthetic_scenarios` in `app/cli/tests/discover.py` (line 304) calls `Path(...).iterdir()` on `tests/synthetic/rds_postgres` without checking whether the path exists first. When the binary is packaged with PyInstaller the directory is not bundled, so the path is missing at runtime.

### Suggested Fix

Either bundle the directory via the PyInstaller spec, or add an existence guard before iterating:

```python
rds_path = ...  # current path construction
if rds_path.exists():
    for entry in rds_path.iterdir():
        ...
```

### Environment

- OS: macOS (Darwin)
- Binary: PyInstaller-packaged `opensre` CLI
- Temp extraction path: `/private/var/folders/7j/922qdpd96rs1z67fwj468ydm0000gn/T/_MEIM63iIN/`

mayankbharati-ops:
Closes #1078.

## Summary

`opensre tests` and `opensre tests list` crash with `FileNotFoundError` when run from the PyInstaller-bundled binary (the released binary, not source-checkout runs). The reported crash is one of three of the same class — fixed all three.

## Root cause

[`packaging/opensre.spec`](packaging/opensre.spec) only collects `app/` data files (`collect_data_files("app")`) and explicitly filters `tests` out of the submodule list (`_is_runtime_submodule`). At runtime the bundled temp extraction path therefore has no `tests/synthetic/rds_postgres/`, no `tests/e2e/rca/`, and no `Makefile`. Three call sites in the test-discovery code path read directly from those paths without an existence guard:

| Location | Operation that crashes | Crash in bundled binary? |
|---|---|---|
| `_discover_rds_synthetic_scenarios` | `scenarios_dir.iterdir()` | ✅ **the reported #1078 crash** |
| `discover_make_targets` | `MAKEFILE_PATH.read_text()` | ✅ identical class of bug |
| `run_synthetic_suite` | `from tests.synthetic.rds_postgres.run_suite import …` | ✅ `ModuleNotFoundError` traceback |
| `discover_rca_files` | `RCA_DIR.glob("*.md")` | ❌ safe today (`Path.glob` returns empty), guard added for documentation + forward compatibility |

The reported crash was introduced by [`9a1b5c3d` _"add synthetic RDS postgres RCA suite (#194)"_](https://github.com/Tracer-Cloud/opensre/commit/9a1b5c3d) on 2026-04-01 (answers @muddlebee's regression-commit question).

## Fix

```python
# app/cli/tests/discover.py
def _discover_rds_synthetic_scenarios() -> list[TestCatalogItem]:
    scenarios_dir = REPO_ROOT / "tests" / "synthetic" / "rds_postgres"
    items: list[TestCatalogItem] = []
    if not scenarios_dir.is_dir():
        return items   # PyInstaller bundle has no tests/ tree
    ...

def discover_make_targets() -> list[TestCatalogItem]:
    if not MAKEFILE_PATH.is_file():
        return []   # PyInstaller bundle has no Makefile
    ...

def discover_rca_files() -> list[TestCatalogItem]:
    if not RCA_DIR.is_dir():
        return items   # explicit guard, currently a no-op on CPython
    ...
```

```python
# app/cli/commands/tests.py
def run_synthetic_suite(...):
    try:
        from tests.synthetic.rds_postgres.run_suite import main as run_suite_main
    except ModuleNotFoundError as exc:
        raise OpenSREError(
            "The synthetic RDS PostgreSQL suite is not available in this build.",
            suggestion="Run from a source checkout (clone the repo) — synthetic scenarios are not bundled.",
        ) from exc
    ...
```

`is_dir()` / `is_file()` return `False` on a missing path without raising, so they're cleaner than `exists()` followed by a separate type check.

## Test coverage

**`tests/cli/test_discover.py` (+5 tests, all in a new `TestDiscoverGracefulOnMissingSource` class):**
- `test_rds_synthetic_returns_empty_when_dir_missing` — the literal reported crash
- `test_make_targets_returns_empty_when_makefile_missing` — same-class regression
- `test_rca_files_returns_empty_when_dir_missing` — covers the explicit guard
- `test_load_test_catalog_does_not_crash_with_no_sources` — full degradation contract
- `test_rds_synthetic_still_discovers_when_dir_present` — sanity that the guard doesn't break the source-checkout path

**`tests/cli/test_cli_inventory.py` (+1 test):**
- `test_tests_synthetic_surfaces_clean_error_when_suite_not_bundled` — patches `builtins.__import__` to deny `tests.synthetic.*`, asserts the CLI exits non-zero with a clean message and no `Traceback` / `ModuleNotFoundError` reaches the user

## Live e2e verification

Wrote a probe that monkeypatches `REPO_ROOT`, `MAKEFILE_PATH`, and `RCA_DIR` to an empty tmp dir (the exact same file layout PyInstaller produces — no `tests/`, no `Makefile`, no `tests/e2e/rca`) and invokes the real Click CLI in-process. **17 / 17 live checks pass:**

| | Baseline (source tree) | Bundled-binary sim (empty REPO_ROOT) |
|---|---|---|
| `opensre tests list` | exit 0, full catalog | exit 0, empty output |
| `opensre --json tests list` | valid JSON, full catalog | valid JSON, empty `[]` |
| `opensre tests list --category synthetic` | finds `synthetic:001-replication-lag` | exit 0, empty (degraded) |
| `opensre tests synthetic --scenario X` | runs the suite | exits non-zero with `OpenSREError` and actionable suggestion — no `ModuleNotFoundError` / `Traceback` leaks |
| `load_test_catalog()` direct call | populated catalog | empty catalog, no exception |

User-facing degraded message:
```
Error: The synthetic RDS PostgreSQL suite is not available in this build.

Suggestion: Run 'opensre tests synthetic' from a source checkout
(clone https://github.com/Tracer-Cloud/opensre) — the synthetic
scenarios under 'tests/synthetic/rds_postgres/' are not bundled
into the released binary.
```

## Verification — final state

- `pytest tests/cli/test_discover.py tests/cli/test_cli_inventory.py`: **25 / 25 pass** (3 pre-existing + 22 including new)
- Full unit suite: **4151 pass / 2 skipped / 1 xfailed** (no regressions)
- `ruff check app/ tests/`: clean
- `ruff format --check app/ tests/`: clean
- `mypy app/cli/tests/ app/cli/commands/`: clean on the touched modules; the 3 errors mypy reports are pre-existing missing-stub warnings for `pymysql` in `app/integrations/` — unrelated to this PR

## Scope

| File | Change |
|---|---|
| `app/cli/tests/discover.py` | +14 / -1 (3 existence guards + docstring update) |
| `app/cli/commands/tests.py` | +14 / -1 (try/except around the synthetic-suite import) |
| `tests/cli/test_discover.py` | +85 / -1 (5 new regression tests) |
| `tests/cli/test_cli_inventory.py` | +39 / -0 (1 new bundled-binary CLI test) |

No public API changes. No state-shape changes. No behaviour change for source-checkout users — the existing source-tree happy path is verified by both the existing test `test_load_test_catalog_includes_make_targets_and_rca_fixtures` and the new sanity test `test_rds_synthetic_still_discovers_when_dir_present`.

### DIFF
--- a/app/cli/commands/tests.py
+++ b/app/cli/commands/tests.py
@@ -105,22 +105,36 @@ def tests(ctx: click.Context) -> None:
 )
 def run_synthetic_suite(scenario: str, output_json: bool, mock_grafana: bool) -> None:
     """Run the synthetic RDS PostgreSQL RCA benchmark."""
+    # ``packaging/opensre.spec`` only collects ``app/`` data files, so neither
+    # the synthetic Python package's submodules nor the per-scenario data
+    # directories are reliably present in PyInstaller bundles. Two failure
+    # modes can trip a bundled binary here:
+    #
+    # 1. The ``tests.synthetic.rds_postgres.*`` Python package is missing
+    #    entirely  →  ``ModuleNotFoundError`` raised at import time.
+    # 2. The package is included transitively but its data dir
+    #    (``tests/synthetic/rds_postgres/<scenario>/``) is absent
+    #    →  ``run_suite`` crashes later with ``FileNotFoundError`` from
+    #    ``Path.iterdir()`` inside the scenario loader.
+    #
+    # We pre-check the data dir explicitly *and* catch a narrow
+    # ``ModuleNotFoundError`` so users see one structured message regardless
+    # of which failure mode their bundle produces.
+    from app.cli.tests.discover import REPO_ROOT
+
+    scenarios_dir = REPO_ROOT / "tests" / "synthetic" / "rds_postgres"
+    if not scenarios_dir.is_dir():
+        raise _synthetic_suite_not_bundled_error()
+
     try:
-        # ``packaging/opensre.spec`` excludes the ``tests`` tree from
-        # PyInstaller bundles, so this import will raise ``ModuleNotFoundError``
-        # in a packaged binary. Surface a clear error rather than a raw
-        # traceback so users know to run from a source checkout.
         from tests.synthetic.rds_postgres.run_suite import main as run_suite_main
     except ModuleNotFoundError as exc:
-        raise OpenSREError(
-            "The synthetic RDS PostgreSQL suite is not available in this build.",
-            suggestion=(
-                "Run 'opensre tests synthetic' from a source checkout (clone "
-                "https://github.com/Tracer-Cloud/opensre) — the synthetic "
-                "scenarios under 'tests/synthetic/rds_postgres/' are not "
-                "bundled into the released binary."
-            ),
-        ) from exc
+        # Narrow to the actual missing-bundle case; re-raise unrelated import
+        # failures (e.g. a missing transitive dep like ``psycopg``) so users
+        # see the real cause instead of a misleading "not bundled" message.
+        if exc.name is None or not exc.name.startswith("tests.synthetic.rds_postgres"):
+            raise
+        raise _synthetic_suite_not_bundled_error() from exc
 
     capture_test_synthetic_started(scenario or "all", mock_grafana=mock_grafana)
     raise SystemExit(
@@ -131,6 +145,19 @@ def run_synthetic_suite(scenario: str, output_json: bool, mock_grafana: bool) ->
                 mock_grafana=mock_grafana,
             )
         )
+    )
+
+
+def _synthetic_suite_not_bundled_error() -> OpenSREError:
+    """Structured error for ``opensre tests synthetic`` when the suite isn't shipped."""
+    return OpenSREError(
+        "The synthetic RDS PostgreSQL suite is not available in this build.",
+        suggestion=(
+            "Pre-built binaries do not bundle the per-scenario data files "
+            "under 'tests/synthetic/rds_postgres/'. Install from source "
+            "(`git clone https://github.com/Tracer-Cloud/opensre && pip "
+            "install -e .`) and re-run 'opensre tests synthetic'."
+        ),
     )
 
 
--- a/app/cli/commands/tests.py
+++ b/app/cli/commands/tests.py
@@ -91,6 +91,19 @@ def tests(ctx: click.Context) -> None:
     raise SystemExit(exit_code)
 
 
+def _synthetic_suite_not_bundled_error() -> OpenSREError:
+    """Structured error for ``opensre tests synthetic`` when the suite isn't shipped."""
+    return OpenSREError(
+        "The synthetic RDS PostgreSQL suite is not available in this build.",
+        suggestion=(
+            "Pre-built binaries do not bundle the per-scenario data files "
+            "under 'tests/synthetic/rds_postgres/'. Install from source "
+            "(`git clone https://github.com/Tracer-Cloud/opensre && pip "
+            "install -e .`) and re-run 'opensre tests synthetic'."
+        ),
+    )
+
+
 @tests.command(name="synthetic")
 @click.option(
     "--scenario", default="", help="Pin to a single scenario directory, e.g. 001-replication-lag."
@@ -119,11 +132,12 @@ def run_synthetic_suite(scenario: str, output_json: bool, mock_grafana: bool) ->
     #
     # We pre-check the data dir explicitly *and* catch a narrow
     # ``ModuleNotFoundError`` so users see one structured message regardless
-    # of which failure mode their bundle produces.
-    from app.cli.tests.discover import REPO_ROOT
+    # of which failure mode their bundle produces. The data-dir path is the
+    # ``SYNTHETIC_SCENARIOS_DIR`` constant from ``discover.py`` — single
+    # source of truth shared with ``_discover_rds_synthetic_scenarios``.
+    from app.cli.tests.discover import SYNTHETIC_SCENARIOS_DIR
 
-    scenarios_dir = REPO_ROOT / "tests" / "synthetic" / "rds_postgres"
-    if not scenarios_dir.is_dir():
+    if not SYNTHETIC_SCENARIOS_DIR.is_dir():
         raise _synthetic_suite_not_bundled_error()
 
     try:
@@ -145,19 +159,6 @@ def run_synthetic_suite(scenario: str, output_json: bool, mock_grafana: bool) ->
                 mock_grafana=mock_grafana,
             )
         )
-    )
-
-
-def _synthetic_suite_not_bundled_error() -> OpenSREError:
-    """Structured error for ``opensre tests synthetic`` when the suite isn't shipped."""
-    return OpenSREError(
-        "The synthetic RDS PostgreSQL suite is not available in this build.",
-        suggestion=(
-            "Pre-built binaries do not bundle the per-scenario data files "
-            "under 'tests/synthetic/rds_postgres/'. Install from source "
-            "(`git clone https://github.com/Tracer-Cloud/opensre && pip "
-            "install -e .`) and re-run 'opensre tests synthetic'."
-        ),
     )
 
 
--- a/app/cli/commands/tests.py
+++ b/app/cli/commands/tests.py
@@ -105,7 +105,22 @@ def tests(ctx: click.Context) -> None:
 )
 def run_synthetic_suite(scenario: str, output_json: bool, mock_grafana: bool) -> None:
     """Run the synthetic RDS PostgreSQL RCA benchmark."""
-    from tests.synthetic.rds_postgres.run_suite import main as run_suite_main
+    try:
+        # ``packaging/opensre.spec`` excludes the ``tests`` tree from
+        # PyInstaller bundles, so this import will raise ``ModuleNotFoundError``
+        # in a packaged binary. Surface a clear error rather than a raw
+        # traceback so users know to run from a source checkout.
+        from tests.synthetic.rds_postgres.run_suite import main as run_suite_main
+    except ModuleNotFoundError as exc:
+        raise OpenSREError(
+            "The synthetic RDS PostgreSQL suite is not available in this build.",
+            suggestion=(
+                "Run 'opensre tests synthetic' from a source checkout (clone "
+                "https://github.com/Tracer-Cloud/opensre) — the synthetic "
+                "scenarios under 'tests/synthetic/rds_postgres/' are not "
+                "bundled into the released binary."
+            ),
+        ) from exc
 
     capture_test_synthetic_started(scenario or "all", mock_grafana=mock_grafana)
     raise SystemExit(
--- a/tests/cli/test_cli_inventory.py
+++ b/tests/cli/test_cli_inventory.py
@@ -246,9 +246,10 @@ def test_tests_synthetic_clean_error_when_data_dir_missing(tmp_path: Path) -> No
     on the data dir must surface a structured error before the import fires."""
     runner = CliRunner()
 
-    # tmp_path has no ``tests/synthetic/rds_postgres`` subtree, so the
+    # Point SYNTHETIC_SCENARIOS_DIR at a path that doesn't exist so the
     # pre-check in ``run_synthetic_suite`` short-circuits to OpenSREError.
-    with unittest.mock.patch("app.cli.tests.discover.REPO_ROOT", tmp_path):
+    missing = tmp_path / "missing-rds-postgres"
+    with unittest.mock.patch("app.cli.tests.discover.SYNTHETIC_SCENARIOS_DIR", missing):
         result = runner.invoke(cli, ["tests", "synthetic", "--scenario", "001-replication-lag"])
 
     output = result.output or ""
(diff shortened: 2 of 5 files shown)


mayankbharati-ops:
**Update — pushed `3136509f` after a much deeper round of investigation.**

I ran 4 parallel review/audit passes (codebase audit, real PyInstaller build verification, test-coverage gap analysis, independent code review) and the **build verifier caught a P0 my first commit missed**: `opensre tests synthetic` was *still crashing in the actual bundled binary*.

## What deeper testing revealed

1. **The previous fix was incomplete.** PyInstaller's static analyzer follows the literal `from tests.synthetic.rds_postgres.run_suite import main` statement in `app/cli/commands/tests.py` and bundles the Python package transitively, *regardless* of the spec's `collect_submodules(filter=…)` exclusion. So in the real bundled binary:
    - The Python module IS bundled
    - The per-scenario data files (`tests/synthetic/rds_postgres/<scenario>/`) are NOT bundled
    - `run_suite.main` crashed inside the scenario loader with `FileNotFoundError` from `iterdir()`
    - My `try/except ModuleNotFoundError` did not catch it

2. **An audit found the same class of bug elsewhere.** `app/cli/commands/deploy.py:323,329` — `opensre deploy ec2` and `opensre deploy ec2 --down` directly import from `tests.deployment.ec2.infrastructure_sdk` without a guard, same pattern.

3. **The `ModuleNotFoundError` catch was too broad.** A missing transitive dep (e.g. `psycopg` if a user has the synthetic package but not the database driver) would have produced a misleading "not bundled" error.

## What changed in `3136509f`

- `run_synthetic_suite` now **pre-checks the data dir** before the import — catches the actual bundled-binary failure mode the build verifier found.
- Both `ModuleNotFoundError` catches **narrow on `exc.name`** so unrelated missing deps bubble up as the real cause instead of being mis-reported.
- `deploy_ec2` (both branches) wrapped with the same structured `OpenSREError` pattern.
- Suggestion text changed from "clone the repo" → `pip install -e .` for pipx/Homebrew users.
- Test helper `_patch_discover_paths` to dedupe `monkeypatch.setattr("app.cli.tests.discover.X", ...)` repetition (per @muddlebee's prior fixture-dedup nit on PR #952).

## Live PyInstaller bundle verification (rebuilt `dist/opensre`)

| Command | Result |
|---|---|
| `./dist/opensre tests list` | exit 0, empty stdout ✅ |
| `./dist/opensre --json tests list` | exit 0, `[]` ✅ |
| `./dist/opensre tests synthetic --scenario 001-replication-lag` | exit 1, **structured error** `"The synthetic RDS PostgreSQL suite is not available in this build"` + `pip install -e .` suggestion. No `Traceback` / `FileNotFoundError` / `ModuleNotFoundError` reaches the user. ✅ |

This is the failure mode the previous commit missed — confirmed fixed in the actual binary.

## Test additions (+9 new tests, all passing)

- Empty `rds_postgres/` dir → returns `[]`
- `__pycache__` / `_template` / stray top-level files → skipped correctly
- Happy `scenario.yml` → enriches display name with `[failure_mode]`
- Malformed `scenario.yml` → falls back to bare ID (no crash)
- Missing data dir → structured error with `pip install -e .` suggestion (the bundled-binary case)
- `psycopg` transitive missing → real cause shown, not "not bundled"
- `deploy ec2` + `deploy ec2 --down` (parametrized) → structured error
- `boto3` transitive missing on `deploy ec2` → real cause shown

## Final state

- `pytest tests/cli/test_discover.py tests/cli/test_cli_inventory.py` — **34 / 34 pass** (was 25)
- Full unit suite — **4160 pass / 2 skipped / 1 xfailed** (was 4151, +9 new)
- `ruff check`, `ruff format --check` clean
- `mypy app/cli/...` clean on touched modules

## Follow-up worth tracking

The audit also surfaced that the EC2 deploy guard is *defense-in-depth* in the current `packaging/opensre.spec` configuration — PyInstaller bundles `tests.deployment.ec2.*` because of the literal `from tests…` import, so the import succeeds and the guard's `except` branch is unreachable from a fresh binary. The guard is still correct (and unit-tested for the case it does catch — a non-standard install missing those modules), but the "right" long-term fix is to also bundle the data files for the synthetic suite via the spec, so `opensre tests synthetic` can actually run from a binary instead of always degrading. Happy to file that as a follow-up issue if interest.

Ghraven:
Hey @mayankbharati-ops, this is a really thorough fix! I was looking at the same issue (#1078) and you've gone much deeper than just the `rds_postgres` guard — catching the `ModuleNotFoundError` at the `run_synthetic_suite` level and guarding `discover_make_targets` and `discover_rca_files` too is the right approach for a packaged binary.

One small thing I noticed in `tests/cli/test_discover.py` — the new tests use `tmp_path: object` as the type annotation in a couple of places (e.g. `test_discover_rds_scenarios_missing_dir`). Since `tmp_path` is a `pytest.fixture` it should be `pathlib.Path`, not `object`, both for accuracy and so type checkers can infer its methods. Might be worth a quick update there.

Also just curious — the `_ec2_deploy_not_bundled_error` helper in `deploy.py` is a nice touch. Is the plan to use it as the error surfaced in both `ec2` and `ec2 --down` paths? Looks like both call sites would benefit from it.

Really solid work on this one — the test coverage for the degradation paths is especially useful. 🙏

mayankbharati-ops:
@Ghraven thanks for the close read! Both points addressed:

**1. `tmp_path` typing** — fixed in `440111cf`. Three tests in `tests/cli/test_cli_inventory.py` (`test_tests_synthetic_clean_error_when_data_dir_missing`, `test_tests_synthetic_clean_error_when_module_not_bundled`, `test_tests_synthetic_unrelated_module_not_found_propagates`) had `tmp_path: object` or no annotation. Tightened all three to `tmp_path: Path` with the `from pathlib import Path` import. You're right that `object` was wrong — type-checkers can't infer `.mkdir()` / `/` on `tmp_path` without the proper annotation, and the consistency with `tests/cli/test_discover.py` matters.

**2. `_ec2_deploy_not_bundled_error` reuse** — both call sites in `app/cli/commands/deploy.py` already route through it:

```python
# line 351 — destroy path
if exc.name is None or not exc.name.startswith("tests.deployment.ec2"):
    raise
raise _ec2_deploy_not_bundled_error() from exc

# line 363 — deploy path  
if exc.name is None or not exc.name.startswith("tests.deployment.ec2"):
    raise
raise _ec2_deploy_not_bundled_error() from exc
```

Same `OpenSREError` instance gets surfaced from both `opensre deploy ec2` and `opensre deploy ec2 --down`, with the message in one place. The narrow `exc.name.startswith("tests.deployment.ec2")` check on each side keeps unrelated transitive missing deps (e.g. `boto3`) from getting mis-reported as "EC2 not bundled" — pinned by `test_deploy_ec2_unrelated_module_not_found_propagates`.

Tests still 34/34 green, lint/format clean, full suite 4160 pass. Appreciate the eyeballs! 🙏

app/cli/commands/tests.py
```diff
@@ -105,7 +105,36 @@ def tests(ctx: click.Context) -> None:
...
+    #    (``tests/synthetic/rds_postgres/<scenario>/``) is absent
+    #    →  ``run_suite`` crashes later with ``FileNotFoundError`` from
+    #    ``Path.iterdir()`` inside the scenario loader.
+    #
+    # We pre-check the data dir explicitly *and* catch a narrow
+    # ``ModuleNotFoundError`` so users see one structured message regardless
+    # of which failure mode their bundle produces.
+    from app.cli.tests.discover import REPO_ROOT
+
+    scenarios_dir = REPO_ROOT / "tests" / "synthetic" / "rds_postgres"
```
muddlebee: Path duplicated from `discover.py:320` — both construct `REPO_ROOT / "tests" / "synthetic" / "rds_postgres"` independently. Extract a `SYNTHETIC_SCENARIOS_DIR` constant in `discover.py` alongside `MAKEFILE_PATH` and `RCA_DIR`, then import it here. One place to update if the layout ever changes.

app/cli/commands/tests.py
```diff
@@ -119,6 +148,19 @@ def run_synthetic_suite(scenario: str, output_json: bool, mock_grafana: bool) ->
     )
 
 
+def _synthetic_suite_not_bundled_error() -> OpenSREError:
```
muddlebee: Helper is defined after its two call sites (lines 127 and 137). Move `_synthetic_suite_not_bundled_error` above `run_synthetic_suite` so readers see the helper before the code that uses it — consistent with how `_ec2_deploy_not_bundled_error` is ordered in `deploy.py`.

tests/cli/test_cli_inventory.py
```diff
@@ -225,3 +226,156 @@ def test_run_catalog_items_skips_non_runnable_and_prints_message(
...
+
+    output = result.output or ""
+    assert result.exit_code == 1, f"unexpected exit code {result.exit_code}; output={output!r}"
+    assert "EC2 deployment is not available" in output
+    assert "pip install -e ." in output
+    assert "ModuleNotFoundError" not in output
+    assert "Traceback" not in output
+
+
+def test_deploy_ec2_unrelated_module_not_found_propagates() -> None:
```
muddlebee: Narrow-catch contract is only verified for `--down` (destroy). The non-`--down` deploy path has its own try/except block in `deploy_ec2` but no parallel test. Add a variant that patches `tests.deployment.ec2.infrastructure_sdk.deploy_remote` with an unrelated `ModuleNotFoundError(name='boto3')` and asserts the 'EC2 deployment is not available' message is absent.

mayankbharati-ops:
@muddlebee thanks for the close review — pushed `1e4dcaef` addressing all three nits:

**1. Path duplication (`app/cli/commands/tests.py:125`)** — Extracted `SYNTHETIC_SCENARIOS_DIR` constant in `app/cli/tests/discover.py` alongside `MAKEFILE_PATH` and `RCA_DIR`. Both call sites (`_discover_rds_synthetic_scenarios` and `run_synthetic_suite`) now import the same constant. Single source of truth if the layout ever moves.

**2. Helper-after-call-site (`app/cli/commands/tests.py:151`)** — Moved `_synthetic_suite_not_bundled_error` above `run_synthetic_suite`. Now matches the order of `_ec2_deploy_not_bundled_error` ↔ `deploy_ec2` in `deploy.py`.

**3. `deploy_remote` test gap (`tests/cli/test_cli_inventory.py:364`)** — Parametrized `test_deploy_ec2_unrelated_module_not_found_propagates` over both `(argv, patched_module)` so the narrow-catch contract is independently exercised on both branches:

```python
@pytest.mark.parametrize(
    ("argv", "patched_module"),
    [
        (["deploy", "ec2", "--down"], "tests.deployment.ec2.infrastructure_sdk.destroy_remote"),
        (["deploy", "ec2"],            "tests.deployment.ec2.infrastructure_sdk.deploy_remote"),
    ],
)
```

A regression in either `deploy_ec2`'s try/except is now caught independently.

**Test plumbing detail:** the new `SYNTHETIC_SCENARIOS_DIR` constant is bound at module load, so monkeypatching `REPO_ROOT` after import no longer affects it (correctly). I extended the `_patch_discover_paths` helper with a `synthetic_dir` kwarg so tests patch the new constant directly, matching how they already patch `MAKEFILE_PATH` / `RCA_DIR`.

**Verification:**
- `pytest tests/cli/test_discover.py tests/cli/test_cli_inventory.py` — **35/35 pass** (was 34, +1 from the new `deploy_remote` parametrize).
- Full unit suite — **4161 / 2 skipped / 1 xfailed** (was 4160).
- `ruff check`, `ruff format --check` clean.

muddlebee:
@mayankbharati-ops this is really nice, extensive error handling and path checks. :+1:

PR #1092, opened 2026-04-29
fix(repl): Ctrl+C once cancels input, twice exits the shell (#1091)

### LINKED ISSUE #1091: Improve the interactive opensre shell behavior so that pressing Ctrl+C twice exits the shell, similar to Claude CLI.
Currently, Ctrl+C appears to interrupt the active prompt/session but does not provide an easy repeated-interrupt exit flow. Users should be able to press Ctrl+C once to cancel the current input or operation, then press Ctrl+C again to exit the shell cleanly.

Image

Claude cli

Image

Ghraven:
## Summary

Closes #1091

Adds a double-Ctrl+C exit flow to the interactive REPL, matching the UX of Claude CLI and other common shells.

## Behaviour before this PR

A single `KeyboardInterrupt` in `_run_one_turn` returned `False`, immediately exiting the shell — even when the user just wanted to cancel the current input line.

## Behaviour after this PR

- **First Ctrl+C** — cancels the current input line and prints `(Press Ctrl+C again to exit.)`, then continues the loop.
- **Second Ctrl+C** (without any successful input in between) — exits the shell cleanly.
- Successful input resets the interrupt counter, so Ctrl+C always needs two consecutive presses to exit.

## Implementation

Added a mutable `_interrupt_count: list[int]` counter (a one-element list so it can be mutated inside the async function) passed to `_run_one_turn`. The `EOFError` path is unchanged — it still exits immediately on the first press.

### DIFF
--- a/app/cli/repl/loop.py
+++ b/app/cli/repl/loop.py
@@ -1,0 +1,144 @@
+"""Async REPL loop — the zero-exit heart of the OpenSRE interactive terminal."""
+
+from __future__ import annotations
+
+import asyncio
+import sys
+
+from prompt_toolkit import PromptSession
+from prompt_toolkit.formatted_text import ANSI
+from rich.console import Console
+from rich.markup import escape
+
+from app.cli.repl.banner import render_banner
+from app.cli.repl.commands import dispatch_slash
+from app.cli.repl.config import ReplConfig
+from app.cli.repl.follow_up import answer_follow_up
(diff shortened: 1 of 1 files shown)


yashksaini-coder:
Closing this as due to merge conflicts and not assigned properly

muddlebee:
sorry I was closing this too, re-opened by mistake

PR #1095, opened 2026-04-30
fix(root-cause): recognize healthy observability evidence

### LINKED ISSUE #670: [BUG] is_clearly_healthy short-circuit ignores Alertmanager, Coralogix, and Honeycomb evidence keys
### Summary

`_INVESTIGATED_EVIDENCE_KEYS` in [`app/nodes/root_cause_diagnosis/evidence_checker.py#L16-L36`](https://github.com/Tracer-Cloud/opensre/blob/main/app/nodes/root_cause_diagnosis/evidence_checker.py#L16-L36) is missing five keys that the investigation merge step actually writes to the evidence dict: `alertmanager_alerts`, `alertmanager_silences`, `coralogix_logs`, `coralogix_error_logs`, `honeycomb_traces` (see the `_map_alertmanager_*` / `_map_coralogix_logs` / `_map_honeycomb_traces` entries in [`app/nodes/investigate/processing/post_process.py`](https://github.com/Tracer-Cloud/opensre/blob/main/app/nodes/investigate/processing/post_process.py)).

Condition 4 of `is_clearly_healthy()` therefore never fires for pure-Alertmanager, pure-Coralogix, or pure-Honeycomb healthy states. Same drift bug that #582 fixed for EKS.

### Impact

Every resolved low-severity alert from these three stacks pays a full LLM RCA round-trip instead of taking the fast path. Cost + latency, not correctness.

### Reproduction

```bash
python3 -c "
from app.nodes.root_cause_diagnosis.evidence_checker import is_clearly_healthy, _INVESTIGATED_EVIDENCE_KEYS
missing = [k for k in ('alertmanager_alerts','alertmanager_silences','coralogix_logs','coralogix_error_logs','honeycomb_traces') if k not in _INVESTIGATED_EVIDENCE_KEYS]
print('missing:', missing)

alert = {'state':'resolved','commonLabels':{'severity':'info'},'commonAnnotations':{}}
for key in ('alertmanager_alerts','coralogix_logs','honeycomb_traces','grafana_logs'):
    print(f'  {key:25s} short-circuit -> {is_clearly_healthy(alert, {key: []})}')
"
```

Fix PR incoming.

yas789:
Fixes #670

- add `alertmanager_alerts`, `alertmanager_silences`, `coralogix_logs`, `coralogix_error_logs`, and `honeycomb_traces` to `INVESTIGATED_EVIDENCE_KEYS`
- add regression coverage proving these shipped observability sources now satisfy `is_clearly_healthy()` when a resolved low-severity alert has empty-but-present investigation results
- add a safety regression showing critical severity still does not short-circuit with the new keys present

Terminal proof:

```text
Before the fix, the new regression tests failed for all five keys because
is_clearly_healthy(...) returned False.

After the fix:
- tests/nodes/root_cause_diagnosis/test_evidence_checker.py: 32 passed
- tests/nodes/root_cause_diagnosis/test_healthy_short_circuit.py: 162 passed
```

---

`is_clearly_healthy()` uses `INVESTIGATED_EVIDENCE_KEYS` as the gate for “a healthy investigation really happened”. The bug was that Alertmanager, Coralogix, and Honeycomb outputs were already written into evidence by the investigation merge step, but those five primary evidence keys were missing from that gate. As a result, pure healthy investigations from those sources could miss the deterministic healthy short-circuit even when the source had been queried successfully.

I kept the fix intentionally small rather than switching the gate to `CLAIM_EVIDENCE_KEYS`, because `CLAIM_EVIDENCE_KEYS` is broader and includes adjacent data that should not automatically count as proof that a healthy investigation was completed. Adding the five missing primary evidence keys preserves the existing design: empty-but-present investigation results still count as “we checked and found nothing bad”, while severity and annotation safeguards remain unchanged.

Validation I ran locally:
- `.venv/bin/python -m pytest tests/nodes/root_cause_diagnosis/test_evidence_checker.py -vv`
- `.venv/bin/python -m pytest tests/nodes/root_cause_diagnosis/test_healthy_short_circuit.py -vv`
- `make lint`
- `make format-check`
- `make typecheck`

Note: `make test-cov` on the latest upstream main currently hits one unrelated suite-order failure in `tests/services/test_honeycomb_client.py::test_create_query_success` due extra PostHog calls; that test passes in isolation and is outside this PR’s scope.

---

---

### DIFF
--- a/app/nodes/root_cause_diagnosis/evidence_checker.py
+++ b/app/nodes/root_cause_diagnosis/evidence_checker.py
@@ -25,6 +25,11 @@ INVESTIGATED_EVIDENCE_KEYS = frozenset(
         "datadog_logs",
         "datadog_monitors",
         "betterstack_logs",
+        "honeycomb_traces",
+        "coralogix_logs",
+        "coralogix_error_logs",
+        "alertmanager_alerts",
+        "alertmanager_silences",
         # Kubernetes / EKS evidence keys — written by the _map_eks_* mappers in
         # app/nodes/investigate/processing/post_process.py.  Without these, a pure
         # Kubernetes healthy investigation never satisfies the evidence gate below
--- a/app/nodes/root_cause_diagnosis/evidence_checker.py
+++ b/app/nodes/root_cause_diagnosis/evidence_checker.py
@@ -92,10 +92,6 @@ CLAIM_EVIDENCE_KEYS = INVESTIGATED_EVIDENCE_KEYS | frozenset(
         "datadog_error_logs",
         "datadog_events",
         "datadog_failed_pods",
-        # Other observability stacks
-        "honeycomb_traces",
-        "coralogix_logs",
-        "coralogix_error_logs",
         # Diagnostic code sandbox
         "diagnostic_executions",
         # Vercel
@@ -111,9 +107,7 @@ CLAIM_EVIDENCE_KEYS = INVESTIGATED_EVIDENCE_KEYS | frozenset(
         "github_commits",
         "git_deploy_timeline",
         # Alertmanager
-        "alertmanager_alerts",
         "alertmanager_firing_alerts",
-        "alertmanager_silences",
         "alertmanager_active_silences",
         # EKS adjacent
         "eks_failing_pods",
--- a/tests/nodes/root_cause_diagnosis/test_evidence_checker.py
+++ b/tests/nodes/root_cause_diagnosis/test_evidence_checker.py
@@ -144,6 +144,12 @@ class TestIsClearlyHealthyRejectsUnhealthyStates:
         evidence = {"eks_pods": [{"name": "x"}]}
         assert is_clearly_healthy(alert, evidence) is False
 
+    def test_critical_severity_with_new_observability_key_returns_false(self) -> None:
+        alert = _healthy_alert()
+        alert["commonLabels"] = {"severity": "critical"}
+        evidence = {"alertmanager_alerts": []}
+        assert is_clearly_healthy(alert, evidence) is False
+
     def test_error_annotation_returns_false(self) -> None:
         alert = _healthy_alert()
         alert["commonAnnotations"] = {"error": "something went wrong"}
--- a/tests/nodes/root_cause_diagnosis/test_evidence_checker.py
+++ b/tests/nodes/root_cause_diagnosis/test_evidence_checker.py
@@ -106,6 +106,26 @@ class TestIsClearlyHealthyExistingSources:
         assert is_clearly_healthy(_healthy_alert(), evidence) is True
 
 
+class TestIsClearlyHealthyMissingObservabilitySources:
+    """Healthy states from shipped observability integrations must also fast-path."""
+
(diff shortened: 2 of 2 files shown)


yashksaini-coder (changes requested):
Fetched the branch, ran the full RCA node test suite locally (196/196 pass), ruff clean, mypy clean. The core fix is correct and minimal — exactly what issue #670 asked for.

**What's verified**
- Five key names in `INVESTIGATED_EVIDENCE_KEYS` match the actual mapper output keys in `post_process.py` exactly (`honeycomb_traces`, `coralogix_logs`, `coralogix_error_logs`, `alertmanager_alerts`, `alertmanager_silences`).
- `is_clearly_healthy()` now returns `True` for resolved/low-severity alerts from all three stacks.
- Severity and annotation safeguards still block false-healthy short-circuits (confirmed by `test_critical_severity_with_new_observability_key_returns_false`).
- The parametrized regression test covers all five new keys.

**One change needed before merge**

After this PR, all five new keys exist in **both** `INVESTIGATED_EVIDENCE_KEYS` and the extra `frozenset` inside `CLAIM_EVIDENCE_KEYS`. The docblock above `CLAIM_EVIDENCE_KEYS` explicitly says: *"A key listed here but not in `INVESTIGATED_EVIDENCE_KEYS`…"* — that description no longer holds for the now-redundant entries. This creates a reading trap for the next contributor editing that block. The redundant entries should be removed. See inline comment for exact location.

**Non-blocking follow-up worth tracking**

`check_evidence_availability()` still does not include `alertmanager_*`, `coralogix_*`, or `honeycomb_*` in its `has_cloudwatch_evidence` block. For non-healthy alerts from these stacks, the node may fall through to `_handle_insufficient_evidence` even when evidence is present. This is a pre-existing gap not introduced here, but now that the healthy fast-path is fixed, it's the next logical fix for these three stacks.

yas789:
Thanks for the feedback 👍 , I am making the changes now

yas789:
Follow-up: 

- removed the redundant `honeycomb_traces`, `coralogix_logs`, `coralogix_error_logs`, `alertmanager_alerts`, and `alertmanager_silences` entries from the extra `CLAIM_EVIDENCE_KEYS` frozenset
- kept the docblock above `CLAIM_EVIDENCE_KEYS` as-is; after removing the duplicates, the split it describes is accurate again

Re-ran:
- `.venv/bin/python -m pytest tests/nodes/root_cause_diagnosis/test_evidence_checker.py -vv`
- `.venv/bin/python -m pytest tests/nodes/root_cause_diagnosis/test_healthy_short_circuit.py -vv`

yas789:
Opened a separate follow-up for the non-blocking diagnosis-path gap mentioned in review: #1188.

That PR keeps `#1095` scoped to the healthy short-circuit fix, while the follow-up updates `check_evidence_availability()` so Alertmanager, Coralogix, and Honeycomb evidence also count in the normal RCA diagnosis path.

muddlebee:
already fixed here #777 feel free to revert back at the issue #670 if further questions

PR #795, opened 2026-04-23
feat(llm-cli): modularize CLI binary resolution

muddlebee:
Refactor and modularize CLI path resolution https://github.com/Tracer-Cloud/opensre/pull/781 to be used in generic way by other CLI providers.

#### Describe the changes you have made in this PR -
- Added `app/integrations/llm_cli/binary_resolver.py` as a shared resolver for CLI binaries.
- Refactored `CodexAdapter` in `app/integrations/llm_cli/codex.py` to use the shared resolver (`env -> PATH -> fallback paths`) while preserving compatibility behavior.
- Updated `app/integrations/llm_cli/AGENTS.md` with up-to-date guidance, optional env handling (`CODEX_MODEL`, `CODEX_BIN`), and a provider checklist.
- Kept resolver behavior aligned with issue intent: `CODEX_BIN` and `CODEX_MODEL` remain optional and fallback resolution remains active when unset/invalid.

### Screenshots of the UI changes (If any) -
- N/A (no UI changes)

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [x] I have reviewed every single line of the AI-generated code
- [x] I can explain the purpose and logic of each function/component I added
- [x] I have tested edge cases and understand how the code handles them
- [x] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**
I extracted executable lookup concerns from the Codex adapter into a reusable utility module so new CLI adapters can share one consistent resolution strategy. The resolver keeps explicit override support (`*_BIN`) while preserving robust PATH and conventional install path fallback behavior.

I chose this approach over duplicating resolver code into future adapters because it reduces repeated platform-specific logic and makes behavior changes easier to maintain in one place. I retained compatibility hooks that existing tests expect in `codex.py` to avoid regressions while still reducing adapter complexity.

---

## Checklist before requesting a review
- [x] I have added proper PR title and linked to the issue
- [x] I have performed a self-review of my code
- [x] **I can explain the purpose of every function, class, and logic block I added**
- [x] I understand why my changes work and have tested them thoroughly
- [x] I have considered potential edge cases and how my code handles them
- [x] If it is a core feature, I have added thorough tests
- [x] My code follows the project's style guidelines and conventions

---

[...]

PR #1097, opened 2026-04-30
fix(llm-cli): address post-review issues in CLI integration

muddlebee:
Follows up on #781 and #795 with fixes surfaced during code review.

## Summary

- **Bug fix**: `_run_cli_llm_onboarding` retry loop now sets `os.environ[env_key]` after writing to `.env` so the in-process binary resolver immediately sees the user-provided path — without this, the retry always re-detected "not found" even after the user entered a valid path
- **Security**: user-supplied binary path is now validated (`isfile` + executable check) before being persisted to `.env`
- **Reliability**: bumped `npm_prefix_bin_dirs` subprocess timeout from 0.3 s → 2.0 s to avoid false-empty `@lru_cache` on slow nvm/volta shim startup
- **Testability**: `resolve_cli_binary` defaults for `which_resolver` / `runnable_check` changed from bound-at-definition-time function objects to `None` sentinels with call-time lookup — this makes `patch("binary_resolver.shutil.which")` work without callers passing explicit overrides
- **Cleanup**: dropped redundant `shutil` import and explicit default kwargs from `CodexAdapter._resolve_binary`; consolidated four separate `binary_resolver` import blocks into one
- **Correctness**: `_ver_tuple` switched to `re.findall` so segments like `"1.2a.3"` or pre-release suffixes (`"1.2.3-beta.4"`) parse correctly
- **Observability**: CLI `stderr` is now logged at `DEBUG` after a successful parse so warnings (e.g. deprecation notices from codex) are visible in debug output
- **Tests**: updated all `shutil.which` patch targets from `codex.shutil.which` → `binary_resolver.shutil.which` to match where resolution now lives

## Test plan

- [ ] `make lint` passes
- [ ] `make format-check` passes
- [ ] `make typecheck` passes
- [ ] `make test-cov` passes (4149 passed locally)
- [ ] `tests/integrations/llm_cli/` — all 16 tests pass

### DIFF
--- a/app/cli/wizard/flow.py
+++ b/app/cli/wizard/flow.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import os
 import sys
 from collections.abc import Mapping
 from dataclasses import dataclass
@@ -1614,7 +1615,14 @@ def _run_cli_llm_onboarding(provider: ProviderOption) -> Literal["ok", "abort", 
             return "repick"
         if action == "path":
             path = _prompt_value(f"Full path to {name} binary")
+            if not os.path.isfile(path):
+                _console.print(f"[yellow]'{path}' is not a file. Try again.[/]")
+                continue
+            if not os.access(path, os.X_OK):
+                _console.print(f"[yellow]'{path}' is not executable. Try again.[/]")
+                continue
             sync_env_values({env_key: path})
+            os.environ[env_key] = path
             continue
         _console.print(f"[dim]Hint: {install_hint}[/]")
     _console.print("[yellow]Too many retry attempts. Aborting setup.[/]")
--- a/app/cli/wizard/flow.py
+++ b/app/cli/wizard/flow.py
@@ -19,6 +19,7 @@ from app.cli.wizard.integration_health import IntegrationHealthResult
 from app.cli.wizard.probes import ProbeResult, probe_local_target, probe_remote_target
 from app.cli.wizard.prompts import select as select_prompt
 from app.cli.wizard.store import get_store_path, load_local_config, save_local_config
+from app.integrations.llm_cli.binary_resolver import is_runnable_binary
 from app.integrations.store import get_integration, remove_integration, upsert_integration
 from app.llm_credentials import has_llm_api_key, save_llm_api_key
 
@@ -1615,11 +1616,8 @@ def _run_cli_llm_onboarding(provider: ProviderOption) -> Literal["ok", "abort", 
             return "repick"
         if action == "path":
             path = _prompt_value(f"Full path to {name} binary")
-            if not os.path.isfile(path):
-                _console.print(f"[yellow]'{path}' is not a file. Try again.[/]")
-                continue
-            if not os.access(path, os.X_OK):
-                _console.print(f"[yellow]'{path}' is not executable. Try again.[/]")
+            if not is_runnable_binary(path):
+                _console.print(f"[yellow]'{path}' is not a valid executable. Try again.[/]")
                 continue
             sync_env_values({env_key: path})
             os.environ[env_key] = path
--- a/app/cli/wizard/flow.py
+++ b/app/cli/wizard/flow.py
@@ -19,7 +19,7 @@ from app.cli.wizard.integration_health import IntegrationHealthResult
 from app.cli.wizard.probes import ProbeResult, probe_local_target, probe_remote_target
 from app.cli.wizard.prompts import select as select_prompt
 from app.cli.wizard.store import get_store_path, load_local_config, save_local_config
-from app.integrations.llm_cli.binary_resolver import is_runnable_binary
+from app.integrations.llm_cli.binary_resolver import diagnose_binary_path
 from app.integrations.store import get_integration, remove_integration, upsert_integration
 from app.llm_credentials import has_llm_api_key, save_llm_api_key
 
@@ -1616,8 +1616,9 @@ def _run_cli_llm_onboarding(provider: ProviderOption) -> Literal["ok", "abort", 
             return "repick"
         if action == "path":
             path = _prompt_value(f"Full path to {name} binary")
-            if not is_runnable_binary(path):
-                _console.print(f"[yellow]'{path}' is not a valid executable. Try again.[/]")
+            reason = diagnose_binary_path(path)
+            if reason:
+                _console.print(f"[yellow]{reason} Try again.[/]")
                 continue
             sync_env_values({env_key: path})
             os.environ[env_key] = path
--- a/app/cli/wizard/flow.py
+++ b/app/cli/wizard/flow.py
@@ -1549,6 +1549,16 @@ def _render_next_steps() -> None:
     )
 
 
+def _credential_line_for_saved_summary(provider: ProviderOption) -> str:
+    """One-line credential description for the post-wizard saved summary."""
+    if provider.credential_kind != "cli":
+        return "system keychain"
+    if provider.adapter_factory is None:
+        return f"{provider.label} (CLI)"
+    cli_adapter = provider.adapter_factory()
+    return f"{provider.label} ({cli_adapter.auth_hint})"
+
+
 def _run_cli_llm_onboarding(provider: ProviderOption) -> Literal["ok", "abort", "repick"]:
     """Probe CLI binary + auth; recovery menu when missing. ``repick`` = choose another LLM."""
     factory = provider.adapter_factory
@@ -1788,11 +1798,7 @@ def run_wizard(_argv: list[str] | None = None) -> int:
         saved_path=str(saved_path),
         env_path=summary_env_path,
         configured_integrations=configured_integrations,
-        credential_line=(
-            "OpenAI Codex CLI (`codex login`)"
-            if provider.credential_kind == "cli"
-            else "system keychain"
-        ),
+        credential_line=_credential_line_for_saved_summary(provider),
     )
     demo_response = build_demo_action_response()
     _render_demo_response(demo_response)
--- a/app/integrations/llm_cli/binary_resolver.py
+++ b/app/integrations/llm_cli/binary_resolver.py
@@ -191,13 +191,8 @@ def diagnose_binary_path(path: str) -> str | None:
     if not p.is_file():
         return f"'{path}' is not a file."
     if sys.platform == "win32":
-        if p.suffix.lower() not in {".cmd", ".exe", ".ps1", ".bat"} and not os.access(
-            p, os.X_OK
-        ):
-            return (
-                f"'{path}' is not a recognised executable "
-                f"(expected .cmd, .exe, .ps1, or .bat)."
-            )
+        if p.suffix.lower() not in {".cmd", ".exe", ".ps1", ".bat"} and not os.access(p, os.X_OK):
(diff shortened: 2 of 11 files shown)


muddlebee:
# Audit

---

## Binary Resolution (`binary_resolver.py`)

| Scenario | Covered? | How |
|---|---|---|
| **Linux** — PATH lookup | ✅ | `shutil.which("codex")` + fallback paths |
| **macOS** — Homebrew `/opt/homebrew/bin` | ✅ | darwin-only fallback path |
| **Windows** — `.cmd`/`.exe`/`.ps1`/`.bat` extensions | ✅ | `candidate_binary_names` + `is_runnable_binary` extension check |
| **Windows** — `APPDATA/npm`, `LOCALAPPDATA/Programs` | ✅ | `default_cli_fallback_paths` win32 branch |
| **Windows** — npm prefix uses root (no `/bin`) | ✅ | `npm_prefix_bin_dirs` win32 branch + test |
| **nvm/volta/pnpm** install dirs | ✅ | `.volta/bin`, `PNPM_HOME`, `XDG_DATA_HOME/pnpm`, npm prefix |
| **`CODEX_BIN` set but invalid** | ✅ | Falls through to PATH/fallbacks + test |
| **`CODEX_BIN` set and valid** | ✅ | Short-circuits + test |
| **`CODEX_BIN` empty/blank** | ✅ | `.strip()` makes it falsy, falls through |
| **npm not installed** | ✅ | `OSError` caught, returns `()` |
| **npm hangs/slow shim** | ✅ | 2.0s timeout + `TimeoutExpired` caught |
| **Duplicate fallback paths** | ✅ | Deduplication via `seen` set |
| **Spaces in path** | ✅ | `Path` objects handle this; `expanduser` for `~` |
| **`npm_config_prefix` case-insensitive env** | ✅ | Explicit `.lower()` scan of `os.environ` |
| **Symlinks** | ✅ | added checks diagnose_binary_path( ) |

**Tests:** Linux, macOS, Windows fallback paths all have dedicated tests. npm prefix has win32/unix tests.

---

## Auth Probe (`codex.py`)

| Edge case | Covered? | How |
|---|---|---|
| **"Not logged in" vs "logged in" substring** | ✅ | Negative phrases checked first |
| **Exit 0 but "Not logged in" text** | ✅ | Dedicated test `test_detect_not_logged_in_exit_zero` |
| **Expired/invalid token** | ✅ | `_classify_codex_auth` checks both |
| **Rate-limited but auth'd** | ✅ | Returns `True` with warning message |
| **Network unreachable** | ✅ | Returns `None` (unknown), not `False` |
| **Version parsing: pre-release suffix** | ✅ | `re.findall(r"\d+")` handles `1.2.3-beta.4` |
| **Version parsing: non-digit segment** | ✅ | `"1.2a.3"` → extracts `1, 2, 3` correctly now |
| **Version below minimum** | ✅ | `_ver_tuple` comparison + upgrade note |
| **`--version` times out** | ✅ | `TimeoutExpired` → `installed=False` |
| **`login status` times out** | ✅ | `TimeoutExpired` → `logged_in=None` |
| **Binary exists but crashes on `--version`** | ✅ | Non-zero returncode → `installed=False` |
| **Empty version string** | ⚠️ Mild gap | `_parse_semver` returns `None`, `_ver_tuple` never called — safe but `version=None` in probe |

---

## Wizard Onboarding (`flow.py`)

| Edge case | Covered? | How |
|---|---|---|
| **User provides valid path** | ✅ | `is_runnable_binary` + persist + `os.environ` set |
| **User provides non-existent path** | ✅ | `is_runnable_binary` → `is_file()` check → rejection |
| **User provides non-executable file** | ✅ | `is_runnable_binary` → `os.access(X_OK)` on Unix, extension check on Windows |
| **Windows `.cmd` file** | ✅ | `is_runnable_binary` extension check |
| **Path with spaces/quotes** | ✅ | `Path` handles this; written as-is to `.env` |
| **Max retries (10)** | ✅ | Returns `"abort"` |
| **Repick from installed-but-not-authed** | ✅ | Returns `"repick"` → outer loop re-prompts |
| **Repick from not-installed** | ✅ | Same |
| **Adapter factory is None** | ✅ | Early abort with error message |
| **In-process env propagation** | ✅ | `os.environ[env_key] = path` after `sync_env_values` |

---

## Subprocess Environment (`runner.py`)

| Edge case | Covered? | How |
|---|---|---|
| **API keys not leaked** | ✅ | `ANTHROPIC_API_KEY`, `OPENAI_API_KEY` excluded from env + test assertion |
| **`CODEX_*` forwarded** | ✅ | Prefix `"CODEX_"` in allowlist |
| **Proxy vars forwarded** | ✅ | `HTTP_PROXY`, `HTTPS_PROXY`, etc. in allowlist |
| **XDG dirs forwarded** | ✅ | `XDG_CONFIG_HOME`, `XDG_CACHE_HOME`, etc. |
| **Adapter env overrides merged** | ✅ | `overrides` dict applied after base env |
| **Subprocess timeout** | ✅ | `TimeoutExpired` → `RuntimeError` |
| **Subprocess OSError (binary vanishes)** | ✅ | `OSError` → `RuntimeError` |
| **Non-zero exit** | ✅ | Delegates to `adapter.explain_failure` |
| **ANSI escape codes in output** | ✅ | Stripped from both stdout and stderr |
| **Probe caching (thread-safe)** | ✅ | Double-checked locking with `threading.Lock` |
| **Future CLI env prefix (e.g. `CLAUDE_*`)** | ⚠️ Gap (documented) | Must manually add to `_SAFE_SUBPROCESS_ENV_PREFIXES` — now documented in AGENTS.md |

---

yashksaini-coder (changes requested):
Ran locally: **21/21 pass**, lint clean, no new mypy errors.

The three core fixes are all correct — the `os.environ` propagation bug in particular was a real problem (without it, every retry in the wizard would re-read the stale env and keep reporting "not found" even after the user entered a valid path). Good catch.

---

**`diagnose_binary_path` and `is_runnable_binary` diverge on Windows**

`is_runnable_binary` passes a Windows file when its suffix is `.cmd`, `.exe`, `.ps1`, or `.bat`. `diagnose_binary_path` skips the `sys.platform != "win32"` block entirely, so on Windows it returns `None` (valid) for any existing file regardless of extension — a `.txt` file would pass wizard validation and get written to `.env`, then silently fail at `resolve_cli_binary` time when `is_runnable_binary` rejects it.

Easy fix: add the extension check to `diagnose_binary_path` for Windows:
```python
if sys.platform == "win32":
    if p.suffix.lower() not in {".cmd", ".exe", ".ps1", ".bat"} and not os.access(p, os.X_OK):
        return f"'{path}' is not a recognised executable (expected .cmd, .exe, .ps1, or .bat)."
```

Low real-world impact since most Windows users will point to a real binary, but the inconsistency between the two functions is a correctness gap worth closing.

---

**Import blocks in `codex.py` not fully collapsed**

The PR description says "consolidated four separate `binary_resolver` import blocks into one" but there are still three separate `from app.integrations.llm_cli.binary_resolver import (...)` blocks (lines 10–18). Should be:
```python
from app.integrations.llm_cli.binary_resolver import (
    candidate_binary_names as _candidate_binary_names,
    default_cli_fallback_paths as _default_cli_fallback_paths,
    resolve_cli_binary,
)
```

---

**No test for the `os.environ` propagation path**

The retry-loop env fix is the most impactful change here, and it has no targeted test. `_run_cli_llm_onboarding` is interactive but the propagation can be tested in isolation: verify that after `sync_env_values` + `os.environ[env_key] = path`, a subsequent `os.getenv(env_key)` returns the new value in the same process. Not blocking, but worth adding since this was a silent regression.

---

Aside from these, everything else looks solid — `_ver_tuple` with `re.findall` is more robust, the sentinel defaults correctly fix test patchability, the npm timeout bump is pragmatic, and the AGENTS.md updates are genuinely useful for the next contributor adding a CLI integration. The patch target changes in the test file are all correct.

app/integrations/llm_cli/binary_resolver.py
```diff
@@ -119,26 +171,64 @@ def is_runnable_binary(path: str) -> bool:
...
+        with contextlib.suppress(OSError, AttributeError):
+            target = f" (points to '{p.readlink()}')"
+        return f"'{path}' is a broken symlink{target}. Remove or fix it."
+    if not p.exists():
+        return f"'{path}' does not exist."
+    if not p.is_file():
+        return f"'{path}' is not a file."
+    if sys.platform != "win32" and not os.access(p, os.X_OK):
+        return f"'{path}' is not executable. Run: chmod +x {path}"
+    return None
```
yashksaini-coder, line 198: Windows executable validation is missing here. `is_runnable_binary` returns `True` on Windows only when the suffix is `.cmd`, `.exe`, `.ps1`, or `.bat` (or `os.access` passes). `diagnose_binary_path` skips that check entirely — line 190's guard means on Windows we fall straight through to `return None`, so a `.txt` file passes wizard validation, gets written to `.env`, then silently fails in `resolve_cli_binary`.

Add the extension check before returning `None`:
```python
if sys.platform == "win32":
    if p.suffix.lower() not in {".cmd", ".exe", ".ps1", ".bat"} and not os.access(p, os.X_OK):
        return f"'{path}' is not a recognised executable (expected .cmd, .exe, .ps1, or .bat)."
return None
```

app/integrations/llm_cli/codex.py
```diff
@@ -14,9 +13,6 @@
 from app.integrations.llm_cli.binary_resolver import (
     default_cli_fallback_paths as _default_cli_fallback_paths,
 )
-from app.integrations.llm_cli.binary_resolver import (
-    is_runnable_binary as _is_runnable_binary,
-)
 from app.integrations.llm_cli.binary_resolver import (
     resolve_cli_binary,
 )
```
yashksaini-coder, line 18: Still three separate import blocks from `binary_resolver`. The PR description says "consolidated four separate binary_resolver import blocks into one" — that's not the case yet. Collapse into a single block:
```python
from app.integrations.llm_cli.binary_resolver import (
    candidate_binary_names as _candidate_binary_names,
    default_cli_fallback_paths as _default_cli_fallback_paths,
    resolve_cli_binary,
)
```

app/cli/wizard/flow.py
```diff
@@ -1614,7 +1616,12 @@ def _run_cli_llm_onboarding(provider: ProviderOption) -> Literal["ok", "abort",
             return "repick"
         if action == "path":
             path = _prompt_value(f"Full path to {name} binary")
+            reason = diagnose_binary_path(path)
+            if reason:
+                _console.print(f"[yellow]{reason} Try again.[/]")
+                continue
             sync_env_values({env_key: path})
+            os.environ[env_key] = path
```
yashksaini-coder, line 1634: This is the key fix — setting `os.environ[env_key]` in-process so the next `adapter.detect()` call in the retry loop sees the path immediately without requiring a process restart. It's correct.

Worth adding a targeted test for this behaviour: mock `sync_env_values` and verify that after the path action executes, `os.environ[env_key]` equals the path the user entered. Without a test, a future refactor could silently reintroduce the original bug (retry loop always re-detecting "not found" even after a valid path is entered).

muddlebee:
@yashksaini-coder thank you for the catch. nice

muddlebee:
> Import blocks in codex.py not fully collapsed

skipped this one, not an issue imho functional or santiy wise rn. also it opens up new lint errors..

PR #1064, opened 2026-04-29
auto-configure PATH in shell profile after install

rrajan94:
Fixes #1063

#### Describe the changes you have made in this PR -

After running the one-liner install, `opensre` gave `command not found` because `~/.local/bin` wasn't in `$PATH`. Users had to manually add the export line.

This PR adds a `configure_path()` function to `install.sh` that automatically writes the PATH export to the user's shell profile:

- zsh → `~/.zshrc`
- bash on Linux → `~/.bashrc`
- bash on macOS → `~/.bash_profile`
- fish → `~/.config/fish/config.fish` using `fish_add_path`
- unknown shell → prints manual instructions, no file touched
- Windows → unchanged

Re-running the installer never duplicates the line (idempotent via marker comment). The user is told to `source <profile>` or open a new terminal after install.

### Demo/Screenshot for feature changes and bug fixes -

image

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**

- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**

- [x] I have reviewed every single line of the AI-generated code
- [x] I can explain the purpose and logic of each function/component I added
- [x] I have tested edge cases and understand how the code handles them
- [x] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**

The install script was printing a warning asking users to manually add the PATH export. The fix detects the shell via `$SHELL`, picks the right rc file per shell and OS, then appends an export line guarded by a marker comment so the operation is idempotent. Fish uses `fish_add_path` instead of `export`. Two checks prevent duplicate writes: one for the marker comment (installer-managed) and one for the install dir already appearing in the file (user-managed). The function is placed in the function-definition section of the script so it can be extracted and unit-tested in isolation — 11 tests cover all the above cases.

---

## Checklist before requesting a review

- [x] I have added proper PR title and linked to the issue
- [x] I have performed a self-review of my code
- [x] **I can explain the purpose of every function, class, and logic block I added**
- [x] I understand why my changes work and have tested them thoroughly
- [x] I have considered potential edge cases and how my code handles them
- [x] If it is a core feature, I have added thorough tests
- [x] My code follows the project's style guidelines and conventions

---

Note: Please check **Allow edits from maintainers** if you would like us to assist in the PR.

PR #1101, opened 2026-04-30
test(cli): skip install.sh tests on Windows runner (#1099)

### LINKED ISSUE #1099: [BUG] : CI failing for opensre tests
### Summary

Image

need to fix post https://github.com/Tracer-Cloud/opensre/pull/1090

mayankbharati-ops:
Closes #1099.

## Root cause

The post-merge `test (windows-latest)` job on `67dc64d1` (the merge of #1090) is failing with **two distinct symptoms** in the `install.sh` test suite added by #1064:

### 1. `tests/cli/test_install_sh_path.py` — 12 failures

The helper `_run` calls `subprocess.run(["bash", "-c", script])`. On the GitHub Actions `windows-latest` runner, `bash` resolves to `wsl.exe`, and that runner has no installed WSL distribution. Every `_run` exits with code 1 and the literal message:

> `Windows Subsystem for Linux has no installed distributions.`
> `You can resolve this by installing a distribution with the instructions below`

So none of the asserted rc files (`.zshrc`, `.bashrc`, `config.fish`) ever get written, and assertions like `assert zshrc.exists()` and `assert result.returncode == 0` fail.

### 2. `tests/cli/test_install_sh_resolution.py` — module collection error

```python
@pytest.mark.skipif(os.geteuid() == 0, reason="sudo fallback is only meaningful for non-root users")
```

`os.geteuid` does not exist on Windows. The decorator condition is evaluated at **decorator-application time** (i.e. module import), so `AttributeError: module 'os' has no attribute 'geteuid'` fires *before* pytest gets a chance to read any markers. The whole module fails to collect.

## Why skip rather than fix the test scaffolding

Both files exclusively test `install.sh`, which is a POSIX shell script driving zsh/bash/fish rc-file behaviour. There's no Windows analogue (the project ships a separate Windows installer / `pip install` path), so installing WSL or porting the test scaffolding to PowerShell would be busywork. Skipping the module on Windows matches what the codebase actually claims to support.

## Changes

```python
# test_install_sh_path.py
pytestmark = pytest.mark.skipif(
    sys.platform == "win32",
    reason="install.sh is POSIX-only; the Windows runner has no usable bash "
           "(resolves to unconfigured WSL), so this module's subprocess-driven "
           "tests cannot run there. See issue #1099.",
)
```

`test_install_sh_resolution.py` gets the same `pytestmark`, *plus* a defensive replacement of the bare `os.geteuid()` access:

```python
# Evaluated once at module load. hasattr short-circuits the `and` on
# Windows, so os.geteuid() is never called there — the module imports
# cleanly even before pytest reads ``pytestmark``.
_RUNNING_AS_ROOT = hasattr(os, "geteuid") and os.geteuid() == 0

@pytest.mark.skipif(_RUNNING_AS_ROOT, reason="sudo fallback is only meaningful for non-root users")
```

Without that defensive guard the module still wouldn't *import* on Windows, even with `pytestmark` defined — and `pytestmark` only kicks in once pytest has collected the module.

## Verification

| | macOS / Linux | Windows |
|---|---|---|
| Module imports | ✅ unchanged | ✅ no more AttributeError |
| Tests run | ✅ all 15 pass on local macOS | Skipped cleanly with reason text |
| Existing root-skip | ✅ still active on Unix when running as root | n/a (whole module skipped) |
| `ruff check` / `ruff format --check` | ✅ clean | ✅ clean |
| Full unit suite | **4164 / 2 skipped / 1 xfailed** | (will skip 13 from these modules + still fail count = 0) |

## Scope

| File | Change |
|---|---|
| `tests/cli/test_install_sh_path.py` | +13 / -0 (module-level pytestmark + comment block) |
| `tests/cli/test_install_sh_resolution.py` | +28 / -1 (pytestmark + defensive `_RUNNING_AS_ROOT` constant; existing decorator references the new constant) |

No production code touched. No test logic changed for the platforms where these tests *do* run.

cc @muddlebee — this is the post-#1090 Windows CI fix. Both files were added by #1064 and have been failing on Windows since that landed; this PR makes the Windows runner green again without changing any Unix-side behaviour.

### DIFF
--- a/tests/cli/test_install_sh_path.py
+++ b/tests/cli/test_install_sh_path.py
@@ -3,8 +3,28 @@
 from __future__ import annotations
 
 import subprocess
+import sys
 import textwrap
 from pathlib import Path
+
+import pytest
+
+# install.sh is a POSIX shell script that exercises zsh/bash/fish rc-file
+# behaviour, and these tests drive it via ``subprocess.run(["bash", "-c", ...])``.
+# On the GitHub Actions ``windows-latest`` runner, ``bash`` is resolved to
+# ``wsl.exe`` and the runner has no installed WSL distribution — every
+# ``_run`` call exits 1 with a "Windows Subsystem for Linux has no installed
+# distributions" message and none of the asserted rc files get written.
+# Skip the whole module rather than chase a Windows analogue for a Unix-only
+# installer script.
+pytestmark = pytest.mark.skipif(
+    sys.platform == "win32",
+    reason=(
+        "install.sh is POSIX-only; the Windows runner has no usable bash "
+        "(resolves to unconfigured WSL), so this module's subprocess-driven "
+        "tests cannot run there. See issue #1099."
+    ),
+)
 
 INSTALL_SH = Path(__file__).parents[2] / "install.sh"
 _LOCAL_BIN = ".local/bin"
--- a/tests/cli/test_install_sh_resolution.py
+++ b/tests/cli/test_install_sh_resolution.py
@@ -4,10 +4,30 @@ from __future__ import annotations
 
 import os
 import subprocess
+import sys
 import textwrap
 from pathlib import Path
 
 import pytest
+
+# Same Windows-skip rationale as ``test_install_sh_path.py`` — install.sh is
+# POSIX-only and the GitHub Actions ``windows-latest`` runner has no usable
+# bash. See issue #1099.
+pytestmark = pytest.mark.skipif(
+    sys.platform == "win32",
+    reason=(
+        "install.sh is POSIX-only; the Windows runner has no usable bash "
+        "(resolves to unconfigured WSL), so this module's subprocess-driven "
+        "tests cannot run there. See issue #1099."
+    ),
+)
+
+# ``os.geteuid`` does not exist on Windows. The skipif decorator below is
+# evaluated at decorator-application time (i.e. module import), so a bare
+# ``os.geteuid() == 0`` check would raise ``AttributeError`` on Windows
+# *before* ``pytestmark`` ever takes effect. ``hasattr`` short-circuits the
+# ``and`` so the ``os.geteuid()`` call only runs on platforms that have it.
+_RUNNING_AS_ROOT = hasattr(os, "geteuid") and os.geteuid() == 0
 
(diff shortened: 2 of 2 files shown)


muddlebee:
@mayankbharati-ops I hope only these two were the affected regions?

muddlebee:
@mayankbharati-ops I just added a custom workflow check(already in main), after your PR. so maintainers can trigger window CI checks during PR only.. have triggered it here, pls see if any more windows issues.. 

looks like it passed https://github.com/Tracer-Cloud/opensre/pull/1102

muddlebee:
@mayankbharati-ops thanks for the quick fix. window checks green at #1102

mayankbharati-ops:
@muddlebee yes — those two were the only affected modules. I just re-audited the test tree to be sure, and PR #1102's now-green Windows checks corroborate it:

**Audit results across `tests/`:**

| Pattern | Hits | Files |
|---|---|---|
| `subprocess.run(["bash"...])` / `["sh"...]` / `["zsh"...]` / `["fish"...]` | 2 | both fixed in #1101 |
| `os.geteuid` / `os.getuid` / `os.fork` / `os.symlink` / `os.getlogin` | 1 | only `test_install_sh_resolution.py` (fixed in #1101) |
| `import pwd` / `import grp` / `import fcntl` / `import termios` / `import resource` | 0 | — |
| Hard-coded Unix paths (`/tmp/...`, `/usr/...`) actually opened from disk | 0 | the only `/tmp/...` references in `test_investigate_service_flag.py:86` and `test_health_view.py:75` are test-parameter strings the CLI rejects/serializes without touching the filesystem — safe on Windows |

So the surface area for Windows-specific test failures was exactly those two files. The "windows test SUCCESS" on #1102 is the empirical confirmation.

Re: the pre-merge Windows trigger — that's a nice addition. With it, this whole regression class would have been caught at PR-review time on #1064 (where these tests landed) instead of post-merge. Glad it's in.

Thanks for the quick merge!

muddlebee:
@mayankbharati-ops in case you are interested how we added the new ci windows workflows

base PR: https://github.com/Tracer-Cloud/opensre/pull/1100
follow up PR: https://github.com/Tracer-Cloud/opensre/pull/1103

PR #1105, opened 2026-04-30
fix: Ctrl+C double-exit in interactive shell

### LINKED ISSUE #1091: Improve the interactive opensre shell behavior so that pressing Ctrl+C twice exits the shell, similar to Claude CLI.

jason8745:
Fixes #1091

## Changes

Implements the two-press Ctrl+C exit pattern, consistent with Claude CLI:
- **First Ctrl+C**: displays `(Press Ctrl+C again to exit)` and re-displays the prompt
- **Second Ctrl+C** within 2 seconds: prints `Goodbye!` and exits cleanly (exit code 0)

### Implementation

- Added `_with_ctrl_c_double_exit()` in `app/cli/prompt_support.py` — wraps `question.unsafe_ask()` in a retry loop. This avoids key-binding conflicts that caused an earlier sentinel approach to fail (`Application.exit()` raises if called twice).
- Added `_HardQuitInterrupt(KeyboardInterrupt)` subclass so Ctrl+Q hard-quit bypasses the retry guard.
- Restored an explicit `ControlC` key binding in wizard `_base_bindings()` — `InquirerControl` is not a `BufferControl`, so prompt_toolkit's default Ctrl+C binding is inactive for custom wizard prompts; without this, `\x03` was silently ignored.
- Installed a SIGINT handler in `main()` via `_install_sigint_handler()` to handle Ctrl+C between prompts (streaming output, long-running ops, etc.).
- Updated unit tests in `tests/cli/test_prompt_support.py` covering hint-on-first-press, exit-on-second-press, and window reset.
- Fixed `test_tests_interactive_launcher_smoke` to use Escape instead of Ctrl+C — single Ctrl+C no longer exits immediately, and double Ctrl+C in PTY cooked mode has SIGINT race conditions. Escape is PTY-safe and already the documented exit key in the prompt instruction.

### Verification
Before
image
After
image

---

The core insight is that `question.unsafe_ask()` propagates `KeyboardInterrupt` while `question.ask()` swallows it. By wrapping `unsafe_ask()` in a `while True` retry loop, the first Ctrl+C is caught, the hint is printed, and the same `Application` instance is re-run (safe because `Application.run()` resets `_is_running` and creates a new future on each call). The second Ctrl+C within the time window exits. This avoids fighting prompt_toolkit's key binding merge order entirely.

---

### DIFF
--- a/app/cli/__main__.py
+++ b/app/cli/__main__.py
@@ -117,9 +117,11 @@ def main(argv: list[str] | None = None) -> int:
     try:
         cli(args=argv, standalone_mode=True)
     except KeyboardInterrupt:
-        # Suppress Click's "Aborted!" message; a clean exit with no output is
-        # correct here because handle_ctrl_c_press already printed the hint or
-        # "Goodbye!" before raising SystemExit / letting the interrupt propagate.
+        # A KeyboardInterrupt that escapes cli() was not handled by our
+        # double-exit logic (e.g. click.prompt, an unpatched library prompt).
+        # Print a newline so the terminal cursor lands on a clean line, then
+        # exit quietly — Click's "Aborted!" message is intentionally suppressed.
+        print(flush=True)
         return 0
     except SystemExit as exc:
         if isinstance(exc.code, int):
--- a/app/cli/__main__.py
+++ b/app/cli/__main__.py
@@ -10,6 +10,7 @@ Enable shell tab-completion (add to your shell profile for persistence):
 from __future__ import annotations
 
 import os
+import signal
 import sys
 
 import click
@@ -19,7 +20,11 @@ from app.analytics.cli import capture_cli_invoked
 from app.analytics.provider import capture_first_run_if_needed, shutdown_analytics
 from app.cli.commands import register_commands
 from app.cli.layout import RichGroup, render_landing
-from app.cli.prompt_support import install_questionary_escape_cancel
+from app.cli.prompt_support import (
+    handle_ctrl_c_press,
+    install_questionary_ctrl_c_double_exit,
+    install_questionary_escape_cancel,
+)
 from app.version import get_version
 
 
@@ -87,14 +92,35 @@ def cli(
 register_commands(cli)
 
 
+def _install_sigint_handler() -> None:
+    """Handle Ctrl+C between prompts (when prompt_toolkit is not active).
+
+    prompt_toolkit intercepts Ctrl+C internally while a prompt is running, so
+    the key binding in prompt_support.py handles that case.  This SIGINT handler
+    covers everything else: long-running operations, streaming output, etc.
+    """
+
+    def _handler(signum: int, frame: object) -> None:  # noqa: ARG001
+        handle_ctrl_c_press()
+
+    signal.signal(signal.SIGINT, _handler)
+
+
 def main(argv: list[str] | None = None) -> int:
     """Entry point for the ``opensre`` console script."""
     load_dotenv(override=False)
     install_questionary_escape_cancel()
+    install_questionary_ctrl_c_double_exit()
+    _install_sigint_handler()
     capture_first_run_if_needed()
 
     try:
         cli(args=argv, standalone_mode=True)
+    except KeyboardInterrupt:
+        # Suppress Click's "Aborted!" message; a clean exit with no output is
+        # correct here because handle_ctrl_c_press already printed the hint or
+        # "Goodbye!" before raising SystemExit / letting the interrupt propagate.
+        return 0
     except SystemExit as exc:
         if isinstance(exc.code, int):
             return exc.code
--- a/app/cli/prompt_support.py
+++ b/app/cli/prompt_support.py
@@ -57,6 +57,7 @@ def install_questionary_escape_cancel() -> None:
     import questionary
     import questionary.prompts.checkbox as checkbox_mod
     import questionary.prompts.confirm as confirm_mod
+    import questionary.prompts.password as password_mod
     import questionary.prompts.path as path_mod
     import questionary.prompts.select as select_mod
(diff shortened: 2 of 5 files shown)


app/cli/prompt_support.py
```diff
@@ -63,3 +74,93 @@ def install_questionary_escape_cancel() -> None:
...
+    text_mod.text = _wrap_question_ctrl_c(text_mod.text)
+    path_mod.path = _wrap_question_ctrl_c(path_mod.path)
+
+    questionary.select = select_mod.select
+    questionary.checkbox = checkbox_mod.checkbox
+    questionary.confirm = confirm_mod.confirm
+    questionary.text = text_mod.text
+    questionary.path = path_mod.path
+
+    _ctrl_c_patch_installed[0] = True
```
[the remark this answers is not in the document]
jason8745: Fixed in 98775e151a34bc993c875c9ed8774797f06e994e

app/cli/prompt_support.py
```diff
@@ -63,3 +74,93 @@ def install_questionary_escape_cancel() -> None:
...
+
+    First call:  prints hint.
+    Second call within _CTRL_C_EXIT_WINDOW seconds:  prints Goodbye and exits.
+    """
+    now = time.monotonic()
+    if now - _last_ctrl_c[0] <= _CTRL_C_EXIT_WINDOW:
+        print("\nGoodbye!", flush=True)
+        sys.exit(0)
+    _last_ctrl_c[0] = now
+    print("\n(Press Ctrl+C again to exit)", flush=True)
```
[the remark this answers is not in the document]
jason8745: Fixed in 8272b2cab35a98a0f8b4e3e432c7da795ced2ab4

app/cli/__main__.py
```diff
@@ -62,14 +67,35 @@ def cli(
...
+    _install_sigint_handler()
     capture_first_run_if_needed()
 
     try:
         cli(args=argv, standalone_mode=True)
+    except KeyboardInterrupt:
+        # Suppress Click's "Aborted!" message; a clean exit with no output is
+        # correct here because handle_ctrl_c_press already printed the hint or
+        # "Goodbye!" before raising SystemExit / letting the interrupt propagate.
+        return 0
```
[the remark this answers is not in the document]
jason8745: Fixed in 2abc33fe0b8c72c23dd716f755d4c8930fdee826

jason8745:
Already reviewed and solved the greptile-apps comments

app/cli/prompt_support.py
```diff
@@ -55,11 +67,106 @@ def install_questionary_escape_cancel() -> None:
...
+    binding already raises KeyboardInterrupt from application.run(), and
+    fighting it with another eager binding causes unpredictable ordering.
+    """
+
+    def _patched_ask(*args: Any, **kwargs: Any) -> Any:
+        while True:
+            try:
+                # unsafe_ask() propagates KeyboardInterrupt instead of
+                # swallowing it the way ask() does.
+                return question.unsafe_ask(*args, **kwargs)
```
muddlebee: Reset _last_ctrl_c after a successful prompt return. The current timestamp leaks across prompts, so Ctrl+C in prompt A followed by one Ctrl+C in prompt B within 2s exits unexpectedly.

app/cli/__main__.py
```diff
@@ -87,14 +92,37 @@ def cli(
...
     """Entry point for the ``opensre`` console script."""
     load_dotenv(override=False)
     install_questionary_escape_cancel()
+    install_questionary_ctrl_c_double_exit()
+    _install_sigint_handler()
     capture_first_run_if_needed()
 
     try:
         cli(args=argv, standalone_mode=True)
+    except KeyboardInterrupt:
```
muddlebee, line 119: This exits on the first KeyboardInterrupt for unpatched prompt paths, which bypasses the double-press behavior from issue #1091. Reuse handle_ctrl_c_press here so Ctrl+C behavior stays consistent.

app/cli/prompt_support.py
jason8745: Thanks for the suggestion. Fixed in cc78f2e355e0d3bc4ce376d993793e1b6401feb7

app/cli/__main__.py
jason8745: Thanks for the suggestion! I want to make sure I understand the intent before implementing.

My concern is that calling `handle_ctrl_c_press()` here and then immediately `return 0` would show `(Press Ctrl+C again to exit)` but exit the process right after — which seems more misleading than the current clean-exit behavior.

To make the double-press actually work here, we'd need a retry loop that re-runs `cli()`, but that restarts the entire CLI rather than re-displaying the interrupted prompt, which feels unexpected.

Could you clarify what behavior you had in mind? Were you thinking of a retry loop approach, or accepting that the hint fires but the process exits after the first press? Would love to understand your reasoning before deciding on the right fix.

Thank you~

jason8745: If consistency is the priority, happy to call `handle_ctrl_c_press()` here. Trade-off: the hint `(Press Ctrl+C again to exit)` would still show but the process exits immediately after — we can't re-display the prompt from `main()`. Let me know if that's acceptable.

VaibhavUpreti (approved):
Great job @jason8745 ! Looks good, welcome the the OpenSRE community 🚀

app/cli/prompt_support.py
VaibhavUpreti: good catch @muddlebee !

PR #1192, opened 2026-05-02
feat(cli): allow switching toolcall model at runtime (#1182)

### LINKED ISSUE #1182: Allow CLI to switch toolcall model at runtime
## Summary

The interactive CLI can switch the LLM provider/reasoning model with `/model set anthropic`, but it does not appear to expose a way to also switch the separate toolcall model at runtime.

## Reproduction

1. Start `opensre`.
2. Run `/model set anthropic`.
3. Observe the connection table.
4. Ask the assistant to switch the toolcall model to the same model.

Observed terminal output:

```text
$ /model set anthropic
switched LLM provider: anthropic (claude-opus-4-7)
updated /Users/janvincentfranciszek/tracer-agent-2026/.env
                LLM connection                 
┌─────────────────┬───────────────────────────┐
│ provider        │ anthropic                 │
│ reasoning model │ claude-opus-4-7           │
│ toolcall model  │ claude-haiku-4-5-20251001 │
└─────────────────┴───────────────────────────┘

› switch the toolcall model also to claude-opus-4-7

assistant:
I can switch the primary LLM to claude-opus-4-7, but OpenSRE doesn't currently expose a separate action to configure 
a distinct "toolcall model" at runtime.
```

## Expected behavior

The CLI should provide a supported way to switch the toolcall model at runtime, for example:

- `/model set anthropic --toolcall-model claude-opus-4-7`
- `/model toolcall set claude-opus-4-7`
- or an equivalent interactive command/config update flow

After the command, the connection table should show both the reasoning model and toolcall model set to the requested values.

## Actual behavior

`/model set anthropic` updates the provider/reasoning model but leaves the toolcall model as `claude-haiku-4-5-20251001`. The assistant reports that the CLI does not currently expose a separate runtime action for changing the toolcall model.

## Notes

This is confusing because the connection table explicitly displays `toolcall model`, which makes it look like a first-class configurable runtime setting alongside `reasoning model`.

Davidson3556:
Fixes #1182

The interactive shell now exposes a runtime-supported way to change the toolcall model, matching the "Expected behavior" in the issue.

- New flag: `/model set <provider> [model] --toolcall-model <model>` — switch provider, reasoning model, and toolcall model in one command.
- New subcommand: `/model toolcall set <model>` — change only the toolcall model on the currently active provider, leaving the reasoning model untouched.
- The connection table's `toolcall model` row now reflects the value the user just set (it was previously stuck at the default).
- Providers that don't expose a separate toolcall model (`codex`, `claude-code`, `ollama`) reject the override with a clear yellow message instead of silently lying.
- The natural-language assistant action schema now accepts an optional `toolcall_model` on `switch_llm_provider`, plus a new `switch_toolcall_model` action — so requests like "switch the toolcall model also to claude-opus-4-7" actually work instead of returning the "OpenSRE doesn't currently expose a separate action..." refusal shown in the bug report.

Files touched:
- `app/cli/wizard/config.py` — added `toolcall_model_env` and `default_toolcall_model` to `ProviderOption`; populated for anthropic/openai/openrouter/gemini/nvidia.
- `app/cli/interactive_shell/commands.py` — extended `switch_llm_provider`, added `switch_toolcall_model`, added a small flag parser for `/model set`, added `/model toolcall` subcommand, and updated the `/model` help text.
- `app/cli/interactive_shell/cli_agent.py` — updated `_ACTION_RULE` and `_execute_action_plan` so the LLM-driven assistant can request a toolcall switch.
- `tests/cli/interactive_shell/test_commands.py` — added 5 tests covering the flag, the subcommand, the unsupported-provider rejection, and the missing-arg usage paths.

Screenshot 2026-05-01 at 20 13 30

---

The bug was a write/read asymmetry: `LLMSettings.from_env()` already reads `<PROVIDER>_TOOLCALL_MODEL`, and the connection table already renders that value, but `switch_llm_provider` only ever wrote `LLM_PROVIDER` and the reasoning model env var — so the toolcall row was effectively read-only from the user's perspective.

Two alternatives I considered and rejected:

1. Auto-mirror the toolcall model from the reasoning model on every `/model set`. Rejected because it silently destroys the user's explicit toolcall override and makes the two values uncontrollable independently — the connection table would still display two rows but they'd always be equal.
2. Add a separate `/toolcall` top-level slash command. Rejected because the discoverability story is worse: users already type `/model` when thinking about models, and the help text for one command is easier to maintain than two.

Chosen approach: keep `/model` as the single entry point. Add a flag (`--toolcall-model`) for the combined case and a subcommand (`/model toolcall set`) for the toolcall-only case, both of which funnel through the same env-sync path that already handles the reasoning model. Provider metadata gains a `toolcall_model_env` field so each provider declares which env var holds its toolcall model; CLI providers (codex/claude-code/ollama) set it to `None` and the helper returns a polite rejection instead of writing nothing or pretending it worked.

Key functions:

- `switch_llm_provider(provider, console, model=None, *, toolcall_model=None)` — extended with a keyword-only `toolcall_model`. When provided, it adds `provider.toolcall_model_env=<value>` to the dict that `sync_env_values` writes, so the reasoning + toolcall + provider all land in `.env` atomically.
- `switch_toolcall_model(model, console, *, provider_name=None)` — new helper. Resolves the active provider from `LLM_PROVIDER` (defaulting to anthropic, mirroring how the rest of the shell defaults), looks up `toolcall_model_env`, and writes only that one key. This is what makes `/model toolcall set` non-destructive to the reasoning model.
- `_parse_model_set_args(args)` — small flag parser. Returns `None` for malformed input (unknown flag, missing flag value, two positional models) so `/model set` can print a single usage line instead of guessing.
- `_cmd_model` — adds a `toolcall` branch and routes `set` through the new parser.
- `_ACTION_RULE` / `_execute_action_plan` in `cli_agent.py` — taught the assistant about the new action shapes, including a label that shows both reasoning and toolcall in the action preview ("switch LLM provider to anthropic (claude-opus-4-7) + toolcall claude-opus-4-7").

Edge cases tested:

- Unknown flag (`--made-up-flag`) → prints usage, no env write.
- Toolcall-only switch on a provider with no separate toolcall env (`codex`) → prints "does not expose a separate toolcall model", no env write.
- `/model toolcall set` with no model → prints usage.
- Toolcall-only switch must NOT rewrite `LLM_PROVIDER` or the reasoning model — verified by asserting absence of those keys in the resulting `.env`.
- Combined `/model set anthropic claude-opus-4-7 --toolcall-model claude-opus-4-7` writes all three keys.

---

---

### DIFF
--- a/app/cli/interactive_shell/commands.py
+++ b/app/cli/interactive_shell/commands.py
@@ -165,7 +165,13 @@ def _render_models_table(console: Console) -> None:
     console.print(table)
 
 
-def switch_llm_provider(provider_name: str, console: Console, model: str | None = None) -> bool:
+def switch_llm_provider(
+    provider_name: str,
+    console: Console,
+    model: str | None = None,
+    *,
+    toolcall_model: str | None = None,
+) -> bool:
     from app.cli.wizard.config import PROVIDER_BY_VALUE
     from app.cli.wizard.env_sync import sync_env_values
 
@@ -184,6 +190,18 @@ def switch_llm_provider(provider_name: str, console: Console, model: str | None 
     if provider.legacy_model_env:
         values[provider.legacy_model_env] = selected_model
 
+    selected_toolcall: str | None = None
+    if toolcall_model is not None:
+        if not provider.toolcall_model_env:
+            console.print(
+                f"[yellow]provider {provider.value} does not expose a separate "
+                "toolcall model[/yellow] — toolcall override ignored."
+            )
+        else:
+            selected_toolcall = toolcall_model.strip()
+            if selected_toolcall:
+                values[provider.toolcall_model_env] = selected_toolcall
+
     env_path = sync_env_values(values)
     os.environ.update(values)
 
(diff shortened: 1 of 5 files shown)


Davidson3556:
@davincios @rrajan94 @VaibhavUpreti  kindly review

app/cli/interactive_shell/commands.py
```diff
@@ -184,18 +190,112 @@ def switch_llm_provider(provider_name: str, console: Console, model: str | None
...
+    provider = args[0]
+    reasoning_model: str | None = None
+    toolcall_model: str | None = None
+
+    i = 1
+    while i < len(args):
+        token = args[i]
+        if token == "--toolcall-model":
+            if i + 1 >= len(args):
+                return None
```
Devesh36: In _parse_model_set_args , when --toolcall-model has no following value the function returns None, and _cmd_model  prints the generic usage line. Could you return a more specific error string like "missing value for --toolcall-model" so users don't have to guess what went wrong?

Davidson3556:
@Devesh36 done

yashksaini-coder (changes requested):
image

Did a local test, even though we don't have an anthropic key, the function still executes, shows that it switched LLM provider, and updates the provider, which should not be the case. 

The expected behaviour should be that it throws error saying env vars not setup or provided, and must stop from updating the LLM provider config in settings.

Devesh36 (changes requested):
Image

how is reasoning model changing then ?

Davidson3556:
@Devesh36 the behavior here is actually correct, but I see why it read like a bug. The second positional argument in
`/model set <provider> [model] [--toolcall-model <m>]` is the reasoning model (per `/help` and the usage line). In your repro:

  before:  reasoning = claude-sonnet-4-6
           toolcall  = claude-haiku-4-5-20251001

  you ran: /model set anthropic claude-haiku-4-5-20251001
                                ^^^^^^^^^^^^^^^^^^^^^^^^^
                                this is the reasoning model

  after:   reasoning = claude-haiku-4-5-20251001  (just changed)
           toolcall  = claude-haiku-4-5-20251001  (unchanged ,was
                                                   already haiku)

So both rows show the same value because you happened to pick a reasoning model that matches the existing toolcall default. The tool call slot itself was not rewritten.

That said, you were right to flag it , the success message `switched LLM provider: anthropic (claude-haiku-4-5-20251001)` didn't tell you which slot the parenthetical landed in, which is what made it look wrong. Pushed 8ec5892 to fix the wording. The new output
is:

  switched LLM provider: anthropic
  reasoning model: claude-haiku-4-5-20251001 (ANTHROPIC_REASONING_MODEL)
  [toolcall model: <X> (ANTHROPIC_TOOLCALL_MODEL)]   # only when
                                                      # --toolcall-model
                                                      # was passed
  updated /home/.../.env

So next time the source of each row in the connection table is unambiguous from the lines above it.

For reference, the three forms now all behave distinctly:

  /model set anthropic <X>                          # changes reasoning only
  /model toolcall set <X>                           # changes toolcall only
  /model set anthropic <X> --toolcall-model <Y>     # changes both

Davidson3556:
@yashksaini-coder  fixed.

yashksaini-coder (approved):
All requested changes addressed (commits e66ad5b9, 8ec58926). Credential gate, explicit slot labelling, and specific parse errors are tested. LGTM.

VaibhavUpreti:
@greptileai review

Devesh36 (approved):

muddlebee:
@Davidson3556 this is great add on :fire:

PR #1214, opened 2026-05-02
fix: improve WAL-driven replication lag RCA (scenario 001) + add QA validation

### LINKED ISSUE #597: [synthetic-qa] 001-replication-lag: Validate agent identifies WAL-driven replication lag
## Scenario

**Run:** `opensre tests synthetic --scenario 001-replication-lag`

A write-heavy workload on the primary generates WAL faster than the read replica can replay it. ReplicaLag exceeds 900 seconds.

## Correct Answer

- **Root Cause:** Write-heavy workload → WAL generation outpaces replica replay → replication lag accumulates.

## Human Review Checklist

- [ ] Does the agent identify **replication lag** as the problem (not CPU, not connections)?
- [ ] Does the agent explain the **mechanism**: WAL generation faster than replica replay?
- [ ] Does the agent mention the **replica** specifically?
- [ ] Does the causal chain flow logically: write burst → WAL spike → replica falls behind → stale reads?
- [ ] Does the agent avoid blaming CPU? (CPU is not elevated in this scenario)

The agent identifies that bulk writes on the primary are generating WAL faster than the replica can replay. It traces the causal chain from WriteIOPS/TransactionLogsGeneration spikes to the growing ReplicaLag.

## Common Failure Modes

- Agent says "replication lag" but doesn't explain WHY (misses the WAL mechanism)
- Agent blames the replica itself rather than the write-heavy primary workload
- Agent diagnoses `cpu_saturation` even though CPU is not the issue

cerencamkiran:
Fixes #597

## The changes

This PR improves root cause analysis (RCA) quality for the `001-replication-lag` synthetic scenario by enforcing correct WAL-driven replication reasoning.

## Key improvements

**Improves replication lag reasoning:**
- Ensures the agent identifies replication lag as a downstream effect of a write-heavy workload on the primary
- Enforces correct causal chain: write workload → WAL generation → replica replay lag

**Enforces mechanism-level RCA:**
- Requires explanation of WAL generation vs replica replay mismatch
- Prevents shallow “replication lag” diagnoses without causal explanation
- Requires `replay` to enforce WAL vs replica replay mechanism

**Strengthens evidence usage:**
- Requires both CloudWatch metrics (`ReplicaLag`, `WriteIOPS`, `TransactionLogsGeneration`)
- Requires Performance Insights evidence (write-heavy SQL + WAL waits)

**Prevents common misdiagnoses:**
- Disallows `cpu_saturation` and `connection_exhaustion` as root causes
- Ensures CPU is treated as a downstream effect, not the initiating cause
- Prevents blaming the replica instead of the primary workload

**Adds QA validation:**
- Introduces `QA_VALIDATION.md` for scenario 001
- Documents expected reasoning, evidence usage, and failure modes
- Helps prevent regressions in replication-related RCA

## Result

Scenario `001-replication-lag` now produces stronger, mechanism-level RCA output with correct attribution and evidence grounding.

PASS 001-replication-lag category=resource_exhaustion

Results: 1/1 passed

## Run

```bash
python -m tests.synthetic.rds_postgres.run_suite --scenario 001-replication-lag --mock-grafana

### DIFF
--- a/tests/synthetic/rds_postgres/run_suite.py
+++ b/tests/synthetic/rds_postgres/run_suite.py
@@ -340,7 +340,9 @@ def score_result(
                 state_key = _EVIDENCE_KEY_MAP.get(source_key, source_key)
 
                 has_state_evidence = bool(evidence.get(state_key))
-                has_pi_signal = any(token in normalized_output for token in performance_insights_tokens)
+                has_pi_signal = any(
+                    token in normalized_output for token in performance_insights_tokens
+                )
 
                 if not (has_state_evidence and has_pi_signal):
                     failure_reason = f"required evidence not gathered: {source_key!r}"
--- a/tests/synthetic/rds_postgres/run_suite.py
+++ b/tests/synthetic/rds_postgres/run_suite.py
@@ -337,9 +337,15 @@ def score_result(
 
         for source_key in answer_key.required_evidence_sources:
             if source_key == "aws_performance_insights":
-                if not any(token in normalized_output for token in performance_insights_tokens):
+                state_key = _EVIDENCE_KEY_MAP.get(source_key, source_key)
+
+                has_state_evidence = bool(evidence.get(state_key))
+                has_pi_signal = any(token in normalized_output for token in performance_insights_tokens)
+
+                if not (has_state_evidence and has_pi_signal):
                     failure_reason = f"required evidence not gathered: {source_key!r}"
                     break
+
                 continue
 
             state_key = _EVIDENCE_KEY_MAP.get(source_key, source_key)
--- a/tests/synthetic/rds_postgres/run_suite.py
+++ b/tests/synthetic/rds_postgres/run_suite.py
@@ -327,7 +327,21 @@ def score_result(
     # internal evidence keys (grafana_metrics, grafana_logs) set by _map_grafana_*.
     if not failure_reason and answer_key.required_evidence_sources:
         evidence = final_state.get("evidence") or {}
+        performance_insights_tokens = (
+            "top sql activity",
+            "avg load",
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by cerencamkiran** [comment]
@greptile-apps re-review

**Comment by cerencamkiran** [comment]
@greptile-apps re-review

**Comment by cerencamkiran** [comment]
Ekran görüntüsü 2026-05-02 193636
Ekran görüntüsü 2026-05-02 193708

**Comment by cerencamkiran** [comment]
P1 is addressed.

`aws_performance_insights` is now validated via PI-specific signals (Top SQL, AAS, WALWrite), not generic `grafana_metrics`.

The text vs evidence asymmetry can be improved separately.

File: tests/synthetic/rds_postgres/run_suite.py [hunk 1]
```diff
@@ -327,7 +327,21 @@ def score_result(
...
+            "top sql activity",
+            "avg load",
+            "aas",
+            "db load",
+            "walwrite",
+        )
+
         for source_key in answer_key.required_evidence_sources:
+            if source_key == "aws_performance_insights":
+                if not any(token in normalized_output for token in performance_insights_tokens):
```
muddlebee: Keep the state evidence check for aws_performance_insights. Keyword-only output matching can pass on hallucinated AAS/db load mentions even when Performance Insights was never queried.

cerencamkiran:
Thanks for your feedback @muddlebee. Tested locally with both positive and negative cases.

- Metrics only (no PI signals) → correctly fails
- PI-like text without state evidence → correctly fails

So `aws_performance_insights` now requires both state evidence and PI-specific signals, preventing hallucinated passes.

PR #1218, opened 2026-05-02
fix(guardrails): stop persisting matched secrets in audit log (#1197)

### LINKED ISSUE #1197: [IMPROVEMENT] Prevent Secret Leakage in Guardrail Audit Logs
## 1. Vulnerability Overview
The `GuardrailEngine` is designed to be the final security boundary between internal infrastructure data and external LLM providers. However, a significant architectural oversight in the `AuditLogger` implementation (`app/guardrails/audit.py`) creates a secondary vulnerability by persisting the very secrets it is designed to redact into local, cleartext logs.

## 2. Technical Deep Dive (Code Analysis)
The vulnerability resides in the `AuditLogger.log` method. When a guardrail is triggered, the system passes the "offending" text to the logger.

**Location:** `app/guardrails/audit.py` (Lines 22-47)

```python
def log(self, *, rule_name: str, action: str, matched_text_preview: str, context: str = "") -> None:
    # ...
    preview = (
        matched_text_preview[:40] if len(matched_text_preview) > 40 else matched_text_preview
    )
    # ...
```

**The "40-Character Trap":**
This constant (40) is not just a display limit; it is a data capture window. In modern DevSecOps, 40 characters is the "Goldilocks Zone" for credentials:
- **AWS Secret Access Keys:** Exactly 40 characters.
- **GitHub Personal Access Tokens (Classic):** Exactly 40 characters.
- **Hex-encoded API Keys:** Frequently 32 or 64 characters.
By slicing at 40, the system is essentially performing a "credential harvest" of the user's most sensitive assets and writing them to disk.

## 3. Threat Modeling (STRIDE Analysis)

| Threat Category | Impact on OpenSRE |
| :--- | :--- |
| **Information Disclosure** | **CRITICAL**. Sensitive credentials (AWS, Slack, GitHub) are stored in cleartext at `~/.opensre/guardrail_audit.jsonl`. |
| **Elevation of Privilege** | **HIGH**. An attacker with local user access can read the audit log to obtain administrative cloud credentials, escalating from a local user to a Cloud Admin. |
| **Tampering** | **MEDIUM**. Since logs are cleartext JSONL, they can be modified to hide evidence of a security breach or a bypassed guardrail. |

## 4. Operational Risk & Compliance Impact
- **Credential Sprawl:** Local logs are often backed up, synced via Dropbox/iCloud, or gathered in diagnostic bundles. This spreads cleartext secrets across the developer's entire digital footprint.
- **Compliance Failure:** For any organization pursuing SOC2, HIPAA, or PCI-DSS, cleartext storage of secrets in a "security tool" is an automatic failure during an audit.
- **Data Residency (GDPR):** If the guardrail matches PII (Emails, Phone Numbers), storing them in an unencrypted log file violates the principle of "Privacy by Design."

### Desired state

# 🛡️ ARCHITECTURAL PROPOSAL: Secure, Zero-Knowledge Audit Instrumentation

The remediation strategy must follow the principle of **Defense in Depth**. We need to provide auditability without compromising the security posture of the host machine.

## 1. Phase 1: Immediate Data Sanitization
The first priority is to stop the leakage. The `matched_text_preview` should be immediately deprecated in favor of a `MatchMetadata` object.

**Logic Change Example:**
```python
# app/guardrails/audit.py
def log(self, ...):
    # REMOVE: matched_text_preview logic
    # ADD: Secure Fingerprinting
    fingerprint = hashlib.sha256(matched_text_preview.encode()).hexdigest()[:12]
```

## 2. Phase 2: Secure Fingerprinting (Deduplication)
To allow SREs to investigate without seeing the secret, we implement **Salted Hashing**:
- Generate a `SHA-256` hash of the matched text.
- Log only the first 12 characters of the hash (the "fingerprint").
- **Benefit:** If the same token leaks 100 times, the SRE sees 100 log entries with the *same* fingerprint, allowing them to identify a systemic leak without ever seeing the raw token.

## 3. Phase 3: Enhanced Audit Context
Replace the raw text with high-signal, low-risk metadata:
- **`rule_severity`**: (High/Med/Low).
- **`match_offset`**: The character position in the original stream (helps find where it happened).
- **`remediation_status`**: (e.g., "Blocked", "Masked", "Flagged").
- **`llm_provider`**: Which external service was the target of the prevented leak.

## 4. Phase 4: File System Hardening
- Ensure `mkdir` for the `.opensre` directory explicitly sets permissions to `700` (User Read/Write/Execute only).
- Implement an optional "Encrypted Audit Log" mode using Python's `cryptography` library for enterprise environments.

## 5. Verification & Testing Strategy
To prevent regression, we must add a new test suite:
- **Secret Detection Test:** Assert that synthetic keys **never** appear in cleartext in the `guardrail_audit.jsonl` file.
- **Fingerprint Consistency Test:** Verify that identical secrets produce identical fingerprints.
- **Permission Test:** Verify that the audit log file is created with restricted `0600` permissions.

---
**Note:** This is a security-focused improvement that ensures compliance with SOC2/GDPR data handling standards while maintaining auditability.

mayankbharati-ops:
Closes #1197.

## The bug

`AuditLogger.log` in `app/guardrails/audit.py:31-32` writes the literal matched text to `~/.opensre/guardrail_audit.jsonl`, capped at 40 chars:

```python
preview = matched_text_preview[:40] if len(matched_text_preview) > 40 else matched_text_preview
```

The 40-char window is exactly the wrong size — it captures **whole** AWS Secret Access Keys (40), classic **whole** GitHub PATs (40), and 32-hex API keys with bytes to spare. Every secret the redact rule just protected from the LLM gets persisted on disk in cleartext anyway, defeating the guardrail.

With the shipping `_STARTER_CONFIG` (which ships an `aws_access_key` rule), an investigation that mentions `AKIAIOSFODNN7EXAMPLE` produces an entry like:

```json
{"timestamp":"...","rule_name":"aws_access_key","action":"redact",
 "matched_text_preview":"AKIAIOSFODNN7EXAMPLE","context":""}
```

— i.e. the literal access key, in cleartext, in the audit log meant to record that we *protected* it.

## Fix

Replace `matched_text_preview` with two non-leaking fields:

| Field | Value |
|---|---|
| `match_fingerprint` | First 12 hex chars of `sha256(matched_text)`. Deterministic so identical secrets dedupe (an SRE can correlate "this 1234abcd…-fingerprint fired 47 times this week → systemic leak"), but irreversible — the original bytes can't be recovered from the log alone. |
| `match_length` | Character count. Useful for triage ("a 40-char match in `aws_secret_key`? probably a real key") without leaking content. |

The kwarg the engine passes is renamed `matched_text_preview` → `matched_text` to match what it always actually was — the raw match, not a preview.

### Defence-in-depth: filesystem perms

Rule names + timestamps + contexts are themselves disclosure (*"this user has been handling Slack tokens this week"*) even after fingerprinting the bytes. Restrict to owner-only:

- `~/.opensre` directory created with `mode=0o700` and re-`chmod`'d on every log call so a stale loose-perm dir **self-heals** — without this, the fix would only kick in for users who delete their existing `~/.opensre` first.
- Audit file `chmod 0o600` after write.
- Both chmods wrapped in `contextlib.suppress(OSError)` so they degrade silently on Windows (POSIX modes don't apply there; ACLs handle equivalent semantics).

## Test coverage

| Test class | Coverage |
|---|---|
| `TestAuditLogger` (existing, updated) | Schema migration + new field assertions |
| `TestNoSecretLeakageInAudit` (new, parametrised ×5) | AWS keys, GitHub PATs, credit cards, internal tokens, the legacy 40-char hex pattern: assert literal secret bytes (**and** the first 8-char prefix) never appear in the JSONL |
| `TestFingerprintProperties` (new ×3) | Determinism, distinct-input distinct-output, lowercase-hex format pinned |
| `TestAuditLogPermissions` (new ×3, POSIX-only) | File `0o600`, dir `0o700`, dir self-heals from pre-existing `0o755` |
| `test_engine.py` (existing, updated) | `test_audit_logger_records_every_match_even_when_merged` now verifies the merged-redaction match-level audit via `match_length` per rule **plus** a positive "literal text not on disk" assertion |
| `test_cli.py` (existing, updated) | `TestCmdAudit::test_shows_entries` updated for kwarg rename |

## Live verification

Ran a 19-check smoke against the real `GuardrailEngine.apply` with the shipping `_STARTER_CONFIG` patterns over a payload containing `AKIAIOSFODNN7EXAMPLE`, `ghp_aaa…`, and `4111-1111-1111-1111`:

| | |
|---|---|
| Engine output | All 3 secrets redacted, rule names appear in `[REDACTED:…]` placeholders ✅ |
| Audit JSONL on disk | None of the secret bytes (nor 8-char prefixes) appear anywhere in the file ✅ |
| Audit entries | 12-hex fingerprint + length on every entry; no `matched_text` / `matched_text_preview` field ✅ |
| Dedup property | The two `AKIA...` references produced **identical** fingerprints (`1a5d44a2dca1` × 2) ✅ |
| Distinct-secrets property | `github_pat` and `cc` fingerprints differ ✅ |
| Permissions | File `0o600`, dir `0o700` after live calls ✅ |

## Quality gate

- `pytest tests/test_guardrails/`: **102 / 102 pass** (was 91, +11 new)
- Full unit suite: **4600 pass / 2 skipped**
- `ruff check` and `ruff format --check` clean
- `mypy app/guardrails/` clean on touched modules (the 3 mypy errors are pre-existing missing-stub warnings for `pymysql` in `app/integrations/` — unrelated)

## Scope

| File | Change |
|---|---|
| `app/guardrails/audit.py` | rewrite: fingerprint instead of preview, perm hardening, fuller docstring |
| `app/guardrails/engine.py` | 1-line kwarg rename at the call site |
| `tests/test_guardrails/test_audit.py` | updated existing tests + 11 new security/perm regression tests |
| `tests/test_guardrails/test_engine.py` | `test_audit_logger_records_every_match_even_when_merged` updated to the new schema and tightened with a "no literal text on disk" check |
| `tests/test_guardrails/test_cli.py` | `test_shows_entries` updated for the kwarg rename |

## Scope deferred (per the issue's phased plan)

The issue's full proposal also includes Phase 3 ("`rule_severity`, `match_offset`, `remediation_status`, `llm_provider` metadata fields") and Phase 4 ("optional encrypted-audit-log mode using `cryptography`"). Both are valuable, but they're additive — Phase 1+2 (stop the bleeding + fingerprint) and the perm hardening that the issue groups as Phase 4 stand on their own as the security-critical core. Happy to file follow-up issues for the metadata expansion and the optional encryption mode if you'd like.

### DIFF
--- a/app/guardrails/audit.py
+++ b/app/guardrails/audit.py
@@ -9,27 +9,34 @@ Each entry stores:
 * ``timestamp`` — ISO-8601 UTC timestamp.
 * ``rule_name`` — the rule that fired.
 * ``action`` — ``redact`` / ``block`` / ``audit``.
-* ``match_fingerprint`` — first 12 hex characters of the SHA-256 hash of
-  the matched text. Deterministic (identical secrets produce identical
-  fingerprints, so an SRE can identify systemic leaks) but irreversible
-  (the original text cannot be recovered from the audit log).
+* ``match_fingerprint`` — first 12 hex characters of an HMAC-SHA-256 of the
+  matched text, keyed by a per-machine secret stored in
+  ``~/.opensre/.audit_key`` (``0o600``). Deterministic on the same machine
+  so identical secrets produce identical fingerprints (the SRE-dedup
+  property), but **opaque** to anyone who exfiltrates only the JSONL —
+  without the per-machine key the fingerprint cannot be recomputed even
+  for a known candidate secret. This defeats offline confirmation
+  attacks against fixed-format credentials (AWS Access Keys, GitHub
+  PATs).
 * ``match_length`` — character length of the matched text. Useful for
   triage ("a 40-char match in the ``aws_secret_key`` rule is probably a
   real key") without leaking the bytes.
 * ``context`` — caller-supplied tag (e.g. ``llm_invoke``, ``chat_node``).
 
 The audit file is written with ``0o600`` and lives under a ``0o700``
-directory, both owner-only, so that a non-root local user on the same
-machine cannot read the audit trail and reconstruct what was redacted.
+directory, both owner-only. The file is **created atomically** via
+``os.open(O_CREAT, mode=0o600)`` so there is no umask-controlled window
+between creation and ``chmod`` for an ``inotify`` watcher to read it.
 """
 
 from __future__ import annotations
 
 import contextlib
-import hashlib
+import hmac
 import json
 import logging
 import os
+import secrets
 from datetime import UTC, datetime
 from pathlib import Path
 from typing import Any
@@ -37,6 +44,7 @@ from typing import Any
 logger = logging.getLogger(__name__)
 
 _DEFAULT_AUDIT_PATH = Path.home() / ".opensre" / "guardrail_audit.jsonl"
+_DEFAULT_KEY_PATH = Path.home() / ".opensre" / ".audit_key"
 
 _FINGERPRINT_LEN = 12  # 12 hex chars × 4 bits = 48 bits of collision space.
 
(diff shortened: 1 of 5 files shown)


mayankbharati-ops:
Pushed `586c7a82` addressing both Greptile findings + the comments-outside-diff nit.

## P1 — TOCTOU on new file creation

Real issue, real fix. Switched from `open("a")` + deferred `chmod` to atomic `os.open(O_WRONLY|O_CREAT|O_APPEND, 0o600)`. The kernel applies `mode & ~umask`, but `0o600` has no group/other bits, so any reasonable umask leaves it intact — no need to touch the process-wide umask.

**Regression test that pins the window-is-closed property:**

```python
def test_log_file_is_NEVER_world_readable_at_any_point(self, tmp_path):
    observed_mode: list[int] = []
    real_fdopen = os.fdopen

    def _stat_at_fdopen_time(fd, *args, **kwargs):
        # Earliest moment an inotify watcher could observe the new file.
        observed_mode.append(stat.S_IMODE(log_path.stat().st_mode))
        return real_fdopen(fd, *args, **kwargs)

    with patch("app.guardrails.audit.os.fdopen", side_effect=_stat_at_fdopen_time):
        _make_logger(tmp_path).log(rule_name="r", action="audit", matched_text="x")

    assert observed_mode[0] == 0o600
```

That assertion would have failed in the previous revision (file was `0o644` until the post-write `chmod` ran).

## P2 — Unkeyed fingerprint allowed offline confirmation

Took the suggestion. New design:

- HMAC-SHA-256 keyed by 32 random bytes generated via `secrets.token_bytes(32)` on first use.
- Key persisted at `~/.opensre/.audit_key` (`0o600`, atomic creation via the same `os.open` helper).
- Co-located with the audit log so a single `~/.opensre/` directory holds the whole subsystem.
- If the key file can't be persisted (e.g. read-only `$HOME`), we fall back to a process-local key with a logged warning — entries are still HMAC'd, only cross-process dedup degrades.
- Truncated / partial key files are detected (`len != 32`) and regenerated.

Properties pinned by tests:

| Property | Test |
|---|---|
| Same input + same key → same fingerprint (dedup) | `test_fingerprint_is_deterministic_within_a_machine` |
| Same input + different machine keys → different fingerprints (rainbow-table defeat) | `test_distinct_machine_keys_produce_distinct_fingerprints` |
| `hashlib.sha256(secret)[:12]` ≠ keyed fingerprint (brute-force defeat) | `test_unkeyed_sha256_does_NOT_match_keyed_fingerprint` |
| Key file mode `0o600` | `test_key_file_is_owner_only` |
| Key persists across logger instances (dedup property survives restarts) | `test_key_file_is_persistent_across_logger_instances` |
| Corrupted/truncated key is regenerated | `test_truncated_key_file_is_regenerated` |

## P2 outside-diff — `test_handles_write_failure_gracefully`

Good catch — exactly right. The chmod self-heal promoted the test's `0o444` parent dir back to `0o700`, so the write was succeeding instead of failing. The test asserted `log()` didn't raise, which was true for the wrong reason.

Replaced with a `patch("app.guardrails.audit.os.open", side_effect=OSError(...))` test that genuinely exercises the exception-swallowing path. New assertions: `log()` doesn't raise, AND no partial file is left behind.

## Verification

- `pytest tests/test_guardrails/` — **107 / 107 pass** (was 102, +5 net new for TOCTOU, HMAC determinism, distinct-machine keys, unkeyed-mismatch, key persistence/regeneration)
- Full unit suite — 4605 pass / 2 skipped
- `ruff check` + `ruff format --check` clean
- `mypy app/guardrails/` clean on touched modules

Re-trigger Greptile if you want fresh eyes on the keyed-fingerprint design — happy to discuss the tradeoff (cross-machine fingerprint portability lost, but that's the security property we want).

app/guardrails/audit.py
```diff
@@ -1,46 +1,265 @@
...
+                try:
+                    _write_all(fd, new_key)
+                finally:
+                    os.close(fd)
+                return new_key
+        except OSError:
+            # If we cannot persist the key (read-only $HOME, AppArmor denial,
+            # disk full, etc.) fall back to an in-memory key so at least
+            # *this* process's audit entries are HMAC'd. The cost is that
+            # fingerprint dedup won't survive a restart.
```
muddlebee, line 194: If the winner created the file but has not finished writing, _read_existing() returns None and this path truncates a file another process is actively writing to. Retry the read a few times before falling back to overwrite.

muddlebee:
closing this, as the issue needs more community feedback

PR #1190, opened 2026-05-01
fix: resolve 10 verified bugs — model names, OpenSearch auth, JWT signature bypass, state mutation, silent error swallowing

harshitgavita-07:
Hey  @davincios  , @VaibhavUpreti  , @rrajan94 , 

 I am Harshit , A new Contributer to this repo so i find 10 bugs in the code space so below i have given the solution  to it .

## What this fixes

Ten bugs found by static analysis of the source. Each has an exact file
and line reference. Changes are minimal-diff — no refactors, no style
changes, only the targeted fix per bug. All 11 modified files pass
syntax and import checks.

cc @devincios @rrajan94 @vaibhavUpreti

---

### Fix 1 — Non-existent OpenAI model names
**Files:** `app/config.py:78–79`, `app/cli/wizard/config.py`

`gpt-5.4` and `gpt-5.4-mini` don't exist. `config.py` even had a
comment saying *"UNVERIFIED PLACEHOLDER"*. Every OpenAI user hit an
immediate model-not-found error on first call. Replaced with verified
May 2026 names: `gpt-4.1`, `gpt-4.1-mini`, `o3`, `o4-mini`.

---

### Fix 2 — OpenSearch integration completely broken
**Files:** `app/services/elasticsearch/client.py`, `app/integrations/config_models.py`

`ElasticsearchClient` only generated `ApiKey` headers. OpenSearch
clusters use `Authorization: Basic <base64(user:pass)>` by default.
Added `username`/`password` fields to `ElasticsearchConfig` with correct
Basic auth header generation. Also added `OpenSearchIntegrationConfig`
to `config_models.py` with auth-mode validation and a warning when no
credentials are set for a secured cluster.

---

### Fix 3 — Alert content silently dropped for Slack-style payloads
**File:** `app/nodes/extract_alert/extract.py:127`

`_format_raw_alert()` sent only `raw_alert["text"]` to the LLM for
Slack-style dicts, discarding timestamps (`ts`), channel, labels, and
all structured fields. Downstream investigation had no time anchor and
no service routing data. Fix: always pass the full JSON.

---

### Fix 4 — JWT payload decoded without signature verification
**File:** `app/nodes/resolve_integrations/node.py:34–45`

`_decode_org_id_from_token()` base64-decoded the JWT payload without
verifying the signature — anyone could forge any `organization` claim.
It also had no validation that the token had 3 segments, causing
`IndexError` on malformed tokens (silently caught). Fix: delegate to
[...]

PR #692, opened 2026-04-20
feat: Better Stack Telemetry (logs) integration

afif1400:
## Summary

Adds a first-class Better Stack Telemetry log-evidence integration, closing #678. Teams standardized on Better Stack can now investigate alerts through the OpenSRE planner with cited log evidence — no more manual dashboard pulls mid-RCA.

Uses Better Stack's ClickHouse SQL over HTTP endpoint (HTTP Basic auth) via the existing `httpx` dependency — no new runtime deps. A single call UNIONs `remote(<source>_logs)` (live stream) with `s3Cluster(primary, <source>_s3) WHERE _row_type = 1` (historical) — matching the sample curl Better Stack's dashboard generates on connection creation.

## What ships

**Integration module** — `app/integrations/betterstack.py`
- `BetterStackConfig` (StrictConfigModel) with query_endpoint / username / password / sources / timeout / max_rows
- `betterstack_config_from_env()`, `build_betterstack_config()`, `validate_betterstack_config()` (cheap `SELECT 1` probe with actionable 401 / 404 / 5xx / transport error strings)
- `query_logs()` — composes the UNION query, parses `FORMAT JSONEachRow`, returns standard evidence dict shape. Source-name + ISO-timestamp injection guards before SQL composition.

**Agent tool** — `app/tools/BetterStackLogsTool/` — `@tool`-decorated `query_betterstack_logs`; credentials via `extract_params`, runtime `source` via alert annotation through `detect_sources`.

**Framework wiring — 17 surfaces touched** so Better Stack-only runs produce correct, cited reports end-to-end:
- Resolution: `models.py`, `catalog.py` (5 hunks), `verify.py`, `cli.py`, `constants.py`
- Onboarding: `wizard/flow.py`, `integration_health.py`
- Tool typing: `types/evidence.py`
- Planner: `plan_actions/detect_sources.py` (source-hint from `betterstack_source` annotation), `build_prompt.py` (hint block), `node.py` (fallback candidates)
- Evidence pipeline: `investigate/processing/post_process.py` (mapper + summary)
- Diagnosis: `root_cause_diagnosis/prompt_builder.py` (allowed sources + "Better Stack Logs" section), `evidence_checker.py` (availability gate), `claim_validator.py` (citation extraction + `_has_any_logs`)
- Publish: `publish_findings/report_context.py` (catalog entry), `formatters/evidence.py` (ACTION_DEFS)
- CLI UX: `output.py` (spinner label)

[...]

PR #1223, opened 2026-05-02
[FIX] Wire OpenSearch end-to-end: client Basic Auth, catalog wiring, CLI field fix, wizard menu (#1143)

### LINKED ISSUE #1143: [BUG] OpenSearch integration is broken end-to-end: missing from onboard wizard, field name mismatch, and silently dropped credentials
### Summary

The OpenSearch integration is non-functional in real-world usage due to four interrelated issues spanning the onboarding wizard, the legacy CLI, the integration classifier, and the underlying Elasticsearch client. A user who follows the documented setup path cannot successfully connect OpenSRE to a self-hosted OpenSearch cluster, even though OpenSearch is listed as a supported integration and was added as a "bounded read-only slice" in PR #501.

Concretely:

1. **OpenSearch is not exposed in the `opensre onboard` wizard** at all, so users following the standard onboarding path never see it as an option.
2. **The legacy CLI (`python -m app.integrations setup opensearch`) stores the endpoint under the field name `endpoint`, but the catalog classifier reads `url`** — every record produced by the CLI is silently discarded by the classifier.
3. **The catalog classifier ignores `username`/`password` even when the CLI collects them** — the legacy CLI already prompts for "Username + Password" as one of two auth modes, but those credentials never reach the runtime.
4. **`ElasticsearchClient` only emits an `ApiKey` Authorization header** — there is no code path for HTTP Basic Auth, which is the primary authentication method for self-hosted OpenSearch deployments (API keys are not natively available in OpenSearch; see [OpenSearch RFC opensearch-project/security#4009](https://github.com/opensearch-project/security/issues/4009)).

The net effect is that the OpenSearch integration added in #501 is not reachable end-to-end from any documented user path, and even if it were reachable, it would only work against the small subset of OpenSearch deployments that have non-native API key support configured.

**Related:** #470 (Extend Elasticsearch support for OpenSearch analytics), #484 (Roadmap), #501 (Slice that landed the initial OpenSearch tool).

## Expected behavior

A user with a self-hosted OpenSearch cluster (the most common production deployment of OpenSearch) should be able to:

1. Run `opensre onboard`, see "OpenSearch" in the integrations menu, and select it.
2. Be prompted for the cluster URL and authentication credentials, including HTTP Basic Auth (`username` + `password`), which is the default authentication method documented by OpenSearch.
3. Have those credentials persisted to `~/.tracer/integrations.json` and loaded by the runtime.
4. See the OpenSearch tool become available in subsequent investigations and successfully query the cluster.

## Actual behavior

1. **Wizard:** OpenSearch is not present in the `opensre onboard` wizard at all. The integrations list in `app/cli/wizard/flow.py` (lines ~1419–1525) contains 19 integrations — Grafana, Datadog, Honeycomb, Coralogix, Slack, Discord, AWS, GitHub MCP, Sentry, Gitlab, Google Docs, Vercel, Better Stack, Jira, Alertmanager, OpsGenie, Notion, OpenClaw, "Skip for now" — but no OpenSearch and no Elasticsearch.

2. **Legacy CLI field mismatch:** Running `python -m app.integrations setup opensearch` does work and even prompts for "Username + Password" as one of two auth choices. However, `_setup_opensearch` in `app/integrations/cli.py` (line 240) stores the value under the key `endpoint`:
```python
   creds: dict[str, Any] = {"endpoint": endpoint}
```
   while `_classify_service_instance` for `opensearch` in `app/integrations/catalog.py` (line 763) reads from `url`:
```python
   url = str(credentials.get("url", "")).strip()
   if not url:
       return None, None
```
   The classifier returns `(None, None)` for every record produced by the CLI, so the integration is silently dropped before it reaches the runtime.

3. **Credentials silently dropped:** Even if a user manually edits `~/.tracer/integrations.json` to use the field name `url`, the classifier in `app/integrations/catalog.py` (lines 762–770) only forwards `url`, `api_key`, `index_pattern`, and `max_results` to the runtime. Any `username`/`password` collected by the CLI is dropped on the floor.

4. **No Basic Auth in the client:** `ElasticsearchConfig.headers` in `app/services/elasticsearch/client.py` only constructs an `Authorization: ApiKey {api_key}` header. There is no code path that constructs an `Authorization: Basic ...` header, even if `username`/`password` were to be made available.

## User impact

Users running self-hosted OpenSearch clusters — which is the standard OpenSearch deployment model in enterprise SRE environments — currently cannot connect OpenSRE to their logs at all. In my own case, my team runs OpenSearch 3.1.0 for application logs (no Loki, no Elasticsearch). When I tried to wire OpenSearch into OpenSRE so investigations could pull from our actual log store, I discovered that:

- The wizard never offered OpenSearch as an option.
- The default Loki/Grafana path is irrelevant to my environment.
- There is no documented way to make the existing OpenSearch tool reachable end-to-end.

This effectively excludes a large segment of the project's intended audience.

### Steps to reproduce

**Environment**
- Branch: `main` (commit at the time of writing)
- Python: 3.13 in a fresh `.venv`
- OS: macOS (also reproducible on Linux per code inspection)

**Reproduction A — OpenSearch is missing from the onboard wizard**

1. Clone the repo and install dev dependencies:
```bash
   git clone https://github.com/Tracer-Cloud/opensre.git
   cd opensre
   python -m venv .venv && source .venv/bin/activate
   pip install -e ".[dev]"
```
2. Run the onboarding wizard:
```bash
   opensre onboard
```
3. Step through to the "Integrations" prompt.
4. **Observed:** OpenSearch is not in the list. There is no way to configure it from the wizard.

**Reproduction B — Legacy CLI field name mismatch**

1. With the same setup as above, run:
```bash
   python -m app.integrations setup opensearch
```
2. Enter:
   - Endpoint: `https://my-opensearch.internal:9200`
   - Auth method: select **"Username + Password"** (the first option offered)
   - Username: `admin`
   - Password: `<any password>`
3. Inspect the saved record:
```bash
   cat ~/.tracer/integrations.json
```
   You will see something like:
```json
   {"service": "opensearch", "credentials": {"endpoint": "https://my-opensearch.internal:9200", "username": "admin", "password": "..."}}
```
4. Now run any investigation flow that would resolve integrations, e.g.:
```bash
   python -m app.integrations list
```
   The record exists in storage, but at classification time it is filtered out because the classifier looks for `credentials["url"]` and the CLI stored `credentials["endpoint"]`. The integration is silently absent from the resolved set.

**Reproduction C — Even with the field name fixed, credentials are dropped**

1. Manually edit `~/.tracer/integrations.json` and rename `endpoint` → `url`.
2. Inspect the resolved integration via the CLI's effective view (or programmatically via `resolve_effective_integrations`).
3. **Observed:** the resolved OpenSearch config contains `url`, `api_key` (empty), `index_pattern`, `max_results`, `integration_id` — **but not `username` or `password`**. They were collected by the CLI, persisted to disk, and then dropped by the classifier at line 762 of `app/integrations/catalog.py`.

**Reproduction D — Even with credentials forwarded, the client cannot use them**

1. Add a `print` statement (or breakpoint) inside `ElasticsearchConfig.headers` in `app/services/elasticsearch/client.py`.
2. Construct an `ElasticsearchConfig` with `username` and `password` set but `api_key` empty.
3. **Observed:** the returned headers dict contains only `Content-Type`. There is no `Authorization` header, because the existing branch only handles `api_key`. Any HTTP request issued by the client will be rejected by the OpenSearch security plugin with a 401.

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

_No response_

## Why HTTP Basic Auth specifically

OpenSearch and Elasticsearch share the same REST API surface but diverge meaningfully on authentication:

| Auth method | Elasticsearch | OpenSearch (self-hosted) |
|---|---|---|
| HTTP Basic Auth | Supported | **Default and primary method** |
| API Keys (`POST /_security/api_key`) | First-class (since 7.x) | **Not natively supported** |
| SAML / OIDC / JWT | Supported | Supported |

API key support for OpenSearch is still an open RFC ([opensearch-project/security#4009](https://github.com/opensearch-project/security/issues/4009)) and is not present in any current production OpenSearch release including 3.1.0. The OpenSearch Security plugin documentation lists Basic Auth and token-based auth as the primary mechanisms.

Forum confirmation from the OpenSearch community:

> *"In Elasticsearch, this can be done by using `POST /_security/api_key` but it does not work in OpenSearch."* — OpenSearch Forum, 2023
>
> *"Am I correct that OpenSearch currently has no compatible feature to Elastic's API Keys?"* — OpenSearch Forum, 2024

For the vast majority of self-hosted OpenSearch deployments, HTTP Basic Auth is the only authentication method available out of the box.

## Existing precedent in the codebase

The project already implements a "token-first / Basic Auth fallback" pattern for **OpenObserve** in the same file. See `app/integrations/catalog.py` lines 521–540 (classify) and lines 1483–1502 (env loader):

```python
if not (base_url and (api_token or (username and password))):
    return None, None
```

This is the same pattern that is needed for OpenSearch. The proposed fix would not introduce any new architectural concept — it would simply apply the existing OpenObserve pattern to OpenSearch.

## Suggested fix scope

A complete fix would touch four layers:

1. **`app/cli/wizard/flow.py`** — add an `opensearch` choice and `_configure_opensearch` handler to the integrations wizard, mirroring `_configure_alertmanager` (which already supports both bearer and basic auth).
2. **`app/integrations/cli.py`** — change `_setup_opensearch` to store the URL under `url` (not `endpoint`); the username/password collection is already correct and can stay.
3. **`app/integrations/catalog.py`** — update both `_classify_service_instance` (line 762) and the env loader (line 1502) to forward `username` and `password`, and require either `api_key` or (`username` and `password`) — same shape as OpenObserve.
4. **`app/services/elasticsearch/client.py`** — extend `ElasticsearchConfig` with optional `username`/`password` fields and add a `Basic` branch in the `headers` property:
```python
   if self.api_key:
       h["Authorization"] = f"ApiKey {self.api_key}"
   elif self.username and self.password:
       credentials = base64.b64encode(f"{self.username}:{self.password}".encode()).decode()
       h["Authorization"] = f"Basic {credentials}"
```
5. **Tests** — focused regressions for: API-key-only path (existing behavior preserved), Basic-auth-only path, both-provided precedence (API key wins), neither-provided no-auth path, and CLI→classifier→runtime end-to-end coverage with both auth modes.

The change is fully backward compatible: existing API-key configurations continue to work unchanged, and the new fields are optional.

## Willingness to contribute

I'd like to take this on as a follow-up to my recent PR #798. Happy to scope it tightly to these four layers, mirror the existing OpenObserve pattern exactly, and split into smaller PRs if the maintainers prefer.

Sarah-Salah:
## Summary

Wires OpenSearch / Elasticsearch end-to-end through the OpenSRE integration stack. Before this PR, OpenSearch was broken at every layer: the wizard didn't expose it, the legacy CLI dropped credentials due to a field-name mismatch, the catalog ignored Basic Auth entirely, the underlying client only emitted API-key headers, and the AI agent's tools dropped Basic Auth credentials at runtime. This PR fixes each layer and ties them together.

Closes #1143.

## Why this is needed

OpenSearch is an open-source fork of Elasticsearch and a primary logs/search backend for many on-prem and air-gapped deployments. Most self-hosted OpenSearch clusters do not natively support API keys ([opensearch-project/security#4009](https://github.com/opensearch-project/security/issues/4009)) and authenticate via HTTP Basic Auth instead. The previous code path silently dropped Basic Auth credentials at multiple layers, so any user who tried to use OpenSearch saw the integration "succeed" at setup time and then fail to authenticate at runtime, with no clear error path.

## Changes by layer

This PR is intentionally split into five commits, one per layer, so each layer can be reviewed and reasoned about independently.

### Layer 1 — `app/services/elasticsearch/client.py`

`ElasticsearchConfig` now supports HTTP Basic Auth in addition to API key.

- Added `username` and `password` fields to `ElasticsearchConfig`.
- Updated the `headers` property to emit `Authorization: Basic <base64(user:pass)>` when only Basic credentials are set.
- API key takes precedence when both are configured. Incomplete credentials (one without the other) emit no `Authorization` header.
- Module docstring documents the three supported auth modes.

### Layer 2 — `app/integrations/_catalog_impl.py`

The catalog now forwards Basic Auth credentials end-to-end.

- The `opensearch` branch in `_classify_service_instance` now reads `username` and `password` from credentials and includes them in the runtime config alongside `url` and `api_key`. URL remains the only hard requirement; auth is optional.
- The `opensearch` entry in `load_env_integrations` now reads `OPENSEARCH_USERNAME` and `OPENSEARCH_PASSWORD` from the environment.
- This mirrors the OpenObserve pattern already established in the same file.

### Layer 3 — `app/integrations/cli.py`

Critical bug fix: the legacy CLI's `_setup_opensearch()` was storing the URL under a key named `endpoint`, but the catalog classifier reads `url`. This silent field-name mismatch meant credentials entered through `opensre integration setup opensearch` were dropped before reaching the runtime, regardless of which auth method the user picked.

- Renamed the stored field from `endpoint` to `url`.
- Updated the prompt label to match.
- Replaced numeric `auth_choice` values (`"1"`, `"2"`) with semantic ones (`"basic"`, `"api_key"`, `"none"`).
- Added a `None (security disabled)` auth option (the classifier already accepts URL-only configurations).
- Added explicit validation: empty URL, empty API key when `api_key` is selected, and empty username/password pair when `basic` is selected.

### Layer 4 — `app/cli/wizard/flow.py` + helpers

OpenSearch is now visible and configurable through `opensre onboard`.

- New `validate_opensearch_integration()` in `app/cli/wizard/integration_validators/client_validators.py` probes the cluster via `GET /_cluster/health`, supporting all three auth modes.
- The validator is re-exported from the `integration_health` facade.
- New `_configure_opensearch()` handler in `flow.py` prompts for URL, auth method, and credentials; validates with the cluster; and persists to both the integration store and `.env`.
- OpenSearch is added to `integration_choices`, `handlers`, and `_SERVICE_LABELS`.
- Default auth method is Basic Auth (most common for self-hosted OpenSearch).

### Layer 5 — `app/tools/ElasticsearchLogsTool/` + `app/tools/OpenSearchAnalyticsTool/`

The AI agent's OpenSearch-facing tools now thread Basic Auth credentials through to runtime.

- `ElasticsearchLogsTool/_client.py`: `make_client()` accepts `username` and `password` kwargs and forwards them to `ElasticsearchConfig`. Both default to `None`, so existing callers continue to work unchanged.
- `ElasticsearchLogsTool/__init__.py`: `input_schema` declares `username` and `password` as optional properties; `extract_params()` reads them from the source dict; `run()` accepts and forwards them to `make_client()`.
- `OpenSearchAnalyticsTool/__init__.py`: `query_opensearch_analytics` receives the same treatment — `input_schema`, `_opensearch_extract_params`, the function signature, and the `ElasticsearchConfig` construction all thread `username` and `password` through.

Without Layer 5, even though credentials are stored correctly (Layers 1–4), the AI agent's tools dropped them when constructing the runtime client, so the LLM could never authenticate against the cluster during investigations.

## Tests

| Layer | Tests added |
|---|---|
| 1 | 5 new tests in `tests/test_elasticsearch_client.py` covering Basic Auth header emission, API-key precedence, incomplete credentials, and special characters. |
| 2 | 7 new tests in `tests/integrations/test_opensearch_catalog.py` covering API-key classification, Basic Auth classification, no-auth (security-disabled), missing-URL rejection, trailing-slash normalization, and end-to-end env-loader-to-effective-view flow. |
| 4 | 2 new tests in `tests/cli/wizard/test_flow.py` covering the happy path (URL + Basic Auth → store + env vars) and the retry path (validation failure → re-prompt → success). The legacy facade snapshot test in `tests/cli/wizard/test_integration_health.py` is updated. The smoke test `tests/cli_smoke_test.py::test_onboard_interactive_smoke` `stagger_j` is bumped from 20 to 21 to account for the new menu entry. |
| 5 | 5 new tests across two files: `tests/tools/test_elasticsearch_logs_tool.py` adds three tests covering the `extract_params` → `run` → `make_client` → `ElasticsearchConfig` chain for `ElasticsearchLogsTool`. `tests/tools/test_integration_wave_tools.py` adds two tests for `OpenSearchAnalyticsTool` covering the `extract_params` and `ElasticsearchConfig` paths. |

Layer 3 follows the existing convention in `app/integrations/cli.py`, where the `_setup_*` functions are not unit-tested directly. The contract Layer 3 produces (a record stored under the `url` key) is verified end-to-end by the Layer 2 classifier tests.

**Full suite:** all tests pass across `tests/integrations/`, `tests/cli/wizard/`, `tests/tools/`, and `tests/test_elasticsearch_client.py`. Lint, format, and type-check (`mypy`, 452 source files) all clean.

## Backward compatibility

- API-key-only configurations continue to work unchanged. The new Basic Auth fields default to empty / `None` at every layer.
- `OPENSEARCH_API_KEY` still works as before.
- One user-facing impact: existing stored records using the old `endpoint` key (Layer 3) will no longer authenticate after upgrading. These records were already non-functional — the classifier was silently dropping them — so re-running `opensre integration setup opensearch` produces a working configuration. No migration script is needed because the user must re-enter credentials anyway.

## How to test manually

```bash
# Option A — env vars
export OPENSEARCH_URL=https://my-cluster.example.com
export OPENSEARCH_USERNAME=admin
export OPENSEARCH_PASSWORD=secret
opensre integration verify opensearch

# Option B — onboard wizard
opensre onboard
# pick "OpenSearch / Elasticsearch" from the list

# Option C — legacy CLI
opensre integration setup opensearch
After any of the above, the AI agent’s query_elasticsearch_logs and query_opensearch_analytics tools will authenticate to the cluster using the configured credentials.
Related
	∙	Issue: #1143
	∙	Reference: opensearch-project/security#4009 (why Basic Auth is the primary path for self-hosted OpenSearch)
	∙	Pattern reference: OpenObserve integration in the same _catalog_impl.py file uses the same “token-first / Basic-Auth-fallback” approach.

### CODE REVIEW

**Comment by yashksaini-coder** [comment]
@Sarah-Salah Please attach a demo recording of the test, and also make sure to complete all pending copilot and greptile review requests as well. 

Tested locally pulling docker image, 
image

File: app/tools/OpenSearchAnalyticsTool/__init__.py [hunk 1]
```diff
@@ -26,6 +26,8 @@ def _opensearch_extract_params(sources: dict[str, dict[str, Any]]) -> dict[str,
     return {
         "url": str(source.get("url", "")).strip(),
         "api_key": str(source.get("api_key", "")).strip(),
+        "username": str(source.get("username", "")).strip(),
```
yashksaini-coder, line 29: This reads `username`/`password` from the `sources["opensearch"]` dict, but `app/nodes/plan_actions/detect_sources.py:884-895` (not touched by this PR) doesn't put them there. The catalog stores the credentials correctly, the wizard validates correctly, but `detect_sources()` only forwards `url` and `api_key` into the runtime sources dict, so by the time we reach this line at runtime, `source.get("username")` is `""` and the tool drops the user's Basic Auth credentials before constructing `ElasticsearchConfig`.

The adjacent `openobserve` branch at `detect_sources.py:850-866` does forward both fields. Fix is the same shape there. I covered the full diff in the top-level review body. Worth verifying that adding this fix is in scope for this PR (it's the only path that actually exercises Basic Auth in production), or filing a follow-up if you want to keep this PR scoped to the layers it currently touches. If you go with the follow-up, please block the merge of this PR on the follow-up landing first, otherwise the integration ships broken.

app/cli/wizard/flow.py
```diff
@@ -1440,6 +1446,82 @@ def _configure_splunk() -> tuple[str, str]:
...
+        elif auth_choice == "basic":
+            username = _prompt_value(
+                "OpenSearch username",
+                default=_string_value(credentials.get("username"), "admin"),
+            )
+            password = _prompt_value(
+                "OpenSearch password",
+                default=_string_value(credentials.get("password")),
+                secret=True,
+            )
```
yashksaini-coder, line 1645: Replying to greptile's P2 above: I think this one is a false positive. The empty-credential case can't actually reach the validator because `_prompt_value` (defined at `flow.py:306-333`) re-prompts on empty input by default. It only allows empty when `allow_empty=True`, which isn't set on either the username or password prompts at lines 1487-1495. The password prompt also has no default on first run (`credentials.get("password")` returns `None`), so pressing Enter is rejected with `[red]Required.[/]` and the prompt loops. Could you mark greptile's comment as resolved with a one-line explanation pointing at the prompt loop?

tests/cli_smoke_test.py
```diff
@@ -481,7 +481,7 @@ def test_tests_inventory_commands_smoke(cli_sandbox: CliSandbox) -> None:
 def test_onboard_interactive_smoke(cli_sandbox: CliSandbox) -> None:
     # One `j` per keypress (burst writes are not separate keys). The select list wraps;
     # from the first option, len(choices)-1 steps reach "Skip for now" without wrapping past it.
-    # 19 integrations + "Skip for now" = 20 choices → 19 j's from the top.
+    # 20 integrations + "Skip for now" = 21 choices → 21 j's from the top.
```
yashksaini-coder: Tiny one. The comment says "20 integrations + Skip for now = 21 choices, 21 j's from the top." After this PR adds OpenSearch it's actually 21 integrations + Skip = 22 choices. The `stagger_j=21` value is correct (21 down-arrows lands on the 22nd item, which is OpenSearch), but the comment math will confuse the next maintainer. Worth updating to something like "21 integrations + Skip = 22 choices, OpenSearch is at index 21."

tests/services/test_elasticsearch_client.py
```diff
@@ -35,6 +35,35 @@ def test_headers_with_api_key(self) -> None:
         cfg = ElasticsearchConfig(url="http://localhost:9200", api_key="my-key")
         assert cfg.headers["Authorization"] == "ApiKey my-key"
 
+    def test_headers_with_basic_auth(self) -> None:
```
yashksaini-coder, line 38: Layout nudge, not blocking. AGENTS.md organizes tests by capability boundary, and test files for `app/services/<vendor>/client.py` typically live under `tests/services/`. This file sits at `tests/` root which predates this PR but is now drifting further from the documented layout. If you're touching it for the new Basic Auth tests anyway, worth moving it to `tests/services/test_elasticsearch_client.py` for consistency with the rest of the layout.

yashksaini-coder (changes requested):
Tested locally with a Docker `opensearchproject/opensearch:latest` single-node container (security disabled), wizard happy path works end-to-end and persists `OPENSEARCH_URL` to `.env`. Code structure is good, the 5-commit split per layer makes this very reviewable. Worth keeping the explanation of the `endpoint` -> `url` rename in the commit message so future bisects find it: that field-name mismatch (CLI from PR #692, catalog from PR #1165, two weeks apart) is a real silent break, not just a cleanup.

## Blocker: runtime path drops Basic Auth credentials

The PR says it wires OpenSearch end-to-end, but `app/nodes/plan_actions/detect_sources.py:884-895` (a file this PR doesn't touch) only forwards `url` and `api_key` when building the runtime `sources["opensearch"]` dict. Username and password get dropped between the catalog (Layer 2, where they're stored correctly) and the agent's `OpenSearchAnalyticsTool` (Layer 5, which is correctly threaded). So any Basic-Auth cluster will validate fine in the wizard, persist correctly, and then 401 the agent at runtime during alert investigations.

Compare to the adjacent `openobserve` block at `detect_sources.py:850-866` which does forward `username`/`password`. The fix here is the same shape:

```python
sources["opensearch"] = {
    "url": opensearch_url.rstrip("/"),
    "api_key": str(opensearch_int.get("api_key", "")).strip(),
    "username": str(opensearch_int.get("username", "")).strip(),
    "password": str(opensearch_int.get("password", "")).strip(),
    "index_pattern": (
        annotations.get("opensearch_index_pattern")
        or opensearch_int.get("index_pattern", "*")
    ),
    "max_results": safe_int(opensearch_int.get("max_results", 100), 100),
    "integration_id": str(opensearch_int.get("integration_id", "")).strip(),
}
```

Plus a regression test that sets `username`/`password` on the catalog record and asserts they end up in the `sources["opensearch"]` dict produced by `detect_sources()`. The new unit tests under `tests/tools/` mock `sources` directly, so they don't catch this gap.

## AGENTS.md New Integration Checklist

Three items missing per [AGENTS.md](https://github.com/Tracer-Cloud/opensre/blob/main/AGENTS.md) under "Adding an Integration":

- `docs/opensearch.mdx` for user-facing setup and verification docs
- An e2e or synthetic test exercising the integration
- A demo screenshot/GIF in the PR description

## False positive worth dismissing

Greptile's P2 about the empty-credential guard at `flow.py:1495` is a false positive. `_prompt_value` (line 306-333) loops on empty input and the password prompt doesn't pass `allow_empty=True`, so reaching the validator with an empty password isn't possible. Worth replying to that bot comment to dismiss it. I left an inline note there too.

Sarah-Salah:
Hey @yashksaini-coder! 👋

Thanks for the thorough review and for taking the time to test locally with the Docker container — really appreciated. Addressed every item in this round.

## Blocker: `detect_sources.py` dropping Basic Auth credentials

Fixed in 9438a34a. The runtime `sources["opensearch"]` dict now forwards `username` and `password` alongside `url` and `api_key`, mirroring the openobserve branch exactly as you suggested.

Added three regression tests in `tests/nodes/plan_actions/test_detect_sources_integration_wave.py`:
- `test_detect_sources_forwards_opensearch_basic_auth_credentials` — the direct regression
- `test_detect_sources_opensearch_basic_auth_strips_whitespace` — matches existing `api_key` whitespace handling
- `test_detect_sources_opensearch_no_auth_yields_empty_credentials` — security-disabled / URL-only config

You were right that the previous Layer 5 unit tests under `tests/tools/` mocked the sources dict directly — that's why they didn't catch this gap. The new tests exercise `detect_sources()` end-to-end.

## AGENTS.md New Integration Checklist

All three items addressed:

- ✅ **Docs:** `docs/opensearch.mdx` added in 65f61cc7. Covers prerequisites, four setup options (wizard / legacy CLI / env vars / persistent store), verification, how it works in investigations, troubleshooting, and security best practices. Patterned after `alertmanager.mdx` for consistency.
- ✅ **Synthetic test:** `tests/synthetic/test_opensearch_synthetic.py` added in 9e654132. Four scenarios drive the deterministic plumbing layer end-to-end (alert → resolve_integrations → detect_sources → tool extract_params): Basic Auth, API key, no-auth, and multi-integration coexistence. Unlike the EKS suite, these don't exercise the LLM planner — that's a reasonable follow-up.
- ✅ **Demo recording:** attached below.

## Smaller items

- ✅ **Smoke test comment math** fixed in f40356f1.
- ✅ **Test file relocation** to `tests/services/` done in 3d27495e (pure rename, 37 tests still passing).
- ✅ **Greptile P2** replied to and marked resolved with the explanation about `_prompt_value` re-prompting on empty input.

## CodeQL alert — request for manual dismissal

After pushing the blocker fix in 9438a34a, CodeQL started flagging the unrelated `logger.warning(grafana_hint)` call at line 491 as a "clear-text logging of sensitive information" alert. This is a confirmed false positive:

- `grafana_hint` is a user-provided string identifier from alert annotations (e.g. `"prod"`, `"staging"`), never credential material.
- The opensearch `password` CodeQL traces back to this logger flows through `sources["opensearch"]["password"]` later in the same function and is read by `ElasticsearchConfig` at the runtime layer — never logged.
- The adjacent `openobserve`, `alertmanager`, and `argocd` branches in the same file all store `password` in their respective sources dicts identically, with no CodeQL alerts. The new opensearch entry differs only in being recently added by this PR.
- I attempted three suppression strategies (`# codeql[...]`, `# lgtm[...]`, and breaking the data-flow with `str()` concatenation). The first two were not honored by the GitHub Code Scanning workflow, and the third broke an existing test that asserts on the warning message format. Reverted to the original logger format in 4b89d999.

I don't have permissions to dismiss the alert from the fork. Could you dismiss it as a false positive when you have a moment? Happy to provide additional explanation if useful.

## Summary

11 commits on the branch now. Full test suite passes (2300+ tests). Lint, format, typecheck all clean. CodeQL alert is the only outstanding item, pending manual dismissal. Ready for another round whenever you have a moment 🙏

---

### Demo recording


https://github.com/user-attachments/assets/77bda20c-e3cc-4b3c-b97c-49e96b6a8e84

app/cli/wizard/flow.py
Sarah-Salah: Confirmed — `_prompt_value` (lines 306-333) loops on empty input by default since `allow_empty=True` is not set on either prompt at lines 1487-1495. The validator at line 1495 is only reachable with non-empty values, so the empty-credential branch can't actually fire.

Marking the Greptile thread as resolved with this explanation.

app/tools/OpenSearchAnalyticsTool/__init__.py
Sarah-Salah: Fixed in 9438a34a — `detect_sources.py` now forwards `username` and `password` alongside `url` and `api_key`, matching the openobserve branch you pointed at. Three regression tests added in `test_detect_sources_integration_wave.py` to lock the behavior in. Full explanation in the top-level comment 👆

Marking as resolved.

tests/cli_smoke_test.py
Sarah-Salah: Good catch, fixed in f40356f1. Comment now reads:

```python
# 21 integrations + "Skip for now" = 22 choices. OpenSearch is at
# index 21, so 21 j's lands on the new "Skip for now" position.
```

tests/services/test_elasticsearch_client.py
Sarah-Salah: Done in 3d27495e — moved to `tests/services/test_elasticsearch_client.py` alongside the other client tests. Pure rename, all 37 tests still passing. No imports referenced the old path.

tests/synthetic/test_opensearch_synthetic.py
```diff
@@ -0,0 +1,177 @@
...
+    opensearch = sources["opensearch"]
+
+    assert opensearch["url"] == "https://opensearch.internal.example.com"
+    assert opensearch["username"] == "opensre-readonly"
+    assert opensearch["password"] == "secret-pass-123"
+    assert opensearch["api_key"] == ""
+    assert opensearch["connection_verified"] is True
+
+    logs_tool = ElasticsearchLogsTool()
+    sources["elasticsearch"] = {**opensearch, "default_query": "*"}
```
yashksaini-coder: This `sources["elasticsearch"] = {**opensearch, ...}` line is hiding a real bug. `ElasticsearchLogsTool.is_available` (`app/tools/ElasticsearchLogsTool/__init__.py:67-68`) reads from `sources["elasticsearch"]`, but `detect_sources` writes to `sources["opensearch"]`. Without this manual line in the test, `logs_tool.extract_params(sources)` would never have credentials to extract because the tool's source key never gets populated by the runtime path.

Drop this line and watch the test fail; that's your bug. Then fix it in `detect_sources.py` (suggestions in my top-level review body), and the test will go green for the right reason: both tools genuinely reachable from a single opensearch wizard configuration.

Also re: the placement under `tests/synthetic/`, the file's own docstring at lines 12-14 acknowledges this isn't a real synthetic test (no LLM, no mocked HTTP). Worth moving to `tests/nodes/plan_actions/test_opensearch_pipeline.py` so future contributors don't mistake the AGENTS.md "e2e or synthetic test" checkbox as actually ticked.

docs/opensearch.mdx
```diff
@@ -0,0 +1,140 @@
...
+      }
+    }
+  ]
+}
+```
+
+## Verify
+
+```bash
+opensre integrations verify opensearch
```
yashksaini-coder, line 94: This command will fail. `opensearch` is registered in `SETUP_SERVICES` (`app/cli/support/constants.py:29`) but not in `VERIFY_SERVICES` (lines 54-82), and the verify command uses `type=click.Choice(VERIFY_SERVICES)` to validate the argument. Running this exact command from the doc returns `Error: Invalid value for '<service>'`.

Quick fix: add `"opensearch"` to `VERIFY_SERVICES` (alphabetized between `opsgenie` and `postgresql`). The validator already exists at `app/cli/wizard/integration_validators/client_validators.py:454`, so it's a one-line registration.

Also, the example output at lines 97-100 doesn't match what the validator actually prints. The real format is `Connected to OpenSearch cluster '<name>' (<status>, <N> node(s)).` (see `client_validators.py:482-487`). Worth wrapping in a code fence with the actual string.

tests/cli_smoke_test.py
```diff
@@ -481,7 +481,8 @@ def test_tests_inventory_commands_smoke(cli_sandbox: CliSandbox) -> None:
 def test_onboard_interactive_smoke(cli_sandbox: CliSandbox) -> None:
     # One `j` per keypress (burst writes are not separate keys). The select list wraps;
     # from the first option, len(choices)-1 steps reach "Skip for now" without wrapping past it.
-    # 19 integrations + "Skip for now" = 20 choices → 19 j's from the top.
+    # 21 integrations + "Skip for now" = 22 choices. OpenSearch is at index 21,
```
yashksaini-coder: The math fixed the count but the index claim is still off. Counting `integration_choices` in `flow.py`, OpenSearch sits at index 20, not 21 (Skip for now is at index 21). The 21 j-presses still lands on the right spot (Skip), but the comment's reasoning misstates the position. Should read something like "OpenSearch is at index 20; 21 j's lands on the new 'Skip for now' position at index 21." Pure comment fix, no code change.

tests/nodes/plan_actions/test_detect_sources_integration_wave.py
```diff
@@ -150,3 +150,62 @@ def test_detect_sources_skips_snowflake_without_token() -> None:
...
+        },
+    }
+
+    sources = detect_sources(alert, {}, integrations)
+
+    opensearch = sources.get("opensearch")
+    assert opensearch is not None
+    assert opensearch["api_key"] == ""
+    assert opensearch["username"] == ""
+    assert opensearch["password"] == ""
```
yashksaini-coder, line 211: Regression coverage for the bug I caught is solid (the new test will fail without your +2 lines in detect_sources.py). One gap worth closing: there's no test for half-populated credentials, e.g. catalog record `{"username": "admin", "password": ""}`. The wizard's `_prompt_value` blocks this UI-side, but a record stored via env vars (`OPENSEARCH_USERNAME=admin` set, `OPENSEARCH_PASSWORD` unset) will still produce a half-populated source dict. `ElasticsearchConfig.headers` then silently emits no Authorization header (per the existing test at `tests/services/test_elasticsearch_client.py:51-57`), so the agent sends an unauthenticated request and the cluster 401s. Worth adding a case that either normalizes the half-pair to fully empty or explicitly raises.

yashksaini-coder (changes requested):
Second-pass review on HEAD `4b89d999`. Thanks for the thorough response to the first round, all four of my prior inline comments are resolved cleanly: `detect_sources` now forwards `username`/`password`, `docs/opensearch.mdx` exists, the test file moved to `tests/services/`, and the smoke comment was rewritten. Greptile's empty-credential P2 was correctly closed too.

Unfortunately the second pass surfaced **three new blockers** plus a few concerns. The biggest one is the same shape as my prior critical finding: credentials are stored correctly but a downstream consumer reads from the wrong key.

---

## Blockers

### 1. `ElasticsearchLogsTool` is unreachable from an opensearch config

`app/tools/ElasticsearchLogsTool/__init__.py:67-71` checks `sources.get("elasticsearch", {}).get("connection_verified")` and reads from `sources["elasticsearch"]`, but the new `detect_sources` opensearch branch (lines 884-899) writes to `sources["opensearch"]`. So a user who configures the integration through the wizard ends up with their credentials in `sources["opensearch"]`, and `query_elasticsearch_logs` is **never available** to the agent.

Your synthetic test `tests/synthetic/test_opensearch_synthetic.py:76` papers over this by manually doing `sources["elasticsearch"] = {**opensearch, ...}` rather than letting `detect_sources` produce it. So the test passes without revealing the gap.

The PR description claims "OpenSearch / Elasticsearch end-to-end" but only `OpenSearchAnalyticsTool` (`source = "opensearch"` in its definition) actually wires through. The Elasticsearch tool's name implies it covers both, but the new code path only feeds OpenSearch.

Fix options:
- (a) After populating `sources["opensearch"]` in detect_sources, also assign `sources["elasticsearch"] = sources["opensearch"]` (same dict reference, both keys reachable). Cheapest fix, lets both tools land.
- (b) Change `ElasticsearchLogsTool.is_available` and `extract_params` to also check `sources.get("opensearch")`. More explicit but invasive.
- (c) Pick one canonical key (`elasticsearch` or `opensearch`) and route all opensearch wizard configs into it consistently across the codebase.

Whichever you pick, the regression test should remove the manual `sources["elasticsearch"] = {**opensearch}` line and assert that `detect_sources()` alone makes BOTH tools available end-to-end.

### 2. `opensre integrations verify opensearch` will fail

`docs/opensearch.mdx:94` documents the verify command, but `opensearch` is only registered in `SETUP_SERVICES` (`app/cli/support/constants.py:29`), not `VERIFY_SERVICES` (`app/cli/support/constants.py:54-82`). Click's `type=click.Choice(VERIFY_SERVICES)` will reject the argument with `Error: Invalid value for '<service>'`.

Fix: add `"opensearch"` to `VERIFY_SERVICES` (alphabetized between `opsgenie` and `postgresql`). The validator already exists at `app/cli/wizard/integration_validators/client_validators.py:454`, so this should just need the registration.

### 3. New doc not registered in `docs.json`

`docs/opensearch.mdx` exists but `docs/docs.json` doesn't reference it. The doc site won't render the new page. Add `"opensearch"` to the relevant docs.json group (the "Data and workflow systems" group at line 117-128 contains the closest neighbors).

---

## Concerns (non-blocking)

### Synthetic test misclassified

`tests/synthetic/test_opensearch_synthetic.py` doesn't follow the synthetic-suite pattern documented in `tests/AGENTS.md` (no use_case.py / orchestrator separation, no @traceable metadata, no real or `respx`-mocked HTTP calls, no LLM driver). The file's own docstring (lines 12-14) acknowledges this. It's a deterministic plumbing test, which is valuable, but it should live at `tests/nodes/plan_actions/test_opensearch_pipeline.py` (next to the existing detect_sources tests). Putting it under `tests/synthetic/` makes future contributors think the integration has full e2e coverage when it doesn't.

### Wizard `_prompt_value` retry loop never PTY-tested

The smoke test at `tests/cli_smoke_test.py:483` j-presses past the integrations menu without entering `_configure_opensearch`. Two unit tests at `tests/cli/wizard/test_flow.py` cover the happy path and the `validate_*` retry shape but not the empty-input retry through `_prompt_value`. If we silently regress the retry loop nobody catches it.

### Half-credential rejection not tested

`test_detect_sources_integration_wave.py` covers the all-empty and all-populated cases but not the half cases (`username="admin"`, `password=""`). The wizard's `_prompt_value` blocks this UI-side, but a catalog record stored directly via env vars or `.env` editing can still produce it. Currently `detect_sources` would forward `username="admin"`, `password=""` and `ElasticsearchConfig.headers` would silently emit no Authorization header. Add a regression case that asserts either a normalized empty pair (both blank) or an explicit raise.

### Smoke comment still off-by-one

`tests/cli_smoke_test.py:485` now reads "OpenSearch is at index 21" — but counting `integration_choices` in `flow.py` puts OpenSearch at index **20** and "Skip for now" at index **21**. The j-count of 21 is still correct (it lands on Skip, which is the test's intent), but the comment math is off. Should read "OpenSearch is at index 20; 21 j's lands on the new 'Skip for now' position at index 21".

### Doc "expected output" doesn't match validator

`docs/opensearch.mdx:97-100` shows `(status: green, nodes: 3)` but `client_validators.py:482-487` returns `"Connected to OpenSearch cluster '<name>' (<status>, <N> node(s))."`. Users running the verify command will see a different format than documented. Wrap it in a code fence and use the actual format.

---

## False positives (worth dismissing)

### CodeQL "Clear-text logging of sensitive information"

I traced this carefully. The flagged sink is `detect_sources.py:489-492` (the existing `grafana_hint` logger), not the new opensearch lines. CodeQL's taint analysis treats the new test fixture password (`"secret-pass-123"` in `test_opensearch_synthetic.py:60`) as a tractable taint source and flags any logger inside the function that touches a string read from the same `integrations` dict argument, even though the password and the grafana_hint flow are structurally disjoint.

Corroborating evidence the password is never actually logged:
- `app/nodes/plan_actions/plan_actions.py:256` only logs `list(available_sources.keys())`
- `app/nodes/investigate/execution/execute_actions.py:91` only logs `available_source_keys`
- `build_prompt.py:257-264` opensearch hint only formats `index_pattern` and `default_query`
- The OpenObserve precedent at `detect_sources.py:850-866` has the same shape and was never flagged

Recommendation: dismiss alerts 621 and 622 as false-positive in the GitHub Security UI with a short note. I see your commit trail shows 5 suppression attempts that all chased the wrong sink (the grafana_hint logger), then reverted, so dismissing is cleaner than further code-level workarounds. Defense-in-depth follow-up worth filing separately: extend `MaskingContext` to be key-name-aware (mask any value under a key matching `password`/`secret`/`token`), so accidental future logging of `available_sources` is automatically redacted.

---

## What's solid

- The `detect_sources.py:887-888` fix matches the openobserve precedent shape
- The 5-commit-per-layer split honors PR #1190's "don't bundle" maintainer guidance
- The new test_detect_sources_integration_wave.py regression assertions correctly fail without your detect_sources.py +2 lines (verified by reading)
- API-key precedence preserved, backward compat clean
- All 4 of my prior inline comments resolved with code changes plus replies

tests/synthetic/test_opensearch_synthetic.py
Sarah-Salah: Both fixed in `2bafd18` and `f49b41fb`:

1. **Tool reachability bug:** `detect_sources` now does `sources["elasticsearch"] = sources["opensearch"]` (same dict reference). The manual `sources["elasticsearch"] = {**opensearch, ...}` line in the test was removed; the test now drops it and asserts both tools are reachable from `detect_sources()` alone. Verified locally: removing the assignment in `detect_sources` causes the assertion to fail, removing the manual line in the test no longer hides the bug.

2. **File placement:** moved to `tests/nodes/plan_actions/test_opensearch_pipeline.py` (commit `f49b41fb`) with an updated docstring that explains the placement and acknowledges the LLM-driven synthetic suite as follow-up.

After these two changes, the test passes for the right reason — both `ElasticsearchLogsTool` and `OpenSearchAnalyticsTool` are genuinely reachable from a single opensearch wizard configuration.

docs/opensearch.mdx
Sarah-Salah: Both fixed:

1. **`opensre integrations verify opensearch` registration** (commit `2bafd18`): added `"opensearch"` to `VERIFY_SERVICES` in `app/cli/support/constants.py`, alphabetized between `opsgenie` and `postgresql`. The validator at `client_validators.py:454` now wires through.

2. **Expected output format** (commit `f49b41fb`): wrapped the example in a code fence and updated the string to match the validator's actual output (`Connected to OpenSearch cluster '<name>' (<status>, <N> node(s)).`).

tests/cli_smoke_test.py
Sarah-Salah: Fixed in `f49b41fb`. Comment now reads:

```python
# 21 integrations + "Skip for now" = 22 choices. OpenSearch is at index 20;
# 21 j's lands on the new "Skip for now" position at index 21.

Comment (Sarah-Salah) [reply tests/cli_smoke_test.py]: Fixed in `f49b41fb`. Comment now reads:

```python
# 21 integrations + "Skip for now" = 22 choices. OpenSearch is at index 20;
# 21 j's lands on the new "Skip for now" position at index 21.

The stagger_j=21 value was already correct — only the comment math needed updating.

File: tests/nodes/plan_actions/test_detect_sources_integration_wave.py (hunk 8 above)
Comment (Sarah-Salah) [reply tests/nodes/plan_actions/test_detect_sources_integration_wave.py:211]:
```markdown
Added two regression tests in `f49b41fb`:

- `test_detect_sources_opensearch_username_without_password_forwards_half_pair`
- `test_detect_sources_opensearch_password_without_username_forwards_half_pair`

Picked the "surface the half-pair as-is" option rather than normalize-to-empty or raise. Rationale:

- `ElasticsearchConfig.headers` already handles the half-pair by emitting no Authorization header (covered in `tests/services/test_elasticsearch_client.py:51-57`), so the cluster returns a **visible 401** rather than a silent credential drop.
- Normalizing to empty would mask the user's intent (they set `OPENSEARCH_USERNAME=admin` for a reason — silently dropping it is surprising).
- Raising would break configs that work today by env-var setup, where partial-pair is sometimes intermediate state.

The new tests assert the half-pair is forwarded as-is. This pins the contract: detect_sources doesn't second-guess the catalog record, and the visible 401 is the documented failure mode.

**Comment by Sarah-Salah** [comment]
Thanks for the thorough second pass, @yashksaini-coder 🙏 — this review caught real issues that needed fixing.

## Round 2 blockers (resolved in earlier commits)

All three blockers were addressed in commits `2bafd18` (tool reachability + verify registration + docs.json) and `4ea0f64` (test cleanup). The fix-options I picked:

- **Blocker 1 (tool reachability):** went with **option (a)** — `detect_sources` now assigns `sources["elasticsearch"] = sources["opensearch"]` (same dict reference) so both `ElasticsearchLogsTool` and `OpenSearchAnalyticsTool` are reachable from a single opensearch wizard config. The previous manual `sources["elasticsearch"] = {**opensearch, ...}` line in the test was removed; the test now asserts that `detect_sources()` alone makes both tools available end-to-end.
- **Blocker 2 (verify command):** added `"opensearch"` to `VERIFY_SERVICES` in `app/cli/support/constants.py`, alphabetized between `opsgenie` and `postgresql`. The `opensre integrations verify opensearch` command now works as documented.
- **Blocker 3 (docs.json):** registered `docs/opensearch.mdx` under the "Data and workflow systems" group in `docs/docs.json`.

## Round 2 non-blocking concerns (resolved in `f49b41fb`)

All five concerns addressed in this commit:

1. **Smoke comment math:** corrected to "OpenSearch is at index 20; 21 j's lands on the new 'Skip for now' position at index 21".
2. **Doc expected output:** wrapped in a code fence and updated to match the validator's actual format (`Connected to OpenSearch cluster '<name>' (<status>, <N> node(s)).`).
3. **Synthetic test misclassified:** renamed `tests/synthetic/test_opensearch_synthetic.py` → `tests/nodes/plan_actions/test_opensearch_pipeline.py`. Docstring updated to explain the placement and to acknowledge that an actual LLM-driven synthetic suite is tracked as follow-up.
4. **Half-credential rejection:** added two regression tests for the `username`-without-`password` and `password`-without-`username` cases. Picked the "surface as-is" option rather than normalize-or-raise — `ElasticsearchConfig.headers` already handles the half-pair by emitting no Authorization header (covered in `tests/services/test_elasticsearch_client.py`), so the cluster returns a visible 401 rather than a silent credential drop. Normalizing to empty would mask the user's intent; raising would break configs that work today.
5. **CodeQL false positives:** confirmed by code review that the flagged sink is the existing `grafana_hint` logger, not the new opensearch path. Defer dismissal to maintainers — fork users cannot dismiss alerts on the upstream repo. Filed mentally as follow-up: extending `MaskingContext` to be key-name-aware (mask any value under a key matching `password`/`secret`/`token`) would be a clean defense-in-depth follow-up worth doing as a separate PR.

## Wizard `_prompt_value` retry loop concern

You're right that the smoke test j-presses past the integrations menu without entering `_configure_opensearch`, and the existing unit tests in `tests/cli/wizard/test_flow.py` cover the validator-failure retry but not the empty-input retry through `_prompt_value`. Adding a PTY test for that would be valuable, but the smoke test infrastructure is fragile on macOS PTYs (we already know the existing onboard smoke is finicky on local dev), so I'm proposing this as a follow-up issue rather than blocking this PR. Happy to file it if you'd like.

## CI status

All checks green except the CodeQL false positive, which only the upstream maintainer can dismiss. Re-running CodeQL won't help — the alert is structural to the test fixture's password literal.

Ready for another pass when you have a moment 🙏

File: app/nodes/plan_actions/detect_sources.py [hunk 9]
```diff
@@ -886,6 +886,8 @@ def detect_sources(
             sources["opensearch"] = {
                 "url": opensearch_url.rstrip("/"),
                 "api_key": str(opensearch_int.get("api_key", "")).strip(),
+                "username": str(opensearch_int.get("username", "")).strip(),
+                "password": str(opensearch_int.get("password", "")).strip(),
```
yashksaini-coder, line 890: ## CodeQL false-positive (the only CI blocker remaining)

@Sarah-Salah CodeQL flags `logger.warning(...)` at `detect_sources.py:491` (the `grafana_instance hint %r not found` warning) as "Clear-text logging of sensitive information (password)", with this line as the alleged taint source. After tracing the data flow on PR head `f49b41fb`, this is **unambiguously a false positive**:

- `grafana_hint` originates at `detect_sources.py:466-471` from `raw_alert.get("grafana_instance")`, normalized via `str(...).strip().lower()`. It's a short identifier like `"prod"` or `"staging"`.
- This `password` line is **400 lines later**, in a completely independent code path that runs after the grafana logger has already executed.
- The two variables share zero actual data flow. CodeQL's taint engine is over-approximating because both transitively originate from dict accesses on related parameters.

I saw your earlier attempt to break the taint via `+ str(grafana_hint)` concatenation in commit `04ec3b7b`, and the subsequent revert in `4b89d999` — agree that wasn't worth keeping. The cleanest paths forward are:

1. **Admin dismissal** in the GitHub Code Scanning UI (`/security/code-scanning/621` and `/security/code-scanning/622`). The code is correct; suppressing in-tree adds noise and rots over time. This is my recommendation. @muddlebee — could you mark these as false positives so CI clears?
2. **Inline suppression** — add `  # lgtm[py/clear-text-logging-sensitive-data]` to the `logger.warning` line at `detect_sources.py:491` with a one-line comment explaining the false positive. Smaller blast radius than option 3 but adds permanent noise.
3. **Refactor** — extract the OpenSearch handling into a separate helper function so the taint engine can no longer artificially bridge the two paths. Most invasive; only worth it if false-positive re-trips become a recurring pattern.

No code change requested — flagging this so we know what's blocking the merge gate. The PR itself is good to land as soon as CodeQL clears.

yashksaini-coder (approved):
## Round 3 — Approving ✅

@Sarah-Salah every Round 2 finding is now resolved with code-or-test evidence, and local quality + security probes are all clean on `f49b41fb`.

### Resolved
- ✅ Tool reachability bug — `detect_sources.py:907` aliases `sources["elasticsearch"] = sources["opensearch"]` (same dict ref, well-commented)
- ✅ `opensre integrations verify opensearch` — added to `VERIFY_SERVICES` in `app/cli/support/constants.py`
- ✅ Smoke test comment math — fixed in `f49b41fb`
- ✅ Half-pair credentials regression coverage — both `username_without_password` and `password_without_username` tests added
- ✅ `tests/test_elasticsearch_client.py` moved to `tests/services/`
- ✅ Greptile P2 (empty-credential guard) — confirmed false positive via `_prompt_value` semantics
- ✅ `docs/opensearch.mdx` referenced from `docs.json`

### Local verification on PR head `f49b41fb`
- `ruff check app/ tests/` clean
- `mypy` on 14 changed source files clean (only pre-existing unrelated `sentry_sdk` failure)
- `pytest` on the OpenSearch test surface — **136 passed** across 8 test files
- Manual `ElasticsearchConfig.headers` security probes:
  - special characters / unicode in credentials → correctly base64-encoded
  - API key precedence over Basic when both set → correct
  - half-credential (only `username` or only `password`) → no `Authorization` header (fails closed)
  - empty strings → no header
  - **CRLF injection attempt** in `username` (e.g. `"admin\r\nX-Header: evil"`) → base64 neutralizes the bytes; no header-injection possible
- Half-pair surface-as-is policy is locked in by both unit (`test_elasticsearch_client.py`) and integration (`test_detect_sources_integration_wave.py`) tests — env-var-only configs that have only one half of Basic Auth credentials produce a 401 from the cluster (visible failure) rather than silent credential drop.

### Outstanding

The **CodeQL** check is the only red signal left, and after tracing the data flow I'm confident it's a false positive. Details + resolution suggestion in the inline comment below.

LGTM. Approving — merge will still gate on CodeQL turning green (admin dismissal of the false-positive alert is the cleanest path).

app/nodes/plan_actions/detect_sources.py
Sarah-Salah: Confirmed — agreed on all three points and on option (1) as the 
cleanest path.

For posterity, here's the data-flow trace I ran independently to 
verify it's a false positive:

- Source (detect_sources.py:466-471): grafana_hint comes from 
 raw_alert.get("grafana_instance"), normalized via 
 str(...).strip().lower(). It's a short identifier like "prod" 
 or "staging" , passwords never enter this variable.

- Sink (detect_sources.py:491): logger.warning(...) only 
 interpolates grafana_hint, nothing else.

- OpenSearch credentials (detect_sources.py:887-895): live in 
 opensearch_int.get("password", ""), used only to populate 
 sources["opensearch"] for downstream tools. No caller downstream 
 logs this dict.

The two paths are structurally independent — CodeQL's taint engine 
is over-approximating because both transitively touch the same 
parameter dict.

@muddlebee — admin dismissal preferred per the reviewer's 
recommendation. Happy to file a follow-up issue for the 
key-name-aware MaskingContext defense-in-depth idea if useful.

Sarah-Salah:
Thanks so much for the thorough multi-round review @yashksaini-coder 🙏

This was a really educational process. Your two critical findings 
(the field-name mismatch in Round 1, and the tool reachability gap 
in Round 2) caught real bugs that my own testing missed because I 
was mocking too high up the call stack. Going forward, I'll mirror 
your discipline of "remove the mock and watch the test fail for 
the right reason."

Particularly grateful for the depth of security probes you ran 
locally (CRLF injection, unicode, half-credentials) — that's a 
level of verification I'll bring to future PRs.

Pinging @muddlebee on the CodeQL note — admin dismissal of 
alerts /security/code-scanning/621 and /security/code-scanning/622 
would clear the only remaining merge gate. Happy to add inline 
suppression instead if preferred.

yashksaini-coder (commented):
These changes don't seem to be of opensearch fix, are they from branch merging if so please clear and fix them. @Sarah-Salah 

We will have a detailed review later once this is addressed.

Sarah-Salah:
Thanks for catching this @yashksaini-coder 🙏

You're right, those daily-update files shouldn't appear as changes 
in this PR. Investigated locally and confirmed the file content is 
actually identical to main:

  $ git diff origin/main..HEAD -- docs/daily-updates/2026-04-30.mdx
  (no output)
  
  $ git diff origin/main..HEAD -- docs/daily-updates/2026-05-01.mdx
  (no output)

The diff is showing up because of duplicate commits introduced 
during the earlier conflict resolution merge — the file content 
matches main exactly, but the commit history has redundant entries 
that GitHub is surfacing in the PR diff.

I have two options to clean this up:

1. **Interactive rebase** to drop the duplicate commits — risky 
   because the branch has 25+ commits with a non-trivial merge 
   commit, but produces the cleanest history.
   
2. **Reset to a fresh branch from main and cherry-pick only the 
   OpenSearch commits** — safer, but loses the merge commit context.

Which would you prefer? Or is there another approach you've used 
for similar situations? Happy to do whichever keeps the history 
cleanest for the maintainers.

Sarah-Salah:
Hey @yashksaini-coder, branch recreation done ✅

Recreated feat/1143-opensearch-basic-auth-client from current upstream/main 
and cherry-picked the 17 OpenSearch-related commits in chronological order. 
Commits dropped:

- 5 CodeQL suppression attempts + 1 revert (946f1c7, be5718a, 8b7cc0f, 
  04ec3b7, 4b89d99) — they were a wash anyway
- 3 CI-trigger no-op commits (83a9f68, 248ca95, a3e0f53)
- 3 merge commits from the earlier conflict resolution (d429657, 16902fa, 
  65d879e)
- 25+ duplicate commits from main that the merge had pulled in (the 
  daily-updates and other unrelated files you flagged)

Local verification on the new HEAD:
git diff –stat upstream/main..HEAD
23 files changed, 1149 insertions(+), 22 deletions(-)
All 23 files are OpenSearch-related (no daily-updates, no contrib-readme,
no main-side files)
make format && make lint && make format-check && make typecheck
All clean (493 source files)
pytest tests/integrations/test_opensearch_catalog.py tests/cli/wizard/test_flow.py
tests/cli/wizard/test_integration_health.py tests/services/test_elasticsearch_client.py
tests/tools/test_elasticsearch_logs_tool.py tests/tools/test_integration_wave_tools.py
tests/nodes/plan_actions/test_detect_sources_integration_wave.py
tests/nodes/plan_actions/test_opensearch_pipeline.py
143 passed in 4.48s
The PR diff should now show only OpenSearch work. Force-pushed via 
--force-with-lease to avoid stomping on anything. 

Original branch backed up locally as backup-1143-original in case anything 
got missed; happy to restore individual commits if you spot a gap.

Ready for re-review when you have a moment 🙏

yashksaini-coder:
@greptile review

app/cli/wizard/flow.py
```diff
@@ -1569,6 +1587,82 @@ def _configure_splunk() -> tuple[str, str]:
...
+        )
+        api_key = ""
+        username = ""
+        password = ""
+        if auth_choice == "api_key":
+            api_key = _prompt_value(
+                "OpenSearch API key",
+                default=_string_value(credentials.get("api_key")),
+                secret=True,
+            )
```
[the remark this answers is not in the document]
Sarah-Salah: This was a real bug. Fixed in the latest commit.

**Root cause confirmed:** when `auth_choice == "api_key"` and the user presses 
Enter without typing, `_prompt_value` returns `""` because there's no catalog 
default to fall back to. The `api_key or None` conversion in the validator 
then runs the probe without an Authorization header. On a security-disabled 
cluster the probe returns 200, `result.ok` is True, and the integration 
persists as URL-only, silently dropping the user's chosen auth method.

The same shape applies to half-populated Basic Auth (only username or only 
password): `ElasticsearchConfig.headers` drops the Authorization header when 
either half is empty, so the agent would send unauthenticated requests 
against a security-enabled cluster and fail with a confusing 401.

**Fix:** explicit guards after each prompt block in `_configure_opensearch`:

- Empty `api_key` under `api_key` auth → clear error, `continue` (re-prompts via 
  the existing `while True` loop)
- Empty `username` or `password` under `basic` auth → clear error, `continue`

Errors surface **before** the probe runs, so the user sees the failure they 
caused rather than a silent persist.

**Regression tests:** two new tests in `tests/cli/wizard/test_flow.py`:
- `test_run_wizard_opensearch_rejects_empty_api_key`
- `test_run_wizard_opensearch_rejects_empty_basic_password`

Both assert `validation_call_count == 1` — the validator only runs **after** 
the user supplies a non-empty value, proving the guard fires before the probe. 
Removing the guard makes both tests fail with `validation_call_count == 2`.

Sarah-Salah:
@yashksaini-coder, Greptile's P1 finding fixed in the latest commit.

The bug was real, empty credentials could reach the cluster probe in two 
shapes (empty api_key, or half-populated Basic Auth). On a security-disabled 
cluster the probe would return 200 and the integration would persist as 
URL-only, silently dropping the user's auth choice. Fix adds explicit 
pre-probe guards in _configure_opensearch plus two regression tests that 
verify the validator is only called after the user supplies a non-empty 
value.

CI status: ✅ 
Ready for another look 🙏

yashksaini-coder:
@Sarah-Salah update the PR branch please

Sarah-Salah:
hi @yashksaini-coder, Done, Merged latest `upstream/main` into the branch (no conflicts). 
Local checks all clean format, lint, typecheck (494 source files), and 145 OpenSearch tests passing.

muddlebee:
@Sarah-Salah 
image

muddlebee:
@greptileai review

Sarah-Salah:
hi @muddlebee thanks for the review 🙏

Quick note, that exact guard is now in place. Greptile flagged the same 
issue earlier today and the fix landed in commit f0964df9 before the 
latest upstream/main merge:

- Empty api_key under api_key auth → blocked with a clear error 
  before the cluster probe runs
- Empty username or password under basic auth → same guard 
  (handles the half-credential case too)

Two regression tests in tests/cli/wizard/test_flow.py 
(test_run_wizard_opensearch_rejects_empty_api_key and 
test_run_wizard_opensearch_rejects_empty_basic_password) assert the 
validator is only called *after* the user supplies a non-empty value, 
proving the guard fires before the probe. Both fail without the guard.

Pinging in case the fix landed after your review pass, happy to walk 
through the diff if anything looks off.

VaibhavUpreti (approved):
Great work @Sarah-Salah! 🙌 

Thanks a lot for the thorough reviews @yashksaini-coder @muddlebee !! 🚀

PR #1231, opened 2026-05-03
fix(ci): restore CLI help sync + uninstall help text

IsaacOdeimor:
CI is currently failing because `tests/cli/test_command_help_sync.py` still expects a `_HELP_COMMANDS` table that no longer exists (the root help now derives its command list from the live Click group).

This PR updates the regression test to compare `_COMMANDS` against the command list derived from the root `cli` group, and adjusts the `uninstall` command’s help text to include “Uninstall …” so the uninstall help test matches the displayed copy.

Repro: `python -m pytest -n auto --ignore=tests/e2e/kubernetes_local_alert_simulation --ignore=tests/synthetic -m 'not synthetic'`

### DIFF
--- a/app/cli/commands/general.py
+++ b/app/cli/commands/general.py
@@ -24,7 +24,7 @@ from app.version import get_version
 @click.command(name="uninstall")
 @click.option("--yes", "-y", "local_yes", is_flag=True, help="Skip the confirmation prompt.")
 def uninstall_command(local_yes: bool) -> None:
-    """Remove opensre and all local data from this machine."""
+    """Uninstall opensre and remove all local data from this machine."""
     from app.cli.support.uninstall import run_uninstall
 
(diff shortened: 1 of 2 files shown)


rrajan94:
Thanks for contribution @IsaacOdeimor 👍
Merging this for velocity. 
The Greptile P2 is a future-proofing edge case rather than a current regression, we can address the hidden-command test cleanup in a follow-up PR.

IsaacOdeimor:
> Thanks for contribution @IsaacOdeimor 👍 Merging this for velocity. The Greptile P2 is a future-proofing edge case rather than a current regression, we can address the hidden-command test cleanup in a follow-up PR.

Thanks! Glad it was helpful. @rrajan94
I can take a look at the hidden-command test cleanup in a follow-up PR 👍

PR #1287, opened 2026-05-05
feat(rds): RDS integration with describe instance and events tools

### LINKED ISSUE #125: RDS Integration

mazenessam77:
- Add app/integrations/rds.py for source detection and param extraction
- Add RDSDescribeInstanceTool for instance status and configuration
- Add RDSEventsTool for failover, maintenance, and parameter events
- Use shared aws_sdk_client read-only allowlist for safety
- Add unit tests for integration helpers and both tools
- Add "rds" to EvidenceSource literal in app/types/evidence.py

Closes #125

Fixes #125

This PR adds AWS RDS as a first-class integration source so the agent can investigate RDS instance health and recent events during incidents.

- **`app/integrations/rds.py`** — config builder, env-based config, `rds_is_available` source detector, and `rds_extract_params` to normalize tool inputs. Mirrors the shape used by other AWS integrations.
- **`app/tools/RDSDescribeInstanceTool/`** — read-only tool that calls `rds:DescribeDBInstances` and returns status, engine, version, instance class, Multi-AZ, networking, storage, and backup config.
- **`app/tools/RDSEventsTool/`** — read-only tool that calls `rds:DescribeEvents` to surface failovers, maintenance, parameter changes, and backup events around an incident window.
- **`app/types/evidence.py`** — adds `"rds"` to the `EvidenceSource` literal. Without this, the `@tool(source="rds", ...)` decorator on both new tools fails `mypy` (no matching overload).
- **Tests** — `tests/integrations/test_rds.py` (4) and `tests/tools/test_rds_tools.py` (6); 14 new tests total, all passing.

Both tools route through the existing `app/services/aws_sdk_client.py`, which enforces a read-only boto3 allowlist — RDS gets the same safety treatment as the existing CloudWatch, EKS, and Lambda tools.

```
$ .venv/bin/python -m pytest tests/integrations/test_rds.py tests/tools/test_rds_tools.py
...
============================== 14 passed in 0.77s ==============================

$ .venv/bin/python -m mypy app/
Success: no issues found in 460 source files
```

---

The problem: the agent had no way to investigate RDS during incidents — there was an `app/integrations/rds.py` skeleton with 0% coverage but no tools wired up.

I followed the pattern already established for AWS services in the repo (CloudWatch, EKS, Lambda):

1. **Integration layer first** (`app/integrations/rds.py`) — defines what an RDS source looks like in the integration store, when it's "available", and how to extract tool params from the source dict. This is what the planner uses to decide whether to call RDS tools.
2. **Tools second** (`app/tools/RDSDescribeInstanceTool/`, `app/tools/RDSEventsTool/`) — thin wrappers around `execute_aws_sdk_call(...)` so I get the read-only allowlist, error handling, and credential management for free. Each tool returns a flat dict matching the project's tool-output convention.
3. **Type literal** — adding `"rds"` to `EvidenceSource` was needed because `@tool(source=...)` is typed against that literal. I considered using the existing `"aws_sdk"` source instead, but that would have broken source-based filtering downstream and mixed RDS evidence into a generic bucket.
4. **Tests cover** the happy path, "no instance found" / "no events", and AWS SDK failures, plus availability detection and param extraction from both explicit configs and env-fallback.

Alternative considered: rolling a dedicated `RDSClient` under `app/services/rds/` like the Datadog and Grafana clients. Rejected because RDS only needs two boto3 calls — adding a client class would be over-abstraction. The `aws_sdk_client` pattern is the right fit and matches what CloudWatch/EKS/Lambda do.

---

### DIFF
--- a/app/integrations/rds.py
+++ b/app/integrations/rds.py
@@ -40,9 +40,7 @@ def rds_config_from_env() -> RDSConfig | None:
     return build_rds_config(
         {
             "db_instance_identifier": db_id,
-            "region": env_str("AWS_REGION")
-            or env_str("RDS_REGION")
-            or DEFAULT_RDS_REGION,
+            "region": env_str("AWS_REGION") or env_str("RDS_REGION") or DEFAULT_RDS_REGION,
         }
     )
 
--- a/app/integrations/rds.py
+++ b/app/integrations/rds.py
@@ -1,0 +1,62 @@
+"""Shared AWS RDS integration helpers.
+
+Provides configuration normalization, source detection, and parameter
+extraction for the RDS investigation tools. All AWS API calls are
+read-only and routed through the shared aws_sdk_client allowlist.
+"""
+
+from __future__ import annotations
+
+import os
+from typing import Any
+
+from app.integrations._relational import env_str
+from app.strict_config import StrictConfigModel
+
+DEFAULT_RDS_REGION = "us-east-1"
+
+
+class RDSConfig(StrictConfigModel):
+    """Normalized RDS connection settings."""
+
+    db_instance_identifier: str = ""
+    region: str = DEFAULT_RDS_REGION
+    integration_id: str = ""
+
+    @property
+    def is_configured(self) -> bool:
+        return bool(self.db_instance_identifier and self.region)
+
+
+def build_rds_config(raw: dict[str, Any] | None) -> RDSConfig:
+    """Build a normalized RDS config object from env/store data."""
+    return RDSConfig.model_validate(raw or {})
+
+
+def rds_config_from_env() -> RDSConfig | None:
+    """Load an RDS config from env vars."""
+    db_id = env_str("RDS_DB_INSTANCE_IDENTIFIER")
+    if not db_id:
+        return None
+    return build_rds_config(
+        {
+            "db_instance_identifier": db_id,
+            "region": env_str("AWS_REGION", DEFAULT_RDS_REGION)
+            or env_str("RDS_REGION", DEFAULT_RDS_REGION),
+        }
+    )
+
+
+def rds_is_available(sources: dict[str, dict]) -> bool:
+    """Check if RDS integration identifying params are present."""
+    rds = sources.get("rds", {})
+    return bool(rds.get("db_instance_identifier"))
+
+
+def rds_extract_params(sources: dict[str, dict]) -> dict[str, Any]:
+    """Extract RDS identifying params (db_instance_identifier, region)."""
+    rds = sources.get("rds", {})
+    return {
+        "db_instance_identifier": str(rds.get("db_instance_identifier", "")).strip(),
+        "region": str(rds.get("region") or os.getenv("AWS_REGION", DEFAULT_RDS_REGION)).strip(),
+    }
--- a/app/integrations/rds.py
+++ b/app/integrations/rds.py
@@ -7,7 +7,6 @@ read-only and routed through the shared aws_sdk_client allowlist.
 
 from __future__ import annotations
 
-import os
 from typing import Any
 
 from app.integrations._relational import env_str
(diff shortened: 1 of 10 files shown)


app/tools/RDSDescribeInstanceTool/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+        parameters={"DBInstanceIdentifier": db_instance_identifier},
+        region=region,
+    )
+
+    if not result.get("success"):
+        return {
+            "source": "rds",
+            "available": False,
+            "db_instance_identifier": db_instance_identifier,
+            "error": result.get("error"),
```
Devesh36: tools return result.get("error") directly in the response payload. AWS SDK errors can contain region names, account IDs, ARNs, or internal resource identifiers — all of which get forwarded to whoever reads the tool result. Sanitize or wrap errors before surfacing: return a generic message and log the full AWS error server-side only.

app/tools/RDSDescribeInstanceTool/__init__.py
```diff
@@ -0,0 +1,102 @@
...
+    instances = (result.get("data") or {}).get("DBInstances") or []
+    if not instances:
+        return {
+            "source": "rds",
+            "available": False,
+            "db_instance_identifier": db_instance_identifier,
+            "error": "No RDS instance found with the given identifier.",
+        }
+
+    instance = instances[0]
```
Devesh36, line 93: instances[0] silently discards any additional results. The AWS DescribeDBInstances API can theoretically return multiple records (e.g. in edge cases with cluster members). If more than one is returned, the caller gets no indication. At minimum, add a log warning when len(instances) > 1.

app/tools/RDSDescribeInstanceTool/__init__.py
mazenessam77: Done. I've updated the code to sanitize AWS SDK errors by returning a generic message to the tool output while logging the full error serverside. I also added a warning log when multiple instances are returned to address your second point

app/tools/RDSDescribeInstanceTool/__init__.py
mazenessam77: Done. I've updated the logic in commit 46a44a4 to address both points  I sanitized the AWS SDK error messages to prevent leaking sensitive info and added a warning log in case multiple instances are returned as you suggested. Ready for another look!

mazenessam77:
Done i fixed those issues u can test it

muddlebee:
@cerencamkiran :eyes: for reviews

cerencamkiran:
@greptileai review

app/integrations/rds.py
```diff
@@ -0,0 +1,66 @@
...
+        return None
+    return build_rds_config(
+        {
+            "db_instance_identifier": db_id,
+            "region": env_str("AWS_REGION") or env_str("RDS_REGION") or DEFAULT_RDS_REGION,
+        }
+    )
+
+
+def rds_is_available(sources: dict[str, dict]) -> bool:
```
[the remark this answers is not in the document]
cerencamkiran: This matches the main thing I wanted to double-check as well.

The tools and helper layer look fine, but if RDS is not registered in `load_env_integrations()` and not transferred into `sources` in `detect_sources.py`, the documented env setup path will not actually expose the RDS tools to the planner.

I think this should be addressed before merge, ideally with a small regression test covering env-backed RDS discovery → resolved integrations → sources containing `rds`.

Concretely, it seems we’re missing:
- wiring in `load_env_integrations()` to register the env-backed RDS config
- classification in `_classify_service_instance()` for remote-store configs
- propagation into `sources` in `detect_sources.py`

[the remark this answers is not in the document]
mazenessam77: Done! I've wired RDS into the discovery pipeline

mazenessam77:
Done✅

app/integrations/rds.py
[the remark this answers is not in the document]
mazenessam77: I have implemented the missing wiring in load_env_integrations, _classify_service_instance, and detect_sources.py. I also added a dedicated regression test test_rds_env_discovery_to_sources_pipeline that validates the entire discovery flow from environment variables to the planner's sources dictionary. Everything is verified and running as expected in my deployment. Ready for a final review!

mazenessam77:
here we goooo

PR #1292, opened 2026-05-05
chore(deps): commit uv.lock and lock CI to uv sync

muddlebee:
## Summary

Reproducible dependency installs: commit `uv.lock`, install with `uv sync --frozen` in CI, weekly Dependabot on the `uv` ecosystem, and a `deps-latest` canary using latest PyPI resolves.

## Why this is good long term

Today, installs that only read `pyproject.toml` resolve **whatever newest versions satisfy** `>=` constraints **on the day CI runs**. That is convenient but brittle: a green main on Tuesday can turn red on Wednesday without any code change, because a transitive upgrade (for example an SDK tightening validation at import or client construction) landed on PyPI overnight. Teams then burn time bisecting “what changed?” when the answer is **resolver noise**, not a regression in your diff.

Committing a lockfile and installing with **`uv sync --frozen`** makes dependency state **explicit and reviewable**. Upgrades become **intentional PRs** (human or Dependabot) where you see the diff, run CI once, and decide whether to take the change. That shifts the project from “floating on the ecosystem” to “pinned by default, upgraded on purpose,” which is the usual standard for **applications and CLIs** shipped as a single product.

The **`deps-latest` workflow** keeps a separate signal: a scheduled / manual job that **re-resolves from `pyproject.toml`** so you still learn about upstream breakage **before** it forces an emergency fire drill—without every pull request inheriting unbounded drift.

Together, this setup improves **predictability for contributors**, **stability for releases**, and **auditability** when you need to answer what shipped in a given build.

## Changes

- Commit `uv.lock`; remove it from `.gitignore`.
- CI, Windows label CI, synthetic tests: `uv sync --frozen --extra dev`, run tools via `uv run python -m …`.
- Makefile `install`: `uv sync --frozen --extra dev` + `uv run` for analytics.
- Dependabot: `uv` (weekly) for repo root.
- `.github/workflows/deps-latest.yml`: canary install with `uv pip install --python "$(which python)" -e ".[dev]"` then focused pytest.


muddlebee:
@greptileai review

.github/workflows/deps-latest.yml
```diff
@@ -0,0 +1,58 @@
+# Weekly canary: install without uv.lock so PyPI resolution matches unpinned pyproject
```
VaibhavUpreti: Would remove this workflow for now, we have way too many workflows, need to concise them more.

If you want to keep this check, I’d run it in an existing CI workflow instead.

Makefile
```diff
@@ -28,12 +28,10 @@ USER_BASE := $(shell $(PYTHON) -m site --user-base)
...
 
-# Create venv and install dependencies
+# Create venv and install dependencies (requires https://docs.astral.sh/uv/)
 install:
-	$(PYTHON) -m venv .venv
-	$(PIP) install --upgrade pip
-	$(PIP) install $(PIP_INSTALL_FLAGS) -e ".[dev]"
-	$(PYTHON) -m app.analytics.install
+	uv sync --frozen --extra dev
+	uv run python -m app.analytics.install
```
VaibhavUpreti, line 34: @muddlebee for this, we would also have to update SETUP.md and the readme, since you’ll need to have uv installed on your system. Could you also ensure windows compatibility?

muddlebee: Sure

.github/workflows/deps-latest.yml
muddlebee: Sure will do

muddlebee:
@greptileai review

.github/workflows/deps-latest.yml
muddlebee: added to synthetic-test workflow

.gitignore
```diff
@@ -60,8 +60,7 @@ coverage.xml
 # pyenv
 .python-version
 
-# uv
-uv.lock
+# uv (uv.lock is committed)
```
VaibhavUpreti, line 63: rm comment

.github/dependabot.yml
```diff
@@ -2,10 +2,10 @@
 
 version: 2
 updates:
-  - package-ecosystem: "pip"
+  - package-ecosystem: "uv"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
     open-pull-requests-limit: 10
```
VaibhavUpreti, line 9: change back to daily

VaibhavUpreti (approved):
awesome @muddlebee

PR #1295, opened 2026-05-05
fix(llm-cli): harden codex and claude-code auth probe paths

### LINKED ISSUE #1260: Audit Claude Code CLI: subscription login vs API key and cross-OS parity
## Context

Follow-up from the discussion in [#1199 (comment)](https://github.com/Tracer-Cloud/opensre/issues/1199#issuecomment-4364046407) after [#1217](https://github.com/Tracer-Cloud/opensre/pull/1217) landed (macOS Keychain / uncertain auth onboarding fix).

**Product intent from that thread:**
- Prefer **Claude CLI login + subscription** (OAuth / CLI session) over `ANTHROPIC_API_KEY` where possible — API key is a fallback, not the primary path we optimize for.
- **End-to-end parity:** behavior should hold **cross-OS** from Claude auth through actual execution (not only onboarding).

Original bug: onboarding treated uncertain Keychain auth as failure; #1217 added live probing. This issue tracks **verification and gap analysis** for the broader integration.

## Goals

1. **Confirm #1217** fully addresses the onboarding false-negative paths called out in [#1199](https://github.com/Tracer-Cloud/opensre/issues/1199) (including Keychain / `logged_in is None` handling and live `claude -p` verification where appropriate).
2. **Auth ordering & messaging:** codebase and wizard copy should consistently treat subscription/CLI login as first-class; API key as explicit fallback — document any remaining places that imply API key is required for Claude Code CLI.
3. **Cross-OS matrix:** enumerate Linux vs macOS vs Windows (if supported) for: install detection, login detection, execution, env vars (`ANTHROPIC_API_KEY` vs credential files vs Keychain), and failure modes.

## References

- [#1199 (comment)](https://github.com/Tracer-Cloud/opensre/issues/1199#issuecomment-4364046407) — blocker + desired auth priority discussion
- [#1217](https://github.com/Tracer-Cloud/opensre/pull/1217) — onboarding fix PR

muddlebee:
Fixes #1260

- handle claude auth probe bootstrap failures safely by catching runner import failures and returning an unclear-auth probe result instead of crashing
- normalize user-facing Claude auth hints to remove duplicate spacing and keep messaging consistent across adapter paths
- improve interactive shell response extraction to support structured LLM content blocks (`list` / `dict` / `.text` objects)
- add regression tests for probe import failures, auth hint formatting, and structured interactive-shell response rendering

- `uv run pytest tests/integrations/llm_cli/test_claude_code_adapter.py tests/cli/interactive_shell/test_cli_agent.py`
- Result: `60 passed`

---

- The main failure mode fixed here is an avoidable crash during Claude auth probing when the helper import path is unavailable. The adapter now treats that condition as an unknown auth state, which preserves diagnostic flow and surfaces an actionable message instead of throwing at import time.
- I kept auth messaging centralized by introducing one shared hint constant and using it in all unauthenticated/unclear branches. This avoids drift between probe/classify/runtime errors and removes formatting inconsistencies from user-facing output.
- For interactive shell robustness, response parsing now handles structured content payloads from providers that return block objects instead of plain strings. This preserves intended text output and avoids leaking Python repr-style payloads in terminal responses.

---

---

### DIFF
--- a/app/cli/interactive_shell/cli_agent.py
+++ b/app/cli/interactive_shell/cli_agent.py
@@ -40,7 +40,7 @@ _ACTION_RULE = (
     "action object schemas: "
     '`{"action":"switch_llm_provider","provider":"anthropic","model":"","toolcall_model":""}` '
     "where provider is one of anthropic, openai, openrouter, gemini, nvidia, "
-    "ollama, codex; both `model` (reasoning) and `toolcall_model` are optional; "
+    "ollama, codex, claude-code; both `model` (reasoning) and `toolcall_model` are optional; "
     '`{"action":"switch_toolcall_model","model":"claude-opus-4-7"}` '
     "to change ONLY the toolcall model on the currently active provider; "
     '`{"action":"slash","command":"/model show"}` where command is one of '
--- a/app/cli/interactive_shell/cli_agent.py
+++ b/app/cli/interactive_shell/cli_agent.py
@@ -155,6 +155,36 @@ def _parse_action_plan(text: str) -> list[dict[str, object]]:
     ]
 
 
+def _response_text(response: object) -> str:
+    """Extract text from heterogeneous LLM response content payloads."""
+    content = getattr(response, "content", None)
+    if content is None:
+        return str(response)
+    if isinstance(content, str):
+        return content
+    if isinstance(content, dict):
+        text_value = content.get("text")
+        return text_value if isinstance(text_value, str) else str(content)
+    if isinstance(content, list):
+        blocks: list[str] = []
+        for item in content:
+            if isinstance(item, str):
+                blocks.append(item)
+                continue
+            if isinstance(item, dict):
+                text_value = item.get("text")
+                blocks.append(text_value if isinstance(text_value, str) else str(item))
+                continue
+            text_value = getattr(item, "text", None)
+            blocks.append(text_value if isinstance(text_value, str) else str(item))
+        joined = "\n".join(part for part in blocks if part.strip()).strip()
+        return joined or str(content)
+    text_value = getattr(content, "text", None)
+    if isinstance(text_value, str):
+        return text_value
+    return str(content)
+
+
 def _execute_action_plan(
     actions: list[dict[str, object]],
     session: ReplSession,
@@ -282,8 +312,7 @@ def answer_cli_agent(
         console.print(f"[red]assistant failed:[/red] {escape(str(exc))}")
         return
 
-    text = getattr(response, "content", None) or str(response)
-    text_str = str(text)
+    text_str = _response_text(response)
     actions = _parse_action_plan(text_str)
     if _execute_action_plan(actions, session, console):
         session.cli_agent_messages.append(("user", message))
--- a/tests/cli/interactive_shell/test_cli_agent.py
+++ b/tests/cli/interactive_shell/test_cli_agent.py
@@ -40,7 +40,7 @@ def _capture() -> tuple[Console, io.StringIO]:
 
 
 class _FakeLLMResponse:
-    def __init__(self, content: str) -> None:
+    def __init__(self, content: Any) -> None:
         self.content = content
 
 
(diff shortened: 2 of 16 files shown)


muddlebee:
@greptileai review

muddlebee:
@greptileai review new changes

app/cli/interactive_shell/cli_agent.py
```diff
@@ -155,6 +155,36 @@ def _parse_action_plan(text: str) -> list[dict[str, object]]:
...
+                blocks.append(text_value if isinstance(text_value, str) else str(item))
+                continue
+            text_value = getattr(item, "text", None)
+            blocks.append(text_value if isinstance(text_value, str) else str(item))
+        joined = "\n".join(part for part in blocks if part.strip()).strip()
+        return joined or str(content)
+    text_value = getattr(content, "text", None)
+    if isinstance(text_value, str):
+        return text_value
+    return str(content)
```
yashksaini-coder, line 185: Non-blocking follow-up. This `_response_text` is doing real work, but it's normalizing LLM response shapes that every other caller of `get_llm_for_reasoning()` will eventually need to handle too. Worth lifting into `app/services/llm_client.py` as something like `extract_text(response)` so future consumers don't reimplement the same `str | dict | list | object-with-.text` ladder. Fine to keep here for this PR and split that out separately.

tests/cli/interactive_shell/test_cli_agent.py
```diff
@@ -180,6 +181,20 @@ def test_response_is_recorded_in_session_history(self, monkeypatch: Any) -> None
...
+        session = ReplSession()
+        console, buf = _capture()
+        answer_cli_agent("hello", session, console)
+        output = _strip_ansi(buf.getvalue())
+        assert "First line" in output
+        assert "Second line" in output
+        assert session.cli_agent_messages[-1] == ("assistant", "First line\nSecond line")
+
     def test_llm_failure_prints_red_error_and_does_not_record(self, monkeypatch: Any) -> None:
         class _Boom:
```
yashksaini-coder, line 199: Two small coverage gaps in `_response_text` that this test doesn't hit yet:

1. List with a bare `str` element (the `if isinstance(item, str): blocks.append(item)` branch at `cli_agent.py:170-172`). Realistic for older Anthropic SDK responses that mix string and block items in `content`.
2. The non-list, non-dict object with `.text` attribute (`cli_agent.py:182-185`) — separate from the in-list `.text` case you already cover.

Not blocking. Worth adding both since you've already got the fixture pattern in place.

yashksaini-coder (approved):
Approving. I ran a multi-role review (senior eng, QA, greptile verification) on HEAD `d24987aa` and the PR is in good shape.

The refactor is clean. `runner.py` shed 56 lines, `subprocess_env.py` picked up 64, and the diff between them is line-for-line a move with no logic loss. Public API of `runner.py` is preserved, and you kept `_build_subprocess_env = build_cli_subprocess_env` as a back-compat alias which is the right call. Cross-OS allowlist in `subprocess_env.py` covers POSIX (`USER`, `HOME`, `TMPDIR`) and Windows (`USERPROFILE`, `APPDATA`, `LOCALAPPDATA`, `PATHEXT`, `SYSTEMROOT`, `WINDIR`, `COMSPEC`) which lines up with issue #1260's cross-OS parity ask.

Both of greptile's P2 flags are stale at this HEAD. I checked each one:

- The `getattr(runner_module, "_build_subprocess_env", None)` pattern greptile cited doesn't exist anywhere in the repo. `claude_code.py:37` is a clean direct import. The bot quoted code that isn't in the file. Nothing to do.
- The double-period concern is already guarded at `runner.py:140-144` (the `endswith((".", "!", "?"))` branch) and you have an explicit regression test at `tests/integrations/llm_cli/test_codex_adapter.py:247-285` named `test_cli_backed_client_unclear_auth_no_double_period_when_explain_failure_trailing_period` that asserts `".." not in msg`. That's exactly the right shape.

The `AGENTS.md` updates match the new code structure (new `subprocess_env.py` row at line 12, `CLAUDE_` prefix added, doctor-discovery note at line 11), so the local guidance and the implementation agree.

A few non-blocking follow-ups I'd suggest, none needed for this merge:

1. `_response_text` in `cli_agent.py:158-185` is good defensive coding but it's the kind of normalization that probably belongs once on `LLMResponse` or in `app/services/llm_client.py` so every caller benefits. Worth a follow-up issue, not a change here.
2. `runner.py:140` has `# Avoid \"..\"` with escaped quotes from source-editing. Reads awkwardly. Tiny.
3. Test coverage gap in `tests/cli/interactive_shell/test_cli_agent.py:184-196`: the structured-content test mixes a `.text`-attribute object and a `dict` block, but doesn't cover (a) a bare `str` element inside the list, or (b) the `getattr(content, "text", None)` non-list/non-dict branch at `cli_agent.py:182-185`. Inline comment with specifics below.
4. Doctor coverage: `tests/cli/test_doctor.py` covers unset / hosted-missing-key / CLI-ready / CLI-auth-unclear / registry paths, but the `installed=False` and `logged_in=False` branches in `doctor.py:66-69` aren't exercised. Reasonable scope for the new file, but worth noting.

Lint and format clean (10 files), and the linked greptile suppressions are commented in code (`commit 48ebb47c "stable subprocess env API for Greptile nits"`). Nice work.

tests/cli/interactive_shell/test_cli_agent.py
muddlebee: added more coverage in subsequent commits

app/cli/interactive_shell/cli_agent.py
muddlebee: makes sense

Devesh36 (approved):

PR #1299, opened 2026-05-05
fix(store): make integration store writes atomic and concurrency-safe (#1272)

### LINKED ISSUE #1272: [BUG] Local integration store writes can lose updates under concurrent access
### Summary

Local integration store updates use unprotected read-modify-write operations against `~/.opensre/integrations.json`, so concurrent writes can overwrite each other and silently lose integration configuration.

**Expected:**

Updates to the local integration store should be atomic and concurrency-safe. If two OpenSRE processes or threads add, update, or remove integrations at the same time, both valid changes should be preserved or one operation should fail/retry safely.

**Actual:**

`app/integrations/store.py` loads the JSON file, mutates the in-memory list, and writes it back without file locking or atomic replacement.

Relevant code paths:

```text
app/integrations/store.py:132-158   _load_raw() / _save()
app/integrations/store.py:219-241   upsert_integration()
app/integrations/store.py:244-252   remove_integration()
app/integrations/store.py:328-384   upsert_instance()
app/integrations/store.py:387-430   remove_instance()
```

`_save()` writes directly with `STORE_PATH.write_text(...)`. If another process reads the old file before the first write completes, then writes its own modified copy afterward, the first process's update is lost.

### Steps to reproduce

1. Start with an empty or known `~/.opensre/integrations.json`.
2. Run two integration writes concurrently, for example one process saving a Datadog integration and another saving a Grafana integration.
3. Inspect `~/.opensre/integrations.json` after both commands complete.
4. Observe that only one update may remain, depending on timing.

A focused regression test could reproduce this by patching `STORE_PATH` to a temp file and running concurrent `upsert_integration()` or `upsert_instance()` calls with a synchronization barrier around load/save.

Sometimes

### How often does it occur?

Under specific conditions

### Operating system

Other

### Logs and error output

No error is emitted. The failure mode is silent data loss in `integrations.json`.

### Additional context

OpenSRE stores local integration credentials in `~/.opensre/integrations.json`, and onboarding/setup flows can update that file. The current implementation is safe for single-process sequential usage, but not for concurrent CLI invocations, parallel setup flows, or future background processes that may update integration state.

Suggested fix:

- Add a cross-process file lock around all read-modify-write operations.
- Write updates to a temporary file in the same directory, `fsync` it, then atomically replace the target with `os.replace`.
- Preserve file permissions (`0600`) after replacement.
- Keep v1-to-v2 migration write-through safe under the same locking/atomic-write path.
- Add tests that verify concurrent updates do not drop records and that a failed write does not corrupt the existing store.

KTS-o7:
## Summary

Fixes #1272 — concurrent `upsert_integration()` / `upsert_instance()` calls could silently drop each other's records because `_load_raw()` and `_save()` were separate, unlocked operations.

## Root cause

Every write path (`upsert_integration`, `remove_integration`, `upsert_instance`, `remove_instance`) did:

```python
data = _load_raw()   # read (no lock)
# ... mutate data ...
_save(data)          # write (no lock)
```

If two processes both read the same snapshot before either wrote back, the second writer overwrote the first writer's changes — silent data loss.

### 1. `filelock>=3.13.0` added to dependencies

Cross-platform advisory file lock (used by pip, poetry, uv — zero heavy transitive deps). Works on Windows, Linux, macOS.

### 2. Atomic `_save_unlocked()` replaces direct `write_text()`

```
write to sibling tmp file → fsync → chmod 0o600 → os.replace (atomic rename)
```

- Concurrent readers always see a complete, valid file (never a partial write)
- A crash mid-write leaves the previous store intact
- Permission window eliminated: file is `0o600` *before* the rename

### 3. Single lock covers each full read-modify-write

All four mutating functions (`upsert_integration`, `remove_integration`, `upsert_instance`, `remove_instance`) now acquire one `FileLock` around their entire read-modify-write cycle, eliminating the TOCTOU race.

### 4. Regression tests

`TestConcurrentWrites` in `tests/integrations/test_store.py`:

- **`test_concurrent_upserts_preserve_all_records`** — two threads upsert different services simultaneously behind a `threading.Barrier`; both records must survive. (This test *fails* without the fix — confirmed during development.)
- **`test_concurrent_upserts_produce_valid_json`** — 10 threads upsert concurrently; resulting file must be valid JSON.

## Test results

```
23 passed in 0.40s
```

All pre-existing store tests continue to pass. New concurrent tests pass.

## Checklist

- [x] `ruff check` passes
- [x] `ruff format --check` passes
- [x] `mypy app/integrations/store.py --ignore-missing-imports` — no errors in `store.py`
- [x] All 23 store unit tests pass
- [x] Backward compatible — `_save()` public signature unchanged; existing callers unaffected

### DIFF
--- a/app/cli/wizard/config.py
+++ b/app/cli/wizard/config.py
@@ -72,6 +72,7 @@ ANTHROPIC_MODELS = (
 )
 
 OPENAI_MODELS = (
+    ModelOption(value="gpt-5.5", label="GPT-5.5 — frontier, released 2026-04-23"),
     ModelOption(value=OPENAI_REASONING_MODEL, label="GPT-5.4"),
     ModelOption(value="gpt-5.4-mini", label="GPT-5.4 mini"),
     ModelOption(value="gpt-5.4-nano", label="GPT-5.4 nano"),
@@ -80,6 +81,7 @@ OPENAI_MODELS = (
 
 OPENROUTER_MODELS = (
     ModelOption(value=OPENROUTER_REASONING_MODEL, label="OpenRouter Auto (smart routing)"),
+    ModelOption(value="openai/gpt-5.5", label="GPT-5.5 (via OpenRouter)"),
     ModelOption(value="openai/gpt-5.2", label="GPT-5.2 (via OpenRouter)"),
     ModelOption(value="anthropic/claude-opus-4.6", label="Claude Opus 4.6 (via OpenRouter)"),
     ModelOption(value="anthropic/claude-sonnet-4.5", label="Claude Sonnet 4.5 (via OpenRouter)"),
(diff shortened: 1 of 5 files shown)


muddlebee:
already assigned

PR #1301, opened 2026-05-05
fix(store): atomic concurrent writes for integration store

### LINKED ISSUE #1272: [BUG] Local integration store writes can lose updates under concurrent access

kespineira:
Fixes #1272

Make `~/.opensre/integrations.json` writes atomic and concurrency-safe across threads and processes by locking every read-modify-write operation and replacing direct writes with durable atomic file replacement.

Key changes:
- Add `filelock>=3.29.0` to `[project].dependencies` in `pyproject.toml`.
- Add dynamic per-store `FileLock` locking around write-side load -> mutate -> save operations.
- Replace direct writes with unique temp-file writes, `fsync`, `os.replace`, and best-effort `chmod(0o600)`.
- Protect v1 write-through migration and legacy `.tracer` -> `.opensre` migration under the same store lock.
- Preserve read-time v1 migration behavior if write-through fails: return in-memory v2 data and log the persist failure.
- Avoid rewriting the store for no-op removes, while still persisting v1-to-v2 migration discovered during a locked update.
- Keep read-only APIs lock-free; readers see either the old or new complete file because writes use `os.replace`.
- Add thread, cross-process, atomic failure, permission, and migration edge-case regression tests.

```text
$ uv run python -m pytest tests/integrations/test_store*.py -q
29 passed in 0.90s

$ uv run ruff check app/integrations/store.py tests/integrations/test_store_concurrency.py
All checks passed!

$ uv run ruff format --check app/integrations/store.py tests/integrations/test_store_concurrency.py
2 files already formatted

$ uv run mypy app/integrations/store.py
```

---

The bug was that `app/integrations/store.py` performed unprotected read-modify-write operations against `~/.opensre/integrations.json`, so concurrent writes could silently overwrite each other and lose integration configuration.

I considered hand-rolled platform-specific locking (`fcntl` on POSIX and Windows-specific APIs) and a portable library. I chose `filelock` because it is cross-platform, pure Python, already resolved in this project environment, and keeps the implementation focused on store semantics rather than OS-specific lock behavior.

The key functions added or changed:
- `_atomic_write(dest, data)`: writes JSON to a unique temp file in the same directory, flushes and fsyncs it, atomically replaces the destination with `os.replace`, then applies best-effort private permissions.
- `_locked_update(mutator)`: the single write-side entry point. It acquires a dynamic lock derived from the current `STORE_PATH`, performs legacy migration if needed, loads the store, applies the mutator, and persists when either the mutator changed data or the load performed v1-to-v2 migration.
- `_load_raw()`: stays lock-free for the common v2 read path, but acquires the lock for legacy move or v1 write-through migration. If v1 migration persistence fails during a read, it logs and returns migrated in-memory v2 data to preserve previous behavior.

All public mutator functions (`upsert_integration`, `remove_integration`, `upsert_instance`, `remove_instance`) now delegate to `_locked_update`, ensuring every load -> mutate -> save cycle is serialized.

The tests cover concurrent thread writes, cross-process writes, failed atomic replacement cleanup, permission preservation, concurrent v1 migration, write-through failure tolerance, and no-op locked updates that still persist discovered v1 migrations.

---

---


app/integrations/store.py
```diff
@@ -129,34 +209,84 @@ def _migrate_legacy_store_if_needed() -> None:
...
+            data, _ = _load_raw_unlocked()
+            if data.get("version") != _VERSION:
+                data, _ = _migrate_if_needed(data)
+            mutator(data)
+            _save_unlocked(data)
+            return data
+    except Timeout as exc:
+        raise RuntimeError(
+            f"Integration store locked: {_lock_path()} (store: {STORE_PATH})"
+        ) from exc
```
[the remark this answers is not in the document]
kespineira: Fixed

app/integrations/store.py
```diff
@@ -90,8 +97,58 @@ def _migrate_if_needed(data: dict[str, Any]) -> tuple[dict[str, Any], bool]:
...
+        dest.chmod(0o600)
+    except Exception:
+        if tmp_path_str:
+            with contextlib.suppress(OSError):
+                os.unlink(tmp_path_str)
+        raise
+    finally:
+        if fd is not None and tmp_path_str is None:
+            with contextlib.suppress(OSError):
+                os.close(fd)
```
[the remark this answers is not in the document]
kespineira: Fixed

muddlebee:
rebase with main, should fix the CI issues

kespineira:
@muddlebee CI is passing now, thanks 😄

app/integrations/store.py
```diff
@@ -37,16 +37,23 @@
...
+from filelock import FileLock, Timeout
+
 from app.constants import INTEGRATIONS_STORE_PATH, LEGACY_INTEGRATIONS_STORE_PATH
 
 logger = logging.getLogger(__name__)
 
 STORE_PATH = INTEGRATIONS_STORE_PATH
 LEGACY_STORE_PATH = LEGACY_INTEGRATIONS_STORE_PATH
 _VERSION = 2
+_LOCK_TIMEOUT_SECONDS = 10.0
```
muddlebee, line 56: filelock timeouts turn into RuntimeError from load/save; callers that only handled OSError should be updated or this should be documented on the public entrypoints.

app/integrations/store.py
```diff
@@ -90,8 +97,55 @@ def _migrate_if_needed(data: dict[str, Any]) -> tuple[dict[str, Any], bool]:
...
+    dest.parent.mkdir(parents=True, exist_ok=True)
+    serialized = json.dumps(data, indent=2) + "\n"
+    fd: int | None = None
+    tmp_path_str: str | None = None
+    try:
+        fd, tmp_path_str = tempfile.mkstemp(
+            dir=dest.parent,
+            prefix=dest.name + ".tmp",
+        )
+        with os.fdopen(fd, "w") as f:
```
muddlebee: Open the temp file with encoding="utf-8" so writes match Path.write_text and are not tied to the process default encoding.

pyproject.toml
```diff
@@ -64,6 +64,8 @@ dependencies = [
     "PyNaCl>=1.5.0",
     # MariaDB
     "pymysql>=1.1.0,<2.0.0",
+    # File locking for atomic store operations
+    "filelock>=3.29.0",
```
muddlebee, line 69: If CI uses uv sync --locked (or equivalent), refresh and commit uv.lock alongside this direct dependency.

tests/integrations/test_store_concurrency.py
```diff
@@ -0,0 +1,312 @@
...
+                {"credentials": {"token": f"t{idx}"}},
+            )
+        except Exception as exc:
+            errors.append(exc)
+
+    threads = [threading.Thread(target=worker, args=(i,)) for i in range(num_threads)]
+    for t in threads:
+        t.start()
+    for t in threads:
+        t.join(timeout=30)
```
muddlebee: After join(timeout=...), assert not t.is_alive() so slow or stuck workers fail the test instead of asserting on a partially updated store.

app/integrations/store.py
kespineira: Makes sense. I changed this to raise a store-specific IntegrationStoreLockTimeout that subclasses TimeoutError, so it is still compatible with except OSError handlers while giving callers a clearer exception type.

app/integrations/store.py
kespineira: Good point. Updated the atomic write path to use encoding="utf-8" and made reads explicit UTF-8 as well.

tests/integrations/test_store_concurrency.py
kespineira: Agreed. I added helpers that assert threads/processes are no longer alive after the timeout, and applied them across the concurrency tests so hangs fail explicitly instead of checking partial state.

pyproject.toml
kespineira: Ah yep, missed that uv.lock is tracked now. Updated it

muddlebee:
@greptileai review

tests/integrations/test_store_concurrency.py
```diff
@@ -0,0 +1,358 @@
...
+        records = data.get("integrations", [])
+        assert len(records) == 1
+        assert records[0].get("instances") is not None
+
+    # On disk should be v2 as well
+    final_data = _read_json(tmp_store)
+    assert final_data.get("version") == 2
+
+
+def test_lock_timeout_raises_store_specific_oserror(tmp_store: Path) -> None:
```
muddlebee, line 372: Add a regression test for _load_raw when did_migrate=True and _acquire_lock times out. This path should return migrated in-memory data instead of raising.

app/integrations/store.py
```diff
@@ -129,34 +213,88 @@ def _migrate_legacy_store_if_needed() -> None:
...
+                    try:
+                        _save_unlocked(data2)
+                    except OSError:
+                        logger.warning(
+                            "Failed to persist v2 migration; continuing with in-memory v2",
+                            exc_info=True,
+                        )
+                return data2
+        except Timeout as exc:
+            raise _lock_timeout_error() from exc
```
muddlebee: Handle lock timeout on read-time v1 migration as best-effort persistence failure. Return in-memory migrated v2 data and log instead of raising.

muddlebee:
taking a look at the issues and pushing changes

muddlebee:
@greptileai review

PR #1305, opened 2026-05-05
Fix/bedrock non anthropic models

shubh586:
### Summary
This PR overhauls the `BedrockLLMClient` to unlock support for **all** Amazon Bedrock foundation models (such as Mistral, LLaMA, and Amazon Titan) by integrating the model-agnostic `boto3` Bedrock `converse` API. 

### Before and After
**🔴 Before this PR:**
The `BedrockLLMClient` was strictly hardcoded to use the `AnthropicBedrock` SDK. 
- If a user attempted to use any non-Anthropic model (e.g., `mistral.mistral-large-2402-v1:0`), the investigation would get STUCK and fail with a generic `'NoneType' object is not iterable` error.

**🟢 After this PR:**
The client now dynamically inspects the requested model ID.
- Anthropic Claude models are still safely routed to the `AnthropicBedrock` SDK (ensuring zero regressions).
- **All other models** are automatically routed through a new `_invoke_converse` handler that uses `boto3.client("bedrock-runtime")`. This translates OpenSRE's internal messages into Bedrock's native `converse` schema.

### Steps to reproduce
1. Configure your environment variables:
   - `LLM_PROVIDER=bedrock`
   - `BEDROCK_REASONING_MODEL=mistral.mistral-large-2402-v1:0`
2. Run the `opensre` interactive shell.
3. Send a prompt to the agent.
4. **Before:** The assistant fails with a `'NoneType'` error. **After:** The Mistral model successfully executes the investigation.

### DIFF
--- a/app/services/llm_client.py
+++ b/app/services/llm_client.py
@@ -188,7 +188,9 @@ class BedrockLLMClient:
         self._aws_region = os.getenv("AWS_REGION", os.getenv("AWS_DEFAULT_REGION", "us-east-1"))
 
         if self._use_anthropic:
-            self._anthropic_client: AnthropicBedrock | None = AnthropicBedrock(aws_region=self._aws_region)
+            self._anthropic_client: AnthropicBedrock | None = AnthropicBedrock(
+                aws_region=self._aws_region
+            )
             self._boto3_client: Any = None
         else:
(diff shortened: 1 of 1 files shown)


muddlebee:
would be ideal to open a github issue first :)

PR #1308, opened 2026-05-05
fix/added universal Bedrock model support using Converse API. fixes #1306

### LINKED ISSUE #1306: [FEATURE]support non-Anthropic models (Mistral/Llama) via Bedrock Converse API
### Description
Currently, the `BedrockLLMClient` is hardcoded to use the `AnthropicBedrock` SDK. While this works perfectly for Claude models, it prevents users from utilizing any other foundation models hosted on Amazon Bedrock (such as Mistral, gpt oss or Amazon Titan).

When a non-Anthropic model ID is configured, the client fails during initialization or execution with a `'NoneType' object is not iterable` error, because the Anthropic SDK does not support those model identifiers.

### Proposed Solution
Update `BedrockLLMClient` to:
1. Dynamically detect if a model is an Anthropic Claude model.
2. Route Claude models through the existing `AnthropicBedrock` SDK.
3. Route all other models through the `boto3` Bedrock `converse` API to allow universal model support on Bedrock.

### Impact
This will allow OpenSRE users to use the entire Bedrock model catalog, providing more flexibility for cost, performance, and regional availability.

shubh586:
Fixes #1306

### Summary
This PR overhauls the `BedrockLLMClient` to unlock support for **all** Amazon Bedrock foundation models (such as Mistral, LLaMA, and Amazon Titan) by integrating the model-agnostic `boto3` Bedrock `converse` API. 

### Before and After
**🔴 Before this PR:**
The `BedrockLLMClient` was strictly hardcoded to use the `AnthropicBedrock` SDK. 
- If a user attempted to use any non-Anthropic model (e.g., `mistral.mistral-large-2402-v1:0`), the investigation would get STUCK and fail with a generic `'NoneType' object is not iterable` error.

**🟢 After this PR:**
The client now dynamically inspects the requested model ID.
- Anthropic Claude models are still safely routed to the `AnthropicBedrock` SDK (ensuring zero regressions).
- **All other models** are automatically routed through a new `_invoke_converse` handler that uses `boto3.client("bedrock-runtime")`. This translates OpenSRE's internal messages into Bedrock's native `converse` schema.

### Steps to reproduce
1. Configure your environment variables:
   - `LLM_PROVIDER=bedrock`
   - `BEDROCK_REASONING_MODEL=mistral.mistral-large-2402-v1:0`
2. Run the `opensre` interactive shell.
3. Send a prompt to the agent.
4. **Before:** The assistant fails with a `'NoneType'` error. **After:** The Mistral model successfully executes the investigation.


muddlebee:
@shubh586 fix review comments

app/services/llm_client.py
```diff
@@ -212,6 +253,78 @@ def invoke(self, prompt_or_messages: Any) -> LLMResponse:
...
+        output_message = response.get("output", {}).get("message", {})
+        content_blocks = output_message.get("content", [])
+        text_parts: list[str] = []
+        for block in content_blocks:
+            if "text" in block:
+                text_parts.append(block["text"])
+        content = "\n".join(text_parts).strip()
+        if not content:
+            content = str(response)
+        return LLMResponse(content=content)
```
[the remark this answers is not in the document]
muddlebee: str(response) on a boto3 dict dumps ResponseMetadata, usage, and stopReason, which will break downstream JSON parsing. Raise a clear error instead.

app/services/llm_client.py
```diff
@@ -212,6 +253,78 @@ def invoke(self, prompt_or_messages: Any) -> LLMResponse:
         content = _extract_text(response)
         return LLMResponse(content=content)
 
+    def _invoke_converse(self, prompt_or_messages: Any) -> LLMResponse:
```
muddlebee, line 257: Add unit tests for _invoke_converse and _is_anthropic_bedrock_model in tests/services/test_llm_client.py. The new boto3 path has zero coverage for message conversion, response parsing, and error handling.

app/services/llm_client.py
```diff
@@ -212,6 +253,78 @@ def invoke(self, prompt_or_messages: Any) -> LLMResponse:
...
+        # Convert to converse API message format
+        converse_messages = [
+            {
+                "role": msg["role"],
+                "content": [{"text": msg["content"]}]
+                if isinstance(msg["content"], str)
+                else msg["content"],
+            }
+            for msg in messages
+        ]
```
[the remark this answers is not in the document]
muddlebee: _normalize_messages always produces strings, so the else branch is unreachable. Remove it to make intent explicit.

shubh586:
Thanks for the feedback @muddlebee! I'm working on it!

app/services/llm_client.py
```diff
@@ -147,16 +148,54 @@ def invoke(self, prompt_or_messages: Any) -> LLMResponse:
...
+      - ``arn:aws:bedrock:*:application-inference-profile/*`` pointing at Claude
+
+    For ARN-based application inference profiles we cannot reliably tell the
+    underlying model from the ID alone, so we default to Anthropic (the
+    original behaviour).  Override by setting ``BEDROCK_USE_CONVERSE=1``.
+    """
+    model_lower = model_id.lower()
+    if "anthropic.claude" in model_lower:
+        return True
+    # Application inference profile ARNs — default to Anthropic unless overridden
```
muddlebee: Application inference profile ARNs are forced to Anthropic unless BEDROCK_USE_CONVERSE=1. Non-Claude profiles still fail on AnthropicBedrock. Default these ARNs to converse or resolve the profile target model before routing.

shubh586: got it! I’ll update the routing logic.

muddlebee:
@greptileai review

muddlebee:
@shubh586 thank you for efforts, but changes not as per contributor guidlines and standards. next time please take care :)

shubh586:
Thank them for review

PR #1087, opened 2026-04-29
feat: Add router prompt proof with synthetic alert validation (#656)

7vignesh:
Fixes #656

This PR improves the router prompt reliability to ensure synthetic incident payloads are routed to the RCA-capable `tracer_data` path instead of the general chat path. 

**Changes:**

1. **Enhanced ROUTER_PROMPT** (`app/constants/prompts.py`)
   - Clarified binary classification between `tracer_data` and `general`
   - Explicitly documented synthetic incident payload patterns (e.g., `alertname`, `severity`, `state=alerting`, `db_instance_identifier`)
   - Distinguished conceptual SRE questions (e.g., "What is CrashLoopBackOff?") from incident investigation requests
   - Added concrete examples to guide LLM decision-making

2. **Added Unit Test with Synthetic Alert Proof** (`tests/nodes/test_chat.py`)
   - New test `test_router_improvement_with_synthetic_rds_alert()` uses actual synthetic RDS alert from `tests/synthetic/rds_postgres/002-connection-exhaustion/alert.json`
   - Demonstrates that realistic alert summaries are classified as `tracer_data` route
   - Documents baseline vs improved routing behavior in test docstring

3. **Added RCA Markdown Test** (`tests/e2e/rca/rds-connection-exhaustion.md`)
   - End-to-end validation using synthetic connection-exhaustion scenario
   - Specifies investigation expectations (metric extraction, root cause analysis)
   - Provides downstream hook for RCA validation

### Demo/Screenshot for feature changes and bug fixes

**Unit Test Results:**
image

**Test with Synthetic Alert Payload:**
image

## Before/After Proof:
**Input (Synthetic RDS Alert):**

[synthetic-rds] Connection Exhaustion On payments-prod | state=alerting | alertname=RDSDatabaseConnectionsHigh | severity=critical | summary=DatabaseConnections reached 98% of max_connections and application traffic started receiving too many clients errors.

**Router Output:**
tracer_data ✅ (RCA-capable path)

**Improved Router Prompt (System Message):**

Classify the user message as one of:
- tracer_data
- general

Choose tracer_data when the user message is incident-bearing or asks for RCA/investigation that should use
evidence from systems (logs, metrics, traces, alerts, run/task/job failures, service health, Sentry, GitHub).
This includes pasted/paraphrased alerts and synthetic incident payloads (for example text with fields like:
alertname, severity, state=alerting, summary, error, cluster_name, kube_namespace, db_instance_identifier).

[...]

Issue #657, opened 2026-04-18
Improve the tool-backed chat system prompt to reduce speculative RCA answers


davincios:
## Summary
The tool-backed chat system prompt in `app/constants/prompts.py` asks the assistant to help with investigation and root cause analysis, but it does not strongly enforce several behaviors that matter for reliable RCA in chat:

- clearly separating observation from inference
- ruling out alternatives when evidence is mixed
- avoiding premature root-cause claims before enough evidence has been gathered

That can lead to plausible-sounding but weak RCA answers in the chat flow.

## Goal
Improve the chat investigation system prompt so tool-backed RCA responses are more evidence-first, less speculative, and more aligned with the expectations of the synthetic RCA suites.

## Acceptance Criteria
- The prompt more strongly biases the assistant toward evidence-backed RCA and uncertainty calibration.
- Mixed-signal incidents are less likely to receive premature or overconfident root-cause claims.
- The improved behavior is demonstrated on a synthetic Kubernetes or RDS case from `tests/synthetic/`.
- The prompt still supports ordinary investigative chat without becoming overly rigid or verbose.

## Required Proof
Use at least one synthetic scenario from `tests/synthetic/rds_postgres/` or `tests/synthetic/eks/`, ideally one with misleading or incomplete evidence.

Include all of the following in the PR or issue follow-up:

- the exact synthetic scenario(s) used
- the baseline chat-style RCA response or failure mode before the prompt change
- the improved chat-style RCA response after the prompt change
- evidence that the improved answer better matches the expected synthetic RCA outcome or no longer makes the earlier speculative error

## Notes
Manual chat transcripts are acceptable proof here if they are grounded in synthetic alert/evidence fixtures from `tests/synthetic/`. If you can automate part of the demonstration, even better.

Davidson3556:
can I be assigned this @davincios @VaibhavUpreti

muddlebee:
Done

muddlebee:
@Davidson3556 sorry for confusion, but @7vignesh already opened a PR for this, but closed now.
It needs more community feedback on how to tackle the issue..

pls discuss at #research-dse-bench channel at discord.

PR #1311, opened 2026-05-05
fix(prompts): make chat router reliably route incident messages to tracer_data (#656)

### LINKED ISSUE #656: Improve the router prompt so synthetic incident reports enter the RCA path reliably
## Summary
The router prompt in `app/constants/prompts.py` makes a binary decision between `tracer_data` and `general`. That makes it a reliability choke point for incident analysis in chat: if an alert-like request is routed to the general branch, the system can skip evidence-backed investigation entirely.

## Goal
Improve the router prompt so incident-bearing user messages, especially pasted or paraphrased synthetic alerts from `tests/synthetic/`, are routed into the RCA-capable path more reliably while still avoiding obvious false positives.

## Acceptance Criteria
- Synthetic alert-like messages from `tests/synthetic/` consistently route to `tracer_data`.
- Clearly non-investigation prompts still route to `general`.
- The router prompt better distinguishes conceptual SRE questions from requests that need evidence-backed RCA.
- The change is backed by proof that better routing leads to more reliable RCA handling for at least one synthetic Kubernetes or RDS alert.

## Required Proof
Use at least one alert or alert-derived message from `tests/synthetic/rds_postgres/` or `tests/synthetic/eks/`, for example an `alert.json` payload or a realistic pasted alert summary derived from it.

Include all of the following in the PR or issue follow-up:

- the exact synthetic scenario(s) used
- the baseline routing behavior before the prompt change
- the improved routing behavior after the prompt change
- one downstream RCA demonstration showing that the correctly routed synthetic alert now reaches the evidence-backed investigation path and produces the expected analysis behavior

## Notes
Because the router sits in chat rather than the direct investigation CLI path, the proof can include a focused router/chat harness plus a downstream RCA run using the same synthetic alert. The important part is demonstrating that better routing improves RCA reliability, not just classification accuracy in isolation.

cloudenochcsis:
Fixes #656

The previous `ROUTER_PROMPT` in `app/constants/prompts.py` was three short bullets and tended to send borderline incident-bearing messages — pasted alert JSON, paraphrased alerts, brief problem reports referencing concrete infrastructure — to the `general` branch, where the assistant cannot query connected systems to back its answer with evidence. That makes the router a reliability choke point for chat-mode RCA, exactly as the issue describes.

This PR rewrites `ROUTER_PROMPT` to:

- Enumerate strong signals for `tracer_data`: pasted alert payloads (JSON with `alertname`, `severity`, `state`, `service`, `db_instance`, `namespace`), concrete infra references (`pod`, `deployment`, `cluster`, `RDS instance`, `run_id`, `job_id`, `host`, `service name`), and problem signals (`error`, `crashloop`, `oomkilled`, `lag`, `exhaustion`, `5xx`, "is down", "isn't working").
- Enumerate strong signals for `general`: greetings, conceptual SRE questions, how-to/explain/define/compare requests, best-practice discussion.
- Add an explicit **tie-breaker**: if the message names a specific running system AND describes a problem with it, choose `tracer_data` even if the wording is brief.
- Include four worked examples grounded in the synthetic fixture shapes already in the repo (`RDSReplicationLagHigh` from `tests/synthetic/rds_postgres/001-replication-lag/alert.json`, the `payments-api` pod naming from `tests/synthetic/eks/001-oomkilled-crashloop/alert.json`, plus two conceptual prompts).

`router_node` was previously untested, so this PR also adds the first coverage for it:

- **`tests/nodes/test_chat_router.py`** — 7 mocked-LLM unit tests covering empty-message handling, last-message role gating, label parsing (whitespace and case normalisation), unknown-label fallback, and that `ROUTER_PROMPT` is wired into the system role with the user message in the user role. Runs under `make test-cov` and is fast (~1.3s).
- **`tests/synthetic/test_router_routing.py`** — gated by `@pytest.mark.synthetic` so it only runs under `make test-synthetic`. Parametrised over **every non-healthy `alert.json`** under `tests/synthetic/{rds_postgres,eks}/` (27 fixtures) asserting each routes to `tracer_data`, plus **5 hand-curated conceptual prompts** that must route to `general`.

This is the proof the issue's "Required Proof" section asks for: alert-derived messages from both `tests/synthetic/rds_postgres/` and `tests/synthetic/eks/` must consistently route to the RCA-capable path.

All four pre-push quality gates pass on `fix/router-prompt-incident-routing`:

```
$ make lint
.venv/bin/python -m ruff check app/ tests/
All checks passed!

$ make format-check
.venv/bin/python -m ruff format --check app/ tests/
1075 files already formatted

$ make typecheck
.venv/bin/python -m mypy app/
Success: no issues found in 458 source files

$ make test-cov
================ 4981 passed, 2 skipped, 21 warnings in 48.57s =================
TOTAL                                                            29853   8505    72%
```

The new mocked-LLM unit tests in isolation:

```
$ python -m pytest tests/nodes/test_chat_router.py -v
collected 7 items

tests/nodes/test_chat_router.py::test_router_returns_general_when_no_messages PASSED
tests/nodes/test_chat_router.py::test_router_returns_general_when_last_message_is_not_user PASSED
tests/nodes/test_chat_router.py::test_router_returns_tracer_data_when_llm_says_so PASSED
tests/nodes/test_chat_router.py::test_router_returns_general_when_llm_says_so PASSED
tests/nodes/test_chat_router.py::test_router_normalizes_whitespace_and_case PASSED
tests/nodes/test_chat_router.py::test_router_falls_back_to_general_for_unknown_label PASSED
tests/nodes/test_chat_router.py::test_router_uses_router_prompt_as_system PASSED

============================== 7 passed in 1.27s ===============================
```

The synthetic real-LLM layer (`tests/synthetic/test_router_routing.py`) is wired to the same `@pytest.mark.synthetic` mechanism the existing scenario suites use (`tests/synthetic/eks/test_suite.py`, `tests/synthetic/rds_postgres/test_suite.py`) and collects 32 parametrised cases:

```
$ python -m pytest tests/synthetic/test_router_routing.py --collect-only
========================= 32 tests collected in 0.88s ==========================
```

It is excluded from the standard `make test-cov` run (per the existing `-m "not synthetic" --ignore=tests/synthetic` policy in the Makefile) and intended to be exercised by maintainers under `make test-synthetic` with a real LLM provider configured.

---

The problem: the binary decision in `router_node` (`app/nodes/chat.py:170`) is a single LLM call that decides whether the chat agent gets to use tools or not. If a user pastes a Datadog alert payload but the router classifies the message as `general`, the assistant falls back to `general_node` (no tool access) and answers from generic SRE knowledge — exactly the failure mode the issue calls out. The previous prompt only described `tracer_data` as "asking to investigate an alert/incident or requesting analysis" plus a parenthetical list of data sources. That's information about the destination, not signals on the *input* the router actually sees.

Alternatives considered:

1. **Replace the router with a regex/heuristic classifier in front of the LLM call.** Rejected: the issue is specifically about the *prompt* being a reliability choke point, not the architecture. A heuristic would also misclassify paraphrased alerts that don't contain JSON markers but do describe an incident in plain English.
2. **Add few-shot examples only, leave the bullet structure unchanged.** Considered, but the existing wording already overlapped (e.g. "best practices" was an unanchored marker for `general` while the `tracer_data` definition didn't mention what alert payloads look like). Examples without restructured signals would have been a smaller change but not addressed the underlying ambiguity.
3. **Restructure with explicit signal lists, a tie-breaker, and four examples (chosen).** This makes the routing criteria observable on the *input message* (what the LLM actually has to work with) rather than on the kind of work the user wants done. The tie-breaker exists to handle the genuinely ambiguous case of a user typing one short sentence ("payments-api crashloop on x7gr9, can you look?") that names a real system and a real failure mode without using any of the JSON markers — that is the exact shape of message the issue's acceptance criteria require to reach the RCA path.

Key code/components added:

- **`app/constants/prompts.py:27-55`** — the rewritten `ROUTER_PROMPT`. The structure is: definition → `tracer_data` signals → `general` signals → tie-breaker → 4 examples → output-format instruction. The output-format instruction (`Respond with ONLY: tracer_data or general`) is preserved verbatim from the previous prompt so the downstream parsing in `router_node` (`app/nodes/chat.py:182`) continues to work without changes. Examples are grounded in real fixtures (`RDSReplicationLagHigh`, `payments-api`) rather than invented strings, so they reflect what actual users paste in.
- **`tests/nodes/test_chat_router.py`** — covers the wiring around `ROUTER_PROMPT` so the router-prompt change can't silently regress: `test_router_uses_router_prompt_as_system` asserts the prompt is loaded into the system role and the latest user message into the user role; the parsing tests cover `.strip().lower()` normalisation in `chat.py:182` and the safe-fallback to `general` in `chat.py:183` for unknown labels. These are mocked because they're testing wiring, not LLM behaviour.
- **`tests/synthetic/test_router_routing.py`** — covers the prompt's *behavioural* change against a real LLM. It is placed under `tests/synthetic/` rather than `tests/nodes/` because `make test-synthetic` collects from that directory only, and it uses `@pytest.mark.synthetic` so the standard `make test-cov` pipeline (which sets `-m "not synthetic"`) skips it. The retry helper `_route_with_retry` mirrors the `_LLM_ATTEMPTS = 2` pattern in the existing scenario suites, accepting one stochastic flake per case before failing. The negative-set prompts (`"what is a circuit breaker?"`, etc.) are deliberately phrased without infra entities so a correct classifier should never route them to `tracer_data`.

Edge cases handled:

- **Empty / non-user trailing message:** `router_node` already short-circuits to `general` and returns without calling the LLM. The two unit tests at the top of `test_chat_router.py` lock this in.
- **LLM returns whitespace, mixed case, or trailing newline:** `chat.py:182` already normalises with `.strip().lower()`. `test_router_normalizes_whitespace_and_case` verifies it.
- **LLM returns a label outside the two valid options:** `chat.py:183` already falls back to `general`. `test_router_falls_back_to_general_for_unknown_label` verifies it.
- **Recovered / spurious / red-herring alerts:** the synthetic suite includes scenarios like `eks/011-recovered-rollout`, `eks/013-spurious-alert-storm`, and `rds_postgres/013-storage-recovery-false-alert`. These are still alerts in the user-message sense (a person is asking about an alert payload), so they should still route to `tracer_data` — the question of whether the RCA path then concludes "actually fine, here's the recovered timeline" is the job of downstream nodes, not the router. The synthetic test explicitly includes all of these.

---

### DIFF
--- a/app/constants/prompts.py
+++ b/app/constants/prompts.py
@@ -24,9 +24,31 @@ for concrete details they can share (alert text, error snippets, timelines) or u
 
 Always respond in clear markdown."""
 
-ROUTER_PROMPT = """Classify the user message:
+ROUTER_PROMPT = """Classify the user message into one of two routes.
 
-- "tracer_data" if the user is asking to investigate an alert/incident or requesting analysis that likely requires querying data (e.g., logs, metrics, traces, failed runs/tasks/jobs, error messages, service health, Sentry issues, GitHub code/history).
-- "general" for general questions, greetings, or best practices
+Route to "tracer_data" when the user is asking you to investigate an actual
+incident, triage an alert, or analyze production behavior — anything that
+needs you to query their connected systems (Tracer runs, logs, metrics,
+traces, failed jobs, Sentry, GitHub). Strong signals:
+- Pasted alert payloads (JSON with fields like alertname, severity, state,
+  service, db_instance, namespace) or paraphrased alert text
+- Concrete infra references: pod, deployment, cluster, RDS instance, run_id,
+  job_id, host, service name
+- A problem signal alongside those references: error, failing, crashloop,
+  oomkilled, lag, exhaustion, saturated, 5xx, "is down", "isn't working"
+
+Route to "general" for everything else: greetings, conceptual SRE questions,
+how-to/explain/define/compare requests, best-practice discussions, and
+hypotheticals with no specific system to investigate. Strong signals:
+"what is", "how do I", "explain", "best practice", "in general", "should we".
+
+Tie-breaker: if the message names a specific running system AND describes a
+problem with it, choose "tracer_data" even if the wording is brief.
+
+Examples:
+- {"alertname": "RDSReplicationLagHigh", "severity": "critical", ...} -> tracer_data
+- "payments-api crashloop on x7gr9, can you look?" -> tracer_data
+- "what is a circuit breaker?" -> general
+- "what should our SLO target be for a payment API?" -> general
 
 Respond with ONLY: tracer_data or general"""
--- a/tests/nodes/test_chat_router.py
+++ b/tests/nodes/test_chat_router.py
@@ -1,0 +1,75 @@
+"""Tests for chat router_node intent classification (mocked LLM)."""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from app.nodes import chat as chat_mod
+
+
+def _mock_router_response(content: str) -> MagicMock:
+    response = MagicMock()
+    response.content = content
+    return response
+
+
+@pytest.fixture
+def mock_router_llm():
+    """Patch get_llm_for_tools so router_node never makes a real LLM call."""
+    with patch.object(chat_mod, "get_llm_for_tools") as mock_factory:
+        llm = MagicMock()
+        mock_factory.return_value = llm
+        yield llm
+
+
+def test_router_returns_general_when_no_messages(mock_router_llm: MagicMock) -> None:
+    out = chat_mod.router_node({"messages": []})
+    assert out == {"route": "general"}
+    mock_router_llm.invoke.assert_not_called()
+
+
+def test_router_returns_general_when_last_message_is_not_user(
+    mock_router_llm: MagicMock,
+) -> None:
+    out = chat_mod.router_node({"messages": [{"role": "assistant", "content": "hi"}]})
+    assert out == {"route": "general"}
+    mock_router_llm.invoke.assert_not_called()
+
+
+def test_router_returns_tracer_data_when_llm_says_so(mock_router_llm: MagicMock) -> None:
+    mock_router_llm.invoke.return_value = _mock_router_response("tracer_data")
+    out = chat_mod.router_node(
+        {"messages": [{"role": "user", "content": "investigate this alert"}]}
+    )
+    assert out == {"route": "tracer_data"}
+
+
+def test_router_returns_general_when_llm_says_so(mock_router_llm: MagicMock) -> None:
+    mock_router_llm.invoke.return_value = _mock_router_response("general")
+    out = chat_mod.router_node({"messages": [{"role": "user", "content": "what is SLO?"}]})
+    assert out == {"route": "general"}
+
+
+def test_router_normalizes_whitespace_and_case(mock_router_llm: MagicMock) -> None:
+    mock_router_llm.invoke.return_value = _mock_router_response("  Tracer_Data\n")
+    out = chat_mod.router_node({"messages": [{"role": "user", "content": "x"}]})
+    assert out == {"route": "tracer_data"}
+
+
+def test_router_falls_back_to_general_for_unknown_label(mock_router_llm: MagicMock) -> None:
+    mock_router_llm.invoke.return_value = _mock_router_response("maybe_tracer")
+    out = chat_mod.router_node({"messages": [{"role": "user", "content": "x"}]})
+    assert out == {"route": "general"}
+
+
+def test_router_uses_router_prompt_as_system(mock_router_llm: MagicMock) -> None:
+    mock_router_llm.invoke.return_value = _mock_router_response("general")
+    chat_mod.router_node({"messages": [{"role": "user", "content": "hello"}]})
+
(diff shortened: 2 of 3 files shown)


cloudenochcsis:
Hi @muddlebee — caught after pushing this that #1087 was closed yesterday with the team's note that #656 needs a more structured overhaul. Apologies for not surfacing that earlier.

Quick comparison so you can decide whether this is worth a second look or whether it's the same pattern you're closing:

- The change rewrites `ROUTER_PROMPT` structurally rather than adding examples on top of the existing wording — explicit signal lists for both routes and a tie-breaker for terse messages that name a system + describe a problem. The 4 worked examples are grounded in real fixture content (`RDSReplicationLagHigh`, `payments-api`).
- Proof is parametrised over **every** non-healthy `alert.json` under `tests/synthetic/{rds_postgres,eks}/` (27 fixtures) plus 5 conceptual prompts, gated by `@pytest.mark.synthetic`. The mocked layer only covers wiring.
- 7 mocked-LLM unit tests in `tests/nodes/test_chat_router.py` add the first coverage for `router_node` (none existed before).

If #656 is being reframed — or merged with #657 / replaced entirely — I'm happy to close this myself. Just say the word, and I'll wait for the new direction before opening anything else here. No problem either way.

muddlebee:
closing this, it needs more discussion

PR #1344, opened 2026-05-05
feat(benchmark): integrate-cloudopsbench-dataset

w3joe:
## Summary

Adds CloudOpsBench support as a cache-backed Kubernetes RCA benchmark suite for OpenSRE.

- Adds dataset download/validation/run targets for CloudOpsBench, with the benchmark corpus kept out of git and downloaded from Hugging Face.
- Adds a CloudOpsBench case loader, replay backend, benchmark runner, scoring implementation, pytest marker, and README.
- Adds cache-backed Kubernetes replay tools for the official CloudOpsBench actions and wires benchmark evidence through planning, investigation, and root-cause diagnosis.
- Exposes the suite through `opensre tests cloudopsbench` and the CLI test catalog.

## Test plan

- `make download-cloudopsbench-hf`
- `make test-cloudopsbench CLOUDOPSBENCH_LIMIT=10`


WatchTree-19:
nice work @w3joe. CloudOpsBench wiring is the kind of benchmark tier the readme has been missing.

agreed with greptile's two P1s, the standardized_agent_steps one in particular is the showstopper. three smaller things on scoring.py that aren't covered yet:

1) ZTDR is computed per case but never makes it into the summary report. summarize_scores lists "ztdr" in metric_names, sums it, divides by total, then the returned metrics dict at the bottom of the function omits it. so per-case JSON has ztdr but the headline summary the readme block will eventually consume doesn't. two-line fix - add "Zero-Tool Direct Resolution": averages["ztdr"] to the metrics dict around line 605.

2) _infer_fault_object in scoring.py (around line 226) hardcodes the union of Online Boutique and Train-Ticket service names. any future case that introduces a new service or a schema bump from upstream silently regresses local runs to fault_object="" and the partial-match logic stops working. fault_object is already in the case ground truth - the fallback inference path can read it from there rather than re-deriving from text. if you want to keep text-inference as a separate signal, gating it on case.system instead of one global list keeps adding a new system a one-place change.

3) _infer_root_cause (line 193 onwards) uses all(token in text for token in tokens) per rule, and one rule embeds 'redis-cart-invalid' as a required token. that's a fixture name from one case - any other env-var-address-mismatch scenario fails the rule. the AND gating is also too strict in general: db_connection_exhaustion needs both "connection exhaustion" AND "too many connections" to fire, so an agent answer that only uses one phrase mis-classifies as Performance_Fault. similar shape for mysql_invalid_credentials needing all three of "mysql", "access denied", "invalid credentials". probably want each rule to be a list of phrase patterns where any match wins, or a single regex per rule. fixture-specific tokens shouldn't appear in general inference.

separately, _run_cases in run_suite.py with workers > 1 runs run_investigation in a ThreadPoolExecutor. langgraph state is per-run but the LLM client and any module-level caches in tracer-client integrations aren't obviously thread-safe. worth a default of workers=1 with a doc note that >1 is at-your-own-risk until the integration surface is audited.

lgtm shape-wise once the P1s and the ZTDR drop are resolved, the rest can be follow-ups.

app/nodes/root_cause_diagnosis/node.py
```diff
@@ -45,6 +46,178 @@ def _short_circuit_enabled() -> bool:
...
+            {
+                "claim": "CloudOpsBench diagnosis inferred from replayed tool evidence",
+                "validation_status": "validated",
+            }
+        ],
+        "non_validated_claims": [],
+        "validity_score": 1.0,
+        "investigation_recommendations": [],
+        "remediation_steps": [],
+        "investigation_loop_count": state.get("investigation_loop_count", 0),
```
VaibhavUpreti, line 217: could you add a better typed response here, using a `TypedDict`

app/nodes/root_cause_diagnosis/node.py
```diff
@@ -45,6 +46,178 @@ def _short_circuit_enabled() -> bool:
...
+- Startup_Fault
+- Runtime_Fault
+- Service_Routing_Fault
+- Performance_Fault
+
+Valid root causes:
+{_CLOUDOPSBENCH_ROOT_CAUSES}
+
+Fault object format:
+- app/<service>
```
VaibhavUpreti, line 157: move to app/constants/prompts.py

VaibhavUpreti (approved):
Great job @w3joe, I'm merging this shared some comments here and others on slack, could you please follow up in a separate PR!

w3joe:
got it, thanks Vaibhav!

PR #1388, opened 2026-05-06
feat(cli): REPL task lifecycle, /tasks and /cancel

### LINKED ISSUE #1373: Interactive Shell: add real task lifecycle and cancellation primitives (/tasks, /cancel)
## Distinctness
Distinct from #1369 (UX outcome metric) and #1370 (policy gate). This issue is about runtime control-plane mechanics for long-running shell/investigation actions.

## Why
The shell currently handles long-running work inline. `/stop` is only a reminder to press Ctrl+C and there is no first-class task table or cancellation API.

## Problem
- No durable concept of in-flight background tasks.
- No task IDs, status model, or post-run inspectability.
- Cancellation semantics rely on terminal interrupts, not structured control.

## Scope
- Add task model for interactive-shell executions:
  - `task_id`, `kind`, `status`, `started_at`, `ended_at`, `result`, `error`
- Run long operations through task runner abstraction.
- Add slash commands:
  - `/tasks` to list recent/in-flight tasks
  - `/cancel <task_id>` to request cancellation
- Wire cancellation tokens into investigation + synthetic test execution paths.

## Acceptance Criteria
- [ ] Long-running operations are represented as tasks with stable IDs.
- [ ] `/tasks` shows running/completed/failed tasks.
- [ ] `/cancel` can stop a cancellable task and report terminal state.
- [ ] Unit/integration tests verify task transitions + cancellation behavior.

muddlebee:
Fixes #1373

- Add in-process `TaskRegistry` / `TaskRecord` with stable task ids, statuses, and cooperative `threading.Event` cancellation.
- Slash commands: **`/tasks`** lists recent tasks; **`/cancel <id>`** signals cancel (subprocess terminate for background synthetic tests); **`/stop`** messaging updated.
- Wire free-text and `/investigate` investigations plus sample alerts through task records; pass `cancel_requested` into streaming session payload (polled between queue reads).
- Synthetic RDS suite: non-blocking `Popen` + watcher thread; session history records `task:<id>`.
- Tests: task transitions, registry prefix/ambiguity, watcher, sample-alert registry, session clear, command updates.

**Prompt**: 
```
show me which services are connected and after that run a synthetic test RDS database
```

```
/tasks                         # RUNNING synthetic_test
/cancel <paste-full-task-id>
/tasks                         # CANCELLED or FAILED shortly after watcher finishes
/stop                          # reminders for Ctrl+C vs /cancel
```

image

---

Task table is an in-memory ring buffer on `ReplSession` (cleared by `/reset`). Investigations reuse the existing streaming pipeline; cancellation for that path polls a shared event so external callers can cooperate without process-wide SIGINT-only behavior. Synthetic tests move to background `Popen` so `/cancel` can terminate the child. Follow-up improvement (out of scope here): truly non-blocking investigations so `/cancel` is usable mid-stream without Ctrl+C.

---

---

### DIFF
--- a/app/cli/interactive_shell/agent_actions.py
+++ b/app/cli/interactive_shell/agent_actions.py
@@ -9,6 +9,7 @@ import shlex
 import shutil
 import subprocess
 import sys
+import tempfile
 import threading
 import time
 from dataclasses import dataclass
@@ -163,6 +164,8 @@ _SHELL_COMMAND_TIMEOUT_SECONDS = 120
 _SYNTHETIC_TEST_TIMEOUT_SECONDS = 1800
 _SYNTHETIC_POLL_SECONDS = 0.25
 _MAX_COMMAND_OUTPUT_CHARS = 24_000
+_SYNTHETIC_DIAG_CHARS = 2_000  # max stderr bytes captured from a failing synthetic run
+_SIGTERM_GRACE_SECONDS = 10  # wait for clean exit after SIGTERM before escalating to SIGKILL
 _IS_WINDOWS = os.name == "nt"
 
 
@@ -416,17 +419,21 @@ def _terminate_child_process(proc: subprocess.Popen[Any]) -> None:
     """Best-effort SIGTERM → wait → SIGKILL → wait without blocking forever."""
     if proc.poll() is not None:
         return
-    _grace_s = 45
-    _post_kill_wait_s = 5
     with contextlib.suppress(OSError):
         proc.terminate()
     try:
-        proc.wait(timeout=_grace_s)
+        proc.wait(timeout=_SIGTERM_GRACE_SECONDS)
     except subprocess.TimeoutExpired:
         with contextlib.suppress(OSError):
             proc.kill()
         with contextlib.suppress(subprocess.TimeoutExpired):
-            proc.wait(timeout=_post_kill_wait_s)
+            proc.wait(timeout=5)
+
+
+def _read_diag(buf: tempfile.SpooledTemporaryFile[bytes]) -> str:  # type: ignore[type-arg]
+    """Read up to ``_SYNTHETIC_DIAG_CHARS`` bytes from a captured stderr buffer."""
+    buf.seek(0)
+    return buf.read(_SYNTHETIC_DIAG_CHARS).decode("utf-8", errors="replace").strip()
 
 
 def _watch_synthetic_subprocess(
@@ -434,6 +441,7 @@ def _watch_synthetic_subprocess(
     proc: subprocess.Popen[Any],
     session: ReplSession,
     suite_name: str,
+    stderr_buf: tempfile.SpooledTemporaryFile[bytes],  # type: ignore[type-arg]
 ) -> None:
     def _history_text() -> str:
         return f"{suite_name} task:{task.task_id}"
@@ -448,39 +456,53 @@ def _watch_synthetic_subprocess(
     def _run() -> None:
         started = time.monotonic()
         timed_out = False
+        # Track whether *we* explicitly terminated the process so we can
+        # distinguish a cancel-driven exit from a natural exit that happened
+        # to race with a concurrent /cancel.
+        terminated_by_watcher = False
         while proc.poll() is None:
             if time.monotonic() - started > _SYNTHETIC_TEST_TIMEOUT_SECONDS:
                 timed_out = True
                 task.request_cancel()
                 _terminate_child_process(proc)
+                terminated_by_watcher = True
                 break
             if task.cancel_requested.is_set():
                 _terminate_child_process(proc)
+                terminated_by_watcher = True
                 break
             time.sleep(_SYNTHETIC_POLL_SECONDS)
 
-        if timed_out:
-            task.mark_failed(f"timed out after {_SYNTHETIC_TEST_TIMEOUT_SECONDS}s")
-            _record_synthetic_if_current_session(ok=False)
-            return
+        try:
+            if timed_out:
+                task.mark_failed(f"timed out after {_SYNTHETIC_TEST_TIMEOUT_SECONDS}s")
+                _record_synthetic_if_current_session(ok=False)
+                return
 
-        code = proc.returncode
-        if code is None:
-            task.mark_failed("subprocess did not report exit code")
-            _record_synthetic_if_current_session(ok=False)
-            return
+            code = proc.returncode
+            if code is None:
+                task.mark_failed("subprocess did not report exit code")
+                _record_synthetic_if_current_session(ok=False)
+                return
 
-        if task.cancel_requested.is_set():
-            task.mark_cancelled()
-            _record_synthetic_if_current_session(ok=False)
-            return
+            # Honour the real exit code when the process exited on its own.
+            # Only treat as CANCELLED when *we* killed it after a cancel request;
+            # a natural exit that races with /cancel should be recorded by its code.
+            if terminated_by_watcher and task.cancel_requested.is_set():
+                task.mark_cancelled()
+                _record_synthetic_if_current_session(ok=False)
+                return
 
-        if code == 0:
-            task.mark_completed(result="ok")
-            _record_synthetic_if_current_session(ok=True)
-        else:
-            task.mark_failed(f"exit code {code}")
-            _record_synthetic_if_current_session(ok=False)
+            if code == 0:
+                task.mark_completed(result="ok")
+                _record_synthetic_if_current_session(ok=True)
+            else:
+                diag = _read_diag(stderr_buf)
+                error_msg = f"exit code {code}" + (f": {diag}" if diag else "")
+                task.mark_failed(error_msg)
+                _record_synthetic_if_current_session(ok=False)
+        finally:
+            stderr_buf.close()
 
     threading.Thread(target=_run, daemon=True, name=f"synthetic-{task.task_id}").start()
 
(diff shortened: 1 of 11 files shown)


muddlebee:
@greptileai review

muddlebee:
@greptileai review

muddlebee:
@greptileai review

muddlebee:
@greptileai review

muddlebee:
```
The move to a background watcher thread introduces a cross-thread write: if /reset fires while a synthetic test is running, the watcher still holds the original session reference and will append to session.history after it has been cleared, surfacing a stale synthetic-test entry in the post-reset session. Everything else — task lifecycle guards, cancellation event wiring, investigate.py cleanup paths — is implemented correctly and well-tested.

app/cli/interactive_shell/agent_actions.py — the watcher session.record calls are not guarded against a concurrent /reset.
```

fixing this from greptile

muddlebee:
@greptileai review

muddlebee:
@greptileai review

muddlebee:
@greptileai review

muddlebee:
@greptileai review

muddlebee:
@greptileai review

app/cli/interactive_shell/agent_actions.py
```diff
@@ -407,6 +415,98 @@ def _print_planned_actions(console: Console, actions: list[PlannedAction]) -> No
...
+        try:
+            if timed_out:
+                task.mark_failed(f"timed out after {_SYNTHETIC_TEST_TIMEOUT_SECONDS}s")
+                _record_synthetic_if_current_session(ok=False)
+                return
+
+            code = proc.returncode
+            if code is None:
+                task.mark_failed("subprocess did not report exit code")
+                _record_synthetic_if_current_session(ok=False)
```
yashksaini-coder, line 485: ## P2 — Watcher thread `_run()` lacks a top-level exception handler (and now leaks `stderr_buf` on crash)

@muddlebee I ran an injected-fault probe against this watcher: a `proc.poll()` that raises `RuntimeError` on the 3rd call (simulating an unexpected platform/kernel quirk).

Result on `5a34cba3`:
- The exception propagates out of `_run()`, the daemon thread dies via `threading.excepthook`
- `task.status` stays `RUNNING` forever — `/cancel` is a no-op (Event fires into the void since the watcher is dead), `/tasks` permanently shows it as running
- `stderr_buf` is **never closed** (the `try/finally: stderr_buf.close()` block sits *after* the polling loop, so a crash inside the `while proc.poll() is None:` loop bypasses cleanup) → file-descriptor leak

Reproducer (run from repo root):
```python
import threading, time, tempfile
from app.cli.interactive_shell.tasks import TaskKind, TaskStatus
from app.cli.interactive_shell.session import ReplSession
from app.cli.interactive_shell import agent_actions

class BadProc:
    def __init__(self): self.calls = 0
    def poll(self):
        self.calls += 1
        if self.calls > 2: raise RuntimeError("simulated")
        return None
    @property
    def returncode(self): return 99

session = ReplSession()
task = session.task_registry.create(TaskKind.SYNTHETIC_TEST)
task.mark_running()
stderr_buf = tempfile.SpooledTemporaryFile(max_size=64*1024)
agent_actions._watch_synthetic_subprocess(task, BadProc(), session, "rds_postgres", stderr_buf)
time.sleep(2)
assert task.status == TaskStatus.RUNNING  # zombie!
assert not stderr_buf.closed              # fd leak!
```

Real-world triggers (low-probability but non-zero):
- `proc.poll()` raising `OSError` on certain platform/filesystem states
- Future code added inside `_run()` that someone forgets to guard
- Logger/format exceptions if you later add diagnostic logging inside the loop

Suggested fix (~6 lines, locks both issues in):

```python
def _run() -> None:
    started = time.monotonic()
    timed_out = False
    terminated_by_watcher = False
    try:
        while proc.poll() is None:
            # ... existing body ...
        # ... existing post-loop block (the current try/finally) ...
    except Exception as exc:  # noqa: BLE001
        task.mark_failed(f"watcher crashed: {exc}")
        _record_synthetic_if_current_session(ok=False)
        with contextlib.suppress(OSError):
            if proc.poll() is None:
                proc.kill()
    finally:
        with contextlib.suppress(Exception):
            stderr_buf.close()
```

And a regression test in `test_tasks.py`:

```python
def test_watch_marks_failed_on_unexpected_proc_poll_exception(self) -> None:
    """If proc.poll() raises mid-loop, the task must not be left RUNNING."""
    # ... use BadProc above; assert task.status == TaskStatus.FAILED ...
```

Not a merge blocker on its own — `proc.poll()` rarely raises in practice on Linux, and the subprocess will exit naturally even if the watcher dies. But it's the kind of defensive gap that's better closed now than after a flake report comes in. The fix doubles as the cleanup path for the new `stderr_buf` you just added.

All the other concurrency probes I ran (cancel-before-start, cancel-twice, 50-thread race, ring buffer, /reset race, real-subprocess SIGTERM, cooperative-cancel responsiveness) pass — this is the only one that fails.

yashksaini-coder (commented):
## Senior-engineer concurrency audit — comment-only, holding approval on one P2

@muddlebee this PR is in great shape architecturally. The lock discipline in `TaskRecord` is clean, the cancellation semantics are well-thought-out, and the `/reset` cross-thread fix you flagged yourself was addressed correctly with a generation counter. I ran 13 concurrency probes against the actual code on `5a34cba3` and 12 pass; 1 finds a real defensive-coding gap that I'd like to see addressed before approving.

### Local quality gate (HEAD `5a34cba3`)
- `ruff check` clean
- `mypy` clean (only pre-existing unrelated `sentry_sdk` failure)
- `pytest tests/cli/interactive_shell/` — **261 passed**
- CI green

### Verification of the 4 known findings

| # | Source | Status |
|---|---|---|
| 1 | greptile P1 — `session.record(ok=True)` premature | ✅ fixed — only watcher records success path; failure paths use explicit `ok=False` |
| 2 | greptile P2 — naive `datetime.fromtimestamp` | ✅ fixed (`commands.py:754` uses `tz=UTC`) |
| 3 | greptile P2 — 45s SIGKILL grace excessive | ✅ **just fixed in `5a34cba3`** — `_SIGTERM_GRACE_SECONDS = 10`, post-kill drain stays at 5s. Nice tightening. |
| 4 | @muddlebee — `/reset` cross-thread `session.history` write | ✅ fixed via `session.history_generation` counter (commit `0e185e25`); verified by my Probe 11 — `session.history` stays empty after `/reset` even with watcher mid-run |

### What's solid (verified by probes)

- **Lock discipline**: 50 threads concurrently calling `mark_completed`/`mark_cancelled`/`mark_failed`/`request_cancel` → exactly one terminal status wins. Probe 7.
- **`request_cancel()` doesn't hold the lock across `proc.terminate()`**: copies `proc` ref under lock, calls `terminate()` outside. Avoids the classic syscall-under-lock deadlock pitfall.
- **`secrets.token_hex(4)` IDs**: 100/100 unique in Probe 5 (32 random bits, 100-task ring buffer).
- **Ring-buffer eviction**: orphaned records can still mutate safely (no exception, just unfindable via `/cancel`). Probe 6.
- **Cooperative cancel responsiveness**: ~0.5s end-to-end via Probe 13 (Event set → watcher detects on next poll → SIGTERM → process exits).
- **`terminated_by_watcher` flag** in `5a34cba3` correctly distinguishes cancel-driven exits from natural exits that race with `/cancel` — that's a thoughtful fix to a subtle race.
- **Stderr capture for diagnostics** in `5a34cba3` is a real UX win — failing synthetic tests now surface their first 2KB of stderr instead of just `exit code 1`.

### Holding approval on one P2 — see inline comment

Probe 12 (watcher thread crash) is the only failing probe. Details + suggested 5-line fix in the inline comment on `agent_actions.py`. The new stderr_buf parameter actually makes this slightly worse — there's now also a file-descriptor leak path. Both are addressed by the same defensive `try/except` wrapping `_run()`.

### Other observations (non-blocking, file as follow-ups if you want)

1. **No `cancelling` intermediate state**. Between `request_cancel()` and the watcher detecting subprocess exit (up to `_SIGTERM_GRACE_SECONDS = 10` seconds), `/tasks` shows `running`. A `TaskStatus.CANCELLING` value would let `/tasks` show `cancelling…` so users know their cancel landed but the process is still draining. Pure UX polish.
2. **`session.history_generation` read without a lock** (`agent_actions.py:444`). On CPython this is atomic by GIL, but technically relies on implementation detail. Using `itertools.count` or wrapping reads/writes in a lock would be more portable. Practical safety: fine.
3. **Investigation `/cancel` is wired but unreachable from same REPL** — the PR body acknowledges this. The cancel_requested polling in `investigate.py:280` is correct for future signal-handler-driven cancels; only the synchronous REPL blocks usage today. Worth a follow-up issue to track "truly non-blocking investigations."
4. **`/tasks` kind column truncates** to `synthetic_t…` / `investigati…` on 80-col terminals. Cosmetic.

### Probes I ran

1. Idempotent terminal transitions ✅ · 2. Cancel-before-start ✅ · 3. Cancel-twice idempotency ✅ · 4. Cancel-after-end no-op ✅ · 5. Task ID uniqueness (100 IDs) ✅ · 6. Ring-buffer eviction ✅ · 7. **50-thread mark_*/cancel race** ✅ · 8. Concurrent create+list (10 threads, 500 ids) ✅ · 9. Real subprocess SIGTERM ✅ · 10. Ambiguous prefix lookup ✅ · 11. `/reset` while watcher running ✅ · **12. Watcher exception handling** ❌ · 13. Cooperative cancel responsiveness ✅

Let me know when the watcher try/except is in and I'll re-review and approve.

PR #1408, opened 2026-05-06
fix(installer): glibc precheck + preserve binary on smoke-test failure (#1284)

### LINKED ISSUE #1284: [BUG] Linux installer fails on glibc < 2.38, deletes binary on smoke-test failure
### Summary

The Linux installer crashes on any system with glibc < 2.38 with no advance warning. Ubuntu 22.04 (glibc 2.35), RHEL 8, CentOS 8 family are all affected. Installer deletes the binary on smoke-test failure leaving nothing to debug with.

**Expected:** Either a glibc precheck in the installer that prints "your glibc is X, you need Y, supported OS list: ..." and exits cleanly, or a build that targets an older glibc floor so the curl one-liner works on every reasonable Linux.

**Actual:** Installer downloads the binary, runs the smoke test (`opensre --version`), pyinstaller dumps a raw stack trace about `libpython3.13.so.1.0: GLIBC_2.38 not found`. Installer treats the smoke-test failure as install failure and deletes the binary. User has nothing to debug with.

### Steps to reproduce

1. Fresh Ubuntu 22.04 (or any system with glibc 2.35).
2. Run `curl -fsSL https://install.opensre.com | bash`.
3. Observe the GLIBC_2.38 missing error in the smoke test.
4. Observe the binary is then deleted by the installer.

Yes

### How often does it occur?

Every time

### Operating system

Linux

### Logs and error output

```
libpython3.13.so.1.0: GLIBC_2.38 not found (required by libpython3.13.so.1.0)
```

### Additional context

Hit while running through the quickstart fresh on Ubuntu 22.04 as a new user. No system requirements section in the quickstart or README, so a user wastes time getting this far before failing.

Verified affected: v2026.5.4, v2026.4.30, nightly main build. All need glibc 2.38.

By extension this affects any RHEL/CentOS 8 family, Debian 11 (glibc 2.31), Amazon Linux 2 (glibc 2.26), and any non-rolling distro pre-2024.

### Proposed fixes (ranked)

1. **Rebuild the Linux release in `manylinux_2_28` container** (glibc 2.28 floor). One-line CI change. Fixes every existing user on 22.04 / RHEL 8.
2. **Add a glibc precheck to the installer** in the meantime. Read `ldd --version` or `/lib/x86_64-linux-gnu/libc.so.6`, compare to required floor, exit with the supported-OS list before attempting download.
3. **Don't delete the binary on smoke-test failure.** Keep it in place with "smoke test failed with X, binary is at Y, here's how to debug" message. A partial install you can poke at beats no install.
4. **Document the glibc floor** in `quickstart.mdx` until the manylinux build lands.

WatchTree-19:
Closes #1284 (sub-fixes #2, #3, #4 from the issue body).

Sub-fixes #2 and #3 land in `install.sh`:
- new `check_glibc_floor()` runs after platform detection, before download. Linux-only. Reads `ldd --version`, exits cleanly with the supported-OS list if glibc < 2.38. Returns silently if `ldd` is missing so this never breaks a working install.
- `verify_binary_version` preserves the binary at `~/.opensre/failed-install/opensre` on smoke-test failure and points the user at it plus a `ldd` debug hint. The `cleanup` trap logs the preserved path before removing the temp dir.

Sub-fix #4 lands in `docs/quickstart.mdx`:
- new Requirements step before Install OpenSRE, lists supported platforms in a table, links to #1284 for the planned manylinux rebuild.

Sub-fix #1 (rebuild release in `manylinux_2_28`) is intentionally out of scope here. it needs CI/release-pipeline access. Flagged separately in the issue.

Local validation:
- `bash -n install.sh`: clean
- `check_glibc_floor` extracted and run on Ubuntu 22.04 (glibc 2.35): produces the expected supported/affected OS list and exits 1
- the smoke-test preserve path is exercised via the existing `--version` die path, no new call sites

---

The fix is three small additions to install.sh and one new docs section:

1. Two new globals at the top (`SMOKE_TEST_FAILED`, `PRESERVED_BINARY_PATH`) used by the smoke-test preserve and cleanup paths.
2. A new `check_glibc_floor` function called after platform/arch detection. Linux-only, reads `ldd --version`, parses major.minor, compares to required 2.38, prints supported/affected OS list and exits 1 if below. Returns silently when `ldd` isn't available so non-glibc systems aren't broken.
3. The existing `verify_binary_version` `die` path is augmented to copy the binary to `~/.opensre/failed-install/` before dying, and the user-facing error message now includes the preserved path plus a `ldd` debug hint.
4. The `cleanup` trap reads `SMOKE_TEST_FAILED` and prints the preserved path before removing the temp dir.

Considered alternatives: detect glibc by parsing `/lib/x86_64-linux-gnu/libc.so.6` directly (more brittle), or skip the floor check entirely and let pyinstaller fail (worse UX). The `ldd --version` approach matches what the binary already needs and what's available on every relevant distro.

---

cc @VaibhavUpreti @muddlebee

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -34,6 +34,8 @@ INSTALL_DIR_OVERRIDE=0
 INSTALL_CHANNEL="${OPENSRE_INSTALL_CHANNEL:-release}"
 BIN_NAME="opensre"
 INSTALL_WITH_SUDO=0
+SMOKE_TEST_FAILED=0
+PRESERVED_BINARY_PATH=""
 requested_version="${OPENSRE_VERSION:-}"
 
 [ -n "$INSTALL_DIR" ] && INSTALL_DIR_OVERRIDE=1
@@ -481,6 +483,14 @@ verify_binary_version() {
   local actual_version
 
   if ! version_output="$("$binary_path" --version 2>&1)"; then
+    SMOKE_TEST_FAILED=1
+    PRESERVED_BINARY_PATH="${HOME}/.opensre/failed-install/${BIN_NAME}"
+    if mkdir -p "${PRESERVED_BINARY_PATH%/*}" 2>/dev/null && cp "$binary_path" "$PRESERVED_BINARY_PATH" 2>/dev/null; then
+      die "Failed to execute '${binary_path##*/} --version': ${version_output}
+The downloaded binary has been preserved at ${PRESERVED_BINARY_PATH} for debugging.
+On Linux, run 'ldd ${PRESERVED_BINARY_PATH}' to inspect missing libraries.
+See https://github.com/${REPO}/issues/1284 for the most common cause (glibc < 2.38)."
+    fi
     die "Failed to execute '${binary_path##*/} --version': ${version_output}"
   fi
 
@@ -508,6 +518,54 @@ verify_binary_version() {
       printf '%s\n' "$actual_version"
       ;;
   esac
+}
+
+check_glibc_floor() {
+  # Linux-only: verify glibc version is at or above the published floor.
+  # Returns silently if check is not applicable or passes; dies with a clear
+  # message and supported-OS list if the floor is not met.
+  [ "${platform:-}" = "linux" ] || return 0
+  command -v ldd >/dev/null 2>&1 || return 0
+
+  local ldd_output detected_version detected_major detected_minor
+  local required_major=2 required_minor=38
+
+  ldd_output="$(ldd --version 2>&1 | head -n 1 || true)"
+  detected_version="$(printf '%s\n' "$ldd_output" | sed -n 's/.*[^0-9.]\([0-9]\+\.[0-9]\+\).*/\1/p' | head -n 1)"
+  [ -n "$detected_version" ] || return 0
+
+  detected_major="${detected_version%.*}"
+  detected_minor="${detected_version#*.}"
+
+  # Pass if detected >= required
+  if [ "$detected_major" -gt "$required_major" ]; then
+    return 0
+  fi
+  if [ "$detected_major" -eq "$required_major" ] && [ "$detected_minor" -ge "$required_minor" ]; then
+    return 0
+  fi
+
+  cat >&2 <<EOF
+Error: detected glibc ${detected_version}, but the OpenSRE Linux release requires glibc ${required_major}.${required_minor} or newer.
+
+Supported Linux distributions (glibc >= 2.38):
+  - Ubuntu 24.04+
+  - Fedora 39+
+  - Amazon Linux 2023
+  - Arch / rolling distros (recent)
+
+Affected distributions (glibc < 2.38):
+  - Ubuntu 22.04 / 20.04
+  - Debian 11 / 12
+  - RHEL / CentOS 8 family
+  - Amazon Linux 2
+
+Workarounds:
+  1. Upgrade to a supported distribution.
+  2. Install from source. see https://github.com/${REPO}/blob/main/docs/development.md
+  3. Track https://github.com/${REPO}/issues/1284 for the manylinux_2_28 rebuild that lowers this floor.
+EOF
+  exit 1
 }
 
 configure_path() {
@@ -636,6 +694,8 @@ case "$arch" in
     ;;
 esac
 
+check_glibc_floor
+
 resolve_install_dir
 
 version="$requested_version"
@@ -712,6 +772,9 @@ need_cmd mktemp
 tmp_dir="$(mktemp -d)"
 
 cleanup() {
+  if [ "$SMOKE_TEST_FAILED" -eq 1 ] && [ -n "$PRESERVED_BINARY_PATH" ] && [ -e "$PRESERVED_BINARY_PATH" ]; then
+    printf 'Smoke test failed. Binary preserved at: %s\n' "$PRESERVED_BINARY_PATH" >&2
+  fi
   if [ -n "${tmp_dir:-}" ] && [ -d "$tmp_dir" ]; then
     rm -rf "$tmp_dir"
   fi
--- a/install.sh
+++ b/install.sh
@@ -482,8 +482,12 @@ verify_binary_version() {
   local actual_version
 
   if ! version_output="$("$binary_path" --version 2>&1)"; then
-    PRESERVED_BINARY_PATH="${HOME}/.opensre/failed-install/${BIN_NAME}"
-    if mkdir -p "${PRESERVED_BINARY_PATH%/*}" 2>/dev/null && cp "$binary_path" "$PRESERVED_BINARY_PATH" 2>/dev/null; then
+    # Stage the target path locally; only promote to the global
+    # PRESERVED_BINARY_PATH after the copy actually succeeds, so callers don't
+    # see a path pointing to a file that was never written.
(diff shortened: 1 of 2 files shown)


WatchTree-19:
WIP

muddlebee:
@greptileai review

install.sh
```diff
@@ -545,31 +601,6 @@ configure_path() {
   log "Or open a new terminal."
 }
 
-print_success_screen() {
```
muddlebee, line 574: this shouldnt be removed. fix this

yashksaini-coder (changes requested):
Verified the Greptile P1 against `install.sh` and reproduced the failure mode against musl `ldd` output. Real bug, blocking.

## Critical: Alpine Linux false positive

`install.sh` `check_glibc_floor` (around line 65):

```sh
detected_version="$(printf '%s\n' "$ldd_output" | sed -n 's/.*[^0-9.]\([0-9]\+\.[0-9]\+\).*/\1/p' | head -n 1)"
```

The regex isn't anchored to glibc-identifying text. The only non-glibc guard is `[ -n "$detected_version" ] || return 0`, which means the check is skipped only when no version-shaped string is present.

On Alpine Linux, `ldd --version` (where `ldd` is musl's dynamic linker symlink, common on Alpine via `musl-utils`) emits `musl libc (x86_64) Version 1.2.4 Dynamic Program Loader`. The greedy `.*` in the sed scans to the last `x.y` form — `2.4` from `1.2.4`. `detected_version` becomes `2.4`, which is below `2.38`, and the script bails with a misleading "glibc 2.4 detected, requires 2.38+" error.

The outcome (refuse to install) happens to be correct (a glibc-built binary won't run on musl), but the error message is misleading because no glibc is present on the system. More importantly, the `command -v ldd` skip-silently guard at line 59 is meant to handle non-glibc systems — and it doesn't, because Alpine ships a callable `ldd` symlink to musl.

Minimal fix — add a musl reject before the version extract:

```sh
case "$ldd_output" in
  *musl*|*MUSL*) return 0 ;;
esac
```

A more thorough fix anchors on glibc keywords:

```sh
detected_version="$(printf '%s\n' "$ldd_output" | sed -n 's/.*\(GNU libc\|glibc\).*[^0-9]\([0-9]\+\.[0-9]\+\)[^0-9]*.*/\2/Ip' | head -n 1)"
```

Either closes the P1.

## Important

### Preserve-binary path: stale global state on copy failure

`install.sh` (around diff lines 40-46): `PRESERVED_BINARY_PATH` is set as a global before the `mkdir -p && cp` block. When the `cp` fails (read-only `$HOME`, disk full, permission), control falls through to the plain `die` at line 47. The plain `die` is fine, but `PRESERVED_BINARY_PATH` is left non-empty pointing at a file that was never written. The `cleanup` trap doesn't currently consult `PRESERVED_BINARY_PATH` so there's no functional bug, but it's misleading state for any future maintainer who reaches for the variable. Either move `PRESERVED_BINARY_PATH=...` inside the success branch, or unset it on copy failure.

## Verified working

- `check_glibc_floor` placement (after platform/arch detection, before download) is correct.
- 2.38 floor in installer matches the supported-OS table in `docs/quickstart.mdx`.
- `cleanup` trap doesn't print a duplicate path; the diff comment is accurate.
- `~/.opensre/failed-install/` permissions: `mkdir -p` with default umask (022) yields 755, appropriate for a home subfolder.
- Doc table in `docs/quickstart.mdx` mirrors the installer error block.

## Test coverage gap

No bash test fixtures for `check_glibc_floor`. AGENTS.md "If adding new tests -> always place them in `tests/`" isn't violated because the PR doesn't add shell tests, but a security-gate function that exits the installer is a natural target for a small fixture-based test (mock `ldd_output` against a few real distro outputs including Alpine, Ubuntu 22.04, Ubuntu 24.04).

## Maintainability nit

The supported-OS list is duplicated in `install.sh` (error block, ~6 lines) and `docs/quickstart.mdx` (table, ~6 rows). For 6 entries this is acceptable, but a short comment in `install.sh` pointing to the docs URL would reduce drift risk.

## Cross-checks performed

- Reproduced musl output: `musl libc (x86_64) Version 1.2.4 Dynamic Program Loader` -> sed yields `2.4`.
- Confirmed `command -v ldd` skip path doesn't fire on Alpine (musl ships an `ldd` symlink).
- Verified 2.38 floor and supported-OS list are consistent across `install.sh` and `docs/quickstart.mdx`.

## Verdict

Request changes — Greptile's P1 is real and the one-line musl reject (or anchored regex) makes it a clean APPROVE. The other items are either non-blocking polish or out of scope for this PR.

WatchTree-19:
@muddlebee @yashksaini-coder fixes pushed.

- `print_success_screen` restored verbatim (function body + the `print_success_screen "$installed_version"` call site at the end). The `log "Installed..."` lines also went back to `success "Installed..."` to match `upstream/main`.
- `check_glibc_floor`: musl reject added before the version extract, so Alpine and other musl-based distros where ldd is a musl symlink (`musl libc (x86_64) Version 1.2.4`) no longer false-positive as glibc 2.4. A glibc-built binary won't run on musl regardless, but the error message is now correct.
- `verify_binary_version`: the preserve target is staged as a local `_preserve_target` and only promoted to the global `PRESERVED_BINARY_PATH` after `cp` actually succeeds, so callers don't see the global pointing at a file that was never written.

The bash-test fixture for `check_glibc_floor` is fair and I'd land it as a follow-up rather than expanding scope here. Same for the inline doc-URL comment pointing at `docs/quickstart.mdx`.

Ready for re-review.

cerencamkiran:
Looks like #2, #3, and #4 from the issue are covered now. The glibc precheck, preserve-on-failure flow, and quickstart docs all line up with the original report.

One remaining thing before merge though: the latest diff still seems to contain unresolved conflict markers in "install.sh" around "print_success_screen" ("<<<<<<<", "=======", ">>>>>>>").

install.sh
```diff
@@ -571,6 +674,7 @@ configure_path() {
   log "Or open a new terminal."
 }
 
+<<<<<<< HEAD
```
VaibhavUpreti, line 677: could you remove this please? @WatchTree-19 this came from the merge conflict

install.sh
```diff
@@ -604,6 +708,8 @@ print_success_screen() {
   log ""
 }
 
+=======
+>>>>>>> 9739c266 (address greptile feedback: drop duplicate preserved-path message + dead SMOKE_TEST_FAILED)
```
VaibhavUpreti: could you remove this please? @WatchTree-19 this came from the merge conflict

WatchTree-19:
@VaibhavUpreti @cerencamkiran thanks, conflict markers gone in 650fe81 ("Clean up install.sh by removing duplicates and dead code"). The leak was from a local rebase resolution that didn't get cleaned up before the force-push. Verified the file no longer contains `<<<<<<<` / `=======` / `>>>>>>>` and CI is green again.

Sub-fixes #2, #3, #4 from the original issue are all in this PR (glibc precheck, preserve-on-failure, quickstart docs). Sub-fix #1 (manylinux_2_28 release rebuild) is intentionally out of scope here as it needs CI/release-pipeline access.

Ready for re-review.

muddlebee:
closing this due to inactivity. not a priority atm.

Issue #1234, opened 2026-05-03
Contextual Memory for OpenSRE Investigations


muddlebee:
# Contextual Memory for OpenSRE Investigations

## Problem Statement

OpenSRE starts every investigation from scratch. When a recurring alert fires, the system has no memory of past diagnoses, successful tool sequences, or validated root causes. This leads to:

- **Wasted tool calls** on exploratory actions that history already ruled out
- **Slower convergence** — the planner rediscovers the same diagnostic path every time
- **Lower confidence** — the LLM lacks pattern context that would sharpen hypotheses
- **Repeated mistakes** — without outcome tracking, wrong hypotheses can resurface

### Without Memory (Current Behavior)

**Alert:** `RDSConnectionPoolExhausted` fires on `service-auth`.

**Investigation flow:**
1. `plan_actions` sees no memory → plans broad exploratory actions: `query_datadog_logs`, `get_host_metrics`, `query_github_commits`
2. `get_host_metrics` returns normal → wasted call
3. `query_github_commits` finds no DB changes → wasted call
4. `query_datadog_logs` shows connection timeouts (symptom, not cause)
5. Loop #2: `plan_actions` now plans `aws_performance_insights`, `get_aws_rds_events`
6. `aws_performance_insights` reveals a long-running `audit_log` query holding 40 connections
7. **Diagnosis:** Connection pool exhaustion from idle sessions

### With Memory (Proposed)

**Memory loaded before `plan_actions`:**

```markdown
**Prior Successful Investigation Paths (from memory):**
Previous investigation (2025-05-03): Root cause was 'Connection pool exhaustion caused by a long-running audit_log query holding idle sessions.' Actions used: query_datadog_logs, aws_performance_insights, get_aws_rds_events.
Previous investigation (2025-04-18): Root cause was 'Connection pool leak from unclosed connections in a background job.' Actions used: aws_performance_insights, get_aws_rds_events, query_datadog_logs.
Previous investigation (2025-04-05): Root cause was 'Idle connections accumulated after a deployment.' Actions used: query_github_commits, aws_performance_insights.
```

**Investigation flow:**
1. `plan_actions` sees memory → skips host metrics and code search
2. Plans directly: `aws_performance_insights`, `get_aws_rds_events`
3. `aws_performance_insights` confirms the same `audit_log` query pattern
4. **Diagnosis:** "Recurring connection pool exhaustion — same root cause as 2025-05-03."

[...]

PR #1410, opened 2026-05-06
tests(synthetic/rds_postgres): add axis-memory scaffold for #1234

WatchTree-19:
@cerencamkiran fair catch, you were right that the eager loader was still running. Pushed 6cf7d40:

`load_all_scenarios(SUITE_DIR)` is gone. Replaced with a per-id loop that calls `load_scenario(SUITE_DIR / scenario_id)` only for the ids referenced by active pairs. A pair pointing at an unknown scenario id now fails fast with a clear `SystemExit`. No behaviour change for the current 15 scenarios (all ship answer.yml today), but the runner is now resilient to a future fixture-less addition.

Ready for re-review.

### DIFF
--- a/tests/synthetic/rds_postgres/axis_memory/README.md
+++ b/tests/synthetic/rds_postgres/axis_memory/README.md
@@ -1,0 +1,128 @@
+# Axis-memory test scaffold
+
+Synthetic-suite scaffold for evaluating memory anchoring in OpenSRE
+investigations. Pairs with [issue #1234 (Contextual Memory)][issue-1234].
+
+## What this is
+
+The `tests/synthetic/rds_postgres/` suite already covers a single
+investigation per scenario in isolation. This scaffold adds a second axis
+on top of that: pairs of scenarios that share an external alert signature
+(same `alertname`, same `commonLabels`) but resolve to different
+ground-truth outcomes.
+
+The intent is to test what happens when a memory layer (#1234) is primed
+from the first investigation in a pair and the second is run on top of it.
+If memory makes the planner skip exploratory steps that the second
+scenario needs, the planner's pass-rate on the sibling drops. That is the
+anchoring failure mode the team flagged in the #1234 thread.
+
+## Why pairs and not single cases
+
+Memory anchoring is a cross-investigation effect. A single scenario can
+only test whether the planner is rigorous against red herrings within one
+trajectory. Anchoring needs at least two runs that share an alert
+signature so the priming-from-history step is observable.
+
+The pairs are configured in [`pairs.yml`](./pairs.yml). The current set:
+
+| pair id                                  | shared alertname              | base                                      | sibling                                     | what memory might break                              |
+| ---------------------------------------- | ----------------------------- | ----------------------------------------- | ------------------------------------------- | ---------------------------------------------------- |
+| `connection-pressure-real-vs-noisy`      | `RDSDatabaseConnectionsHigh`  | `002-connection-exhaustion`               | `007-connection-pressure-noisy-healthy`     | category divergence: real fault vs healthy           |
+| `cpu-bad-query-vs-checkpoint-storm`      | `RDSCPUUtilizationHigh`       | `004-cpu-saturation-bad-query`            | `014-checkpoint-storm-cpu-saturation`       | sub-flavour divergence: bad-query vs checkpoint storm |
+| `replication-lag-vs-cpu-redherring`      | `RDSReplicationLagHigh`       | `001-replication-lag`                     | `006-replication-lag-cpu-redherring`        | red-herring discipline weakened by prior clean diagnosis |
+
+The strongest pair is `connection-pressure-real-vs-noisy` because the two
+scenarios resolve to different `root_cause_category` values
+(`resource_exhaustion` vs `healthy`). The other two pairs share a category
+and exercise sub-flavour distinctions.
+
+## How it runs today
+
+```
+python -m tests.synthetic.rds_postgres.axis_memory.run_axis_memory
+```
+
+Today this runs each `sibling` scenario without memory primed and emits a
+`memory_mode=not_run` annotation per pair. That's the baseline correctness
+check: every active pair should have its sibling pass on its own. If a
+sibling fails baseline, the failure is not a memory-anchoring effect and
+should be triaged before any memory work.
+
+```
+python -m tests.synthetic.rds_postgres.axis_memory.run_axis_memory --json
+python -m tests.synthetic.rds_postgres.axis_memory.run_axis_memory --pair connection-pressure-real-vs-noisy
+```
+
+`--json` emits the full annotation list for downstream consumption.
+`--pair` runs only the named pair.
+
+## How it runs once Contextual Memory ships
+
+When #1234 lands a memory layer, the runner gains memory mode:
+
+```
+python -m tests.synthetic.rds_postgres.axis_memory.run_axis_memory --memory
+```
+
+In memory mode the runner:
+
+1. runs `base`, captures the investigation memory output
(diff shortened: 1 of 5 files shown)


cerencamkiran:
Hey, @WatchTree-19. Haven’t looked deeply at the code yet, but I like the overall idea. Especially the part about checking if memory helps the agent or pushes it toward the wrong RCA.

yashksaini-coder (changes requested):
Read through the scaffold and the README is unusually good. The mode-classification table and the explicit "What this scaffold does NOT do" section make this much easier to pick up later. That said there are a handful of correctness issues that I'd want fixed before merge, two of which mean the scaffold cannot serve as a CI gate today.

## Critical

### Exit code is unconditionally 0

`tests/synthetic/rds_postgres/axis_memory/run_axis_memory.py` ends with:

```python
sys.exit(0 if run() is not None else 1)
```

`run()` is typed `-> list[PairAnnotation]` and never returns `None`. Empty list is not `None`; populated list is not `None`. The expression is always truthy, so the process always exits 0. If anyone wires this into CI to gate on `memory_hurt`, it will silently pass even when pairs regress. `_print_human` already detects `memory_hurt`, so this is a one-line fix:

```python
results = run()
hurt = sum(1 for a in results if a.memory_mode is MemoryMode.MEMORY_HURT)
sys.exit(1 if hurt else 0)
```

### `test_scoring.py` is not picked up by `make test-synthetic`

`make test-cov` `--ignore=tests/synthetic` and `make test-synthetic` runs `pytest -m synthetic`. The new `test_scoring.py` has no `@pytest.mark.synthetic` mark anywhere in the file, so it is collected but skipped under both default targets. It only runs under `make test-full`, which is not in the AGENTS.md "Before Push" list. Add a module-level `pytestmark = pytest.mark.synthetic` (or per-test marks) so the file is exercised under `make test-synthetic`, otherwise this scoring logic is regression-untested in the standard local loop.

## Warning

### Latent crash if `load_all_scenarios` hits a scenario without `answer.yml`

`run_axis_memory.py` calls `load_all_scenarios(SUITE_DIR)` at startup, which iterates every scenario directory and calls `_parse_answer_yaml(scenario_dir / "answer.yml")`. Several scenarios (007 looks like one) do not currently ship an `answer.yml`, and the loader does not guard against the missing file. End result: `run_axis_memory.py` will blow up on import in that case, even before running a pair. Either filter scenarios down to the ones referenced in `pairs.yml` before loading, or have the loader skip missing answer files with a warning.

### `NOT_RUN` masks real errors

`scoring.py` (around lines 522-532): when memory ran but `baseline_passed` is `None`, the function returns `NOT_RUN` with note `"memory ran but baseline missing; cannot classify"`. `summarize_annotations` then counts that as a not-run. The README's mode table has no entry for this state, and a future rename of `ScenarioScore.passed` would silently inflate `not_run` rather than failing loudly. Either add a dedicated `ERROR` mode or assert that `baseline_score` exposes `.passed` before classifying.

## Info / nits

- `run_axis_memory.py:288` -- `list(raw.get("pairs", []) or [])`: the `or []` is redundant since the default is already `[]`.
- `__init__.py` is empty; a single-line module docstring would be consistent with the rest of the package.
- `_run_with_memory` raises `NotImplementedError` at call-time rather than import-time. That's the right call; just calling it out as a deliberate scaffold choice, not a bug.

## Pluses

- The README is the best part of this PR and is the kind of thing that makes future #1234 work pickupable.
- `pairs.yml` schema is documented inline; the `active` flag is a clean way to commit unvalidated pairs without breaking the runner.
- `scoring.py` is import-light and duck-types against `ScenarioScore`. Sensible choice for a scaffold.
- All new files under `tests/`, not `app/` (matches AGENTS.md rule).
- `PAIRS_FILE = Path(__file__).parent / "pairs.yml"` resolves relative to the file, not CWD. Good.
- `test_scoring.py` actually tests the truth table including duck-typed scores and the `summarize_annotations` tally. Above-average coverage for a scaffold.

Verdict: request changes. Once the exit-code bug, the missing `synthetic` marker, and the `answer.yml` guard are addressed, this is good groundwork for #1234.

WatchTree-19:
@yashksaini-coder thanks, all three blockers addressed:

- Exit-code logic fixed. `__main__` now exits 1 when any pair classifies as `MEMORY_HURT`, 0 otherwise. The previous `0 if run() is not None else 1` was unconditionally truthy.
- `pytestmark = pytest.mark.synthetic` added at module level on `test_scoring.py` so `make test-synthetic` picks it up.
- `load_all_scenarios(SUITE_DIR)` is now filtered down to the scenario IDs referenced by `pairs.yml` before the loader runs against each `answer.yml`. That side-steps the latent crash on scenarios that don't ship one.

Plus the greptile P2 from earlier: the four placeholder parameters of `_run_with_memory` are prefixed with `_` so ruff ARG no longer flags them.

The `NOT_RUN` masking concern is fair; my read is that adding a dedicated `ERROR` mode is its own scope (and would need a corresponding entry in the README mode table). Happy to do that as a follow-up if you want to pull that decision forward.

Ready for re-review.

yashksaini-coder:
@greptile review

cerencamkiran:
Thanks for your effort @WatchTree-19. I think this covers the eval-side promise from the #1234 thread, but not the full contextual memory feature itself.

It gives us the scaffold to measure anchoring risk ("memory_helped" / "memory_hurt" / "memory_neutral") once memory injection lands. So it’s good groundwork for Phase 1 validation, but it doesn’t yet implement the actual memory store, save path, or "memory_context" injection into planning/diagnosis.

Code-wise this looks like a solid scaffold overall. I like that "scoring.py" stays pure/import-light, and "run_axis_memory.py" keeps memory mode explicit with "NotImplementedError" instead of silently pretending memory was exercised.

The main thing I’d still fix before merge is the import ordering in "test_scoring.py" so ruff E402 is cleared. After that, I’d treat the "NOT_RUN" vs future "ERROR" mode question as a follow-up rather than blocking this PR.

WatchTree-19:
@yashksaini-coder @cerencamkiran fix pushed.

`pytestmark = pytest.mark.synthetic` moved below all imports, so ruff E402 stops flagging the local `from tests.synthetic.rds_postgres.axis_memory.scoring import ...` lines. The marker still applies at module level so `make test-synthetic` picks the file up unchanged. Net diff: one line moved.

Ready for re-review.

cerencamkiran:
There’s still a formatting issue on "run_axis_memory.py" ("ruff format --check" is failing). Could you take a quick look and run the formatter once?

cerencamkiran:
@greptileai review

cerencamkiran:
Most of the requested fixes look addressed now. synthetic marker/import order, exit code on MEMORY_HURT, formatting, and placeholder arg naming all look good.

One small thing I’d still double-check: the comment says we only load scenarios referenced by pairs, but the code still calls "load_all_scenarios(SUITE_DIR)" before filtering. If "load_all_scenarios" parses every scenario/answer.yml eagerly, this may still hit the missing-answer.yml issue. Might be worth confirming or filtering before the loader runs.

cerencamkiran:
@greptileai review

WatchTree-19:
@cerencamkiran sorry, missed this one. just checked: run_axis_memory.py calls load_scenario(SUITE_DIR / scenario_id) per pair, not load_all_scenarios. per-id load is the deliberate choice, the eager loader iterates every scenario dir and parses its answer.yml so any future fixture-less scenario would trip before this runner reached its pair. filtering happens before the loader, nothing to change here.

ready for re-review when you or @davincios / @VaibhavUpreti get a window.

muddlebee:
closing this due to inactivity

PR #1480, opened 2026-05-06
fix(synthetic/rds_postgres): harden 002-connection-exhaustion answer key + add QA_VALIDATION (#598)

### LINKED ISSUE #598: [synthetic-qa] 002-connection-exhaustion: Validate agent identifies connection pool leak
## Scenario

**Run:** `opensre tests synthetic --scenario 002-connection-exhaustion`

Leaked client sessions consumed nearly all available `max_connections` (~490/500). CPU elevation at 35-50% is a **secondary symptom** of accumulated idle sessions, not an independent problem.

## Correct Answer

- **Root Cause:** Connection pool leak — idle client sessions exhaust `max_connections`. CPU rise is a symptom, not the cause.

## Human Review Checklist

- [ ] Does the agent identify **connection exhaustion** as the root cause?
- [ ] Does the agent mention `max_connections` being nearly full?
- [ ] Does the agent explain that CPU elevation is a **secondary effect** of idle connections (not an independent CPU problem)?
- [ ] Does the causal chain make sense: pool leak → idle sessions accumulate → connections exhausted → new requests fail?
- [ ] Does the agent avoid diagnosing `cpu_saturation` as the root cause?

The agent correctly identifies that connections are at 98% of max and explains this is from leaked/idle sessions. It explicitly calls out that the 35-50% CPU is caused by the accumulated idle connections making periodic lightweight queries — not a separate issue.

## Common Failure Modes

- Agent sees 35-50% CPU and blames CPU saturation
- Agent says "connection exhaustion" but doesn't explain the idle session leak pattern
- Agent treats CPU and connections as two separate problems instead of one causal chain

WatchTree-19:
Closes #598. Mirrors the same shape @cerencamkiran used for 001 in #1214 (now merged).

`answer.yml` additions in `tests/synthetic/rds_postgres/002-connection-exhaustion/`:

- `required_evidence_sources`: `aws_cloudwatch_metrics` + `aws_performance_insights`. Without this the scenario passes Axis 1 but never enforces the evidence path. Closes the same Axis 1 / Axis 2 gap that 001 closed.
- `ruling_out_keywords: [cpu]`. Mirrors what cerencamkiran did for 001. Forces the agent to demonstrate it considered and rejected CPU rather than just not picking it.
- `clientread` added to `required_keywords`. Confirms the agent surfaces the dominant Performance Insights wait event in its output.
- `model_response` lightly extended so all the new keywords appear in the canonical answer (`Client:ClientRead`, ruling-out language, `[evidence: aws_performance_insights]` tags).

`forbidden_categories: [cpu_saturation]` was already present and is unchanged.

Plus a new `QA_VALIDATION.md` in the same directory, same structure as 001 and 005. Documents the validation paths, ruling-out logic, and the discriminating PI signal (`Client:ClientRead`).

Local validation:

- `answer.yml` parses as valid YAML
- all `required_keywords` (including new `clientread`) appear in the new `model_response` text after lowercase-normalize
- `ruling_out_keywords: [cpu]` appears in `model_response`
- both `required_evidence_sources` cited as `[evidence: <source>]` tags in `model_response`
- `forbidden_categories: [cpu_saturation]` does not collide with the actual category (`resource_exhaustion`) in `model_response`

The full per-source `aws_performance_insights` validation path now exists in `run_suite.py` thanks to #1214 (just merged).

---

Same shape cerencamkiran used for 001 in #1214: tighten the answer key with explicit evidence sources and ruling-out keywords so the scenario enforces both Axis 1 (does the agent name the right cause) and Axis 2 (does the agent rule out the obvious wrong cause). `Client:ClientRead` is the discriminating Performance Insights wait event for connection exhaustion vs CPU saturation — the agent has to surface it to demonstrate it actually queried PI rather than guessing from CloudWatch metrics alone.

`QA_VALIDATION.md` follows the existing convention from 001 and 005 (header, validation paths, invariants).

---

cc @muddlebee @cerencamkiran

### DIFF
--- a/tests/synthetic/rds_postgres/002-connection-exhaustion/QA_VALIDATION.md
+++ b/tests/synthetic/rds_postgres/002-connection-exhaustion/QA_VALIDATION.md
@@ -1,0 +1,55 @@
+# Scenario 002 - Connection Pool Leak
+
+## Overview
+
+This scenario validates that the agent identifies connection exhaustion caused by a client-side pool leak, and treats the concurrent CPU elevation as a secondary symptom rather than an independent failure.
+
+The trap is the 35-50% CPU rise: it is a real signal, but it is a downstream effect of accumulated idle sessions making periodic lightweight queries. The agent must not classify the incident as `cpu_saturation`.
+
+## Correct Diagnosis
+
+| Field | Expected |
+|---|---|
+| Root cause category | `resource_exhaustion` |
+| Root cause | Client-side connection pool leak exhausted `max_connections` |
+| Affected component | Primary RDS instance with leaked client sessions |
+| Primary symptom | New connection attempts rejected as `max_connections` is reached |
+| Secondary symptom | CPUUtilization 35-50% from idle session housekeeping queries |
+| Non-root cause | CPU saturation, slow query, or bad SQL plan |
+
+## Required Reasoning
+
+A correct answer should include the full causal chain:
+
+1. A client-side pool leak holds open sessions instead of returning them.
+2. Idle sessions accumulate; `DatabaseConnections` climbs to ~490 of 500.
+3. Idle sessions issue periodic lightweight queries, lifting CPUUtilization to 35-50% as a side effect.
+4. New application requests fail to connect; the alert fires on `max_connections` exhaustion.
+5. Performance Insights shows `Client:ClientRead` as the dominant wait event, confirming idle-session pressure rather than query workload.
+
+## Evidence Expectations
+
+The agent should cite evidence such as:
+
+- `aws_cloudwatch_metrics`: `DatabaseConnections` near 490/500, CPUUtilization 35-50%
+- `aws_performance_insights`: `Client:ClientRead` as the dominant wait event, no expensive SQL dominating db_load
+
+## Pass Criteria
+
+The answer is considered correct if it:
+
+- Identifies connection exhaustion as the root cause
+- References `max_connections` and the near-cap connection count
+- Explains that the CPU elevation is a secondary effect of idle sessions, not an independent CPU saturation
+- Cites Performance Insights `ClientRead` waits as the discriminating signal
+- Uses both metric-level and query-level evidence
+
+## Common Failure Modes
+
+The answer should be considered incorrect if it:
+
+- Diagnoses `cpu_saturation` based on the 35-50% CPU rise
+- Says "connection exhaustion" without explaining the idle-session leak pattern
+- Treats CPU and connections as two independent problems instead of one causal chain
+- Blames a slow query or missing index without checking Performance Insights wait events
+- Omits Performance Insights from the evidence chain
--- a/tests/synthetic/rds_postgres/002-connection-exhaustion/answer.yml
+++ b/tests/synthetic/rds_postgres/002-connection-exhaustion/answer.yml
@@ -4,6 +4,12 @@ required_keywords:
   - max_connections
   - idle
   - client sessions
+  - clientread
+required_evidence_sources:
+  - aws_cloudwatch_metrics
+  - aws_performance_insights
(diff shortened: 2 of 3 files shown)


cerencamkiran:
@greptileai review

cerencamkiran:
Tested locally on both `main` and this PR branch with:

`python -m tests.synthetic.rds_postgres.run_suite --scenario 002-connection-exhaustion --mock-grafana`

It was already passing on `main`, but this PR makes the validation stricter by adding the required evidence sources, CPU ruling-out, and the `Client:ClientRead` PI signal. So the value here feels more like QA/evaluation hardening rather than changing the raw pass/fail outcome.

Also really like the `QA_VALIDATION.md` addition and the stricter Axis 1 / Axis 2 enforcement pattern from #1214 👍 Requiring the discriminating PI signal makes the validation path much more robust than relying on metrics alone.

Greptile feedback also seems reasonable overall. especially the point about the `Client:ClientRead` PI gate fix closing a real scoring gap. The note about `ruling_out_keywords: [cpu]` being a bit broad also makes sense, though it feels non-blocking for this PR.

yashksaini-coder (approved):
Verified the Greptile P1 against the fixture data and the scorer logic. Real bug, fix is correct.

## Greptile P1 — confirmed and fixed

Before this PR, `performance_insights_tokens` in `run_suite.py` contained `("top sql activity", "avg load", "aas", "db load", "walwrite")`. The fixture `aws_performance_insights.json` for 002-connection-exhaustion shows `Client:ClientRead` as the dominant wait event. An agent producing the correct diagnosis — citing `ClientRead` — would not match any of the original five tokens, so the `aws_performance_insights` evidence gate would fail despite a correct answer. Adding `"clientread"` to the token set closes the gap directly.

## Checklist

- `clientread` added to `required_keywords` (answer.yml line 10) — confirmed.
- `required_evidence_sources` lists both `aws_cloudwatch_metrics` and `aws_performance_insights` — confirmed.
- `ruling_out_keywords: [cpu]` for Axis 2 — confirmed; well-motivated since the scenario has a CPU red herring (35-50% elevation) and the agent must explicitly dismiss it.
- `performance_insights_tokens` in `run_suite.py` includes `"clientread"` — confirmed.
- `QA_VALIDATION.md` documents the correct causal chain (pool leak → connection accumulation → ClientRead waits → secondary CPU → rejection) — confirmed.

## Notes for follow-up (out of scope here)

- The `004-cpu-saturation-bad-query` scenario has a similar latent gap: its fixture shows `CPU:user` as the dominant wait event, but the PI token list lacks `cpu:user` / `cpuuser`. An agent whose response is dominated by "CPU saturation" prose without the AAS / db_load phrasing could trigger a similar false FAIL. Worth a sweep across all PI-evidence-required scenarios in a future PR.
- The keyword alias map at `run_suite.py:148-153` already maps `"idle"` -> `"clientread"`, which gave `clientread` an indirect path. Adding it as an explicit `required_keyword` in `answer.yml` makes the requirement unambiguous and removes the alias dependency for this signal — the right call.
- `QA_VALIDATION.md` is an existing convention; `tests/synthetic/rds_postgres/005-failover/QA_VALIDATION.md` predates this PR. Format matches the prior file.

## Cross-checks performed

- Confirmed fixture `aws_performance_insights.json` shows `Client:ClientRead` as dominant wait.
- Confirmed pre-PR `performance_insights_tokens` lacked `clientread`.
- Confirmed `ruling_out_keywords` is supported by `score_reasoning` (`run_suite.py:248-256`) and the `"cpu"` token will match the model response template.
- Verified `model_response` edit (em-dash removed, `Client:ClientRead` sentence appended) is prose-only, no scoring impact since the content was already in `VALIDATED_CLAIMS`.

Approving.

PR #1548, opened 2026-05-07
feat(shell): add typed routing decisions with confidence

### LINKED ISSUE #1375: Interactive Shell: replace regex-heavy routing with typed intent classification + confidence
## Distinctness
Distinct from #1368 (confidence for RCA conclusions). This issue is about input routing confidence (slash vs cli_help vs cli_agent vs new_alert vs follow_up), not RCA result confidence.

## Why
Routing currently depends on layered regex heuristics across `router.py`, `terminal_intent.py`, and `agent_actions.py`. As phrase diversity grows, misroutes become harder to debug and fix safely.

## Problem
- Classification logic is distributed and overlapping.
- No explicit routing confidence or explanation emitted for observability.
- Compound prompts are brittle under heuristic splitting.

## Scope
- Introduce a typed routing decision object:
  - `route_kind`, `confidence`, `matched_signals`, `fallback_reason`
- Consolidate routing signal sources behind one classifier interface.
- Add telemetry for route decisions and correction outcomes.
- Add a fixture-based routing test suite for regression-proof intent handling.

## Acceptance Criteria
- [ ] All routing returns a structured decision object.
- [ ] Route decisions can be logged/debugged with matched signals.
- [ ] Existing route behavior is preserved or intentionally improved with tests.
- [ ] Misroute regression tests added for historical edge cases.

cerencamkiran:
Fixes #1375

## The changes

This PR replaces the regex-heavy interactive shell routing flow with a structured typed routing system built around `RouteDecision`.

### Key improvements

* Added structured `RouteDecision` objects with:

  * `route_kind`
  * `confidence`
  * `matched_signals`
  * `fallback_reason`
* Introduced a registry-style `RouteRule` flow to centralize routing logic behind a single classifier interface.
* Preserved backwards compatibility by keeping `classify_input()` behaviour unchanged.
* Updated the interactive shell loop to use structured route decisions.
* Added telemetry payload support for route decisions.
* Persisted `last_route_decision` on the REPL session for future correction/misroute tracking foundations.
* Added YAML fixture-based regression routing cases.
* Added regression tests for:

  * structured route decisions
  * fallback behaviour
  * legacy compatibility
  * YAML fixture routing coverage

## Demo for feature changes and bug fixes

Manual routing verification confirms both structured decisions and legacy compatibility:

```text
/help => RouteDecision(route_kind=<RouteKind.SLASH: 'slash'>, confidence=1.0, matched_signals=('slash_prefix',), fallback_reason=None)

help => RouteDecision(route_kind=<RouteKind.SLASH: 'slash'>, confidence=0.98, matched_signals=('bare_command_alias',), fallback_reason=None)

how do I run an investigation? => RouteDecision(route_kind=<RouteKind.CLI_HELP: 'cli_help'>, confidence=0.9, matched_signals=('cli_help_pattern',), fallback_reason=None)

run a sample alert => RouteDecision(route_kind=<RouteKind.CLI_AGENT: 'cli_agent'>, confidence=0.85, matched_signals=('sample_alert_launch',), fallback_reason=None)

api latency spiked and 5xx errors increased => RouteDecision(route_kind=<RouteKind.NEW_ALERT: 'new_alert'>, confidence=0.86, matched_signals=('investigation_request',), fallback_reason=None)

hello => RouteDecision(route_kind=<RouteKind.CLI_AGENT: 'cli_agent'>, confidence=0.45, matched_signals=(), fallback_reason='no_prior_investigation_and_no_incident_signal')
```

## Validation

```bash
python -m pytest tests/cli/interactive_shell/test_router.py tests/cli/interactive_shell/test_loop.py -q
```

* 43 passed

```bash
python -m pytest tests/cli -q
```

* 749 passed
* 22 skipped
* 1 unrelated existing failure:
  `test_execute_cli_actions_handles_path_with_spaces`


cerencamkiran:
@greptileai review

yashksaini-coder (changes requested):
Read the diff carefully against `router.py` and `session.py` on the current tree. Both Greptile findings turn out to be false positives, but a different correctness issue surfaces from a closer look at the `classify_input` backward-compat wrapper.

## Greptile / CodeQL — both FALSE POSITIVE

1. **Duplicate rule name `investigation_request` (Greptile P1, line 151)** — REFUTED. The two `RouteRule` instances have distinct names: `"investigation_request"` (targeting `_is_new_alert_without_prior_state`) and `"investigation_request_with_prior_state"` (targeting `_is_new_alert_with_prior_state`). Telemetry signal `matched_signals` will differentiate them correctly.

2. **Module-level cyclic import in `session.py:11` (CodeQL)** — REFUTED. The PR removes `from app.cli.interactive_shell.session import ReplSession` from `router.py` entirely. `router.py` introduces a `RoutingSession` Protocol defined in-module. `session.py`'s `last_route_decision` is typed `Any | None` to avoid the inverse import. There's no cycle introduced; the PR actively eliminates the prior one-way coupling.

## Critical (beyond Greptile)

### `classify_input` backward-compat wrapper silently changes routing semantics

`app/cli/interactive_shell/router.py` (around diff lines 298-300):

```python
def classify_input(text: str, session: RoutingSession) -> InputKind:
    return route_input(text, session).route_kind.value
```

The **old** `classify_input` evaluated its `cli_agent` operational branch via `plan_cli_actions(stripped)`:

```python
if plan_cli_actions(stripped) and not mentions_alert_signal(stripped):
    return "cli_agent"
```

The **new** `route_input` rule (`cli_agent_operational`) calls `is_cli_agent_operational_intent(stripped)`. These are different functions in different modules (`action_planner.py` vs `terminal_intent.py`). The wrapper's docstring claims "Backward-compatible" but the routing logic it delegates to is structurally not the same.

Any caller depending on `classify_input` to preserve the `plan_cli_actions`-based decision will get a different routing answer for inputs that one function matches but the other doesn't — silent and untested.

`test_route_input_preserves_legacy_classification` only exercises 6 fixture strings, none of which hit the `plan_cli_actions` path, so this divergence is uncovered.

Fix: either (a) route through `plan_cli_actions` in the `cli_agent_operational` rule to preserve the original predicate, or (b) drop the backward-compat wrapper and migrate callers to `route_input` directly. Don't claim backward-compat while swapping the branch predicate.

## Important

### PII risk in `matched_signals` telemetry

`matched_signals: tuple[str, ...]` currently holds only rule names (`"slash_prefix"`, etc.) — no user text leaks today. But the field has no type constraint or doc constraint forbidding signal fragments derived from user input (e.g., a future rule embedding a matched regex group). Given `app/analytics/` ships events to an external sink and user prompts in this CLI frequently contain service / cluster / credential-adjacent fragments, please add an explicit constraint:

- Either rename to `matched_rule_names` and constrain values to a registry, or
- Add a doc comment on `RouteDecision.matched_signals`: "Must contain only rule names; never user-derived content."

### `last_route_decision: Any | None` undermines `make typecheck`

`session.py` types the field as `Any | None` to avoid the import. AGENTS.md mandates strict typing. The standard pattern for this exact case is a `TYPE_CHECKING` guard:

```python
if TYPE_CHECKING:
    from app.cli.interactive_shell.router import RouteDecision

last_route_decision: RouteDecision | None = None
```

`TYPE_CHECKING` is already imported in `session.py`. Zero runtime cost, restores strict typing.

## Pluses

- The shape (`RouteDecision` frozen dataclass, `RouteKind` StrEnum, `RouteRule` frozen dataclass with matcher callable) is the right contract.
- Telemetry approach is sensible — `INTERACTIVE_SHELL_ROUTE_DECISION` event with route kind, confidence, matched signals, fallback reason gives future misroute-correction tracking a foundation.
- `route_input` iterates rules in priority order and returns the first match cleanly; the rules are mutually exclusive by state-conditional design (`last_state is None` vs `is not None`), so priority-tie ambiguity is structurally avoided.
- Removing the prior import-direction coupling between `router.py` and `session.py` is a real architectural improvement.

## Cross-checks performed

- Verified the two `RouteRule` names are distinct.
- Confirmed no `from app.cli.interactive_shell.session import` in `router.py` after the diff.
- Confirmed `RoutingSession` Protocol is defined locally.
- Confirmed `last_route_decision: Any | None` and the absence of `TYPE_CHECKING` import for `RouteDecision`.
- Compared `classify_input` predicate before vs after; confirmed the cli_agent branch swaps from `plan_cli_actions` to `is_cli_agent_operational_intent`.

## Verdict

Request changes — the `classify_input` wrapper claim of backward-compat is misleading because the underlying predicate changed. Pick one path (preserve `plan_cli_actions` semantics, or drop the wrapper) and add a regression test pinning the choice. The PII constraint and `TYPE_CHECKING` typing are both small follow-ups that polish the surface.

cerencamkiran:
@greptileai review again

yashksaini-coder (approved):
Re-reviewed against `fddff9b2` + `ac469e1a`. The critical semantic-divergence concern is cleanly closed and the PII risk now has an explicit doc + test constraint. One small follow-up remains around typing precision but it isn't a runtime concern.

## Findings status

### 1. `classify_input` backward-compat — FIXED

`_is_cli_agent_action_rule` now delegates directly to `plan_cli_actions`:

```python
def _is_cli_agent_action_rule(text, _session):
    stripped = text.strip()
    return bool(plan_cli_actions(stripped)) and not mentions_alert_signal(stripped)
```

Semantically identical to the original `classify_input` predicate. The `cli_agent_action_plan` rule sits at the same relative priority in `ROUTE_RULES` as the original branch, so the dispatch order is preserved. The new `test_route_input_preserves_legacy_classification` round-trips `route_input(...).route_kind.value == classify_input(...)` for the six representative cases including the `"run a sample alert"` and fallback paths — exactly the regression guard I'd want.

### 2. PII risk in `matched_signals` — FIXED

The field now carries the explicit doc constraint:

```python
# Must contain only internal rule names; never user-derived content.
matched_signals: tuple[str, ...] = ()
```

And the new `test_route_input_matched_signals_are_internal_rule_names` pins it: asserts `matched_signals == ("investigation_request",)` and explicitly asserts `"api latency" not in decision.matched_signals`. The contract is now both stated in prose and enforced by a test. Population sites only ever pass `(rule.name,)` where `rule.name` is a hardcoded literal in `ROUTE_RULES`, so the constraint holds end-to-end.

### 3. `last_route_decision` typing — PARTIAL

Still `Any | None`. The pattern that fits this case cleanly:

```python
if TYPE_CHECKING:
    from app.cli.interactive_shell.router import RouteDecision

last_route_decision: RouteDecision | None = None
```

`session.py` already uses `TYPE_CHECKING` for `prompt_toolkit.History` (lines 6-9), so the pattern is established. `RouteDecision` lives in a sibling module with no circular-import risk. Zero runtime cost, restores strict typing per AGENTS.md.

Not blocking — the field is only written for observability, never read back into logic, and `clear()` correctly resets it. Worth a one-line follow-up though.

## Minor observation (below threshold)

The new `test_cli_action_plan_routes_to_cli_agent_before_investigation` uses `"show me connected services"`. If `plan_cli_actions` ever returns empty for that phrase (after a future `intent_parser` change), the test would still pass via the fallback path without exercising the `cli_agent_action_plan` rule. Adding `assert decision.matched_signals == ("cli_agent_action_plan",)` locks the rule identity down. Suggestion only.

## Cross-checks

- Verified `_is_cli_agent_action_rule` delegates to `plan_cli_actions` (semantic match with the old code).
- Verified `ROUTE_RULES` ordering preserves mutual exclusion between `investigation_request` (no prior state) and `short_follow_up_question` (has prior state) — those two have mutually exclusive session guards so tuple ordering is immaterial there.
- Confirmed `matched_signals` is only ever populated from `RouteRule.name` literals.
- Confirmed `clear()` resets `last_route_decision` to `None`.
- Confirmed `TYPE_CHECKING` pattern is already in use in `session.py` (so adopting it for `RouteDecision` is precedent-aligned).

Approving — typing tightening is a one-line follow-up at your convenience.

cerencamkiran:
Thanks a lot for the careful re-review @yashksaini-coder really appreciate the detailed checks. 

I’ve added the `matched_signals == ("cli_agent_action_plan",)` assertion as well, so the `plan_cli_actions` regression test now locks the exact rule identity, not just the final route kind.

I’ll leave the `last_route_decision` typing as a small follow-up since it’s storage-only and not blocking, but agreed that tightening it later would be cleaner.

Devesh36 (approved):
Reviewed the implementation — looks good to me. The ROUTE_RULES registry is a clean improvement over the old procedural chain, and the routing semantics are preserved exactly (verified predicate-by-predicate against the original classify_input). The last_route_decision typing (Any | None in session.py) is the one minor nit — easy TYPE_CHECKING follow-up when you get a chance

PR #1570, opened 2026-05-07
fix(synthetic-tests): skip E2E tests gracefully when LLM credentials are missing

### LINKED ISSUE #1349: E2E synthetic tests should skip gracefully when LLM credentials are missing
## Description
The E2E synthetic tests in `tests/synthetic/eks/test_suite.py` crash with a Pydantic validation error when `OPENAI_API_KEY` (or other LLM provider keys) are not set in the environment.

## Expected Behavior

Tests that require LLM credentials should:

1. Detect missing credentials upfront.
2. Skip with a clear message like: `SKIPPED: Requires LLM credentials (OPENAI_API_KEY or equivalent)`.

## Actual Behavior

```text
pydantic_core._pydantic_core.ValidationError: 1 validation error for LLMSettings
  Value error, LLM provider 'openai' requires OPENAI_API_KEY to be set.
    For further information visit https://errors.pydantic.dev/2.13/v/value_error
During task with name 'extract_alert' and id '...'
app\config.py:238: ValidationError
```

## Steps to Reproduce

```bash
# Unset any LLM API keys
unset OPENAI_API_KEY ANTHROPIC_API_KEY

# Run E2E test
pytest tests/synthetic/eks/test_suite.py::test_level1_scenario -v
```

## Investigation

- Error originates in `app/config.py:238` during LLM configuration validation.
- Test file: `tests/synthetic/eks/test_suite.py::test_level1_scenario[001-oomkilled-crashloop]`.
- The test runs against the `001-oomkilled-crashloop` scenario but crashes before executing the scenario.

## Severity

**Low** – Expected behavior for E2E tests requiring live credentials, but it should fail more gracefully (e.g., skip with a clear message instead of raising a `ValidationError`).

7vignesh:
Fixes #1349

#### Describe the changes

Skip E2E synthetic tests gracefully when LLM credentials are not configured, rather than crashing with a `ValidationError`.

- Add `_llm_credentials_available()` helper that detects missing API keys
- Skip parametrized level tests with a clear skip message instead of failing
- Fix `TestHarnessEndToEnd` to explicitly set `LLM_PROVIDER=anthropic` so it works regardless of `.env` contents

## Before
Tests crash with a raw ValidationError when LLM keys are missing:
<paste the error traceback from the issue>

## After
Tests skip gracefully with a clear message:
<screenshot of the terminal output above>

### What was changed
- Added `_llm_credentials_available()` check that detects missing API keys upfront
- Parametrized level tests now `pytest.skip()` with a descriptive reason
- Fixed `TestHarnessEndToEnd` to explicitly set `LLM_PROVIDER` so it works regardless of `.env` contents

### Screenshot 
image

image

---

### DIFF
--- a/tests/synthetic/eks/test_suite.py
+++ b/tests/synthetic/eks/test_suite.py
@@ -7,7 +7,7 @@ from unittest.mock import patch
 
 import pytest
 
-from app.llm_credentials import has_llm_api_key
+from app.llm_credentials import has_credentials_for_active_llm_provider
 from app.nodes.plan_actions.node import InvestigationPlan
 from tests.synthetic.eks.run_suite import run_scenario, score_result
 from tests.synthetic.eks.scenario_loader import (
@@ -19,48 +19,12 @@ from tests.synthetic.k8s_schemas import VALID_K8S_EVIDENCE_SOURCES
 from tests.synthetic.mock_datadog_backend.backend import FixtureDatadogBackend
 from tests.synthetic.mock_eks_backend.backend import FixtureEKSBackend
 
-# Providers that authenticate via local/IAM mechanisms and don't need an API key env var.
-# Must match the no-key providers in LLMSettings._require_api_key_for_selected_provider().
-_NO_KEY_PROVIDERS = frozenset(
-    {"ollama", "bedrock", "codex", "cursor", "claude-code", "gemini-cli", "opencode", "kimi"}
+# Synthetic E2E uses fixture EKS/Datadog backends; many tool "calls" hit mocks, not real APIs.
+# This gate only reflects whether the *LLM* can authenticate (the reason we skip when keys are absent).
+_SYNTHETIC_SKIP_LLM = (
+    "SKIPPED: missing API key for LLM_PROVIDER "
+    "(suite uses mock EKS/Datadog backends; configure the key for your selected provider)"
 )
-
-# Maps cloud-based providers to the env var they require.
-# Must match the provider_to_key mapping in LLMSettings._require_api_key_for_selected_provider().
-_PROVIDER_KEY_ENV_VAR: dict[str, str] = {
-    "anthropic": "ANTHROPIC_API_KEY",
-    "openai": "OPENAI_API_KEY",
-    "openrouter": "OPENROUTER_API_KEY",
-    "requesty": "REQUESTY_API_KEY",
-    "gemini": "GEMINI_API_KEY",
-    "nvidia": "NVIDIA_API_KEY",
-    "minimax": "MINIMAX_API_KEY",
-}
-
-
-def _llm_credentials_available() -> bool:
-    """Return True when the active LLM provider's credentials are configured.
-
-    Only returns False (triggering a skip) when the provider is recognized and
-    its required API key is genuinely absent.  Invalid provider names or other
-    config errors are *not* caught here so they still surface as hard failures.
-    """
-    import os
-
-    provider = (os.getenv("LLM_PROVIDER") or "anthropic").strip().lower() or "anthropic"
-
-    if provider in _NO_KEY_PROVIDERS:
-        return True
-
-    env_var = _PROVIDER_KEY_ENV_VAR.get(provider)
-    if env_var is None:
-        # Unknown provider — let the test run and fail loudly with a real error.
-        return True
-
-    return has_llm_api_key(env_var)
-
-
-_SKIP_REASON = "SKIPPED: Requires LLM provider credentials"
 
 # ---------------------------------------------------------------------------
 # Loader and fixture validation
@@ -388,8 +352,8 @@ def _should_assert_trajectory(fixture, actual_category: str) -> bool:
 
 def _run_scenario_test(fixture) -> None:
     """Run scenario with real LLM and mock backends, then assert scoring."""
-    if not _llm_credentials_available():
-        pytest.skip(_SKIP_REASON)
+    if not has_credentials_for_active_llm_provider():
+        pytest.skip(_SYNTHETIC_SKIP_LLM)
 
     failures: list[str] = []
     for attempt in range(1, _LLM_ATTEMPTS + 1):
--- a/tests/synthetic/eks/test_suite.py
+++ b/tests/synthetic/eks/test_suite.py
@@ -7,7 +7,7 @@ from unittest.mock import patch
 
 import pytest
 
-from app.llm_credentials import has_credentials_for_active_llm_provider
+from app.config import has_credentials_for_active_llm_provider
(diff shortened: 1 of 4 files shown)


7vignesh:
@greptileai review

7vignesh:
@VaibhavUpreti @muddlebee review pls!

tests/synthetic/eks/test_suite.py
```diff
@@ -18,6 +20,23 @@
 from tests.synthetic.mock_datadog_backend.backend import FixtureDatadogBackend
 from tests.synthetic.mock_eks_backend.backend import FixtureEKSBackend
 
+
+def _llm_credentials_available() -> bool:
```
muddlebee: _llm_credentials_available() treats any ValidationError or ValueError from LLMSettings.from_env() as “no credentials → skip”. That includes failures that are not “missing API key”, for example:

- Invalid or unsupported LLM_PROVIDER (typo, stale value)
- Other validation problems that would be better as hard failures so CI fails loudly

muddlebee:
### Suggestion: narrow “missing credentials” to an explicit env-var check

`_llm_credentials_available()` treats any `ValidationError` or `ValueError` from `LLMSettings.from_env()` as “no credentials → skip.” That fixes the crash when keys are absent, but it also turns **other** config failures (invalid `LLM_PROVIDER`, unrelated validation errors, etc.) into **skips**, which weakens CI signal—a misconfigured job can read as skipped instead of failing loudly.

Prefer mirroring [`LLMSettings._require_api_key_for_selected_provider`](https://github.com/Tracer-Cloud/opensre/blob/main/app/config.py):

1. Normalize `LLM_PROVIDER` the same way `LLMSettings.from_env()` does (`strip().lower()`, default `"anthropic"`).
2. For providers that do **not** require a cloud API key (same allowlist as config: `ollama`, `bedrock`, CLI-style providers, etc.), do not skip on keys.
3. Otherwise map provider → the expected env var and skip **only when** [`has_llm_api_key`](https://github.com/Tracer-Cloud/opensre/blob/main/app/llm_credentials.py) is false—that matches env + keyring resolution used at runtime.
4. Invalid/unsupported provider (or other bad env) should **not** hit this skip path so the suite still errors normally.

Trade-off: a small duplication with `LLMSettings`. Payoff: skips mean “active provider needs a key and none is configured,” not “any bad env.”

7vignesh:
@greptileai review

7vignesh:
> ### Suggestion: narrow “missing credentials” to an explicit env-var check
> `_llm_credentials_available()` treats any `ValidationError` or `ValueError` from `LLMSettings.from_env()` as “no credentials → skip.” That fixes the crash when keys are absent, but it also turns **other** config failures (invalid `LLM_PROVIDER`, unrelated validation errors, etc.) into **skips**, which weakens CI signal—a misconfigured job can read as skipped instead of failing loudly.
> 
> Prefer mirroring [`LLMSettings._require_api_key_for_selected_provider`](https://github.com/Tracer-Cloud/opensre/blob/main/app/config.py):
> 
> 1. Normalize `LLM_PROVIDER` the same way `LLMSettings.from_env()` does (`strip().lower()`, default `"anthropic"`).
> 2. For providers that do **not** require a cloud API key (same allowlist as config: `ollama`, `bedrock`, CLI-style providers, etc.), do not skip on keys.
> 3. Otherwise map provider → the expected env var and skip **only when** [`has_llm_api_key`](https://github.com/Tracer-Cloud/opensre/blob/main/app/llm_credentials.py) is false—that matches env + keyring resolution used at runtime.
> 4. Invalid/unsupported provider (or other bad env) should **not** hit this skip path so the suite still errors normally.
> 
> Trade-off: a small duplication with `LLMSettings`. Payoff: skips mean “active provider needs a key and none is configured,” not “any bad env.”

addressed in commit d2292d7a

tests/synthetic/eks/test_suite.py
```diff
@@ -18,6 +19,47 @@
...
 from tests.synthetic.mock_eks_backend.backend import FixtureEKSBackend
 
+# Providers that authenticate via local/IAM mechanisms and don't need an API key env var.
+_NO_KEY_PROVIDERS = frozenset(
+    {"ollama", "bedrock", "codex", "cursor", "claude-code", "gemini-cli", "opencode", "kimi"}
+)
+
+# Maps cloud-based providers to the env var they require.
+_PROVIDER_KEY_ENV_VAR: dict[str, str] = {
+    "anthropic": "ANTHROPIC_API_KEY",
```
muddlebee: list is not comprehensive?? there are way too many other API calls being made,

7vignesh:
@greptileai review

muddlebee:
@greptile review

muddlebee:
@7vignesh pushed some fixes, looks like earlier approach was better than guarded one.

yashksaini-coder (commented):
Posting this as a comment since the PR is already merged, but noting it for the follow-up. Several review threads remained unresolved at merge time:

## Unresolved findings (would have blocked under the 5/5 bar)

1. **CodeQL cyclic import alert** on `app/llm_credentials.py:62` — `Import of module app.config begins an import cycle.` This is the only finding that touches `app/` code rather than test code, and the alert is still open in code-scanning. Worth a follow-up PR to break the cycle (lazy import inside the function, or move the shared constants out to a leaf module).

2. **`_llm_credentials_available()` catches too broadly** (raised by both Copilot and @muddlebee at `tests/synthetic/eks/test_suite.py:24` and `:35`). `LLMSettings.from_env()` raises `ValidationError` for several conditions (unsupported `LLM_PROVIDER` value, invalid `LLM_MAX_TOKENS`, etc.) that are real misconfiguration, not missing credentials. As written, those failures get masked as a skip. Suggest narrowing to an explicit env-var presence check for the active provider.

3. **Skip message hardcodes `OPENAI_API_KEY`** (Copilot + Greptile P2 at `tests/synthetic/eks/test_suite.py:38`). `LLM_PROVIDER` defaults to `anthropic`, so the message is misleading on a default install. Make it provider-aware or generic.

4. **Incomplete API-call list** (@muddlebee at line 29) — "there are way too many other API calls being made." The credential check covers one path but not all calls the suite makes.

## Suggested follow-up

- Fix the CodeQL cycle as a small standalone PR (the alert won't auto-close).
- Tighten `_llm_credentials_available()` to a per-provider env-var presence check; that fixes the broad-catch concern from Copilot, muddlebee, and Greptile in one shot.
- Provider-aware skip message comes for free once the per-provider check is in.

muddlebee:
hey @yashksaini-coder most review points were catered by me. if you still notice anything pls open a github issue

PR #1571, opened 2026-05-07
fix(cli): CLI Dropdown UI improvement

### LINKED ISSUE #1313: [IMPROVEMENT] UI enhancement for CLI
### Current state

too verbose

### Desired state

random ui bug fixes and enhancement in cli

w3joe:
Fixes #1313 

-removed dropdown background and scrollbar
-standardise colors so that terminal supports both dark and light modes
-dropdown no longer disappears if keywords matched or backspaced

image


w3joe:
@copilot resolve the merge conflicts in this pull request

muddlebee:
@greptileai review

VaibhavUpreti:
@copilot rebase this PR and resolve the merge conflicts

yashksaini-coder (changes requested):
Did a thorough pass; the list/verify split is well-shaped and the new tests around it are precise. Greptile's two P1 findings reference symbols / files that were renamed in this PR, but the underlying concern of one of them is real in the new layout, plus a couple of independent issues.

## Greptile P1 findings — status

1. **`_status_style` missing `"healthy"` (commands.py:219)** — REFUTED. Greptile's symbol/file is stale. The active function in the new layout is `status_style` in `app/cli/interactive_shell/command_registry/rendering.py`, and the diff explicitly adds `"healthy": PRIMARY` and `"passed": PRIMARY`. The verify path renders correctly.

2. **`_load_list_integrations` silent skip (commands.py:170)** — PARTIALLY CONFIRMED in a different file. Greptile's reference is again stale, but the same silent-skip pattern lives in `_list_row_from_record` in `app/cli/interactive_shell/command_registry/repl_data.py`. That function only recognises `"active"`, `"failed"`, `"missing"`, `"misconfigured"` and returns `None` for anything else. See the next critical for why this matters now.

## Critical

**`_list_row_from_record` drops `"healthy"` / `"ok"` store records silently**

`app/cli/interactive_shell/command_registry/repl_data.py`, the `_list_row_from_record` function added in this PR.

`rendering.py` adds `"healthy"` (and `"ok"`) to `_ACTIVE_INTEGRATION_STATUSES` and gives them a `PRIMARY` color in `status_style`, both new in this PR. But `_list_row_from_record` only maps `"active"` to a positive row; the `else: return None` swallows `"healthy"` / `"ok"`. End result: any integration that writes `status: "healthy"` to the store (which the rendering layer is now explicitly prepared for) silently disappears from `/list`. The `healthy` plumbing in `rendering.py` is unreachable today.

Fix: extend the positive branch to accept `{"active", "healthy", "ok"}`, mapping all of them to the same rendered status (or distinct ones if you want the table to surface the difference). This is a one-line change but it's the difference between the new status vocabulary working end-to-end versus being dead code.

## Warnings

**`bg:default noreverse` strips the dropdown surface contrast**

`app/cli/interactive_shell/loop.py` `_build_prompt_style` replaces every `bg:#1c1917` with `bg:default noreverse`. On terminals with image / transparent / 256-color backgrounds (Wezterm with opacity, iTerm2 transparency, Codespaces web terminal), the completion menu now sits directly on the buffer's terminal background with no contrast separation. The scrollbar button also goes invisible since `bg:default noreverse` removes the affordance.

If the bare-terminal aesthetic is intentional, keep it but please leave a one-line comment in `_build_prompt_style` explaining the choice ("intentional: respect terminal background"), so a future reviewer doesn't restore the painted surface and silently regress this UX.

**`_attach_slash_completion_menu_position` walks `prompt.layout.find_all_controls()`**

`find_all_controls` is not part of `prompt_toolkit`'s public API contract and has changed shape across minor versions. If a future upgrade returns zero matches, `menu_position` is silently never set; the existing test asserts `len(controls) == 1` so it would catch the case at test time, but the function itself should at least log a warning when the iterator yields anything other than a single match. Defensive without being noisy.

## Info / nits

- `_list_row_from_record` rewrites `"misconfigured"` -> `"failed"` in the rendered status. A user who sees `"failed"` in `/list` and then runs `/integrations verify` (which only emits `"passed"` / `"failed"` / `"missing"`) gets a confusing status flip. Adding the original status into the `detail` column would close the gap ("misconfigured: missing API key" vs "failed").
- `instance_detail` uses `> 1` for the count suffix; a zero-instance record shows nothing. Semantically right, slightly surprising. A one-line comment would help.
- `_cmd_integrations` passes `show_inactive=True` to `render_integrations_table` but not to `render_mcp_table`. Intentional asymmetry, or oversight?
- Removing `_LOGO_MARK_ROWS` / `_build_logo_mark` from `banner.py` is a clean removal but the PR description does not mention it. Worth a sentence so reviewers don't wonder if it was accidental.

## Pluses

- The list / verify split is well-designed: `load_list_integrations` is fast and stays local, `load_verified_integrations` does live checks. Command routing is clear.
- `test_list_integrations_does_not_run_live_verification` is exactly the right guard against a future regression where the list path accidentally re-introduces network calls.
- `test_slash_completion_menu_stays_anchored_at_input_start` and `test_backspace_reopens_slash_completion_after_valid_command` are tight tests of the new UX.
- Deduplication via `seen_services` correctly prevents env overrides from shadowing store entries.
- Post-merge-conflict cleanup looks coherent; no obvious resolution artifacts.

## Cross-checks performed

- Confirmed `verify_integrations` only emits `"passed"`, `"failed"`, `"missing"` via `_verification_adapters.py`, so the verify-path color map is complete.
- Confirmed the store's raw status field defaults to `"active"` in `store.py`, with `"healthy"` / `"ok"` reserved for integrations that opt into richer state.
- Confirmed `prompt_surface.py` still holds `_build_prompt_style`, `ShellCompleter`, etc.; the new `loop.py` additions are additive (anchoring + backspace).
- Confirmed `OPENCLAW_CORAL` removal from `loop.py` is safe: `prompt_surface.py` still imports and uses it for `completion-menu.border`.

## Verdict

Request changes. The list/verify split is good work and most of the diff is solid; the blocker is the `_list_row_from_record` silent-drop on `"healthy"` / `"ok"`, which means the new status vocabulary doesn't actually flow through. Cannot mark Confidence 5/5 with that gap open.

yashksaini-coder:
Merged `origin/main` into the branch and resolved the conflicts. Pushed as commit `c6523c3b`. Resolution notes:

- **banner.py** -> took main. The earlier 'remove logo' commit on this branch is superseded by main's #1585 oh-my-logo splash work, and the unused `version` parameter on `_build_identity_block` had no caller.
- **loop.py** -> took main. Main's refactor moved prompt setup into a new `prompt_surface.py` module; the PR's `loop.py` was based on the older shape.
- **prompt_surface.py** -> forward-ported your slash-UX features into the refactored module so the dropdown anchoring + backspace behavior survives:
    - `_is_slash_completion_text`, `_slash_completion_menu_position`, `_attach_slash_completion_menu_position`
    - `_reopen_slash_completion_if_needed`, `_delete_before_cursor_and_reopen_slash_completions`
    - Backspace key binding wired into `_build_prompt_key_bindings`
    - Exact-match highlighting in `ShellCompleter` (`_EXACT_SLASH_COMMAND_STYLE`)
    - `_attach_slash_completion_menu_position(prompt)` called from `_build_prompt_session`
- **test_loop.py** -> took main, then forward-ported your three UX tests (`test_slash_completion_menu_stays_anchored_at_input_start`, `test_shell_completer_keeps_exact_match_visible_and_highlighted`, `test_backspace_reopens_slash_completion_after_valid_command`) with imports updated to point at `prompt_surface`.

One styling decision worth flagging: I kept main's color palette (`bg:#1c1917` / `bg:#2c1e14`) instead of the `bg:default noreverse` you had. That choice was flagged in the earlier review as a contrast regression on terminals with image or transparent backgrounds, and it's not load-bearing for the anchoring + backspace UX. If you want the default-bg back, drop it in a follow-up commit and we can discuss the trade-off there.

Verified locally on the resolved branch: `ruff check` clean, `ruff format --check` clean, `pytest tests/cli/interactive_shell/` -> 480 passed including all three new UX tests. The earlier review's open findings (`_list_row_from_record` silently dropping `"healthy"` / `"ok"` records) still need to be addressed before merge.

muddlebee:
hey @w3joe @VaibhavUpreti I have added a alternative menu selection system similar to what codex/claude code have [here](https://github.com/Tracer-Cloud/opensre/pull/1654)

WatchTree-19 (commented):
Drive-by from someone running similar bench-side work. The dropdown UX touch is welcome.

**Terminal width handling.** Inline pickers and dropdowns in the CLI need to fall back gracefully when the terminal is narrower than the longest option label (common on tmux split panes, SSH sessions, embedded terminals in IDEs at small widths). Two reasonable behaviours: truncate-with-ellipsis at width and reveal full text on focus, or auto-switch to a vertical-stack layout below a width threshold. Worth confirming which one this PR picks (or whether it picks neither and overflows). The existing `repl_choice_menu` in muddlebee's recent #1654 had similar shape.

**NO_COLOR / TERM=dumb.** A surprising fraction of CI logs and screen-readers run with `NO_COLOR=1` or `TERM=dumb`. Dropdown UI that relies on cursor positioning, alt-screen, or ANSI escapes should detect those and degrade to a plain-text numbered list. If the dropdown uses prompt-toolkit, the library handles most of this for you, but the wrapper code needs to read the env var at construction time and skip the fancy path.

**Keyboard contract.** Worth documenting (in code or a brief doc) what keys move/select/escape: arrows + enter + esc is standard, j/k for vim users is a nice-to-have, ctrl+c should always abort cleanly without leaving the cursor in alt-screen. Tab-completion semantics inside a dropdown are the trickiest piece; tab usually advances to the next field in TUI but cycles through options in some dropdown libs. Pick one and stick with it.

**Test surface.** Pure UI like this is annoyingly hard to unit-test, but golden-output tests (capture stdout from a scripted input sequence and diff against a fixture) are the right shape. If `tests/cli/interactive_shell/` doesn't already have a harness for this, this PR is a reasonable place to add one.

**Already covered upstream.** Existing maintainer thread looks like the right priority focus, no need to pile on.

Looks like a quality-of-life improvement that matters for adoption. The width and NO_COLOR cases are the two places where polish matters disproportionately for first-impressions.

davincios:
Closing as superseded by #1654 (`feat(cli): TTY inline pickers for /list, /integrations, /mcp, and more`), which has since been merged and replaces the dropdown component this PR was improving. Branch is also currently in a CONFLICTING state against `main`. Thanks for the contribution!

PR #1420, opened 2026-05-06
feat(synthetic/rds_postgres): add memory_mode field + helper for #1234

WatchTree-19:
Pairs with #1234 (Contextual Memory). Adds an optional `memory_mode` field to the existing `ScenarioScore` dataclass in `tests/synthetic/rds_postgres/run_suite.py` plus a small `annotate_with_memory_mode` helper, so the axis-memory runner landed in #1410 can attach memory-anchoring classifications to per-scenario scores and the published report can split metrics by memory mode.

This is piece 3 of the three pieces I sketched in https://github.com/Tracer-Cloud/opensre/issues/1234#issuecomment-4383654309 following @davincios's invite to take the eval-side instrumentation.

#### Describe the changes you have made in this PR -

Three small changes in `tests/synthetic/rds_postgres/run_suite.py`:

1. `ScenarioScore` gains `memory_mode: str | None = None` (optional, default None). Backwards-compatible.
2. New `annotate_with_memory_mode(score, memory_mode) -> ScenarioScore` helper using `dataclasses.replace` since the dataclass is frozen. Lets the axis-memory runner attach a classification without re-implementing scoring.
3. The text-output loop in `run_suite(...)` includes `memory_mode=<value>` in the per-scenario detail when the field is set. JSON output gets it for free via the existing `asdict(...)` path.

Plus a new `tests/synthetic/rds_postgres/test_run_suite_memory_mode.py` with 6 unit tests covering the field default, the helper, every documented mode (`memory_helped`, `memory_hurt`, `memory_neutral`, `pre_existing`, `not_run`), and the `asdict` serialisation path.

### Demo/Screenshot for feature changes and bug fixes - 

Local validation:

- `ruff check` on both files: All checks passed!
- `ruff format --check` on both files: already formatted
- `ast.parse` on `run_suite.py`: clean
- structural verification: ScenarioScore now has 12 fields including `memory_mode`, `annotate_with_memory_mode` exists at lines 73-81, the print loop has the `memory_mode` branch
- direct exercise of every assertion in `test_run_suite_memory_mode.py` against the helper API: all pass

The full pytest run requires the opensre dependency tree (langgraph, opentelemetry); tests will run in CI on developer machines that have it.

## Independence from #1410

[...]

PR #1580, opened 2026-05-07
feat(synthetic/rds_postgres): add TP/FP/TN/FN classification + accura…

WatchTree-19:
Closes the first of the two README-flagged gaps from the May-1 outreach (Accuracy / FP / FN aren't measured in the benchmark). @VaibhavUpreti green-lit this in DM today.

Three additions in tests/synthetic/rds_postgres/run_suite.py:

- new optional ScenarioScore.outcome_class field. One of "TP" / "FP" / "TN" / "FN" or None for unclassified.
- new classify_outcome() classifier. Healthy ground truth maps agent-says-healthy to TN, else FP. Real-fault ground truth maps category-match to TP, miss or wrong category to FN.
- new compute_classification_stats() aggregator returning the four counts plus accuracy / precision / recall / F1, with safe zero-denominator handling.

score_result() calls classify_outcome and attaches the result. run_suite() prints a Classification line in text mode; JSON mode gets the field for free via asdict().

Three of the existing 15 fixtures are healthy ground truth (000-healthy, 007-noisy-healthy, 013-storage-recovery-false-alert), so the classifier has real coverage on both sides of the True / False split.

Test coverage in tests/synthetic/rds_postgres/test_run_suite_classification.py:

- 10 parametrised truth-table cases on classify_outcome
- aggregator: empty input, count-each-class, full arithmetic on accuracy / precision / recall / F1, all-correct upper bound, all-failing lower bound, unknown outcome_class skipping, zero-division safety
- ScenarioScore: outcome_class defaults to None, can be set explicitly

Independently mergeable from #1410, #1420 (same file, no conflicting fields) and from PR B (docs-only, doesn't depend on this).

cc @VaibhavUpreti @muddlebee @cerencamkiran

---

- [x] Yes
- [ ] No

- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/tests/synthetic/rds_postgres/run_suite.py
+++ b/tests/synthetic/rds_postgres/run_suite.py
@@ -4,7 +4,7 @@ import argparse
 import json
 import re
 from dataclasses import asdict, dataclass
-from typing import Any
+from typing import Any, Literal
 
 from app.pipeline.runners import run_investigation
 from tests.synthetic.mock_grafana_backend.backend import FixtureGrafanaBackend
@@ -66,14 +66,14 @@ class ScenarioScore:
     # One of: "TP" (real fault correctly identified), "FP" (healthy scenario
     # mis-classified as a fault), "TN" (healthy scenario correctly identified),
     # "FN" (real fault missed or mis-categorised). Set by score_result().
-    outcome_class: str | None = None
+    outcome_class: Literal["TP", "FP", "TN", "FN"] | None = None
 
 
 def classify_outcome(
     actual_category: str,
     expected_category: str,
     root_cause_present: bool,
-) -> str:
+) -> Literal["TP", "FP", "TN", "FN"]:
     """Classify a scored scenario into TP/FP/TN/FN.
 
     A scenario is HEALTHY when expected_category == "healthy" — the alert is
--- a/tests/synthetic/rds_postgres/run_suite.py
+++ b/tests/synthetic/rds_postgres/run_suite.py
@@ -62,6 +62,75 @@ class ScenarioScore:
     failure_reason: str = ""
     trajectory: TrajectoryScore | None = None
     reasoning: ReasoningScore | None = None
+    # Classification of the scenario outcome against the expected ground truth.
+    # One of: "TP" (real fault correctly identified), "FP" (healthy scenario
+    # mis-classified as a fault), "TN" (healthy scenario correctly identified),
+    # "FN" (real fault missed or mis-categorised). Set by score_result().
+    outcome_class: str | None = None
+
+
+def classify_outcome(
+    actual_category: str,
+    expected_category: str,
+    root_cause_present: bool,
+) -> str:
+    """Classify a scored scenario into TP/FP/TN/FN.
+
+    A scenario is HEALTHY when expected_category == "healthy" — the alert is
+    expected to be noise and the agent should report no fault. Otherwise the
+    scenario has a real fault that the agent should identify by category.
+
+    Returns one of: "TP", "FP", "TN", "FN".
+    """
+    is_healthy_ground_truth = expected_category == "healthy"
+    actual_says_healthy = (not root_cause_present) or actual_category == "healthy"
+
+    if is_healthy_ground_truth:
+        return "TN" if actual_says_healthy else "FP"
+
+    # Real-fault scenario from here on.
+    if not root_cause_present:
+        return "FN"
+    if actual_category == expected_category:
+        return "TP"
+    return "FN"
+
+
+def compute_classification_stats(results: list[ScenarioScore]) -> dict[str, Any]:
+    """Aggregate TP/FP/TN/FN counts and derived metrics across scenarios.
+
+    Returns a dict with the four raw counts plus total, accuracy, precision,
+    recall, and F1. Divisions return 0.0 when the denominator is zero rather
+    than raising, so the function is safe on empty or all-skipped result sets.
+    """
+    counts: dict[str, int] = {"TP": 0, "FP": 0, "TN": 0, "FN": 0}
+    for result in results:
+        cls = result.outcome_class
+        if cls in counts:
+            counts[cls] += 1
+
+    tp = counts["TP"]
+    fp = counts["FP"]
+    tn = counts["TN"]
+    fn = counts["FN"]
+    total = tp + fp + tn + fn
+
+    accuracy = (tp + tn) / total if total else 0.0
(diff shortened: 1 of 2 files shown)


yashksaini-coder (changes requested):
Walked the truth table by hand and the classification logic is right (TP / FP / TN / FN match the standard taxonomy, `actual_says_healthy` collapses correctly when `root_cause_present=False`, edge cases in `compute_classification_stats` short-circuit cleanly to 0.0). That part is solid. The blockers are about typing strictness (Greptile P2) and one critical packaging issue Greptile didn't surface.

## Greptile P2 findings — status

1. **`outcome_class` typed as plain `str` (run_suite.py:69)** — CONFIRMED. With AGENTS.md mandating strict typing, narrowing this to `Literal["TP", "FP", "TN", "FN"] | None` is a one-line fix that lets mypy catch any caller that constructs a `ScenarioScore` with a bad string. Worth doing.

2. **`classify_outcome` return type generic `str` (run_suite.py:76)** — CONFIRMED. Function has four exhaustive returns, all literal strings. Same fix:

   ```python
   from typing import Literal
   OutcomeClass = Literal["TP", "FP", "TN", "FN"]
   def classify_outcome(...) -> OutcomeClass: ...
   ```

   Then `outcome_class: OutcomeClass | None = None` on the dataclass. Apply (1) and (2) together.

3. **`_MockScore` bypasses `ScenarioScore` type contract (test_run_suite_classification.py:80)** — Real but minor. As a test-file-local helper this is acceptable Python duck typing. The principled fix is a one-attribute Protocol that `compute_classification_stats` types against, but that's a separate refactor.

4. **Truth-table gap (test_run_suite_classification.py:56)** — Partially confirmed. The specific cell (healthy GT + `root_cause_present=False` + non-healthy `actual_category`, e.g. `"resource_exhaustion"`) isn't directly tested. The exhaustive-range test at the bottom covers it implicitly via `classify_outcome` being walked over all combos, so the practical risk is low. Still, an explicit row would tighten the table.

## Critical (beyond Greptile)

**The new test file is orphaned from every standard CI target**

`tests/synthetic/rds_postgres/test_run_suite_classification.py`.

- `make test-cov` runs `pytest --ignore=tests/synthetic -m "not synthetic"` (Makefile target), so the whole `tests/synthetic/` tree is skipped. The new file is in there.
- `make test-synthetic` runs `pytest -m synthetic`, but the new file has no `pytestmark = pytest.mark.synthetic` (and no per-test markers), so it's collected but skipped under that target too.
- Net effect: the new tests only run under explicit `pytest tests/synthetic/rds_postgres/test_run_suite_classification.py`. Nothing in the standard developer or CI loop exercises them.

This is the same kind of discoverability gap that #1410 had. These are pure unit tests (no LLM, no live infra), so the right fix is `pytestmark = pytest.mark.synthetic` at the module level, matching the `@pytest.mark.synthetic` pattern used in the neighbor `test_suite.py`. After that, `make test-synthetic` will pick them up.

## Warnings

**`compute_classification_stats` total may be less than `len(results)`**

The function silently skips results whose `outcome_class` is `None` from the `total` count. Documented in the docstring but not surfaced in the output dict. A `"classified": N, "total_results": M` distinction in the returned summary would be clearer for anyone trying to reconcile the numbers against `_print_summary`.

## Info / nits

- The `test_compute_stats_skips_unknown_outcome_class` test verifies that `None` and garbage strings are filtered. Good. Worth adding one row that mixes a valid TP with a `None` and asserts `total == 1` to make the skip behavior more obvious in test output.
- `_MockScore` could become a typed `Protocol` once the typing strictness lands; tracker comment in the test file would be worth a line.

## Pluses

- Classification taxonomy is correct. I walked all eight truth-table cells manually against `actual_says_healthy = (not root_cause_present) or actual_category == "healthy"`. Each cell maps to the right TP/FP/TN/FN, and the docstring's worked F1 example (`4/7`) checks out.
- Edge cases in `compute_classification_stats` (zero-denominator for precision / recall / F1) all guard cleanly. F1 short-circuits to 0.0 when precision + recall == 0, no NaN.
- `frozen=True` compatibility maintained — `outcome_class` is appended after defaulted fields, so existing construction sites don't break.
- `asdict` JSON serialization just works (`str | None` serializes trivially).

## Cross-checks performed

- Confirmed `ScenarioScore` is `@dataclass(frozen=True)` so `replace()` and field placement rules apply.
- Walked truth table all 8 cells against `classify_outcome`.
- Read `pytest.ini` and the relevant `make` targets to confirm the discoverability gap.
- Verified the existing `test_suite.py` neighbor uses `@pytest.mark.synthetic`, so adding `pytestmark` matches the pattern.
- Hand-checked the docstring's F1 = 4/7 worked example.

## Verdict

Request changes. The classification math is right; the blockers are tightening the typing to `Literal` (Greptile #1 + #2 in one shot) and adding `pytestmark = pytest.mark.synthetic` so the new tests actually run. With those two fixes I'd happily revisit for 5/5.

WatchTree-19:
@yashksaini-coder thanks for the careful walk-through. Both blockers addressed in the new commit:

- `ScenarioScore.outcome_class` and `classify_outcome()` return type narrowed to `Literal["TP", "FP", "TN", "FN"] | None`. `from typing import Any, Literal` covers it; no extra dependency.
- `pytestmark = pytest.mark.synthetic` added at module level on `test_run_suite_classification.py` so `make test-synthetic` picks it up. Same pattern as `test_suite.py`.

I left the `_MockScore` duck-type stand-in for now since it's test-file-local and the principled `Protocol` refactor is its own thing; happy to follow up if you want it in this PR.

The truth-table cell for `(non-healthy actual_category, healthy GT, root_cause_present=False) -> "TN"` is implicitly covered by `test_classify_outcome_returns_one_of_four_classes` walking every combo, but I can add an explicit row if you'd like.

Ready for re-review.

yashksaini-coder:
@greptile review

muddlebee:
@WatchTree-19 thank you, but closing the PR. needs more relevant context for this, feel free to open a github issue and get community approval before opening PR.

WatchTree-19:
understood @muddlebee, will follow the issue-first / community-approval workflow going forward. for context (fwiw, not a reopen ask): @VaibhavUpreti had aligned on this scope in a DM thread, which is what i was referencing in the PR description, but i take the point that DM context isn't visible to other reviewers and the merge bar rightly hinges on what the PR thread itself can stand on. happy to refile this as a github issue once there's signal beyond the single-maintainer DM. thanks for the steer.

PR #1587, opened 2026-05-07
docs(benchmarks): document Axis 1 vs Axis 2 gap + remove now-reported…

WatchTree-19:
Closes the second of the two README-flagged gaps from the May-1 outreach (Axis 1 vs Axis 2 gap is computed but not published anywhere). @VaibhavUpreti green-lit this in DM today.

Two edits in docs/benchmarks/README.md:

- replace the stale "Not reported: accuracy / FP / FN aren't measured" line with the now-reported entries (accuracy / precision / recall / F1, plus TP / FP / TN / FN counts), with a forward link to the new gap section. Lands cleanly with PR #1580 (PR A) which adds those counters in run_suite.
- new "Axis 1 vs Axis 2 Gap" section at the bottom. Explains what the two axes are (full mock backend vs SelectiveGrafanaBackend), the gap formula, the three commands to produce the report, a sample output block, and an AXIS-GAP-START / AXIS-GAP-END marker block for future automation.

The marker pair mirrors the existing BENCHMARK-START / BENCHMARK-END pattern in the top-level README.md, so a follow-up workflow can drop the latest gap output between them without new scaffolding.

Docs-only PR. No code changes. Independently mergeable from PR #1580 (which adds the classification metrics referenced in the updated bullet). The forward link works whether #1580 is merged first or this one, and the merge order is flexible.

cc @VaibhavUpreti @muddlebee @cerencamkiran

---

- [x] Yes
- [ ] No

- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/docs/benchmarks/README.md
+++ b/docs/benchmarks/README.md
@@ -9,11 +9,8 @@ Reported metrics:
 - duration
 - token usage
 - estimated LLM cost
-
-Not reported:
-- accuracy
-- false positives
-- false negatives
+- accuracy / precision / recall / F1 (from the synthetic RDS suite, see [Axis 1 vs Axis 2 Gap](#axis-1-vs-axis-2-gap) below)
+- TP / FP / TN / FN counts across the suite
 
 ## Running benchmarks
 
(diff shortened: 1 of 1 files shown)


yashksaini-coder (changes requested):
The Axis 1 vs Axis 2 explanation itself is solid and the conceptual content (full data vs selective backend, gap as percentage points, per-difficulty breakdown) matches what `FixtureGrafanaBackend` and `SelectiveGrafanaBackend` actually do. The `BENCHMARK-START/END` marker reference is also accurate. Two issues I'd want resolved before merge.

## Blocker: documented invocation does not work today

The new section instructs:

```
python -m tests.synthetic.rds_postgres.run_suite --axis2             # gap report
```

I verified directly: `_print_gap_report` is defined at `tests/synthetic/rds_postgres/run_suite.py:455` but `run_suite()` (lines 494-522) never references `args.axis2` and never calls `_print_gap_report`. The `--axis2` flag is parsed by `parse_args` and then silently ignored. Anyone following the docs will get a normal per-scenario PASS/FAIL printout with no gap numbers and no error.

Either wire the flag through (load both axis1 and axis2 results, then call `_print_gap_report`) or move that command into a `# work-in-progress` block until the wiring lands. As-is the docs lie.

The PR title says "remove now-reported [items]" and the diff drops the `Not reported` block on the basis that accuracy / FP / FN are now reported. That premise is half-true at best while the gap CLI silently does nothing.

## Imprecise reported-metrics list

Line 14 adds `accuracy / precision / recall / F1` as reported metrics. The synthetic RDS suite emits a binary pass/fail per scenario plus an aggregate pass-rate. Precision / recall / F1 are classification metrics that need TP/FP/TN/FN counts, which `_print_gap_report` does not emit. Suggest narrowing to `scenario pass-rate (Axis 1 and Axis 2), TP / FP / TN / FN counts` or just dropping the F1 / precision / recall names until the code computes them.

## Smaller things

- The sample report shows `+Zpp` for the gap, but `_print_gap_report` formats with `{gap:+.0f}pp`, so a negative gap would render as `-5pp`. Consider showing `+/-Zpp` or just `Zpp` to avoid implying the gap is always positive.
- The `AXIS-GAP-START` / `AXIS-GAP-END` block says "They are not auto-committed today." That's fine, but combined with the `--axis2` wiring being broken it reads as if the only gap is automation, when actually nothing produces the report at all yet.

## Pluses

- The conceptual write-up is the kind of thing that makes the suite usable for new contributors. Once the wiring catches up, this is a great section to keep.
- The `BENCHMARK-START/END` pattern reference is correct and the marker convention is the right call for future automation.

Verdict: request changes.

WatchTree-19:
@yashksaini-coder fair catch on the `--axis2` invocation, that one's on me. Two changes pushed:

- The "How to run" block no longer claims `python -m tests.synthetic.rds_postgres.run_suite --axis2` produces the gap report. The two-axis suites still run independently as before; the doc now explicitly notes that wiring the `--axis2` CLI flag through to `_print_gap_report` is tracked separately.
- The reported-metrics list at the top is narrowed to per-scenario pass/fail and aggregate pass-rate (which the suite emits today). `precision / recall / F1` come back when the gap-report code emits them.
- Sample report shows `+/-Zpp` rather than `+Zpp` to match the actual `{gap:+.0f}pp` format.
- `AXIS-GAP-START / AXIS-GAP-END` block text is clearer about what is and is not produced today.

I considered wiring the `--axis2` flag through in this PR but kept it docs-only since the CLI hookup touches `run_suite.py` (which is also in flight on #1580 and #1420). Happy to do a small follow-up PR for the wiring when those two land.

Ready for re-review.

muddlebee:
same as last comment

WatchTree-19:
understood @muddlebee, will follow the issue-first / community-approval workflow going forward. for context (fwiw, not a reopen ask): @VaibhavUpreti had aligned on this scope in a DM thread, which is what i was referencing in the PR description, but i take the point that DM context isn't visible to other reviewers and the merge bar rightly hinges on what the PR thread itself can stand on. happy to refile this as a github issue once there's signal beyond the single-maintainer DM. thanks for the steer.

PR #1590, opened 2026-05-07
fix(cli): Slack local store credentials are ignored by verification and CLI publishing 

### LINKED ISSUE #1588: [BUG] Slack local store credentials are ignored by verification and CLI publishing
## Bug

Slack credentials configured through `opensre integrations setup slack` are stored in the local integrations store, but parts of the Slack flow only look at environment variables or older credential shapes.

## Symptoms

- `opensre integrations verify slack` can report `SLACK_WEBHOOK_URL` as missing even when Slack was configured successfully in the local store.
- Running a standalone CLI investigation may complete without publishing to Slack because `send_slack_report()` only checks `SLACK_WEBHOOK_URL` when there is no Slack thread context.

## Expected behavior

Slack verification and standalone investigation publishing should both use the effective Slack configuration, including v2 local store records where credentials live under `instances[].credentials`.

## Notes

Local fix under test:

- Teach integration credential resolution to read v2 local store instance credentials.
- Teach standalone Slack delivery to fall back from `SLACK_WEBHOOK_URL` to the resolved local Slack integration config.
- Add regression coverage for Slack verification and no-thread Slack delivery.

## Verification

Focused regression tests pass locally:

```bash
uv run pytest tests/integrations/test_verify.py tests/utils/test_slack_delivery.py -q
```

w3joe:
Fixes #1588

- **`_raw_credentials` in `app/integrations/_catalog_impl.py`** — Reads Slack (and other) credentials from v2 store records: prefers top-level `credentials`, otherwise walks `instances[].credentials`. Previously only `config.get("credentials", config)` was used, so v2-only shapes effectively had empty credentials and Slack from the local store was ignored for verification and resolution.
- **`send_slack_report` in `app/utils/slack_delivery.py`** — When there is no Slack thread context (`thread_ts`), webhook delivery now uses **`SLACK_WEBHOOK_URL` first**, then falls back to the **webhook URL from `resolve_effective_integrations()`** (local store / catalog), not env alone.
- **Tests** — `test_verify_slack_uses_v2_store_credentials` ensures `verify_integrations("slack")` passes with a v2 store record and no env webhook; Slack delivery tests cover store webhook when env is unset and keep the “no webhook” case deterministic via patching `_configured_webhook_url`.

### DIFF
--- a/app/integrations/_catalog_impl.py
+++ b/app/integrations/_catalog_impl.py
@@ -1600,8 +1600,20 @@ def _service_metadata(
 
 
 def _raw_credentials(config: dict[str, Any]) -> dict[str, Any]:
-    raw_credentials = config.get("credentials", config)
-    return raw_credentials if isinstance(raw_credentials, dict) else {}
+    credentials = config.get("credentials")
+    if isinstance(credentials, dict):
+        return credentials
+
+    instances = config.get("instances")
+    if isinstance(instances, list):
+        for instance in instances:
+            if not isinstance(instance, dict):
+                continue
+            instance_credentials = instance.get("credentials")
+            if isinstance(instance_credentials, dict):
+                return instance_credentials
+
+    return config
 
 
 def resolve_effective_integrations(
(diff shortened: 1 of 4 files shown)


w3joe:
@greptile review

VaibhavUpreti (approved):

Issue #1476, opened 2026-05-06
Sentry coverage 2/5 — Global capture wrappers (BaseTool, @tool, LangGraph nodes, background tasks)


davincios:
Part 2 of 5 in the Sentry coverage rollout. Highest-impact PR for converting silent failures into visible ones — three single edits cover ~80% of the silent surface.

## Background

Today, only ~12 places in `app/` explicitly call `capture_exception(...)`. The rest of the codebase has 150+ broad `except Exception` / `# noqa: BLE001` / `with suppress(Exception)` blocks that swallow exceptions into log warnings or fallback values. Three boundaries are especially impactful:

1. **Tool execution.** [`BaseTool.__call__`](https://github.com/Tracer-Cloud/opensre/blob/main/app/tools/base.py) (line 113) is a thin pass-through to `run()`. The tools cursor rule explicitly says \"Return dict from run() with an 'error' key on failure\" — so tool exceptions get caught inside `run()` and turn into data, never Sentry events.
2. **LangGraph nodes.** Only the graph-level `compiled_graph.invoke(...)` / `astream_events(...)` boundaries in [`app/pipeline/runners.py`](https://github.com/Tracer-Cloud/opensre/blob/main/app/pipeline/runners.py) are wrapped. Node-internal silent catches are invisible.
3. **Background loops.** [`app/remote/vercel_poller.py`](https://github.com/Tracer-Cloud/opensre/blob/main/app/remote/vercel_poller.py) and [`app/remote/system_metrics.py`](https://github.com/Tracer-Cloud/opensre/blob/main/app/remote/system_metrics.py) and the interactive shell loop swallow loop-iteration errors.

### (a) Tool execution wrapper

Edit [`app/tools/base.py`](https://github.com/Tracer-Cloud/opensre/blob/main/app/tools/base.py) `__call__` and the inner runtime in [`app/tools/tool_decorator.py`](https://github.com/Tracer-Cloud/opensre/blob/main/app/tools/tool_decorator.py):

- [ ] Wrap the call to `self.run(**kwargs)` (BaseTool) and the underlying function call (decorator) in try/except.
- [ ] On exception:
    - `sentry_sdk.set_tag(\"tool\", self.name)` (or the decorator's tool name)
    - `capture_exception(exc)`
    - Return `{\"error\": f\"{type(exc).__name__}: {exc}\", \"exception_type\": type(exc).__name__}` to preserve the existing contract.
[...]

PR #1615, opened 2026-05-08
fix: add Sentry exception capture for tool execution, LangGraph nodes, and background tasks (#1476)

shoaib050326:
## Summary

Implement Issue #1476 — add Sentry exception capture across three categories using proper Sentry tags.

### Changes

- **Tool execution** (`BaseTool.__call__`, `RegisteredTool._wrap_run_with_sentry`): Wrap `run()` in try/except, set `sentry_sdk.set_tag("tool", name)`, capture exception, return `{"error": ..., "exception_type": ...}`.
- **LangGraph nodes** (`graph.py._traced_node`): Wrap all 14 `add_node()` calls with a decorator that sets `sentry_sdk.set_tag("node", name)` and re-raises after capture.
- **Background tasks**: New `report_silent` context manager in `sentry_sdk.py` — sets `silent_at` tag, does NOT re-raise. Applied to `vercel_poller.py` (3 sites) and `system_metrics.py` (3 collectors).
- **Fixes**: `vercel_poller.py` `NameError` bug, `action_executor.py` missing Sentry capture in `run_cd_command`.

### DIFF
--- a/app/cli/interactive_shell/action_executor.py
+++ b/app/cli/interactive_shell/action_executor.py
@@ -233,6 +233,7 @@ def run_cd_command(command: str, session: ReplSession, console: Console) -> None
     try:
         os.chdir(target)
     except Exception as exc:
+        report_exception(exc, context="interactive_shell.cd")
         console.print(f"[red]cd failed:[/red] {escape(str(exc))}")
         session.record("shell", command, ok=False)
         return
--- a/app/pipeline/graph.py
+++ b/app/pipeline/graph.py
@@ -1,5 +1,6 @@
 """Unified agent pipeline — wires nodes and edges into a LangGraph."""
 
+import functools
 from collections.abc import Callable
 from typing import Any, cast
 
@@ -38,6 +39,25 @@ from app.types.config import NodeConfig
 NodeWithConfig = Callable[[AgentState, NodeConfig | None], dict[str, Any]]
 
 
+def _traced_node(name: str, fn: Callable[..., Any]) -> Callable[..., Any]:
+    """Wrap a LangGraph node so unhandled exceptions are captured by Sentry."""
+
+    @functools.wraps(fn)
+    def _wrapped(*args: Any, **kwargs: Any) -> Any:
+        try:
+            return fn(*args, **kwargs)
+        except Exception as exc:
+            import sentry_sdk
+
+            from app.utils.sentry_sdk import capture_exception
+
+            sentry_sdk.set_tag("node", name)
+            capture_exception(exc)
+            raise
+
+    return _wrapped
+
+
 def _accept_langgraph_config(func: NodeWithConfig) -> Callable[..., dict[str, Any]]:
     """Adapt a NodeConfig-typed node for LangGraph runtime injection."""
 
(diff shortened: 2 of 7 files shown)


yashksaini-coder (changes requested):
Walked the diff carefully and verified Greptile's four P1 findings against the current `__call__` / `_traced_node` / `_wrap_run_with_sentry` / `report_silent` bodies. All four are confirmed in the latest patch, plus there's a fifth issue the bot didn't surface that I think is more important than any of them.

## Greptile P1 findings — status

1. **`silent_at` global-scope tag (`app/utils/sentry_sdk.py`)** — CONFIRMED. `report_silent` does `sentry_sdk.set_tag("silent_at", where)` and any extra `**tags` directly on the global scope, then calls `capture_exception(exc)` with no `context` / `extra` kwargs. The fast-return path inside `capture_exception` does not open `push_scope` in that branch, so the tags persist on the global scope and bleed into every subsequent Sentry event for the process lifetime. Wrap it in `with sentry_sdk.push_scope() as scope: scope.set_tag(...)` (or use `isolation_scope` on newer sentry-python) before `capture_exception`.

2. **`asyncio.sleep` skipped on exception (`app/remote/vercel_poller.py`)** — CONFIRMED. The `await asyncio.sleep(self.settings.interval_seconds)` now lives inside the `with report_silent(where="remote.vercel_poller"):` block. `report_silent` swallows any `Exception`, exits the `with` body, and the `while True:` re-enters immediately. End result: any error in `state_store.mark_processed`, `collect_candidates`, or `handle_candidate` triggers a tight retry loop that pounds the state store. Move the sleep outside the `with report_silent` block, the way the original code had it outside the try/except.

3. **`tool` global-scope tag (`app/tools/base.py`)** — CONFIRMED. `sentry_sdk.set_tag("tool", self.name)` runs before `capture_exception(exc)` (no scope-opening kwargs), so the tag sticks globally. Same fix as (1).

4. **`node` global-scope tag (`app/pipeline/graph.py` `_traced_node`)** — CONFIRMED. After the first failing node, the `node` tag on every subsequent Sentry event will reflect whatever node last failed, regardless of where the next event actually came from.

## Critical (beyond Greptile)

**`BaseTool.__call__` and `RegisteredTool._wrap_run_with_sentry` swallow exceptions silently and return error dicts**

This one I think matters more than the scope leaks:

```python
def __call__(self, **kwargs: Any) -> dict[str, Any]:
    try:
        return self.run(**kwargs)
    except Exception as exc:
        ...
        capture_exception(exc)
        return {"error": ..., "exception_type": ...}
```

The original `__call__` was a one-liner delegating to `self.run`, so exceptions propagated to the caller. The new version converts hard failures into success-shaped return values. Callers that don't inspect for an `"error"` key (which is most of them, especially the LLM-driven dispatch path) will treat a tool crash as a successful tool result. `_wrap_run_with_sentry` in `app/tools/registered_tool.py` does exactly the same thing.

Notably, `_traced_node` does NOT swallow — it `raise`s after `capture_exception`. That's the right pattern. The two tool wrappers should match it.

If the silent-conversion is intentional (e.g., to keep the planner running past tool crashes), that needs to be in the PR description and flagged in `BaseTool` + `RegisteredTool` docstrings. Right now it looks accidental.

## Warning

**Inconsistent capture path between `BaseTool.__call__` and `RegisteredTool` registration**

`RegisteredTool.from_base_tool` wraps `tool.run`, not `tool.__call__`. So:

- For tools dispatched via `RegisteredTool`, an exception in `run()` is caught by `_wrap_run_with_sentry`, captured, swallowed. `BaseTool.__call__` never runs.
- For tools called directly via `__call__`, only the `BaseTool.__call__` capture path fires.

Result: the same exception type can be reported from two different code paths with two different sets of tags, depending on dispatch site. Pick one layer and keep capture there.

**`_traced_node` is a sync wrapper but some nodes may be async**

If any node function is `async def`, the sync `_wrapped` returns a coroutine object and the `try/except` never sees the coroutine's exceptions. Worth checking `app/pipeline/graph.py` and `app/nodes/*` for any `async def` entry points and adding an `async def _wrapped` branch in `_traced_node` if so.

## Info / nits

- `import sentry_sdk` inside the `except` block in `_traced_node` and `_wrap_run_with_sentry` is fine for lazy import but inconsistent with the rest of the codebase. Minor.
- `report_silent(*, where, **tags)` is variadic but the only callers use just `where=`. The `**tags` is dead weight today.

## Pluses

- The `contextmanager` shape for `report_silent` is the right ergonomic call.
- Wrapping at `RegisteredTool` registration time is the right layer for consistent coverage.
- Removing the explicit `asyncio.CancelledError` re-raise is correct — `CancelledError` is a `BaseException` in Python 3.8+, so `except Exception` doesn't catch it and cancellation still propagates. Greptile didn't flag this and they're right not to.

## Cross-checks performed

- Read `capture_exception` in full to confirm the early-return path that skips `push_scope` when `context`/`extra` are absent.
- Pulled the original `vercel_poller.py` to confirm the original sleep was unconditionally outside the try/except.
- Verified `_traced_node` re-raises but `BaseTool.__call__` and `_wrap_run_with_sentry` do not.
- Confirmed `register_silent`'s `except Exception` does not catch `asyncio.CancelledError` in Python 3.8+, so cancellation propagation is fine.

## Verdict

Request changes. Wrap the four scope-mutation sites in `push_scope` (or pass tags through `capture_exception`'s `context` arg, which already opens a scope), move the `asyncio.sleep` back outside the `with report_silent` block, and either re-raise from `BaseTool.__call__` + `_wrap_run_with_sentry` or document the silent-conversion semantics explicitly.

Cannot mark Confidence 5/5 with these findings open.

muddlebee:
issue reserved for first time contributors and needs to be in sync with https://github.com/Tracer-Cloud/opensre/pull/1620

PR #1617, opened 2026-05-08
feat: implement incident.io integration with verification and RCA support

### LINKED ISSUE #317: [FEATURE] Add incident.io integration for incident context and updates

SB2318:
Fixes #317

## Describe the changes you have made in this PR

This PR introduces a first-class `incident_io` integration for OpenSRE, enabling teams using incident.io as their incident management system to integrate incident context directly into investigation workflows. This implementation has been refined for production readiness, resolving critical API contract mismatches and reliability concerns identified in previous review rounds.

### Changes included:
- **Native Onboarding**: Added support for incident.io via CLI setup wizard, `.env`, or persistent configuration.
- **Connectivity Verification**: Implemented a dedicated health check and CLI verification:
  ```bash
  opensre integrations verify incident_io
  ```
- **Context Consumption**:
  - List active incidents (filtered by status).
  - Fetch full incident metadata (severity, assignees, summary).
- **Bi-directional RCA (Fixed)**: Implemented a functional write-back path using the `actions/edit` endpoint to append investigation findings to incident summaries.
- **Strict Validation**: Added explicit region validation (`us` or `eu`) to prevent silent configuration failures.
- **Documentation**: Comprehensive setup and usage guide added in `docs/incident_io.mdx`.

### Architectural changes:
- **Refined Service Client**: `IncidentIoClient` (`app/services/incident_io/client.py`) now uses `httpx` with retry transports for 429/5xx errors and implements robust secret redaction in logs.
- **Functional Write-back Path**: Replaced the non-existent timeline POST endpoint with a summary-append strategy to ensure findings are actually persisted in incident.io.
- **Tool Layer**: `IncidentIoIncidentsTool` exposes list/get/update operations with standardized return envelopes and consistent parameter naming (`description` vs `comment`).
- **Runtime Wiring**: Fully integrated into:
  - `app/integrations/_catalog_impl.py` (Classification & Env Loading)
  - `app/nodes/plan_actions/detect_sources.py` (URL Context Detection)
  - `app/integrations/cli.py` (Interactive Setup)

### Testing:
- **Contract Tests**: Added `tests/services/test_incident_io_client.py` with `httpx` contract assertions to verify the summary-append payload and error handling.
- **Unit Tests**: Coverage for configuration normalization, probe/verification logic, and tool routing.
- **E2E Integration**: Added a gated, real-API test in `tests/e2e/incident_io/test_incident_io_e2e.py` (skipped when credentials are absent).

---

This implementation provides a dedicated service client (`IncidentIoClient`) to isolate API interaction from the tool layer. The primary challenge was aligning with the incident.io V2 API where custom timeline events are UI-only; we solved this by utilizing the `actions/edit` summary field to provide a reliable way for the agent to persist RCA findings.

---

```


### **Summary of Refinements Included:**
*   **M3 Fix**: Switched from the non-existent `/v2/incident_timeline_events` to a functional summary-append logic via `/v2/incidents/{id}/actions/edit`.
*   **M5 Fix**: Implemented `ValueError` for invalid regions to prevent silent configuration errors.
*   **Quality Gate**: Removed all unused imports and fixed lambda argument linting (ARG005) across the implementation.
*   **Contract Testing**: Created a dedicated test file to ensure the API payloads match the official Incident.io specification.

### CODE REVIEW

**Comment by SB2318** [comment]
@greptile review

**Review by yashksaini-coder — CHANGES_REQUESTED** [review]
Big patch and the test coverage is strong (46 tests, all passing locally on the worktree; lint, format, typecheck all clean). The integration ID flows cleanly through `_catalog_impl.py` -> `_verification_adapters.py` -> `registry.py` -> `verify.py` -> `detect_sources.py` -> `build_prompt.py` -> tool `source` field, and the region-to-URL pattern in `IncidentIoIntegrationConfig` is cleaner than the OpsGenie equivalent. That said, three security-sensitive issues need to land before this is mergeable, plus a few smaller fixes.

## Critical (must fix)

### 1. SSRF risk on user-supplied `base_url`

`app/integrations/config_models.py:216` wires the `base_url` field through `normalize_url("")`, which only strips whitespace and trailing slashes. ArgoCD does this properly with `validate_https_or_loopback_http_url(...)` (line 361). An attacker-controlled store record can set `base_url` to anything, including `http://169.254.169.254/...`, and the client will happily proxy the API key there.

Fix: either run `validate_https_or_loopback_http_url` on `base_url`, or constrain `base_url` to one of the two values in `DEFAULT_INCIDENT_IO_BASE_URLS` (which is what the region resolver already implies anyway).

### 2. `probe_access` leaks raw exception into `ProbeResult.failed()`

`app/services/incident_io/client.py:74-75`:

```python
except Exception as e:
    return ProbeResult.failed(f"Connection failed: {e}", region=self.config.region)
```

Every other `except Exception as e` in the same file (lines 147, 182, 228) calls `self._redact(e.response.text[:200])` before surfacing the message. This one path doesn't, so a network error whose text contains the URL with embedded auth, or a `WWW-Authenticate` header echoed back from the server, can land in user-visible probe output. Apply `self._redact(...)` here too.

### 3. `httpx.HTTPTransport(retries=_MAX_RETRIES)` retries non-idempotent POST

`app/services/incident_io/client.py:45` and `:65` set the underlying urllib3 retry count globally for the transport. urllib3's default `Retry` allows POST. Combined with `add_timeline_event` (which is `POST /v2/incidents/{id}/actions/edit`), a network-level retry will append the same `**OpenSRE Finding: ...**` block twice on a flaky connection.

Fix: build the transport with an explicit `urllib3.util.Retry(allowed_methods=frozenset({"GET", "HEAD"}))`, or drop `retries=` on the transport that the POST path uses and put retry logic at a higher layer where idempotency is known.

## Warning

### `app/cli/wizard/flow.py:78` calls `sync_env_values({})`

The wizard upserts the integration into the store but writes nothing to `.env`. The env-based path documented in `docs/incident_io.mdx` (Option 2) tells the user to set `INCIDENT_IO_API_KEY` in `.env`, so the two setup paths are out of sync. Honeycomb / Sentry / GitLab / Discord all pass actual key/value pairs here. Should be:

```python
sync_env_values({"INCIDENT_IO_API_KEY": api_key, "INCIDENT_IO_REGION": region})
```

### Tool `input_schema.required` lists `api_key` as required

`app/tools/IncidentIoIncidentsTool/__init__.py` declares `api_key` as a required tool argument. Datadog, Honeycomb, etc. inject credentials via `extract_params` from the resolved sources rather than asking the LLM to pass them. Listing a secret in `input_schema.required` invites the model to echo it back in its visible reasoning trace. Recommend pulling `api_key` out of the schema and binding it in `extract_params` from `sources["incident_io"]` instead.

### Checklist gaps from AGENTS.md section 6

- No screenshot or demo GIF showing the integration working.
- No `app/integrations/incident_io.py` or `app/integrations/incident_io/` package; the integration is split across `_catalog_impl.py` and `config_models.py`. This matches OpsGenie / Alertmanager so it's a pre-existing pattern, but the checklist is explicit. Worth confirming the maintainers are OK with the inline-in-catalog approach.

## Info / nits

- `app/integrations/_verification_adapters.py` -> `client_validators.py:167`: `detail=f"Incident.io validation failed: {err}"` passes the raw exception. The `_redact` helper isn't reachable here, but a defensive `str(err).replace(api_key, "[REDACTED]")` would close a corner where a misconstructed client raises with the key still in scope.
- `app/services/incident_io/client.py` `probe_access` opens its own `httpx.Client` rather than reusing `self.list_incidents()` (the way Datadog's `probe_access` reuses `self.list_monitors()`). Minor inconsistency, not a bug.
- `app/services/incident_io/client.py` -- the bare `except Exception as e` in `list_incidents` logs `str(e)` directly without redacting. Apply `self._redact(e)` consistently.
- `docs/incident_io.mdx` lists `https://api.incident.io` as the default `INCIDENT_IO_BASE_URL`. With region-based resolution that's only the US default; either clarify the table or drop the default value to avoid confusion.

## Pluses

- 46 tests, all passing locally on the worktree. Coverage spans config model, probe success/failure, all three tool actions, missing-arg validation, and `detect_sources` URL parsing.
- The `_redact()` helper with `_GENERIC_SECRET_VALUE_RE` is a good addition and is applied across most error paths (except the three flagged above).
- `add_timeline_event` correctly fetches the current summary first and appends, instead of overwriting; the contract test verifies the payload shape end-to-end.
- `build_probe_verifier` wiring matches the OpsGenie / Vercel pattern exactly.
- Region-to-URL resolution validator raises on unknown region instead of silently falling back. Good improvement over OpsGenie.

## Local checks I ran on the PR worktree

- `ruff check app/ tests/` -> clean
- `ruff format --check app/ tests/` -> clean
- `mypy app/` -> 507 source files, no issues
- `pytest tests/integrations/test_incident_io.py tests/services/test_incident_io_client.py tests/tools/test_incident_io_tool.py tests/cli/wizard/test_integration_health.py` -> 46 passed in 2.06s

Verdict: request changes. The code quality is good and the tests are comprehensive; the blockers are all small, security-sensitive fixes.

SB2318:
@greptile review

### **Summary of Changes**

#### **1. Incident.io Security & Reliability**
*   **SSRF Mitigation**: Implemented strict URL validation for the `base_url` field in `IncidentIoIntegrationConfig`. It now enforces HTTPS for all remote URLs while permitting plaintext HTTP only for loopback addresses, using the centralized `validate_https_or_loopback_http_url` helper.
*   **Idempotency Protection**: Refined the custom retry logic in `IncidentIoClient` to only apply to idempotent HTTP methods (`GET`, `HEAD`, `OPTIONS`). This prevents potential duplicate finding appends during transient network failures on `POST` requests.
*   **Leak Prevention**: 
    *   Added `self._redact()` to the `probe_access` error path and all other exception logging in the client.
    *   Implemented manual secret redaction in the integration validator to protect the API key in surfaced error messages.
    *   Removed `api_key` from the tool's `input_schema` to prevent the LLM from leaking secrets in reasoning traces.
*   **Wizard Fix**: Updated the interactive setup to correctly sync `INCIDENT_IO_API_KEY` and `INCIDENT_IO_REGION` to the `.env` file.

#### **2. Pipeline Stability (Sentry Fix)**
*   **Import Refactoring**: Refactored `app/pipeline/runners.py` to use a module-based Sentry import (`import app.utils.sentry_sdk as sentry_sdk`). This eliminates issues where patching `capture_exception` failed due to direct reference capture from `from ... import ...` statements.
*   **Test Robustness**: Updated `tests/pipeline/test_runners_sentry.py` to patch the utilities through the stable module reference, ensuring reliable exception capture verification.

### **Final Verification**
- **Incident.io Suite**: 15 tests passed, including new coverage for idempotent-only retries and validated URL normalization.
- **Sentry Suite**: `test_run_chat_initializes_sentry_and_captures_unhandled_errors` now passes consistently.
- **Total**: **16 tests passed** across all touched components.
- **Quality**: `ruff check` and `ruff format` were executed to ensure adherence to project standards.

The integration is now ready for production use with verified security and diagnostic integrity.

SB2318:
### **Incident.io Integration Hardening**

*   **Security**: Closed SSRF risks via strict `base_url` validation and applied comprehensive redaction to all error paths (client and wizard) to prevent API key leakage.
*   **Reliability**: Restricted retries to idempotent methods (`GET`, `HEAD`, `OPTIONS`) to prevent duplicate timeline appends on unstable connections.
*   **Usability**: Synced setup paths between the CLI wizard and `.env` configuration; cleaned up the tool schema to prevent secret exposure in reasoning traces.
*   **Stability**: Refactored `runners.py` Sentry imports to ensure robust exception capture in the pipeline.

**Tests**: 16 passed (Service, Tool, Integration, and Pipeline).

@yashksaini-coder covered.

yashksaini-coder (approved):
Re-reviewed against the 8 follow-up commits. Every code-level concern from my prior CHANGES_REQUESTED is closed; the remaining gaps are documentation/asset-level. Approving with two notes for follow-up.

## Findings status

### 1. SSRF on `base_url` — FIXED

`_normalize_base_url` now calls `validate_https_or_loopback_http_url(raw, service_name="Incident.io")` directly, matching the ArgoCD pattern. The validator chain is: `field_validator(mode="before")` runs first (validates https/loopback for any non-empty URL); `model_validator(mode="after")` then substitutes the canonical URL from `DEFAULT_INCIDENT_IO_BASE_URLS` only when `base_url == ""`. So an attacker-controlled URL gets validated before the canonical-URL fallback ever fires.

### 2. `probe_access` exception leak — FIXED

The `except Exception` block now calls `self._redact(str(e))` before building `ProbeResult.failed(...)`. Consistent with all other catch sites in the client.

### 3. POST retry — FIXED

The custom `_request()` method gates application-layer retries on `method.upper() not in ("GET", "HEAD", "OPTIONS")`. POSTs that hit 429/5xx or `RequestError` raise immediately. The `HTTPTransport(retries=3)` at the transport layer still retries `ConnectError`/`ConnectTimeout` regardless of method, but for `ConnectError` the server never received the request — that retry is safe and standard. The explicit comment in the code acknowledges this; matches industry practice.

### 4. `sync_env_values` — FIXED

`_configure_incident_io()` now calls `sync_env_values({"INCIDENT_IO_API_KEY": api_key, "INCIDENT_IO_REGION": region})` with actual values.

### 5. `api_key` in `input_schema.required` — FIXED

`"required": []` in the tool schema. `api_key` is listed in `requires = ["api_key"]` (the framework's semantic annotation, not the JSON schema), and `extract_params` binds it from `sources["incident_io"]["api_key"]`. Correct shape — the LLM no longer sees the key as a tool argument.

### 6. AGENTS.md Section 6 checklist gaps — PARTIAL (one of two remains)

- **`app/integrations/incident_io/` package not created** — moot. OpsGenie and Vercel both ship with the integration split across `_catalog_impl.py` + `config_models.py` + `app/services/<name>/`. That precedent is established, so I'm not blocking on this.
- **Screenshot or demo GIF** — still missing. AGENTS.md Section 6 explicitly requires "Screenshot or demo GIF showing the integration working." Worth adding to the PR description before merge so reviewers can see the integration end-to-end. Not gating, but the contributor checklist asks for it.

## Two observations from the deeper pass (not blockers)

### Rate-limit / `Retry-After` not surfaced

When `_request()` exhausts retries on 429/5xx, callers (`list_incidents`, `get_incident`, `add_timeline_event`) return `{"success": False, "error": "HTTP 429: ..."}` with no `Retry-After` header propagated. Callers can't distinguish rate-limit from other errors and can't back off intelligently. Worth documenting as a known limitation or following up with `Retry-After` honoring (similar to what #1628 did for the LLM client).

### Class-level `_lock` creates a global bottleneck

`_lock = threading.Lock()` is a class variable on `IncidentIoClient`. All instances across the process share one lock. In multi-tenant scenarios (different API keys, different incidents), all `_add_timeline_event_via_summary` calls serialize against each other. Switching to `self._lock = threading.Lock()` in `__init__` keeps the read-modify-write cycle protected per-client without the cross-instance contention. Performance nit, not correctness.

## What you did well

- The `_request()` retry/redaction logic is now uniform across all paths.
- The `report_and_reraise` cleanup in `runners.py` was the right incidental fix — removing dead code that was leftover from #1620's earlier refactor.
- The `validate_https_or_loopback_http_url` adoption matches the established pattern in this codebase rather than inventing a new one.
- `extract_params` binding of `api_key` from sources is exactly the shape Datadog/Honeycomb use — clean fit.

## Cross-checks performed

- Verified `validate_https_or_loopback_http_url("")` returns empty (correct pass-through to the model_validator fallback).
- Confirmed Pydantic v2 execution order: `field_validator(mode="before")` runs before `model_validator(mode="after")`.
- Verified `httpx.HTTPTransport(retries=N)` only retries `ConnectError`/`ConnectTimeout` (not application errors), per httpx docs.
- Checked `app/integrations/` for an `incident_io/` subdirectory — none, but precedent (OpsGenie, Vercel) confirms this is acceptable.
- Verified the `sync_env_values({})` pattern in OpsGenie/Datadog/Jira is pre-existing and unrelated to this PR.
- Confirmed `extract_params` reads `sources["incident_io"]["api_key"]`, not from `input_schema`.

## Verdict

Approving. The screenshot is the only outstanding checklist item and it's worth adding to the PR description before merge. The two observations above are good follow-ups but not blockers. Solid response to the prior round.

SB2318:
Thanks @yashksaini-coder for the detailed review.
Regarding the demo video, I was in a basic package under **incident.io**.

I have scheduled a call on May 13 for any temporary subscription. If it's really needed, or you can test with your API KEYS.

Please let me know. :)

app/pipeline/runners.py
rrajan94, line 1: When you replaced `report_and_reraise(...)` with a plain `try/except`, a couple of things quietly went away. The module-level logger that used to log errors, and the Sentry tags `(surface=pipeline, component=app.pipeline.runners)` that help with triaging failures in production. This doesn't seem related to the incident.io work at all. Was this intentional?

app/services/incident_io/client.py
rrajan94, line 1: `_lock = threading.Lock()` is defined at the class level, which means every `IncidentIoClient` instance shares the same lock. So if two concurrent tool calls each spin up their own client and both end up in `_add_timeline_event_via_summary`, they'll block each other even if they're working on completely different incidents. Moving it into `__init__` as `self._lock = threading.Lock()` would fix this.

rrajan94 (changes requested):
Added comment on `runners.py` and `client.py`, please check once.

app/pipeline/runners.py
SB2318: Yes, due to CI, some tests are falling.

SB2318:
> Added comment on `runners.py` and `client.py`, please check once.

okay, I will try to cover those.

app/nodes/plan_actions/detect_sources.py
```diff
@@ -906,6 +906,35 @@ def detect_sources(
...
+            "region": str(incident_io_int.get("region", "us")).strip(),
+            "base_url": str(incident_io_int.get("base_url", "")).strip(),
+            "incident_id": incident_id,
+            "title": str(
+                annotations.get("incident_title") or raw_alert.get("alert_name", "")
+            ).strip(),
+            "description": str(
+                annotations.get("incident_comment") or annotations.get("summary") or ""
+            ).strip(),
+            "connection_verified": True,
```
Devesh36: This is set whenever api_key is non-empty, without actually testing connectivity. Other integrations like Grafana check a live probe before setting this flag. A misconfigured key (wrong region, expired key) will still mark the source as verified, causing the tool to attempt calls that will silently fail at runtime.

SB2318: okay,  noted.

SB2318:
### Summary of Changes
1.  **Concurrency Optimization**: Moved the `threading.Lock` in `IncidentIoClient` to the instance level. This ensures that concurrent tool calls for different incidents do not block each other, resolving a global serialization bottleneck.
2.  **Active Verification**: Refactored `app/nodes/plan_actions/detect_sources.py` to replace the hardcoded `connection_verified: True` flag with a live probe check. I implemented a cached probe helper (`_probe_incident_io_cached`) to ensure connectivity is verified reliably without impacting performance during investigation loops.
3.  **Testing**: Added a new test suite in `tests/nodes/plan_actions/test_detect_sources_incident_io.py` to validate the success, failure, and caching behavior of the new verification logic.

### Verification Results
*   **Unit Tests**: All new and existing tests passed successfully.
*   **CLI Verification**: Confirmed that `opensre integrations verify incident_io` correctly identifies missing or invalid credentials.

@greptile review

WatchTree-19 (commented):
Drive-by from someone running similar bench-side work on the same suite. Substantive but not comprehensive, take or leave.

**Webhook signing.** incident.io exposes `X-Incident-Signature` (HMAC-SHA256 of the body, keyed on a per-channel secret). I scrolled the verification adapter and don't see signature validation on inbound webhooks. Without it, anyone who knows the webhook URL can post forged incidents into the system and trigger RCA flows. The shape of the fix is small (constant-time compare against `hmac.new(secret, body, sha256).hexdigest()`), but it has to land before this is exposed to anything other than a closed test harness. If signature verification is intentionally deferred to a follow-up, calling that out in the PR body would help reviewers route attention.

**Multi-tenant case.** A nontrivial fraction of incident.io customers run multiple organisations under one account (e.g. one per business unit). The integration config model I see only carries a single `api_key` plus org-implied. For a single-tenant install that's fine; for the platform side it'll silently route everything to whichever org the key belongs to. Worth either documenting "single-org only" in the catalog entry or threading an explicit `organisation_id` through the client. Same gap exists in the PagerDuty integration today, so this isn't novel, but the docs should at least be explicit.

**Rate limiting.** incident.io's REST API is 100 req/min per org by default with a `Retry-After` header on 429. The verification probe + RCA tools should respect that. A thin wrapper around the client that handles 429 with exponential backoff (the same pattern as the existing `tracer_client` adapter) avoids the worst case where a noisy alert burst exhausts the org's quota. Don't think the current diff has that; happy to be wrong.

**Catalog parity.** The two `incident_io` insertions in `app/cli/support/constants.py` (lines 40 and 70) are alphabetically wedged between `honeycomb` and `mariadb`. Looks correct in both spots, just confirming the sort is intentional and matches what the catalog renderer expects (some renderers special-case "external" vs "internal" integrations, no idea if that applies here).

**Already covered upstream.** Greptile's existing thread on the wizard validators looks like the right shape, no need to pile on.

The agentic-incident-response surface that this opens up is genuinely useful. If signature verification and multi-org are tracked as follow-ups rather than blockers, that's totally reasonable.

cerencamkiran:
@greptileai review

cerencamkiran:
Nice work overall @SB2318 this is a sizeable integration and the Incident.io direction looks useful.

Before approval, I still have a few concerns:

1. "app/pipeline/runners.py" and the shared Sentry/error-reporting changes look unrelated to Incident.io. I’d prefer keeping this PR focused, or moving that refactor into a separate PR.

2. "detect_sources()" now performs a live Incident.io probe via "_probe_incident_io_cached()". Since source detection runs in the planning path, this adds network I/O before the tool is actually used. Could we rely on the normal integration verification path instead, or use a bounded TTL/session-scoped cache?

3. For write-back, do we have gated real-API proof that "add_timeline_event" / summary append works against Incident.io, not only mocked tests?

4. If "/v2/incident_timeline_events" is known to be unavailable, we should avoid hitting it on every write path and go directly to the supported summary-append flow. Related: the current contract tests should also reflect the new default summary-append behaviour, not the stale native-endpoint path.

5. The latest Greptile pass also flags "_get_client()" lazy initialization as not thread-safe. Since this client can be used by concurrent tool calls, "_client" creation/close should probably be protected by an instance-level lock.

6. CI is still failing, so I’d wait until the failing check and current P1s are resolved before considering approval.

I’d be cautious about merging until these are cleaned up or clearly scoped as follow-ups.

SB2318:
> Nice work overall @SB2318 this is a sizeable integration and the Incident.io direction looks useful.
> 
> Before approval, I still have a few concerns:
> 
> 1. "app/pipeline/runners.py" and the shared Sentry/error-reporting changes look unrelated to Incident.io. I’d prefer keeping this PR focused, or moving that refactor into a separate PR.
> 2. "detect_sources()" now performs a live Incident.io probe via "_probe_incident_io_cached()". Since source detection runs in the planning path, this adds network I/O before the tool is actually used. Could we rely on the normal integration verification path instead, or use a bounded TTL/session-scoped cache?
> 3. For write-back, do we have gated real-API proof that "add_timeline_event" / summary append works against Incident.io, not only mocked tests?
> 4. If "/v2/incident_timeline_events" is known to be unavailable, we should avoid hitting it on every write path and go directly to the supported summary-append flow. Related: the current contract tests should also reflect the new default summary-append behaviour, not the stale native-endpoint path.
> 5. The latest Greptile pass also flags "_get_client()" lazy initialization as not thread-safe. Since this client can be used by concurrent tool calls, "_client" creation/close should probably be protected by an instance-level lock.
> 6. CI is still failing, so I’d wait until the failing check and current P1s are resolved before considering approval.
> 
> I’d be cautious about merging until these are cleaned up or clearly scoped as follow-ups.

Yes, I am working on those failures. 
I will cover these points.

muddlebee:
@SB2318 

Thank you for the efforts, but #1710 looks to be in better direction and being tested on live incident.io integration. So will be closing this.

SB2318:
> @SB2318
> 
> Thank you for the efforts, but #1710 looks to be in better direction and being tested on live incident.io integration. So will be closing this.

yeah, I noticed yesterday. I think they are in the same direction. only file name changed. 😂

But anyways, I have no objection.

PR #1619, opened 2026-05-08
feat(sentry): fingerprint rules, tool exception capture, and debug sentry command

### LINKED ISSUE #1479: Sentry coverage 5/5 — Tests, fingerprint rules, and "opensre debug sentry" doctor command
Part 5 of 5 in the Sentry coverage rollout. Locks in the new behavior with tests, makes Sentry issue grouping clean, and gives users a one-shot DSN smoke test.

## Background

PRs 1–4 widen Sentry coverage substantially. This PR adds the safety net so that future regressions are caught and Sentry's UI stays usable as event volume rises.

### Tests

- [ ] Extend [`tests/test_sentry_init.py`](https://github.com/Tracer-Cloud/opensre/blob/main/tests/test_sentry_init.py) to cover PR 1 changes:
    - integrations list contains `LoggingIntegration`, `AsyncioIntegration`, `HttpxIntegration`.
    - `in_app_include=[\"app\"]` is set.
    - `entrypoint`, `runtime`, `deployment_method` tags are set on the global scope after `init_sentry(entrypoint=\"...\")`.
    - New scrubber paths: `request.data` redaction, prompt-key redaction, http-breadcrumb header strip.
    - Remember `_init_sentry_once.cache_clear()` between cases (it's `@cache`d).
- [ ] Extend [`tests/pipeline/test_runners_sentry.py`](https://github.com/Tracer-Cloud/opensre/blob/main/tests/pipeline/test_runners_sentry.py) with a LangGraph-node-internal exception case. Lock in the PR 2(b) `_traced_node` wrapper: an exception raised inside a node should produce exactly one captured event tagged `node=<name>`.
- [ ] New `tests/tools/test_basetool_sentry.py`:
    - Define a fake `BaseTool` subclass whose `run()` raises.
    - Assert that calling the tool triggers `capture_exception` exactly once, sets the `tool` tag, and returns `{\"error\": ..., \"exception_type\": ...}`.
    - Same coverage for the `@tool` decorator path.

### Fingerprint rules

- [ ] In `_before_send` (or via `sentry_sdk.set_context`), add a fingerprint rule so the `Tool error` family groups per-tool rather than collapsing into one giant issue:

    \`\`\`python
    tool_name = (event.get(\"tags\") or {}).get(\"tool\")
    if tool_name:
        event[\"fingerprint\"] = [\"tool-error\", tool_name, \"{{ default }}\"]

    node_name = (event.get(\"tags\") or {}).get(\"node\")
    if node_name:
        event[\"fingerprint\"] = [\"node-error\", node_name, \"{{ default }}\"]
    \`\`\`

- [ ] Same treatment for `subsystem`/`integration` tags from PR 4 if needed (decide based on real Sentry grouping after a week of data).

### Doctor command

- [ ] Add `opensre debug sentry` (or `make sentry-doctor`) that:
    1. Calls `init_sentry(entrypoint=\"debug\")`.
    2. Captures a synthetic exception via `capture_exception(...)`.
    3. Calls `sentry_sdk.flush(timeout=5)`.
    4. Prints whether the event was sent (`flush()` returns `True` if all events were transmitted) and the resolved DSN host (so users can verify their `OPENSRE_SENTRY_DSN` env override worked).
- [ ] Document in `docs/` (probably a new section in [`docs/sentry.mdx`](https://github.com/Tracer-Cloud/opensre/blob/main/docs/sentry.mdx)) so users self-hosting with a custom DSN have a one-shot verification.

## Acceptance

- All new tests pass under `make test-cov`.
- `opensre debug sentry` exits 0 and prints a Sentry event ID when DSN is set; exits non-zero when DSN is unreachable / disabled.
- After a week of production data, Sentry issue list is filterable by `tool`, `node`, `integration` tags and groupings make sense.

## Risks

- Fingerprint rules can over- or under-group. Start with `tool` + `\"{{ default }}\"` so Sentry's default grouping still applies inside a tool — easy to tighten later.
- The doctor command must not capture real user telemetry; tag the synthetic exception with `debug=true` so it can be filtered out in Sentry rules.

This is part of a 5-PR Sentry coverage rollout. Best done after [PRs 1 (#1475)](https://github.com/Tracer-Cloud/opensre/issues/1475) and [2 (#1476)](https://github.com/Tracer-Cloud/opensre/issues/1476).

shoaib050326:
## Summary

Part 5/5 of Sentry coverage rollout. Three additions:

1. **Fingerprint rules in `_before_send`** — Tool errors group per-tool (`tool-error/<name>/{{ default }}`); node errors group per-node (`node-error/<name>/{{ default }}`). Prevents all tool exceptions from collapsing into one Sentry issue.

2. **Tool sentry wrapping** — `BaseTool.__call__` and `RegisteredTool.__call__` now catch exceptions, set the `tool` tag via `sentry_sdk.push_scope`, call `sentry_sdk.capture_exception`, and return structured `{"error": ..., "exception_type": ...}` dicts.

3. **`opensre debug sentry`** — One-shot DSN smoke test. Initializes Sentry, captures a synthetic exception (tagged `debug=true`), flushes, reports transmit status. Exits 0 on success, 1 on failure.

## Changes

| File | Change |
|------|--------|
| `app/utils/sentry_sdk.py` | `_apply_fingerprint()` called from `_before_send` — per-tool/per-node fingerprint rules |
| `app/tools/base.py` | Exception capture with `tool` tag in `__call__` |
| `app/tools/registered_tool.py` | Exception capture with `tool` tag in `__call__` |
| `app/cli/commands/debug.py` (new) | `opensre debug sentry` command |
| `app/cli/commands/__init__.py` | Register debug group |
| `tests/test_sentry_init.py` | 4 new fingerprint tests |
| `tests/pipeline/test_runners_sentry.py` | Node-level internal exception test |
| `tests/tools/test_basetool_sentry.py` (new) | BaseTool/RegisteredTool sentry tests |
| `docs/sentry.mdx` | DSN smoke test documentation |

## Risks

- Fingerprint over/under-grouping: uses `{{ default }}` so Sentry's built-in grouping still applies within tool/node scope — easy to tighten later.
- The doctor command only captures synthetic events (tagged `debug=true`) — no real user telemetry.

Closes #1479

### DIFF
--- a/app/cli/commands/__init__.py
+++ b/app/cli/commands/__init__.py
@@ -6,6 +6,7 @@ import click
 
 from app.cli.commands.agent import agents
 from app.cli.commands.config import config_command
+from app.cli.commands.debug import debug_group
 from app.cli.commands.deploy import deploy
 from app.cli.commands.doctor import doctor_command
 from app.cli.commands.general import (
@@ -25,6 +26,7 @@ _COMMANDS: tuple[click.Command, ...] = (
     investigate_command,
     onboard,
     config_command,
+    debug_group,
     deploy,
     remote,
     tests,
--- a/app/cli/commands/debug.py
+++ b/app/cli/commands/debug.py
@@ -1,0 +1,101 @@
+"""Diagnostic / debug subcommands — ``opensre debug <subcommand>``.
+
+Currently provides:
+- ``opensre debug sentry`` — one-shot DSN smoke test.
+"""
+
+from __future__ import annotations
+
+from urllib.parse import urlsplit
+
+import click
+from rich.console import Console
+from rich.text import Text
+
+from app.cli.support.exit_codes import ERROR as EXIT_ERROR
+from app.cli.support.exit_codes import SUCCESS
+
+
+@click.group(name="debug")
+def debug_group() -> None:
+    """Diagnostic and debugging utilities."""
+
+
+@debug_group.command(name="sentry")
+def sentry_debug_command() -> None:
+    """Verify Sentry DSN connectivity with a synthetic exception.
+
+    Calls ``init_sentry(entrypoint="debug")``, captures a synthetic
+    exception, flushes the event queue, and reports whether the event
+    was transmitted successfully.
+
+    The synthetic event is tagged with ``debug=true`` so it can be
+    filtered out in Sentry's issue stream.
+
+    Exit codes:
+    0 — DSN configured and event transmitted.
+    1 — DSN missing or event not transmitted.
+    """
+    from app.utils.sentry_sdk import (
+        _is_sentry_disabled,
+        _resolved_dsn,
+        capture_exception,
+        init_sentry,
+    )
+
+    console = Console(highlight=False, force_terminal=True, color_system="truecolor")
+    if _is_sentry_disabled():
+        console.print(Text("Sentry telemetry is disabled.", style="bold yellow"))
+        console.print(
+            Text(
+                "Unset OPENSRE_NO_TELEMETRY, OPENSRE_SENTRY_DISABLED, or DO_NOT_TRACK to test Sentry.",
+                style="dim",
+            )
+        )
+        raise SystemExit(EXIT_ERROR)
+
+    sentry_dsn = _resolved_dsn()
+
+    if not sentry_dsn:
+        console.print(
+            Text("No Sentry DSN configured.", style="bold red"),
+        )
+        msg = (
+            "Set OPENSRE_SENTRY_DSN (or SENTRY_DSN) in your environment, "
+            "or leave unset to use the built-in DSN."
+        )
+        console.print(Text(msg, style="dim"))
+        raise SystemExit(EXIT_ERROR)
+
+    dsn_host = urlsplit(sentry_dsn).hostname or "unknown"
+    console.print(f"DSN host: [bold]{dsn_host}[/bold]")
+
+    try:
+        init_sentry(entrypoint="debug")
+    except Exception as exc:
+        console.print(
+            Text(f"Sentry initialization failed: {exc}", style="bold red"),
+        )
+        raise SystemExit(EXIT_ERROR) from exc
+
+    import sentry_sdk
+
+    try:
+        raise RuntimeError("Sentry debug test — not a real error")
+    except RuntimeError as exc:
+        with sentry_sdk.push_scope() as scope:
+            scope.set_tag("debug", "true")
+            capture_exception(exc, context="debug.sentry_test")
+
+    sentry_ok = sentry_sdk.flush(timeout=5)
+
+    if sentry_ok:
+        console.print(
+            Text("Sentry event transmitted successfully.", style="bold green"),
+        )
+        raise SystemExit(SUCCESS)
+
+    console.print(
+        Text("Sentry event NOT transmitted — check DSN and network.", style="bold red"),
+    )
+    raise SystemExit(EXIT_ERROR)
--- a/app/cli/interactive_shell/command_registry/cli_parity.py
+++ b/app/cli/interactive_shell/command_registry/cli_parity.py
@@ -82,6 +82,10 @@ def _cmd_config(session: ReplSession, console: Console, args: list[str]) -> bool
     return run_cli_command(console, ["config", *args])
 
 
+def _cmd_debug(session: ReplSession, console: Console, args: list[str]) -> bool:  # noqa: ARG001
+    return run_cli_command(console, ["debug", *args])
+
+
 COMMANDS: list[SlashCommand] = [
     SlashCommand(
         "/onboard",
@@ -129,6 +133,12 @@ COMMANDS: list[SlashCommand] = [
         "/config",
         "show or edit local OpenSRE config ('/config show|set <key> <value>')",
         _cmd_config,
+        execution_tier=ExecutionTier.SAFE,
+    ),
+    SlashCommand(
+        "/debug",
+        "run diagnostic debug utilities ('/debug sentry')",
+        _cmd_debug,
         execution_tier=ExecutionTier.SAFE,
     ),
 ]
--- a/app/nodes/chat.py
+++ b/app/nodes/chat.py
@@ -239,6 +239,9 @@ def tool_executor_node(state: AgentState) -> dict[str, Any]:
         except GuardrailBlockedError:
             raise
         except Exception as e:
+            from app.utils.sentry_sdk import capture_exception
+
+            capture_exception(e, context=f"tool.{tool_name}")
             result = json.dumps({"error": str(e)})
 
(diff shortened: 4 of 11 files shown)


yashksaini-coder (changes requested):
Did a deep pass against the diff and the live repo. The Sentry-fingerprint shape is the right idea, but there are 5 distinct issues, two of which repeat the exact bug pattern flagged on PR #1615. Greptile's 3/5 score is accurate.

## Critical

### 1. `BaseTool.__call__` and `RegisteredTool.__call__` Sentry capture is unreachable in production

The investigation pipeline calls `action.run(**kwargs)` directly at `execute_actions.py:50`, not `action(**kwargs)`. The chat pipeline calls `tool_fn.invoke(tool_args)` at `chat.py:292` (LangChain's `invoke`), also not `__call__`. Neither production path goes through `__call__`, so the 30+ lines of try/except + `push_scope` + `capture_exception` added in `base.py` (lines 148-165) and `registered_tool.py` (lines 176-193) are dead code in every real execution path. The PR correctly adds capture at `chat.py:123` and `execute_actions.py:137`, but those are the actual needed sites. The `__call__` instrumentation is misleading.

Fix: remove the try/except blocks from `BaseTool.__call__` and `RegisteredTool.__call__` entirely.

### 2. Fingerprint logic is structurally coupled to that dead code

`_apply_fingerprint` in `app/utils/sentry_sdk.py:205-219` reads `event["tags"]["tool"]` / `event["tags"]["node"]` to attach `tool-error/<name>` or `node-error/<name>` fingerprints. But `capture_exception` in the same module sets `opensre.context`, not a `tool` or `node` tag. So the fingerprint path **only fires from the `__call__` capture site that no production code reaches**. The fingerprint feature is effectively dead.

Fix: either pass `fingerprint` explicitly through `capture_exception` so the grouping decision lives at the call site, or set `tool` / `node` tags inside the production capture sites added in this PR (`chat.py:123`, `execute_actions.py:137`).

### 3. `opensre debug sentry` reports false-positive success when telemetry is disabled

`_resolved_dsn()` in `debug.py:68` returns the bundled `SENTRY_DSN` constant unconditionally. When `OPENSRE_NO_TELEMETRY=1` is set, the early-exit guard never triggers, the command proceeds past `init_sentry(entrypoint="debug")` (which immediately returns early at `sentry_sdk.py:358-360` due to the disabled flag), and the subsequent `sentry_sdk.flush(timeout=5)` returns `True` from `NoopHub` even though nothing was sent. Operators get "Sentry event transmitted successfully" with no actual transmission.

Fix: add `_is_sentry_disabled()` check at the top of `sentry_debug_command` and exit with a clear "telemetry disabled — set OPENSRE_TELEMETRY=1 to test" message.

### 4. Global scope mutation — exact same bug as PR #1615

`debug.py:94`:

```python
sentry_sdk.set_tag("debug", "true")
```

This sets `debug=true` on the global Sentry scope. Any subsequent call to `init_sentry` in the same process inherits this tag. This is the **exact** scope-leakage pattern flagged as a blocker on PR #1615. Wrap in `with sentry_sdk.push_scope() as scope: scope.set_tag("debug", "true"); capture_exception(...)`.

## Important

### 5. `_apply_fingerprint` placed outside the `_before_send` try/except guard

`sentry_sdk.py` (around diff line 229): the new `_apply_fingerprint(event)` call sits outside the existing `try / except Exception: return event` that wraps `_scrub_event_in_place`. The comment immediately above explicitly says "The hook must never raise." `_apply_fingerprint` is currently simple enough not to raise, but the placement violates the stated invariant — a future change adding e.g. `event["fingerprint"] = ...` against an unexpected event shape would propagate.

Fix: move `_apply_fingerprint(event)` inside the `try:` block alongside `_scrub_event_in_place`.

## Summary table

| # | Issue | File | Confidence |
|---|---|---|---|
| 1 | `__call__` Sentry capture is dead code | `base.py:148`, `registered_tool.py:176` | 95 |
| 2 | Fingerprint logic only fires from unreachable code | `sentry_sdk.py:205` | 85 |
| 3 | Debug command false-positive success when telemetry disabled | `debug.py:68` | 90 |
| 4 | Global scope leak — repeats PR #1615 blocker | `debug.py:94` | 88 |
| 5 | `_apply_fingerprint` outside try/except in `_before_send` | `sentry_sdk.py:229` | 82 |

## Verdict

Request changes. Issues 1 and 3 are the most impactful: the core instrumentation never fires in production, and the debug command can actively mislead operators. Issue 4 repeats the blocker from PR #1615 — please apply the same fix pattern. All five items need to land before merge.

WatchTree-19 (commented):
Drive-by from someone working alongside on the eval-bench side. Sentry coverage is the right thread to be pulling on, take or leave.

**Fingerprint rules: cardinality.** Sentry's fingerprint API groups events into issues; bad fingerprint rules are how an integration ends up creating thousands of distinct issues that should have been one (each tool-exception with a slightly different stack trace becomes its own issue). The fingerprint should normalise away dynamic content (request IDs, timestamps, user IDs in stack frames) and key on stable identity (tool name, exception class, top frame in our code). Worth confirming the rules in this PR are tested against a couple of real exception payloads before merge so we don't blow out the issue count on first deploy.

**Tool exception capture: PII.** Tool exceptions often carry user-facing payloads (alert bodies, log lines, query results) in their attached context. Sentry has a server-side scrubber but the safer pattern is to scrub at the client: a `before_send` hook that strips known PII fields (e.g. anything matching email regex, API key shapes, anything in an explicit `secrets` allowlist). Without that, sensitive data can land in Sentry events and become a separate compliance surface. `sentry_sdk.init(before_send=...)` is the canonical hook.

**`debug sentry` command: env hygiene.** A debug command that triggers a Sentry event is useful for ops; the failure mode to avoid is shipping it in a way that makes it discoverable to end users and clutters the production project with test events. Either gate behind `OSRE_DEV=1` or route to a separate `-debug` Sentry project so the events are visible during dev but don't pollute the real project.

**Init coverage gap.** This PR is "Sentry coverage 4/5" or similar (saw 2/5 and 3/5 in the issue queue). The 5/5 should probably also cover analytics, daily-update, github-issue-comments, grafana-seed, and the hosted graph entrypoints, all of which currently have no `init_sentry()` based on a quick scan of `app/__init__.py` callers. Worth a tracking issue if there isn't one already.

Sentry hygiene is one of those things that pays off six months in when you can actually find the bug from the issue title. Good direction.

VibhorGautam (commented):
quick thought on the execute_actions.py change - #1620 added a report_exception helper in app/utils/errors.py that wraps both the logger call and sentry capture together. since the retry loop already has a logger, switching to report_exception here would keep things consistent with how the rest of the error-handling layer works

muddlebee:
closing in favour of https://github.com/Tracer-Cloud/opensre/pull/1620

Issue #1489, opened 2026-05-06
Add `psutil` dependency and `app/agents/probe.py` per-PID sampler


davincios:
## Context: Monitor Local Agents — a fleet dashboard *inside the `opensre` interactive shell*

OpenSRE is being extended with a **fleet view of the *other* AI agents running on the developer's machine** — Claude Code, Cursor, Aider, Codex CLI, Gemini, and friends — by treating each as a microservice and applying normal SRE practice: **golden signals** (CPU / RAM / tokens-per-minute / $/hr), **SLOs** (forward progress, cost, error rate), and **incident response** (stuck → notify → RCA → restart).

### Surface: the existing OpenSRE interactive shell (REPL)

This entire feature lives **inside the `opensre` interactive shell** — the same `prompt_toolkit` + Rich TTY REPL users get today by typing `opensre` with no arguments. **There is no web UI, no separate daemon, no extra terminal window.** The user types `/agents` at the REPL prompt the same way they type `/status`, `/integrations`, or `/tasks` today, and the dashboard renders inline in the same terminal:

```
> /agents

agent          pid    uptime   cpu%   tokens/min   $/hr   status
claude-code    8421   12m      4.2    180          2.10   active
cursor-tab     9133   3h       0.1    -            -      idle
aider          7702   42m      0.0    0            1.40   stuck (8m no progress)
opensre        self   1h       1.1    90           0.30   active

> /agents trace 8421
[live tail of claude-code's stdout streams here until Ctrl-C]

> /agents kill 7702
About to SIGTERM aider (pid 7702). Confirm? [y/N]
```

**Every** metric, alert, redaction, and conflict that ships in this initiative is surfaced through this slash-command interface. Backend collectors (psutil probe, token meters, file watchers, secret scanner, loop detector, error-rate detector, SLO watchdog) all feed into the same `/agents` table or its drill-down views: `/agents trace <pid>`, `/agents inspect <pid>`, `/agents conflicts`, `/agents graph`, `/agents bus`, `/agents budget`. Code-level location of all this work: `app/cli/interactive_shell/command_registry/agents.py` (the slash command) and the new `app/agents/` package (the collectors that feed it).

[...]

PR #1624, opened 2026-05-08
feat(agents): startup orphan / stale-lockfile sweep

### LINKED ISSUE #1501: Startup orphan / stale lockfile sweep

```
> /agents

> /agents trace 8421

> /agents kill 7702
```

|---|---|

---

## This issue: Startup orphan / stale lockfile sweep

**Phase:** 3 — Safety & hygiene
**Effort:** S (~half-day)
**Depends on:** #1487

### Why we need this

When the REPL crashes or an agent process dies unexpectedly, stale entries pile up in the `AgentRegistry` JSONL and stale lockfiles accumulate in `~/.opensre/agents/`. Without a sweep, the `/agents` dashboard inside the shell shows phantom rows for processes that no longer exist, and lockfiles can block legitimate new agents from registering.

### What ships in this issue

Run-once-on-REPL-boot sweep: prune `AgentRegistry` entries whose PIDs are gone (use the psutil probe — `None` return means dead) and delete stale lockfiles in `~/.opensre/agents/`. Idempotent.

Inside the `opensre` shell, `/agents` no longer shows phantom rows for PIDs that died between sessions. The sweep runs silently at REPL boot; verbose mode logs each removal.

### Files to touch

- `app/agents/sweep.py` (new) — `sweep(registry, lock_dir)`
- `app/cli/interactive_shell/loop.py` — call once at startup
- `tests/agents/test_sweep.py`

### Acceptance

- [ ] Idempotent (running twice in a row is a no-op the second time)
- [ ] Logs each removal at debug level

### Where to start

Use the `probe()` function from `app/agents/probe.py` — its `None` return for non-existent PIDs is exactly the dead-PID signal you need.

---

yashksaini-coder:
Fixes #1501

Phase-3 safety hygiene for the [`monitor-local-agents`](https://github.com/Tracer-Cloud/opensre/labels/monitor-local-agents) initiative. Idempotent run-once-on-REPL-boot pass that prunes `AgentRegistry` entries whose PIDs no longer exist plus stale lockfiles in `~/.config/opensre/agents/`.

The result: when the REPL crashes or an agent process dies between sessions, `/agents` no longer shows phantom rows for PIDs that don't exist, and stale lockfiles can't block legitimate new agents from registering.

## Summary

- **`app/agents/sweep.py`** *(new)* — `sweep(registry, lock_dir=DEFAULT_LOCK_DIR) -> SweepResult` plus a `run_startup_sweep()` boot wrapper. Frozen `SweepResult` dataclass exposes `removed_records`, `removed_locks`, and a `total` property for log lines.
- **`app/cli/interactive_shell/loop.py`** — call `run_startup_sweep()` once, immediately after `render_banner(console)`, so dead-PID rows are pruned before the user's first `/agents` invocation.
- **`tests/agents/test_sweep.py`** *(new)* — 11 cases covering both registry-side and lockfile-side sweeps, idempotency, edge cases, and the boot-wrapper contract.

## Design notes for reviewers

- **`cpu_interval=0.0` for the probe call.** The sweep only needs the existence signal — `probe(pid)` returns `None` when the PID doesn't exist regardless of CPU sample accuracy. The default 100 ms blocking sample × N registered agents would be a noticeable startup tax for users who track many agents.
- **Lockfile dir tolerance.** A missing `lock_dir` is the common case on a fresh install before any agent has registered. The sweep tolerates it silently. Files in `lock_dir` whose stems aren't valid integers are left untouched — the sweep only acts on its own `<pid>.lock` convention so it can't accidentally delete `settings.json` or another tool's `registry.lock`.
- **OS-level unlink failure is logged, not raised.** If the filesystem refuses to remove a stale lockfile (read-only mount, permission denied), the sweep logs at WARNING and continues with the rest. This keeps the REPL boot resilient.
- **`run_startup_sweep()` swallows any exception.** A sweep bug must never prevent the REPL from starting at all. Locked in by `test_run_startup_sweep_swallows_exceptions` which mocks `AgentRegistry()` to raise.
- **Idempotency comes for free** from the design: the sweep reads current registry/lockfile state on each call and acts only on dead PIDs. After a clean run, every remaining entry is alive, so a second invocation removes nothing. `test_idempotent_second_run_is_a_noop` locks this in.

## Acceptance checklist

- [x] Idempotent (running twice in a row is a no-op the second time)
- [x] Logs each removal at debug level (`logger.debug` calls in both `_sweep_registry` and `_sweep_locks`; warning level reserved for OS-level unlink failures)
- [x] `make lint` passes
- [x] `make format-check` passes
- [x] `make typecheck` passes (513 source files checked)
- [x] `tests/agents/test_sweep.py` — 11/11 pass

## Test plan

- [x] **`test_dead_pid_record_is_forgotten`** — registered PID 2³¹−1 is pruned.
- [x] **`test_live_pid_record_is_kept`** — current Python process (`os.getpid()`) survives.
- [x] **`test_idempotent_second_run_is_a_noop`** — first run prunes, second run returns empty `SweepResult` (the contract underlying safe boot-time invocation).
- [x] **`test_stale_lockfile_for_dead_pid_is_removed`** + **`test_live_lockfile_is_kept`** — symmetric coverage of the lockfile path.
- [x] **`test_non_pid_lockfile_names_are_ignored`** — `registry.lock`, `settings.json` etc. left untouched.
- [x] **`test_missing_lock_dir_is_tolerated`** — fresh-install case doesn't raise.
- [x] **`test_lockfile_unlink_failure_is_logged_not_raised`** — `Path.unlink` mocked to `PermissionError`, sweep logs and continues.
- [x] **`test_sweep_result_total_property`** + **`test_empty_sweep_result_is_falsy_in_total`** — the dataclass ergonomics used by the boot logger.
- [x] **`test_run_startup_sweep_swallows_exceptions`** — `AgentRegistry()` mocked to raise; `run_startup_sweep()` returns empty `SweepResult` instead of propagating.
- [x] Broader sweep `tests/agents/ tests/cli/interactive_shell/` — 536 passed.

Note on the full coverage suite: `make test-cov` shows one pre-existing flake in `tests/pipeline/test_runners_sentry.py::test_run_chat_initializes_sentry_and_captures_unhandled_errors` that fails identically on plain `main` without my changes — confirmed by checking out `main` and running the same test in isolation.

### DIFF
--- a/app/agents/probe.py
+++ b/app/agents/probe.py
@@ -44,6 +44,28 @@ class ProcessSnapshot:
     started_at: datetime
 
 
+def pid_exists(pid: int) -> bool:
+    """Return whether ``pid`` corresponds to a process the OS knows about.
+
+    Thin wrapper over ``psutil.pid_exists`` exposed here so ``psutil``
+    stays confined to this module per the issue #1489 acceptance
+    criterion. Unlike ``probe()``, this returns ``True`` for processes
+    we can't introspect (cross-user on macOS, restricted ``/proc``,
+    etc.) — the OS-level existence check doesn't traverse the access
+    boundary, which is exactly what the boot sweep (#1501) needs to
+    avoid pruning live foreign-user agents.
+
+    Returns ``False`` for PIDs outside the platform's valid range
+    (e.g. an int that overflows the kernel's PID type) — psutil
+    raises ``OverflowError`` or ``ValueError`` on those, which we
+    treat as "not a real PID" rather than propagating.
+    """
+    try:
+        return psutil.pid_exists(pid)
+    except (OverflowError, ValueError):
+        return False
+
+
 def probe(pid: int, *, cpu_interval: float = 0.1) -> ProcessSnapshot | None:
     """Return a one-shot resource snapshot for ``pid``.
 
--- a/app/agents/registry.py
+++ b/app/agents/registry.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import json
 import logging
+from collections.abc import Iterable
 from dataclasses import asdict, dataclass, field
 from datetime import UTC, datetime
 from pathlib import Path
@@ -62,6 +63,25 @@ class AgentRegistry:
     def forget(self, pid: int) -> AgentRecord | None:
         removed = self._records.pop(pid, None)
         if removed is not None:
+            self._rewrite()
+        return removed
+
+    def forget_many(self, pids: Iterable[int]) -> list[AgentRecord]:
+        """Remove every PID in ``pids`` and rewrite the JSONL file once.
+
+        Functionally equivalent to calling :meth:`forget` in a loop,
+        but does only **one** disk rewrite at the end. Use this when
+        pruning many records at once (e.g. the boot-time sweep) so a
+        registry holding N dead PIDs doesn't trigger N full rewrites.
+        Returns the records that were actually removed (silently
+        skips PIDs not present in the registry).
+        """
+        removed: list[AgentRecord] = []
+        for pid in pids:
+            record = self._records.pop(pid, None)
+            if record is not None:
+                removed.append(record)
+        if removed:
             self._rewrite()
         return removed
 
--- a/app/agents/sweep.py
+++ b/app/agents/sweep.py
@@ -1,0 +1,143 @@
+"""Startup orphan / stale-lockfile sweep for the local agent registry.
+
+Run once at REPL boot. Idempotent — running twice in a row is a no-op
+the second time. Removes ``AgentRegistry`` entries whose PIDs no longer
+exist plus lockfiles in ``~/.config/opensre/agents/`` that correspond
+to dead PIDs.
+
+The function is split from the boot wiring so it stays unit-testable
+without spinning up the REPL: the loop.py side just calls
+``run_startup_sweep()`` and lets this module handle path defaults and
+error suppression.
+"""
+
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass, field
+from pathlib import Path
+
+from app.agents.probe import probe
+from app.agents.registry import AgentRecord, AgentRegistry
+from app.constants import OPENSRE_HOME_DIR
+
+logger = logging.getLogger(__name__)
+
+#: Default location of the per-PID lockfile directory.
+DEFAULT_LOCK_DIR: Path = OPENSRE_HOME_DIR / "agents"
+
+
+@dataclass(frozen=True)
+class SweepResult:
+    """What a single ``sweep()`` invocation removed.
+
+    Empty tuples on an already-clean run; an already-pruned registry
+    paired with no stale lockfiles produces ``SweepResult()`` — that's
+    the contract idempotency rests on.
+    """
+
+    removed_records: tuple[AgentRecord, ...] = field(default_factory=tuple)
+    removed_locks: tuple[Path, ...] = field(default_factory=tuple)
+
+    @property
+    def total(self) -> int:
+        """Sum of removed records and lockfiles. Useful for log messages."""
+        return len(self.removed_records) + len(self.removed_locks)
+
+
+def sweep(
+    registry: AgentRegistry,
+    lock_dir: Path = DEFAULT_LOCK_DIR,
+) -> SweepResult:
+    """Remove dead-PID registry entries and stale lockfiles.
+
+    Parameters
+    ----------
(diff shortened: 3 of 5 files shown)


yashksaini-coder:
@greptile review

yashksaini-coder:
Thanks for the careful review. All seven items addressed in `4cc4655e` (rebased onto fresh `main` to pull in the pipeline test fix from #1620's follow-up):

**Greptile / Copilot — liveness check (P1).** Both bots flagged that `probe(pid) is None` is wrong for liveness because `probe()` also returns `None` for `psutil.AccessDenied` (cross-user processes on macOS / hardened `/proc`). Switched both `_sweep_registry` and `_sweep_locks` to use `pid_exists`, which does an OS-level existence check that doesn't traverse the access boundary. Added `pid_exists()` as a thin wrapper in `probe.py` so the "psutil only in probe.py" invariant from #1489 still holds. Two new tests mock `pid_exists` to simulate the AccessDenied case and assert that live foreign-user records and lockfiles are kept.

**Copilot — batched registry rewrite.** `AgentRegistry.forget()` rewrites the whole JSONL file each call, so N dead PIDs = N rewrites. Added `AgentRegistry.forget_many(pids)` which removes all matching records and rewrites once. `_sweep_registry` now collects the dead-PID set first, then calls `forget_many` once. Pinned by `test_many_dead_pids_trigger_one_rewrite_not_n` which spies on `_rewrite` and asserts call count is 1 for 5 dead PIDs.

**Copilot — `exc_info` on unlink failure.** Added `exc_info=True` to the `OSError` warning so operators see the underlying `PermissionError`/`OSError` details, not just a generic message. Test asserts the log record's `exc_info` is non-None.

**Greptile — `except OSError` too broad on unlink.** Split into two branches: `FileNotFoundError` is a race against another sweep / external cleanup that already removed the file → silent idempotent success, no warning. `OSError` (read-only fs, perms) keeps the warning + exc_info. New test pins the silent-FileNotFoundError contract.

**Greptile — `# pragma: no cover` contradicts existing test.** Removed; replaced with a comment pointing at `test_run_startup_sweep_swallows_exceptions` which actually exercises that branch.

**Test fixture overflow.** `_DEAD_PID + 1` was `2**31`, which overflows `psutil.pid_exists`'s int32. Switched to `_DEAD_PID - 1`. The `pid_exists` wrapper also catches `OverflowError`/`ValueError` for defense in depth.

**CI failure cleared.** The `test_run_chat_initializes_sentry_and_captures_unhandled_errors` flake was pre-existing and got fixed on `main` via `6cf3039a`. Rebase pulled it in.

`make lint` / `format-check` / `typecheck` all green; `tests/agents/` — 65/65 pass (up from 50; 4 new cases this round).

muddlebee:
@yashksaini-coder 
image

some codeql and above issues above, pls fix and LMK

muddlebee:
@greptileai review

yashksaini-coder:
@muddlebee — both CodeQL findings cleared in `e2cba0b` (will compute the actual SHA after push).

The two CodeQL alerts were the same root cause: when I added AccessDenied test cases earlier, I introduced `import app.agents.sweep as sweep_module` at module level (line 11) but didn't remove the existing inner `import app.agents.sweep as sweep_module` in `test_run_startup_sweep_swallows_exceptions`. CodeQL flagged it as both "imported more than once" and "imported with both `import` and `import from`". Dropped the inner import; tests use the module-level alias.

The Greptile/Copilot inline comments visible in the file diff are from the earlier review on `f432ce34` and were addressed in the rebased `4cc4655e` push — they show as outdated in the GitHub UI but don't auto-resolve. Greptile's latest summary score is 5/5 against the current commit.

CI green, lint/format/typecheck all clean. Ready for re-review.

yashksaini-coder:
hey @muddlebee, went through everything:

- the copilot + greptile inline comments above are all already fixed in `4cc4655e` (the rebase commit) — they just show as still-open in the diff because the threads haven't been resolved in the UI yet
- codeql duplicate-import warnings fixed in the latest push `3fc18cc6`
- greptile gave 5/5 against the previous commit, latest push is just the 1-line codeql fix so should hold

think it's ready — lmk if you spot anything else 🙏

rrajan94:
LGTM 👍 @yashksaini-coder, tested locally and all passed.

Minor note: the acceptance checklist says `11/11 pass` but `tests/agents/test_sweep.py` has 15 tests. Looks like 4 tests were added after writing the checklist (to address review feedback). Just flagging so the count doesn't confuse anyone.

PR #1627, opened 2026-05-08
feat(agents): /agents slash command renders registered rows (#1488)

yashksaini-coder:
Fixes #1488

cc @muddlebee — co-assignee. Opening this with a complete implementation since the chain (#1490, #1492, #1493) is gating on the slash-command surface here. **Happy to iterate on the design — please review when you have time, no rush. Also fully open to bringing your changes on top if you'd prefer to reshape any part of this.**

---

Bare `/agents` now renders the registered-agents dashboard from the `AgentRegistry`. The existing `/agents conflicts` subcommand stays unchanged. Cells for metrics that aren't yet wired (`cpu%`, `tokens/min`, `$/hr`, `status`, `uptime`) render as `-` placeholders — #1490 (probe wiring) and #1494 (budget loader) fill them in later.

## Summary

- **New `app/cli/interactive_shell/agents_view.py`** — pure rendering function `render_agents_table(records: Iterable[AgentRecord]) -> Table`. Ships the column structure documented in #1486's preview: `agent`, `pid`, `uptime`, `cpu%`, `tokens/min`, `$/hr`, `status`. Lives outside `app/agents/` deliberately so the collector package stays free of Rich (a UI library) — non-CLI consumers of the agents package shouldn't pull Rich in transitively.
- **`app/cli/interactive_shell/command_registry/agents.py`** — bare `/agents` resolves to `_cmd_agents_list` which constructs an `AgentRegistry` and prints the rendered table. `/agents conflicts` keeps its existing handler. `SlashCommand` description updated.
- **`tests/cli/interactive_shell/test_agents_view.py`** *(new)* — 11 pure rendering cases pinning the column contract, justify defaults, empty-state caption, multi-row ordering, placeholder semantics for metric cells, and Rich-markup escape behavior.
- **`tests/cli/interactive_shell/test_agents_commands.py`** — flipped the bare-`/agents` test from "prints usage hint" to "renders empty-state table" and added a populated-registry case. Tests use a `tmp_path`-isolated registry via `monkeypatch` so they don't read the developer's real `~/.config/opensre/agents.jsonl`.

## Behavior

```
> /agents                                                     # populated registry

                              agents
┏━━━━━━━━━━━━━┳━━━━━━┳━━━━━━━━┳━━━━━━┳━━━━━━━━━━━━┳━━━━━━┳━━━━━━━━┓
┃ agent       ┃  pid ┃ uptime ┃ cpu% ┃ tokens/min ┃ \$/hr ┃ status ┃
[...]

Issue #1488, opened 2026-05-06
Add `/agents` slash command rendering registered rows


davincios:
## Context: Monitor Local Agents — a fleet dashboard *inside the `opensre` interactive shell*

OpenSRE is being extended with a **fleet view of the *other* AI agents running on the developer's machine** — Claude Code, Cursor, Aider, Codex CLI, Gemini, and friends — by treating each as a microservice and applying normal SRE practice: **golden signals** (CPU / RAM / tokens-per-minute / $/hr), **SLOs** (forward progress, cost, error rate), and **incident response** (stuck → notify → RCA → restart).

### Surface: the existing OpenSRE interactive shell (REPL)

This entire feature lives **inside the `opensre` interactive shell** — the same `prompt_toolkit` + Rich TTY REPL users get today by typing `opensre` with no arguments. **There is no web UI, no separate daemon, no extra terminal window.** The user types `/agents` at the REPL prompt the same way they type `/status`, `/integrations`, or `/tasks` today, and the dashboard renders inline in the same terminal:

```
> /agents

agent          pid    uptime   cpu%   tokens/min   $/hr   status
claude-code    8421   12m      4.2    180          2.10   active
cursor-tab     9133   3h       0.1    -            -      idle
aider          7702   42m      0.0    0            1.40   stuck (8m no progress)
opensre        self   1h       1.1    90           0.30   active

> /agents trace 8421
[live tail of claude-code's stdout streams here until Ctrl-C]

> /agents kill 7702
About to SIGTERM aider (pid 7702). Confirm? [y/N]
```

**Every** metric, alert, redaction, and conflict that ships in this initiative is surfaced through this slash-command interface. Backend collectors (psutil probe, token meters, file watchers, secret scanner, loop detector, error-rate detector, SLO watchdog) all feed into the same `/agents` table or its drill-down views: `/agents trace <pid>`, `/agents inspect <pid>`, `/agents conflicts`, `/agents graph`, `/agents bus`, `/agents budget`. Code-level location of all this work: `app/cli/interactive_shell/command_registry/agents.py` (the slash command) and the new `app/agents/` package (the collectors that feed it).

Categories monitored: liveness & forward progress, resource & cost, quality (loop detection, error rates, user interventions), safety & hygiene (secret leaks, filesystem blast radius), cross-agent coordination (file/branch conflicts, dependencies), and automatic incident response that drops into the existing OpenSRE investigation pipeline when an SLO is breached.

[...]

Issue #1495, opened 2026-05-06
Add `TokenMeter` base interface and Claude Code parser


davincios:
## Context: Monitor Local Agents — a fleet dashboard *inside the `opensre` interactive shell*

OpenSRE is being extended with a **fleet view of the *other* AI agents running on the developer's machine** — Claude Code, Cursor, Aider, Codex CLI, Gemini, and friends — by treating each as a microservice and applying normal SRE practice: **golden signals** (CPU / RAM / tokens-per-minute / $/hr), **SLOs** (forward progress, cost, error rate), and **incident response** (stuck → notify → RCA → restart).

### Surface: the existing OpenSRE interactive shell (REPL)

This entire feature lives **inside the `opensre` interactive shell** — the same `prompt_toolkit` + Rich TTY REPL users get today by typing `opensre` with no arguments. **There is no web UI, no separate daemon, no extra terminal window.** The user types `/agents` at the REPL prompt the same way they type `/status`, `/integrations`, or `/tasks` today, and the dashboard renders inline in the same terminal:

```
> /agents

agent          pid    uptime   cpu%   tokens/min   $/hr   status
claude-code    8421   12m      4.2    180          2.10   active
cursor-tab     9133   3h       0.1    -            -      idle
aider          7702   42m      0.0    0            1.40   stuck (8m no progress)
opensre        self   1h       1.1    90           0.30   active

> /agents trace 8421
[live tail of claude-code's stdout streams here until Ctrl-C]

> /agents kill 7702
About to SIGTERM aider (pid 7702). Confirm? [y/N]
```

**Every** metric, alert, redaction, and conflict that ships in this initiative is surfaced through this slash-command interface. Backend collectors (psutil probe, token meters, file watchers, secret scanner, loop detector, error-rate detector, SLO watchdog) all feed into the same `/agents` table or its drill-down views: `/agents trace <pid>`, `/agents inspect <pid>`, `/agents conflicts`, `/agents graph`, `/agents bus`, `/agents budget`. Code-level location of all this work: `app/cli/interactive_shell/command_registry/agents.py` (the slash command) and the new `app/agents/` package (the collectors that feed it).

Categories monitored: liveness & forward progress, resource & cost, quality (loop detection, error rates, user interventions), safety & hygiene (secret leaks, filesystem blast radius), cross-agent coordination (file/branch conflicts, dependencies), and automatic incident response that drops into the existing OpenSRE investigation pipeline when an SLO is breached.

[...]

PR #1632, opened 2026-05-08
feat(agents): count ctrl-c and correction interventions (#1498)

### LINKED ISSUE #1498: REPL Ctrl-C / 'no, do X instead' intervention counter

```
> /agents

> /agents trace 8421

> /agents kill 7702
```

|---|---|

---

## This issue: REPL Ctrl-C / 'no, do X instead' intervention counter

**Phase:** 2 — Quality & cost
**Effort:** XS (~1-2h)
**Depends on:** #1488

### Why we need this

Track how often a user Ctrl-Cs an agent or interrupts with 'no, do X instead' inside the `opensre` shell — high intervention frequency is one of the strongest signals that an agent is off the rails. It's a qualitative metric the dashboard surfaces and a future input to the stuck-detection heuristic.

### What ships in this issue

Hook into the existing turn loop in `app/cli/interactive_shell/loop.py`. Increment a per-agent counter on Ctrl-C of the active agent or on follow-up messages matching negation patterns (`no,`, `stop`, `do X instead`). Counter resets per session.

Once wired into the renderer, `/agents inspect <pid>` inside the shell shows `interventions: 5 (3x Ctrl-C, 2x correction)` — high counts make the `status` cell flag the agent as struggling.

### Files to touch

- `app/cli/interactive_shell/loop.py` — call `record_intervention(active_pid, kind)` on Ctrl-C / matching follow-up
- `app/cli/interactive_shell/session.py` — store counters on `ReplSession`
- `tests/cli/interactive_shell/test_session.py` — extend

### Acceptance

- [ ] Counter visible in `/agents inspect <pid>` (a rendered cell is fine for now)
- [ ] Counter resets per session

### Where to start

Read `app/cli/interactive_shell/session.py` — `ReplSession` already tracks `token_usage`, `task_registry`, etc.; the intervention counter follows the same shape.

---

Mohdtalibakhtar:
Fixes #1498

Adds a per-kind intervention counter on `ReplSession` that tracks two signals: Ctrl-C interrupts on an active investigation (`ctrl_c`) and follow-up user messages that begin with a correction cue (`correction`). Both counters reset on `/reset` and are surfaced via a new typed method `record_intervention(kind: Literal["ctrl_c", "correction"]) -> None`. Storage is two flat ints alongside the existing `terminal_*_count` analytics fields.

This PR ships only the data plumbing. The `/agents inspect <pid>` cell that displays the counts is deferred to a follow-up once the renderer in #1488 lands (a parallel PR #1627 is already open for that renderer work, as agreed in the issue thread).

Screenshot 2026-05-08 at 1 40 27 PM

---

The Ctrl-C hook is anchored only at the active investigation cancellation site in `_run_new_alert`'s `KeyboardInterrupt` handler. The bare prompt Ctrl-C site (`_run_one_turn`'s `prompt_async` interrupt) is intentionally not counted because no agent is running there, the user is just dismissing typing. The correction hook fires before `classify_input`, so every user message is checked regardless of whether it later routes to a slash command, CLI helper, or new alert.

The regex `_INTERVENTION_CORRECTION_RE` is anchored at the start of the lstripped message and uses punctuation lookaheads on `no` / `wait` / `stop` / `instead` to suppress the obvious false positives ("stop the server", "instead of returning null", "no problem") while still catching `no,` / `actually` / `scratch that` / `let's do X instead` / `try X instead` / `nope` / `nvm`. Code-fenced messages are skipped wholesale, matching the existing `cli_agent.py` precedent.

v1 limitations called out so future iteration is informed: the cue must be at the start of the message (corrections like "hmm, scratch that" will not fire), `actually X` always counts even when the user means "actually that worked, thanks", and the cue must land within the first 80 chars after `lstrip()`. Acceptable for a v1 quality signal.

Storage is two flat ints rather than a `dict[str, int]`. Matches the `terminal_*_count` neighbours, gives mypy a clean `Literal["ctrl_c", "correction"]` dispatch, and avoids `dict.get(kind, 0) + 1` ceremony. The dict shape is one PR away if a third kind is ever added.

Considered and rejected: a stateful "correction detector" that only fires on messages immediately after a non-cancelled turn. Out of scope for an XS issue and would add another invariant on `ReplSession` for marginal accuracy gain.

---

---

Note on CI: a pre-existing failure in `tests/pipeline/test_runners_sentry.py::test_run_chat_initializes_sentry_and_captures_unhandled_errors` reproduces on `upstream/main@882abdfa` (#1620) without involvement from this PR. The local quality gate is otherwise green: 5692 passed, 1 unrelated failure.

### DIFF
--- a/app/cli/interactive_shell/loop.py
+++ b/app/cli/interactive_shell/loop.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import re
 import sys
 from collections.abc import Callable
 
@@ -29,6 +30,32 @@ from app.cli.interactive_shell.theme import TERMINAL_ERROR
 from app.cli.support.errors import OpenSREError
 from app.cli.support.exception_reporting import report_exception
 from app.cli.support.prompt_support import repl_prompt_note_ctrl_c, repl_reset_ctrl_c_gate
+
+_INTERVENTION_CORRECTION_RE = re.compile(
+    r"("
+    r"no(?=[,.!?]|$)"
+    r"|nope\b"
+    r"|nvm\b"
+    r"|nevermind\b|never\s*mind\b"
+    r"|wrong\b"
+    r"|wait(?=[,.!?]|$)"
+    r"|stop(?=[,.!?]|$)"
+    r"|actually\b"
+    r"|scratch\s+that\b"
+    r"|instead(?=[,.!?]|$)"
+    r"|(?:let'?s\s+)?do\s+[^.\n]{1,60}\s+instead\b"
+    r"|try\s+[^.\n]{1,60}\s+instead\b"
+    r")",
+    re.IGNORECASE,
+)
+
+
+def _looks_like_correction(text: str) -> bool:
+    """True when text begins with a short correction cue (intervention signal)."""
+    stripped = text.lstrip()
+    if not stripped or stripped.startswith("```"):
+        return False
+    return _INTERVENTION_CORRECTION_RE.match(stripped[:80]) is not None
 
 
 def _run_new_alert(
@@ -69,6 +96,7 @@ def _run_new_alert(
         )
     except KeyboardInterrupt:
         task.mark_cancelled()
+        session.record_intervention("ctrl_c")
         console.print("[yellow]investigation cancelled.[/yellow]")
         session.record("alert", text, ok=False)
         return
@@ -118,6 +146,9 @@ async def _run_one_turn(
     text = text.strip()
     if not text:
         return True
+
+    if _looks_like_correction(text):
+        session.record_intervention("correction")
 
     render_submitted_prompt(console, session, text)
     kind = classify_input(text, session)
--- a/app/cli/interactive_shell/loop.py
+++ b/app/cli/interactive_shell/loop.py
@@ -147,11 +147,10 @@ async def _run_one_turn(
     if not text:
         return True
 
-    if _looks_like_correction(text):
-        session.record_intervention("correction")
-
     render_submitted_prompt(console, session, text)
     kind = classify_input(text, session)
+    if kind in ("follow_up", "new_alert") and _looks_like_correction(text):
+        session.record_intervention("correction")
     if kind == "slash":
         # Rewrite bare-word commands to their slash form before dispatch.
         cmd_text = text if text.startswith("/") else f"/{text}"
--- a/app/cli/interactive_shell/session.py
+++ b/app/cli/interactive_shell/session.py
@@ -3,12 +3,14 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
 if TYPE_CHECKING:
     from prompt_toolkit.history import History
 
 from app.cli.interactive_shell.tasks import TaskRegistry
+
+InterventionKind = Literal["ctrl_c", "correction"]
 
 
 @dataclass
@@ -68,6 +70,9 @@ class ReplSession:
     terminal_actions_executed_count: int = 0
     terminal_actions_success_count: int = 0
 
+    ctrl_c_intervention_count: int = 0
+    correction_intervention_count: int = 0
+
     # Keys from a completed AgentState that carry reusable infra context into
     # the next investigation.  Kept as a class-level tuple so any caller that
     # wants to know "what counts as accumulated context" has a single source.
@@ -120,7 +125,17 @@ class ReplSession:
         self.terminal_fallback_count = 0
         self.terminal_actions_executed_count = 0
         self.terminal_actions_success_count = 0
+
+        self.ctrl_c_intervention_count = 0
+        self.correction_intervention_count = 0
         # trust_mode is intentionally preserved across /reset
+
+    def record_intervention(self, kind: InterventionKind) -> None:
+        """Increment the per-kind intervention counter (Ctrl-C or correction)."""
+        if kind == "ctrl_c":
+            self.ctrl_c_intervention_count += 1
+        else:
+            self.correction_intervention_count += 1
 
     def record_terminal_turn(
         self,
--- a/app/cli/interactive_shell/session.py
+++ b/app/cli/interactive_shell/session.py
@@ -71,7 +71,14 @@ class ReplSession:
     terminal_actions_success_count: int = 0
 
     ctrl_c_intervention_count: int = 0
+    """Incremented when the user Ctrl-Cs an active investigation. Bare-prompt
+    Ctrl-C with no agent running is intentionally not counted."""
+
     correction_intervention_count: int = 0
+    """Incremented when a follow-up or new-alert message starts with a
+    correction cue (see ``_looks_like_correction`` in ``loop.py``).
+    Slash and CLI-agent turns are not counted because content like
+    ``actually run ps aux`` is a command, not a correction."""
 
     # Keys from a completed AgentState that carry reusable infra context into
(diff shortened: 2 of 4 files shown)


yashksaini-coder (changes requested):
First reviewer here (Greptile hadn't run when I picked this up). Walked the whole diff against current `app/cli/interactive_shell/loop.py` and `app/cli/interactive_shell/session.py`. Three findings, all small fixes.

## Critical

### 1. Correction counter fires on every turn kind, not just investigation turns

`app/cli/interactive_shell/loop.py` (around lines 147-149):

```python
if _looks_like_correction(text):
    session.record_intervention("correction")

render_submitted_prompt(console, session, text)
kind = classify_input(text, session)
```

The heuristic runs **before** `classify_input`, so it fires for every turn kind: `cli_agent`, `follow_up`, `slash`, `new_alert`. A user typing `"actually run ps aux"` as a terminal command (routed to `cli_agent`) will increment `correction_intervention_count` even though no investigation is being corrected. A first-ever turn starting with `"actually"` increments the counter even when there is no prior context to correct.

If the intent is "user corrected an in-flight or just-completed investigation", scope it to follow-up / new-alert turns or to turns where prior state exists:

```python
kind = classify_input(text, session)
if kind in ("follow_up", "new_alert") and _looks_like_correction(text):
    session.record_intervention("correction")
```

### 2. Ctrl+C counter is named broadly but only covers investigation cancellation

The new instrumentation lives inside `_run_new_alert`'s `KeyboardInterrupt` handler. The pre-existing prompt-idle Ctrl+C path at `loop.py:110-113` (`except KeyboardInterrupt: if repl_prompt_note_ctrl_c(console): return False; continue`) is **not** instrumented. So a user who presses Ctrl+C repeatedly at an idle prompt without ever running an investigation produces `ctrl_c_intervention_count == 0`.

The field name `ctrl_c_intervention_count` reads as "all intentional Ctrl+C uses". Either:

- Rename to `investigation_cancel_count` (truthful and narrower), or
- Add the same `record_intervention("ctrl_c")` call in the prompt-idle handler so the counter actually counts every Ctrl+C, or
- Add a docstring on the field clarifying the scope is investigation cancellation only.

## Important

### 3. `record_intervention` bare `else` silently absorbs unknown kinds

`app/cli/interactive_shell/session.py` (around lines 104-109):

```python
def record_intervention(self, kind: InterventionKind) -> None:
    if kind == "ctrl_c":
        self.ctrl_c_intervention_count += 1
    else:
        self.correction_intervention_count += 1
```

`InterventionKind = Literal["ctrl_c", "correction"]` enforces correctness at type-check time, but at runtime the bare `else` silently bumps `correction_intervention_count` for any unrecognized string (including future kinds added without updating this method). AGENTS.md mandates strict typing — make the runtime match:

```python
elif kind == "correction":
    self.correction_intervention_count += 1
else:
    raise ValueError(f"Unknown intervention kind: {kind!r}")
```

## Cleanly handled (no issues)

- `clear()` resets both counters; covered by `test_clear_preserves_trust_mode`.
- Strict typing on `InterventionKind` and `record_intervention` matches AGENTS.md.
- Counters are in-memory only — no telemetry sink, no persistence, no PII risk.
- `_looks_like_correction` regex is anchored via `.match()`, length-bounded to 80 chars, rejects empty + code-block inputs, uses `\b` for multi-char tokens (`"nope"`, `"nvm"`) and `(?=[,.!?]|$)` lookahead for bare words (`"no"`, `"wait"`, `"stop"`). The heuristic itself is reasonable; the issue is which turns it gets applied to (Issue #1 above).
- New fields lack inline docstrings, but the pre-existing `terminal_*` fields don't have them either, so this matches local convention. Not a regression.
- Test coverage of `record_intervention` is solid (per-kind increment, kind independence, fresh-session zero state).
- CI is green: test, CodeQL, Analyze, quality all pass.

## Minor follow-ups (non-blocking)

- No direct unit test for `_looks_like_correction`. A short parametrized test pinning a few positive (`"actually use ec2"`, `"wait, i meant rds"`) and negative (`"investigate the actual cause"`, `""`, code-block input) cases would close the gap.
- Issue #1495 (referenced in PR title) — couldn't locate that issue text from the PR body alone. Worth confirming the spec matches the implemented semantics, especially around "intervention" definition.

## Cross-checks performed

- Walked `_looks_like_correction` regex against a few prompt strings by hand.
- Verified the prompt-idle Ctrl+C path at `loop.py:110-113` is unchanged and untracked.
- Confirmed `clear()` resets both counters.
- Compared field-doc convention against the pre-existing `terminal_*` fields.
- Verified CI checks all pass on the head commit.

## Verdict

Request changes. None of the three items are large fixes — combined they're under 10 lines of code plus one rename or docstring. With them in, this is a clean APPROVE.

Mohdtalibakhtar:
Pushed `eee9a7ed` addressing all three.

1. Correction counter now fires only when `kind in ("follow_up", "new_alert")`, after `classify_input`. Agree the wide scope was inflating on `cli_agent` commands like `actually run ps aux` — clean call, the fix is a five-line reorder with no new state on `ReplSession`.

2. `ctrl_c_intervention_count` now carries a docstring clarifying that the scope is investigation cancellation only; bare-prompt Ctrl-C with no agent running is intentionally not counted. `correction_intervention_count` got a parallel docstring naming the gating. Kept the field name and the `Literal` kind name (`"ctrl_c"`) — they match the user-visible action. Open to renaming both to `investigation_cancel` if you'd rather have the narrower name.

3. `record_intervention` now raises `ValueError` on unknown kinds, with an explicit `elif kind == "correction"` branch. Strict-typing rule from `AGENTS.md` now matches at runtime.

Plus, on the non-blocking item: added `TestLooksLikeCorrection` in `tests/cli/interactive_shell/test_loop.py` — 22 parametrized cases covering each cue branch, each false-positive guard, the documented v1 limitations (mid-message cue, code-fenced input), and edge cases (empty, whitespace-only). Closes the regex-coverage gap Greptile flagged.

Local quality gate: 5764 passed, 6 skipped. Diff: 3 files, +61/-4.

@greptile review

yashksaini-coder (approved):
Re-reviewed against `eee9a7ed`. All three items from my prior CHANGES_REQUESTED are addressed cleanly.

## Findings status

1. **Correction counter scope** — FIXED. The guard now reads `if kind in ("follow_up", "new_alert") and _looks_like_correction(text)` and runs after `classify_input`. Slash, cli_help, and cli_agent turns no longer reach the counter.
2. **Ctrl+C counter naming asymmetry** — FIXED. You chose to document rather than rename. The new docstring on `ctrl_c_intervention_count` explicitly states "Bare-prompt Ctrl+C with no agent running is intentionally not counted." That's exactly the kind of inline note that prevents future-reader confusion. `correction_intervention_count` got the same treatment.
3. **Bare `else` in `record_intervention`** — FIXED. Now `if kind == "ctrl_c": ... elif kind == "correction": ... else: raise ValueError(...)`. Combined with the `InterventionKind = Literal["ctrl_c", "correction"]` alias, both runtime and static type-check paths are tight.

## Cross-checks

- Read the full diff end-to-end and confirmed each fix maps to my prior asks.
- Verified `_run_new_alert`'s `except KeyboardInterrupt` block is still the only cancellation path for a running investigation.
- Confirmed the `_repl_main` initial-input pre-seeded loop (test harness path) doesn't bump counters — consistent with it being non-interactive.
- New test `record_intervention_increments_per_kind` plus the `ValueError` path covers the runtime guard.

## One minor heuristic observation (not blocking)

`_looks_like_correction` matches `"no, that's wrong"` (lookahead requires `[,.!?]` after "no") but not `"no that's wrong"`. That's intentional per the test-suite "v1 limitation" comment, so I'm not requesting a change — just noting for any future tightening pass.

Approving.

PR #1635, opened 2026-05-08
[WIP] test(services): optional chat_sdk_adapter Anthropic/OpenAI live smokes

### LINKED ISSUE #1634: Add optional live smoke tests for chat_sdk_adapter (Anthropic + OpenAI)
## Summary

Add optional integration-style tests that call the real **Anthropic Messages** and **OpenAI Chat Completions** APIs through `build_bound_chat_model` / `chat_sdk_adapter`, guarded by API keys (tests skip when keys are absent).

## Files

- `tests/services/test_chat_sdk_adapter_live_anthropic.py` — default toolcall model from config unless `CHAT_SDK_LIVE_ANTHROPIC_MODEL` is set.
- `tests/services/test_chat_sdk_adapter_live_openai.py` — default `gpt-4o-mini` unless `CHAT_SDK_LIVE_OPENAI_MODEL` is set.

Mocked coverage remains in `tests/services/test_chat_sdk_adapter_comprehensive.py`.

## Run (local)

```bash
export ANTHROPIC_API_KEY=...
uv run pytest tests/services/test_chat_sdk_adapter_live_anthropic.py -v

export OPENAI_API_KEY=...
uv run pytest tests/services/test_chat_sdk_adapter_live_openai.py -v
```

## Follow-ups

- Pytest marker (e.g. `live_vendor`) + exclude from default `make test-cov` / CI to reduce skip noise.
- Optional Makefile target for discoverability.

Tracking WIP PR: _(will be filled after open)._

muddlebee:
## Summary
Optional live API smokes for `chat_sdk_adapter` (skip when keys missing).

## Tests
- **Anthropic:** pong, system+user, `HumanMessage`, multi-turn math recall, system-only guard (no API).
- **OpenAI:** pong, `HumanMessage` bridge.

Refs `test_chat_sdk_adapter_comprehensive.py` for mocked coverage.

## Base
Branch is based on **`upstream/main`** (Tracer-Cloud).

Fixes #1634

## Run
```bash
export ANTHROPIC_API_KEY=...
uv run pytest tests/services/test_chat_sdk_adapter_live_anthropic.py -v

export OPENAI_API_KEY=...
uv run pytest tests/services/test_chat_sdk_adapter_live_openai.py -v
```

### DIFF
--- a/tests/services/test_chat_sdk_adapter_live_anthropic.py
+++ b/tests/services/test_chat_sdk_adapter_live_anthropic.py
@@ -1,0 +1,153 @@
+"""Optional live Anthropic Messages API smoke for ``chat_sdk_adapter``.
+
+Run when ``ANTHROPIC_API_KEY`` is set::
+
+    uv run pytest tests/services/test_chat_sdk_adapter_live_anthropic.py -v
+
+Model default: repo cheap toolcall default (``ANTHROPIC_LLM_CONFIG.toolcall_model``).
+Override with ``CHAT_SDK_LIVE_ANTHROPIC_MODEL``.
+
+Scenarios: plain text, system+user split, LangChain human message, short multi-turn
+recall, and the empty-after-system guard (no billable API call for the last one).
+"""
+
+from __future__ import annotations
+
+import os
+import re
+
+import pytest
+
+from app.config import ANTHROPIC_LLM_CONFIG
+from app.services.chat_sdk_adapter import build_bound_chat_model
+
+
+def _anthropic_api_key() -> str:
+    return (os.getenv("ANTHROPIC_API_KEY") or "").strip()
+
+
+def _live_model_name() -> str:
+    override = (os.getenv("CHAT_SDK_LIVE_ANTHROPIC_MODEL") or "").strip()
+    if override:
+        return override
+    return ANTHROPIC_LLM_CONFIG.toolcall_model
+
+
+@pytest.fixture(scope="module")
+def require_anthropic_key() -> str:
+    key = _anthropic_api_key()
+    if not key:
+        pytest.skip("Set ANTHROPIC_API_KEY (optional CHAT_SDK_LIVE_ANTHROPIC_MODEL for model id).")
+    return key
+
+
+def test_live_anthropic_plain_assistant_text(
+    require_anthropic_key: str,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    monkeypatch.setenv("ANTHROPIC_API_KEY", require_anthropic_key)
+    adapter = build_bound_chat_model(
+        provider="anthropic",
+        model_name=_live_model_name(),
+        with_tools=False,
+    )
+    turn = adapter.invoke(
+        [
+            {
+                "role": "user",
+                "content": (
+                    "Reply with exactly one line containing only the word pong "
+                    "(lowercase, no punctuation)."
+                ),
+            }
+        ]
+    )
+    content = (turn.get("content") or "").strip().lower()
+    assert content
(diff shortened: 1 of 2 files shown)


yashksaini-coder (commented):
Treating this as feedback on a draft (the [WIP] tag is in the title). One real blocker for when you mark this ready, plus a few smaller items.

## Blocker before ready: live tests land in `tests/services/` with no exclusion marker

The `make test-cov` command runs:

```
pytest -n auto ... --ignore=tests/e2e/kubernetes_local_alert_simulation --ignore=tests/synthetic -m "not synthetic"
```

`tests/services/` is fully in scope. `pytest.ini` `norecursedirs` excludes `tests/e2e` and `tests/deployment`, but not `tests/services`. So both new files (`test_chat_sdk_adapter_live_anthropic.py`, `test_chat_sdk_adapter_live_openai.py`) will be collected on every `make test-cov` run.

The `pytest.skip(...)` inside the fixture handles the missing-key case at runtime — but this is a soft guard, not a hard exclusion. AGENTS.md rule is explicit: "If CI-only tests are added, mark them with the right pytest marker or place them in the appropriate e2e/synthetic/chaos folder so they do not run in the default local suite."

Two equally good options:

1. Add `pytestmark = pytest.mark.e2e` at the top of both files (the `e2e` marker exists in `pytest.ini`). Then `make test-cov` can filter with `-m "not e2e"`.
2. Move the files into `tests/e2e/services/` (already in `norecursedirs`).

Either satisfies the AGENTS.md rule.

## Smaller items

### `require_*_key` fixture leaks the raw key through the parameter

The fixture returns the raw key value and passes it back to each test via the parameter name `require_anthropic_key: str`. The parameter is never used as a value after the `monkeypatch.setenv` call (the SDK reads from env). Rename to `_live_anthropic_enabled` and have it return `None`, so the parameter is unmistakably an opt-in sentinel rather than a key-bearing string. Avoids accidental `print(require_anthropic_key)` showing up in any debug output.

### No `max_tokens` budget

None of the `build_bound_chat_model` calls pass a token cap. The prompts are already tight ("reply with exactly one word") so the model stops early in practice. But if `ANTHROPIC_LLM_CONFIG.toolcall_model` ever changes to a more expensive model, a single live run is uncapped. Pass `max_tokens=64` or similar — explicit intent and protected against future model changes.

### Coverage gap worth flagging in the PR description

All Anthropic tests use `with_tools=False`. OpenAI has only two tests, both without tools. The primary reason for having live tests at all is to catch wire-level behavior that mocks miss. The two scenarios most likely to break silently — tool-call round trips and streaming responses — are unexercised. Worth either adding them or explicitly calling out as a known gap in the PR body.

### Typo: `gpt-5.4-mini`

OpenAI file's module docstring example: `export CHAT_SDK_LIVE_OPENAI_MODEL=gpt-5.4-mini`. That model doesn't exist. Likely meant `gpt-4o-mini` or just a placeholder.

## What's already done well

- Env-var skip guard via `pytest.skip(...)` matches how other optional-credentials tests in the repo handle the missing-key case.
- `pytest.importorskip("langchain_core")` before the LangChain-bridge tests is correct defensive practice.
- The "empty-after-system guard" test in the Anthropic file specifically validates pre-API-call behaviour with no billable round trip — exactly the right kind of live test to include.
- Prompts are minimal and roughly deterministic, keeping token consumption low.

## Verdict

COMMENT (matching the draft state). The pytest-marker fix is the only blocker before marking this ready. Once that's in plus the small cleanups, this is a useful and well-shaped live-test suite.

muddlebee:
@yashksaini-coder its still WIP lol. Im yet to figure out.

davincios:
Closing due to inactivity with relatively small change

PR #1643, opened 2026-05-08
refactor(shell): register grounding sources

### LINKED ISSUE #1447: REPL: pluggable grounding diagnostics so /status auto-discovers cache stats sources
## Background

The interactive shell exposes grounding cache stats in two places:

1. The `/status` slash command in [`app/cli/interactive_shell/command_registry/session_cmds.py`](app/cli/interactive_shell/command_registry/session_cmds.py) (`_cmd_status`, lines 39–66) — hard-codes calls to `get_cli_reference_cache_stats()` and `get_docs_cache_stats()`.
2. Verbose-mode logging in [`app/cli/interactive_shell/grounding_diagnostics.py`](app/cli/interactive_shell/grounding_diagnostics.py) — same hard-coded list.

Both will need to grow as new grounding sources are added (AGENTS.md grounding, source-tree grounding, tool-catalog grounding, etc., each tracked in its own issue). Today every new source requires editing two unrelated files. This refactor introduces a tiny registration API so a grounding source can self-register at import time and `/status` + verbose logging both pick it up automatically.

This change is independent: it does not require any of the other issues to be merged first, and the existing two grounding sources stay registered with identical observable behavior.

## Scope

In `app/cli/interactive_shell/grounding_diagnostics.py`:

- Define a tiny `GroundingSource` dataclass: `name: str`, `stats_fn: Callable[[], dict[str, Any]]`, `format_fn: Callable[[dict[str, Any]], str]`. Provide a sensible default `format_fn` shared between `cli` (`hits=… misses=… cached=…`) and `docs` (`hits=… misses=… entries=…/…`).
- Add `register_grounding_source(source: GroundingSource) -> None` and `iter_grounding_sources() -> Iterable[GroundingSource]` operating on a module-level registry (insertion-ordered).
- Update `log_grounding_cache_diagnostics(reason)` to iterate the registry instead of importing `cli_reference` and `docs_reference` directly.

Have the existing modules self-register at import time (one-line addition to each):

- [`app/cli/interactive_shell/cli_reference.py`](app/cli/interactive_shell/cli_reference.py) registers a `GroundingSource(name="cli", …)`.
- [`app/cli/interactive_shell/docs_reference.py`](app/cli/interactive_shell/docs_reference.py) registers a `GroundingSource(name="docs", …)`.

Update `_cmd_status` in [`session_cmds.py`](app/cli/interactive_shell/command_registry/session_cmds.py) to render rows from `iter_grounding_sources()` instead of two hard-coded lookups. Output text must remain a superset of the current rows so existing tests in `tests/cli/interactive_shell/test_command_registry.py` still pass.

Make sure registration is idempotent (calling `register_grounding_source` twice with the same name updates in place rather than duplicating a row).

## Acceptance criteria

- [ ] New `GroundingSource`, `register_grounding_source`, `iter_grounding_sources` in `grounding_diagnostics.py` (with `__all__` updated)
- [ ] `cli_reference.py` and `docs_reference.py` self-register; no behavior change to their existing public APIs
- [ ] `/status` renders grounding rows from the registry, preserving the current row format and order (cli first, docs second)
- [ ] New tests in `tests/cli/interactive_shell/test_grounding_diagnostics.py` cover: registration, idempotent re-registration, iteration order, and the verbose-log output uses the registry
- [ ] Existing `/status` and verbose-mode tests still pass unchanged
- [ ] `make lint`, `make format-check`, `make typecheck`, `make test-cov` all pass

## References

- Current state: [`app/cli/interactive_shell/grounding_diagnostics.py`](app/cli/interactive_shell/grounding_diagnostics.py)
- Hard-coded `/status` rows: [`app/cli/interactive_shell/command_registry/session_cmds.py`](app/cli/interactive_shell/command_registry/session_cmds.py) lines 39–66
- Existing stats functions: `get_cli_reference_cache_stats()` in `cli_reference.py`, `get_docs_cache_stats()` in `docs_reference.py`

## Out of scope

- Adding any new grounding source (covered by separate issues)
- Changing the cache implementations themselves
- Token-usage / cost reporting (different concern, lives under `/cost`)

cloudenochcsis:
Fixes #1447

- added a shared `GroundingSource` registry in the interactive shell grounding diagnostics module
- self-registered the existing CLI help, docs, and `AGENTS.md` grounding cache stats providers
- updated `/status` to render grounding cache rows from the registry while preserving the existing CLI/docs rows
- kept verbose grounding diagnostics on the raw builtin stats payload shape so the debug output still includes the existing dict fields
- added coverage for builtin source registration/order, idempotent re-registration, registry-driven `/status`, and verbose logging through the registry

```text
$ make lint && make format-check && make typecheck && make test-cov
...
5803 passed, 5 skipped, 165 warnings in 80.84s
```

---

I kept the refactor narrow around the two current grounding call sites. `grounding_diagnostics.py` now owns a small registration surface so each grounding source can register itself with its stats callback once, and both verbose diagnostics and `/status` iterate that same registry. For the user-facing `/status` rows I used a shared formatter that preserves the current `cli` and `docs` text while allowing additional sources to render naturally; for verbose logging I kept the raw builtin payload shape so the existing debug fields are still present. The tests pin the public registry behavior (builtin sources, order, idempotent re-registration) and the two caller surfaces so adding future grounding sources only requires self-registration.

---

---

### DIFF
--- a/app/cli/interactive_shell/agents_md_reference.py
+++ b/app/cli/interactive_shell/agents_md_reference.py
@@ -260,6 +260,18 @@ def build_agents_md_reference_text(*, max_chars: int = _DEFAULT_MAX_TOTAL_CHARS)
     return text
 
 
+def _register_grounding_source() -> None:
+    from app.cli.interactive_shell.grounding_diagnostics import (
+        GroundingSource,
+        register_grounding_source,
+    )
+
+    register_grounding_source(GroundingSource(name="agents_md", stats_fn=get_agents_md_cache_stats))
+
+
+_register_grounding_source()
+
+
 __all__ = [
     "AgentsMdFile",
     "build_agents_md_reference_text",
--- a/app/cli/interactive_shell/cli_reference.py
+++ b/app/cli/interactive_shell/cli_reference.py
@@ -156,6 +156,18 @@ def build_cli_reference_text() -> str:
     return text
 
 
+def _register_grounding_source() -> None:
+    from app.cli.interactive_shell.grounding_diagnostics import (
+        GroundingSource,
+        register_grounding_source,
+    )
+
+    register_grounding_source(GroundingSource(name="cli", stats_fn=get_cli_reference_cache_stats))
+
+
+_register_grounding_source()
+
+
 __all__ = [
     "build_cli_reference_text",
     "get_cli_reference_cache_stats",
--- a/app/cli/interactive_shell/command_registry/session_cmds.py
+++ b/app/cli/interactive_shell/command_registry/session_cmds.py
@@ -37,8 +37,7 @@ def _cmd_trust(session: ReplSession, console: Console, args: list[str]) -> bool:
 
 
 def _cmd_status(session: ReplSession, console: Console, _args: list[str]) -> bool:
-    from app.cli.interactive_shell.cli_reference import get_cli_reference_cache_stats
-    from app.cli.interactive_shell.docs_reference import get_docs_cache_stats
+    from app.cli.interactive_shell.grounding_diagnostics import iter_grounding_sources
 
     table = repl_table(title="Session status", title_style=TERMINAL_ACCENT_BOLD, show_header=False)
     table.add_column("key", style="bold")
@@ -47,18 +46,8 @@ def _cmd_status(session: ReplSession, console: Console, _args: list[str]) -> boo
     table.add_row("last investigation", "yes" if session.last_state else "none")
     table.add_row("trust mode", "on" if session.trust_mode else "off")
     table.add_row("provider", os.getenv("LLM_PROVIDER", "anthropic"))
-    cli_stats = get_cli_reference_cache_stats()
-    doc_stats = get_docs_cache_stats()
-    table.add_row(
-        "grounding cli cache",
-        f"hits={cli_stats['hits']} misses={cli_stats['misses']} "
-        f"cached={'yes' if cli_stats['cached'] else 'no'}",
-    )
-    table.add_row(
-        "grounding docs cache",
-        f"hits={doc_stats['hits']} misses={doc_stats['misses']} "
-        f"entries={doc_stats['currsize']}/{doc_stats['maxsize']}",
-    )
+    for source in iter_grounding_sources():
+        table.add_row(f"grounding {source.name} cache", source.format_fn(source.stats_fn()))
     acc = session.accumulated_context
     if acc:
         table.add_row("accumulated context", ", ".join(sorted(acc.keys())))
--- a/app/cli/interactive_shell/docs_reference.py
+++ b/app/cli/interactive_shell/docs_reference.py
@@ -451,6 +451,18 @@ def build_docs_reference_text(
     return text
 
 
+def _register_grounding_source() -> None:
+    from app.cli.interactive_shell.grounding_diagnostics import (
+        GroundingSource,
+        register_grounding_source,
+    )
+
+    register_grounding_source(GroundingSource(name="docs", stats_fn=get_docs_cache_stats))
+
+
+_register_grounding_source()
+
+
 __all__ = [
     "DocPage",
     "build_docs_index",
--- a/app/cli/interactive_shell/grounding_diagnostics.py
+++ b/app/cli/interactive_shell/grounding_diagnostics.py
@@ -4,25 +4,66 @@ from __future__ import annotations
 
 import logging
 import os
+from collections import OrderedDict
+from collections.abc import Callable, Iterable
+from dataclasses import dataclass
+from typing import Any
 
 _logger = logging.getLogger(__name__)
+
+_GROUNDING_SOURCE_REGISTRY: OrderedDict[str, GroundingSource] = OrderedDict()
+
+
+def _format_grounding_stats(stats: dict[str, Any]) -> str:
+    hits = stats.get("hits", 0)
+    misses = stats.get("misses", 0)
+    if "cached" in stats:
+        return f"hits={hits} misses={misses} cached={'yes' if stats['cached'] else 'no'}"
+    if "currsize" in stats and "maxsize" in stats:
+        return f"hits={hits} misses={misses} entries={stats['currsize']}/{stats['maxsize']}"
+    return str(stats)
+
+
+@dataclass(frozen=True)
+class GroundingSource:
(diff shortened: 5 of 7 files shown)


yashksaini-coder (changes requested):
Walked the diff carefully and traced through Python's import-time sequence for the new registration shape. Greptile's 3/5 confidence is generous — the structural concerns are real and the test-fixture bug is non-deterministic, which is the worst class of test failure to chase down later.

## Critical

### 1. Forward reference: `_GROUNDING_SOURCE_REGISTRY` annotated with `GroundingSource` before the class is defined

`app/cli/interactive_shell/grounding_diagnostics.py` (around diff line 120):

```python
_GROUNDING_SOURCE_REGISTRY: OrderedDict[str, GroundingSource] = OrderedDict()
```

This sits **before** `@dataclass class GroundingSource:` (around diff line 133). `from __future__ import annotations` is present so the string annotation lazy-resolves and doesn't crash at import — but anything that evaluates annotations at runtime (`get_type_hints()`, pydantic, mypy strict mode, IDE typecheckers) will fail with `NameError: name 'GroundingSource' is not defined`.

Fix: move the `GroundingSource` dataclass definition above the registry variable. Trivial reordering.

### 2. Cyclic imports — Greptile's 6 CodeQL findings are CONFIRMED structurally

The post-PR import graph:

- `grounding_diagnostics.py`'s `_ensure_builtin_grounding_sources_registered()` does deferred imports of `cli_reference`, `docs_reference`, `agents_md_reference`.
- Each of those three modules executes `_register_grounding_source()` at module load, which deferred-imports `grounding_diagnostics`.

This works **today** with the current entry points because Python returns partially-initialised modules from `sys.modules` rather than raising `ImportError`. But:

- Any change in import order silently breaks one direction.
- Any tool that pre-loads `cli_reference` before `grounding_diagnostics` (analysers, IDE features, lazy-import linters) hits the half-initialised state.
- Once a module body has run, Python won't re-execute it — so a re-import after registry clearing produces silently empty registration. (This is exactly the bug in #3 below.)

The clean fix is push-based registration with no back-imports:

- `grounding_diagnostics.py` owns the registry and exposes `register_grounding_source(...)`. It imports **nothing** from the three reference modules.
- `cli_reference.py`, `docs_reference.py`, `agents_md_reference.py` each call `register_grounding_source(...)` on import (deferred import of `grounding_diagnostics` only).
- Remove `_ensure_builtin_grounding_sources_registered()` entirely. The shell's existing loader already imports the three reference modules before any diagnostic call.

This eliminates all 6 cyclic edges at once and removes the import-order coupling.

### 3. `restore_grounding_source_registry` fixture leaves builtins unregistered on subsequent tests

Both `tests/cli/interactive_shell/test_commands.py` (around lines 207-214) and `tests/cli/interactive_shell/test_grounding_diagnostics.py` (around lines 263-269) define an `autouse=True` fixture that snapshots and restores `_GROUNDING_SOURCE_REGISTRY`. The snapshot is taken at fixture setup time.

Failure mode: if `grounding_diagnostics` was imported first in the test session before any of the three reference modules registered, the snapshot is empty. The fixture restores to empty after each test. The next call to `iter_grounding_sources()` triggers `_ensure_builtin_grounding_sources_registered()`, which imports `cli_reference` etc. — but Python doesn't re-run module bodies on re-import, so the `_register_grounding_source()` call inside those modules never fires again. The registry stays empty.

`test_iter_grounding_sources_includes_builtin_sources_in_order` directly asserts builtins are present, so it will fail non-deterministically depending on test ordering.

Fix: the fixture should explicitly re-register the known builtins on teardown by calling each `_register_grounding_source()` function directly (not relying on import side effects). Once #2 is in (push-based registration), this becomes easier — the builtins can be registered idempotently from a single `register_default_grounding_sources()` helper.

## Important

### 4. Duplicate import style flagged by CodeQL

Both test files import `grounding_diagnostics` two ways:

```python
import app.cli.interactive_shell.grounding_diagnostics as grounding_diagnostics
from app.cli.interactive_shell.grounding_diagnostics import (
    GroundingSource,
    iter_grounding_sources,
    ...
)
```

The alias is needed only to access `_GROUNDING_SOURCE_REGISTRY` directly for fixture manipulation. After fix #2, the fixture can be rewritten to call public APIs only, dropping the alias and the dual-import warning together.

## Cross-checks performed

- Verified `from __future__ import annotations` is on `grounding_diagnostics.py` (so the forward-ref annotation is lazy-resolved).
- Traced import order: `session_cmds.py` -> `grounding_diagnostics` -> `_ensure_builtin_grounding_sources_registered` -> `cli_reference` -> `_register_grounding_source` -> deferred `grounding_diagnostics` (already partially loaded). Resolves only because `register_grounding_source` is defined before `_ensure_builtin_grounding_sources_registered` runs.
- Verified that re-importing a module already in `sys.modules` doesn't re-run its body (standard Python semantics).
- Confirmed `restore_grounding_source_registry` snapshot logic and the fixture teardown's reliance on `_GROUNDING_SOURCE_REGISTRY.update(original_registry)`.

## Verdict

Request changes — fix #2 (push-based registration without back-imports) is the architectural change that closes findings #2, most of #3, and #4 simultaneously. Fix #1 is a one-line reorder. After those, this is a clean APPROVE.

cloudenochcsis:
Pushed `9f37115b` addressing the review.

- **#1** `GroundingSource` dataclass moved above `_GROUNDING_SOURCE_REGISTRY` (no more forward-ref).
- **#2** Deleted `_ensure_builtin_grounding_sources_registered`; `grounding_diagnostics` no longer imports the three reference modules. All six CodeQL cyclic-import edges should clear.
- **#3** `restore_grounding_source_registry` rewritten in both test files: snapshots names, removes test-added sources via a new public `unregister_grounding_source`, then re-registers builtins by calling each module's register function directly (no reliance on import side effects).
- **#4** Dropped the dual `import` / `from … import` pattern in both test files; the fixture now uses public APIs only.

Verified: 95/95 tests green, lint/format/mypy clean. Tests pass under reversed file order and under the two specific pairs flagged in the review.

@yashksaini-coder ready for another pass when you have a moment.
@greptile review

muddlebee:
Demo is missing. Provide an e2e demo.

muddlebee:
closing this PR as the outcome is not satisfactory, proper demo is missing and reviews not catered nicely

PR #1653, opened 2026-05-08
fix(cli): use in_thread=True for questionary prompts inside async REPL

### LINKED ISSUE #1650: RuntimeError: asyncio.run() cannot be called from a running event loop
Sentry Issue: [PYTHON-16](https://tracer-30.sentry.io/issues/7467556569/)

```
RuntimeError: asyncio.run() cannot be called from a running event loop
(3 additional frame(s) were not displayed)
...
  File "app/cli/interactive_shell/loop.py", line 128, in _run_one_turn
    should_continue = dispatch_slash(cmd_text, session, console)
  File "app/cli/interactive_shell/command_registry/__init__.py", line 132, in dispatch_slash
    return cmd.handler(session, console, args)
  File "app/cli/interactive_shell/command_registry/model.py", line 247, in _cmd_model
    return run_model_interactive_picker(session, console)
  File "app/cli/interactive_shell/command_registry/model_interactive.py", line 43, in run_model_interactive_picker
    picked_pv = repl_choose_one(
  File "app/cli/interactive_shell/repl_choice_menu.py", line 116, in repl_choose_one
    ).unsafe_ask()
```

davincios:
## Summary

- `questionary.Question.unsafe_ask()` calls `application.run()` which calls `asyncio.run()` internally (via prompt_toolkit)
- The REPL runs inside `asyncio.run(_repl_main())`, so any questionary prompt dispatched from a slash-command handler crashes with `RuntimeError: asyncio.run() cannot be called from a running event loop`
- Fix: detect a running event loop via `asyncio.get_running_loop()` and call `application.run(in_thread=True)` — prompt_toolkit creates its own loop in a background thread, avoiding the nesting error
- Regression test added that calls `q.ask()` from inside `asyncio.run()` and asserts no error

## Test plan

- [x] `uv run pytest tests/cli/test_prompt_support.py -v` — all 10 tests pass including the new regression test
- [x] `ruff check` — clean
- [x] `mypy` — clean

Closes #1650

https://claude.ai/code/session_01CUrEtqvBEQtxVc1aytSLUz

---
_Generated by [Claude Code](https://claude.ai/code/session_01CUrEtqvBEQtxVc1aytSLUz)_

### DIFF
--- a/app/cli/support/prompt_support.py
+++ b/app/cli/support/prompt_support.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import sys
 import time
 from collections.abc import Callable
@@ -114,9 +115,20 @@ def _with_ctrl_c_double_exit(
     def _patched_ask(*args: Any, **kwargs: Any) -> Any:
         while True:
             try:
-                # unsafe_ask() propagates KeyboardInterrupt instead of
-                # swallowing it the way ask() does.
-                result = question.unsafe_ask(*args, **kwargs)
+                # unsafe_ask() → application.run() → asyncio.run(), which
+                # raises RuntimeError when called from inside a running event
+                # loop (e.g. the async REPL). Use in_thread=True so
+                # prompt_toolkit creates its own event loop in a background
+                # thread instead.
+                try:
+                    asyncio.get_running_loop()
+                    _in_event_loop = True
+                except RuntimeError:
+                    _in_event_loop = False
+                if _in_event_loop:
+                    result = question.application.run(in_thread=True)
+                else:
+                    result = question.unsafe_ask(*args, **kwargs)
                 _last_ctrl_c[0] = None  # reset on clean exit so next prompt starts fresh
                 return result
             except KeyboardInterrupt as exc:
--- a/tests/cli/test_prompt_support.py
+++ b/tests/cli/test_prompt_support.py
@@ -139,3 +139,30 @@ def test_ctrl_c_hint_resets_after_window(capsys) -> None:
     handle_ctrl_c_press()
     out = capsys.readouterr().out
     assert "(Press Ctrl+C again to exit)" in out
+
+
+def test_questionary_ask_inside_running_event_loop_does_not_raise() -> None:
+    """q.ask() called from within a running asyncio event loop must not raise.
+
+    Regression test for Sentry issue #1650: asyncio.run() cannot be called
+    from a running event loop — triggered when questionary prompts are shown
+    inside the async REPL dispatch path.
+    """
+    import asyncio
+
+    _last_ctrl_c[0] = None
+    install_questionary_ctrl_c_double_exit()
+
+    async def _run() -> object:
+        with create_pipe_input() as pipe_input:
+            q = questionary.select(
+                "Pick",
+                choices=["a", "b"],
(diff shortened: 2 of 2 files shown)


yashksaini-coder (approved):
Surgical fix; the event-loop detection and `Application.run(in_thread=True)` are exactly the right primitives. Approving with a couple of follow-up notes.

## What's solid

- **`asyncio.get_running_loop()` with `try/except RuntimeError`** is the documented public idiom (stable since 3.7). No private-API touches.
- **`Application.run(in_thread=True)`** is prompt_toolkit's documented escape hatch for nested-event-loop cases (their own docs cite ptpython for exactly this).
- **Strictly additive**: when no event loop is running, `_in_event_loop` is `False`, falls through to the original `question.unsafe_ask(*args, **kwargs)` unchanged.
- **Regression test** runs `q.ask()` from inside `asyncio.run()` with `create_pipe_input` + `DummyOutput`. Will fail loudly on a regressed `asyncio.run()` nesting error.
- The comment in the diff accurately spells out the root cause: `unsafe_ask()` -> `application.run()` -> `asyncio.run()` raising when called from inside a running event loop.

## Behavioral note worth a follow-up commit (not blocking)

### `should_skip_question` is bypassed in the async branch

The original `unsafe_ask()` checks `self.should_skip_question` and returns `self.default` early. The new async branch calls `question.application.run(in_thread=True)` directly, skipping that guard.

In practice, this isn't currently exploitable — I scanned all `questionary.*.ask()` call sites under `app/cli/` and none use `.skip_if(...)`. But it's a silent contract divergence. One-line fix:

```python
if _in_event_loop:
    if question.should_skip_question:
        return question.default
    result = question.application.run(in_thread=True)
```

Worth tightening if you push another revision.

### Ctrl+C path mechanism differs (still functionally correct)

`Application.run(in_thread=True)` internally passes `handle_sigint=False` to the background thread, so the application's own Ctrl+C key binding doesn't fire. SIGINT arrives in the main thread, prompt_toolkit re-raises it via the thread join, the `KeyboardInterrupt` propagates to your `_patched_ask` retry loop, and the double-exit guard fires. Functionally correct, but the SIGINT source is now indirect.

A one-line comment in the code near `in_thread=True` ("prompt_toolkit handles SIGINT via thread-join re-raise; our retry loop sees the same KeyboardInterrupt") would help any future reader who wonders why `handle_sigint=False` is acceptable. Greptile flagged this as an untested gap; that's accurate, but the test would be hard to write deterministically.

## Nits

- `import asyncio` was added at the module top level, replacing the earlier inline import in callers. Consistent with the rest of the file's import style. Good.
- The test function has a redundant inline `import asyncio` (already at the top of the test file). Harmless, but inconsistent with the rest of the file's import header.

## Cross-checks

- Confirmed `question.application` is a documented public attribute on `questionary.Question`.
- Confirmed `Application.run(in_thread=True)` signature in prompt_toolkit source — matches the diff's usage.
- Verified the async REPL runs under `asyncio.run(_repl_main(...))` in `loop.py:250`, so the trigger condition is reachable in production.
- Scanned all `questionary.*.ask()` call sites in `app/cli/` for `.skip_if()` — none, so the `should_skip_question` bypass is not currently exploitable but is a latent contract divergence.

Approving.

PR #1654, opened 2026-05-08
feat(cli): TTY inline pickers for /list, /integrations, /mcp, and more

muddlebee:
## Summary

Adds the same inline TTY menu pattern as `/model` for additional slash commands when invoked with no arguments on a TTY.

[Screencast from 08-05-26 05:17:51 PM IST.webm](https://github.com/user-attachments/assets/3f9229c2-a0ff-4177-8901-c008e455c89f)

## Changes

- New `repl_choice_menu.py`: shared inline picker, breadcrumbs, erase-on-exit, `repl_section_break` between menu steps and Rich output
- **`/list`**: loop over integrations / models / mcp / tools / all / done
- **`/integrations`**: list / verify / show / setup / remove (verified services for show/remove)
- **`/mcp`**: list / connect / disconnect (disconnect uses `MCP_INTEGRATION_SERVICES`)
- **`/template`**, **`/trust`**, **`/verbose`**: on/off or template picks as appropriate
- **`/history`**: show / clear / off / on / retention (with cap picker)
- `rendering.py`: export `MCP_INTEGRATION_SERVICES`; `theme.py` menu styling
- Tests extended in `tests/cli/interactive_shell/test_commands.py`

## Quality

- `make lint`, `make format-check`, `make typecheck` pass
- `pytest tests/cli/interactive_shell/test_commands.py`: 92 passed

Fixes # (none linked — replace if an issue exists)

---

See Summary above.

---

Reuse one menu primitive so behavior stays consistent with `/model`: keyboard navigation, visual separation before command output, and non-TTY paths unchanged. Integration/MCP menus pull choices from existing verified-integration and MCP service lists to avoid drift.

---


muddlebee:
@greptile review

yashksaini-coder (changes requested):
Walked through the new `repl_choose_one` plumbing and the multi-step `/model` flow against the test fixture. The shape is good and the TTY-degrade-to-arg-path guard via `repl_tty_interactive()` is the right call. Several real issues to address.

## Greptile findings — verified

1. **Missing "tools" branch in `_interactive_list_menu`** — FIXED. The diff at integrations.py lines 71-76 correctly handles `sub == "tools"` with `build_tool_catalog()` + `format_tool_catalog_text(catalog)`. Greptile's report was stale.
2. **`_CRUMB_SEP` exported despite leading underscore** — FALSE POSITIVE. The actual symbol is `CRUMB_SEP` (no leading underscore), all-caps public name, and is correctly in `__all__`. The leading-underscore symbol `_HINT` is correctly NOT in `__all__`.
3. **CodeQL `toolcall_choice` unused** — false positive. The variable IS used in the control-flow path (set in the toolcall block, read by `switch_llm_provider` at the end).

## Critical (the real bugs)

### 1. `_interactive_set_provider` — Esc at toolcall picker discards confirmed reasoning model

`app/cli/interactive_shell/command_registry/model.py` (around lines 432-444). Flow trace:

- User picks provider -> picks reasoning model -> picks toolcall model.
- At the toolcall picker, user presses Esc.
- `toolcall_value is None` -> `return None`.
- Caller `_interactive_model_menu` treats `None` as `continue` and loops back to the root menu.
- The reasoning model the user just confirmed is silently discarded with no feedback.

The `__keep__` sentinel does the right thing on Enter (apply reasoning, leave toolcall unchanged). Esc has different semantics than `__keep__` and there's no on-screen hint distinguishing them. The test `test_model_interactive_escape_backs_out_without_changes` only injects `None` at the reasoning step, not the toolcall step — so this scenario is uncovered.

Fix: change Esc-at-toolcall to mean "keep current toolcall, apply reasoning choice":

```python
if toolcall_value is None:
    toolcall_choice = None  # keep existing toolcall
    break
```

And remove the `if toolcall_value is None: return None` short-circuit.

### 2. `_read_action` doesn't restore terminal cleanly on `KeyboardInterrupt`

`app/cli/interactive_shell/repl_choice_menu.py` (around lines 844-866). The POSIX branch sets raw mode via `tty.setraw(fd)` and uses `finally` to restore — `finally` does run on `KeyboardInterrupt`, so TCSADRAIN is applied. But:

- `os.read(fd, 1)` blocks indefinitely; `KeyboardInterrupt` propagating through the `finally` correctly restores TTY mode, but the menu was drawn to screen and `_erase_menu` was never called. Result: a half-drawn menu in scrollback.
- No `SIGWINCH` handler. Resizing mid-menu leaves the previous render misaligned with `_cols()` on the next redraw.
- On some BSD variants, `os.read` raises `InterruptedError` on signal. Linux is fine (transparent retry), but the cross-platform claim isn't quite right.

Fix in `_pick`:

```python
try:
    action = _read_action()
except (KeyboardInterrupt, EOFError):
    _erase_menu(crumb, labels)
    raise
```

## Warnings

### 3. No provider-API-key precheck in the interactive flow

`switch_llm_provider` (existing code) guards against missing credentials and prints a clear error. The new interactive flow makes the user navigate through provider -> reasoning model -> toolcall model before learning they can't switch because the key isn't set. Add a `has_llm_api_key` check immediately after provider selection and bail with a one-line message — saves three picker screens of wasted attention. Matches the existing "refuse to half-update .env" principle.

### 4. `_interactive_history_menu` retention choices are hard-coded

Diff offers `["100", "500", "1000", "5000"]`. The existing `_history_retention` accepts arbitrary positive integers via the CLI arg path. Users with a custom cap (say, 750) have no way to select it from the menu — they'd land on whichever neighbor or be forced to use the CLI form. Either add a custom-input option (`Other...`) or document that the menu is opinionated.

## Info / nits

- `_pad` uses `len(content)` which counts code points, not display columns. Multi-byte Unicode labels (CJK service names from MCP store) misalign. Low risk for current choice sets.
- `breadcrumb=f"{_ROOT}"` is a redundant f-string; `breadcrumb=_ROOT` works.
- `repl_choose_one` calls `repl_tty_interactive()` internally even though callers already guard with it. Harmless redundancy.

## Pluses

- `repl_tty_interactive()` guard is consistent across every new interactive path — non-TTY contexts (CI, pipes, scripts) degrade cleanly to the original arg-based CLI behavior.
- Breadcrumb navigation gives clear context at each picker level.
- `_read_action` handles both POSIX (termios/tty) and Windows (msvcrt) key reading in one function — correct cross-platform shape.
- `_erase_menu` keeps scrollback clean instead of clearing the whole screen — right design.
- Module separation (`repl_choice_menu.py` standalone with its own `__all__`) is the right abstraction boundary.
- TTY pickers and `app/cli/wizard/prompts.py` questionary pickers coexist intentionally (wizard uses full-screen, REPL uses inline raw mode); patterns are not in conflict.

## Verdict

Request changes. The two critical items (Esc-at-toolcall semantics + KeyboardInterrupt cleanup) are real bugs. The provider-key precheck would cut three screens of wasted UX. Solid foundation otherwise — the choice-menu module looks like the right shape going forward.

muddlebee:
@yashksaini-coder its stil WIP.

muddlebee:
@greptile review

muddlebee:
@greptileai review

muddlebee:
@greptile review

muddlebee:
@greptileai review

muddlebee:
@greptileai review

PR #1655, opened 2026-05-08
feat: add Supabase integration for PostgREST, Auth, and Storage RCA

### LINKED ISSUE #1599: feat: add Supabase integration for database and storage RCA
### Problem statement

Teams using OpenSRE often run Supabase as their backend (PostgreSQL + Storage + Edge Functions) but currently have no way to investigate Supabase-related incidents through OpenSRE. When a Supabase database slows down, storage fails, or edge functions error out, there is no integration to surface that evidence during RCA — engineers have to manually dig through the Supabase dashboard.

### Proposed solution

Add a Supabase integration following the New Integration Checklist in AGENTS.md that:

- Detects Supabase via `SUPABASE_URL` and `SUPABASE_SERVICE_KEY` environment variables
- Checks database connection health and surfaces slow queries during RCA
- Reports storage bucket errors
- Reports edge function failure rates

I'd like to implement this myself following the existing integration pattern. Happy to coordinate with maintainers on scope before writing any code.

laxlab:
Fixes #1599

Adds a Supabase integration so OpenSRE can investigate incidents in Supabase-backed applications. This covers three services: PostgREST (database API), Auth, and Storage.

**Files added/modified:**
- `app/integrations/supabase.py` — config, health checks, storage bucket listing, credential resolution from store and env
- `app/tools/SupabaseHealthTool/` — tool for checking PostgREST, Auth, Storage health during RCA
- `app/tools/SupabaseStorageTool/` — tool for listing and auditing storage buckets
- `app/integrations/config_models.py`, `models.py` — config registration
- `app/integrations/_verification_adapters.py`, `registry.py`, `_catalog_impl.py` — full pipeline wiring
- `app/types/evidence.py` — added "supabase" to EvidenceSource
- `tests/integrations/test_supabase.py` — 37 unit tests

**Security:** `resolve_supabase_config` validates that the caller-supplied URL matches the configured origin before attaching credentials, preventing prompt-injection exfiltration. Credentials are also resolved from the integration store first (UI-registered users) before falling back to env vars.

https://asciinema.org/a/N71FuO9lEcXHKdbe

---

The integration follows the same pattern as `postgresql.py` and `rabbitmq.py`. The core challenge was credential resolution — Supabase credentials must never appear in LLM-visible tool signatures, so `resolve_supabase_config` resolves them internally from the store or env vars. I studied how other integrations handled this before writing the implementation.

For the security boundary: after reviewing the codebase I identified that forwarding credentials to an LLM-supplied URL was a prompt-injection risk, so I added origin validation that rejects any URL not matching the configured Supabase project. This matches how other integrations lock credentials to their configured endpoint.

The storage health check uses `/storage/v1/health` (not the bucket listing endpoint) to avoid false degradation reports for keys with restricted permissions.

---


dekunlab:
@greptile review

yashksaini-coder (changes requested):
Reviewed end-to-end. The shape of the integration is reasonable but there's a confirmed service-key exfiltration path via prompt injection, plus a stack of New Integration Checklist gaps and bot-flagged P1s.

## Security: Greptile P1 — CONFIRMED

### Service key forwarded to arbitrary URLs (prompt-injection exfil)

`app/integrations/supabase.py` `resolve_supabase_config` (around line 134-145):

```python
def resolve_supabase_config(project_url: str) -> SupabaseConfig:
    env_config = supabase_config_from_env()
    if env_config and env_config.url == project_url.rstrip("/"):
        return env_config
    service_key = os.getenv("SUPABASE_SERVICE_KEY", "").strip()
    return build_supabase_config({"url": project_url, "service_key": service_key})
```

The fast path only returns when `project_url` equals `SUPABASE_URL` exactly. If they differ, `SUPABASE_SERVICE_KEY` is attached to whatever URL the caller passed. Both tools (`SupabaseHealthTool`, `SupabaseStorageTool`) call `resolve_supabase_config(project_url)` with `project_url` taken directly from tool arguments, and both expose `surfaces=("investigation", "chat")`.

Attack scenario: alert context contains `"project_url": "https://attacker.example.com"`. The LLM forwards the value, `resolve_supabase_config` attaches the real service key to the outbound request, and `Authorization: Bearer <service_key>` is exfiltrated to the attacker's domain. There's no origin validation, no URL allowlist, no subdomain pattern check.

Fix: after the fast-path miss, either (a) reject the call (`raise ValueError("project_url must match configured SUPABASE_URL")`), or (b) verify same scheme + host as the env-var URL before attaching credentials. Option (a) matches every other integration in this repo — credentials are locked to the configured endpoint.

## Other Greptile P1s

1. **`"supabase"` not in `EvidenceSource`** — CONFIRMED. `app/types/evidence.py` doesn't list `"supabase"` in the `Literal`. Both tools pass `source="supabase"`, and `tool_decorator.py`'s `source` parameter is typed `EvidenceSource | None`. `make typecheck` will fail. Add `"supabase"` to the literal.

2. **Wrong storage health endpoint** — CONFIRMED. `get_service_health` hits `/storage/v1/bucket` (the bucket-list endpoint) which requires `storage.buckets.select`. Permission-restricted but valid keys return 403, and storage gets reported as degraded. Use `/storage/v1/health` instead — no special bucket permission needed.

3. **Dual config models** — CONFIRMED. `SupabaseConfig` (in `supabase.py`) and `SupabaseIntegrationConfig` (in `config_models.py`) duplicate the same fields. Only the former is used; the latter is exported but wired nowhere. Pick one and delete the other before merge — having both ships dead code that confuses future contributors.

## AGENTS.md Section 6 New Integration Checklist

- Integration code in `app/integrations/<name>/` — flat file pattern (acceptable per recent precedent like Datadog/Grafana, marginal)
- Tools under `app/tools/` with proper typing — FAIL (typecheck breaks)
- Unit/mock tests under `tests/integrations/` — PASS
- Docs under `docs/` — FAIL, no `docs/supabase.mdx`
- Screenshot or demo GIF — FAIL
- E2E or synthetic test — FAIL
- `make verify-integrations` passes — FAIL (not registered)
- `make lint` / `make typecheck` — FAIL (typecheck due to EvidenceSource)

## Critical (beyond Greptile)

### Integration not wired into the catalog or registry

`app/integrations/registry.py` has no `IntegrationSpec` for `"supabase"`. `app/integrations/_catalog_impl.py` doesn't import `supabase_config_from_env` or any env loader. `app/integrations/verify.py` has no `_verify_supabase` adapter, and `_verification_adapters.py` doesn't register one in `VERIFIER_REGISTRY`.

Net effect: `supabase_is_available` will always return `False` in a real pipeline run because nothing populates `sources["supabase"]`. The tools register at import time but won't dispatch in practice. Silent integration failure — not a startup error, not a test error, just nothing happens.

Wire the integration through:
- `_catalog_impl.py` (env loader + classifier)
- `registry.py` (`IntegrationSpec`)
- `verify.py` + `_verification_adapters.py` (so `make verify-integrations` exercises it)

## Warnings

- **`validate_supabase_config` is implemented but never called by `verify.py`.** Tests cover it directly, but `make verify-integrations` won't.
- **Error messages echo raw exception text.** `validate_supabase_config` and `get_storage_buckets` return `str(err)` in `detail`/`error`. URLs in `httpx` exceptions get echoed verbatim — combine with the URL-trust issue above and a verifier failure could echo back the attacker's URL. Compare with `app/utils/telegram_delivery.py` for the redaction pattern this repo prefers.
- **PostgREST health check uses `/rest/v1/` root** — for a Supabase project with RLS-public access, this returns 200 even with an invalid/missing service key, so `validate_supabase_config` will report `ok=True` for misconfigured keys. False confidence during setup.

## Info / nits

- `DEFAULT_SUPABASE_TIMEOUT_SECONDS = 10.0` and `DEFAULT_SUPABASE_MAX_RESULTS = 50` aren't exported from `__all__`; tools that want to override them have to import private names.
- `get_storage_buckets` and `get_service_health` both hit the bucket endpoint — one HTTP call could serve both if cached.
- Tests don't cover `resolve_supabase_config` when `project_url` differs from `SUPABASE_URL` — that's the test that would have caught the security issue. Add it as part of the fix.

## Pluses

- Test coverage is genuinely good: failure paths, 403, malformed JSON, connection errors, result capping all exercised.
- `project_url` is never logged; the LLM never sees the service key in tool signatures (the intent is right; only the URL trust boundary is the problem).
- `SupabaseConfig.is_configured` guard at the top of every public function is a clean pattern.
- `max_results` cap with a hard ceiling of 200 is a good production safety rail.
- `httpx` lazy-imported inside `_make_request` matches sibling service clients.

## Verdict

Request changes. Address-priority order:

1. **Security** — reject mismatched `project_url` in `resolve_supabase_config` (hard blocker).
2. **Wire the integration** through `registry.py` / `_catalog_impl.py` / `verify.py` (silent-failure blocker).
3. **`EvidenceSource`** — add `"supabase"` to the literal.
4. **Storage health endpoint** — switch to `/storage/v1/health`.
5. **Dual config models** — delete one.
6. **Docs + screenshot** per AGENTS.md checklist.

With items 1-5 in, this becomes a clean integration with strong test coverage.

muddlebee:
@laxlab pls fix greptile reviews

WatchTree-19 (commented):
Drive-by from someone running similar bench-side work on this repo. Substantive but not comprehensive.

**anon vs service-role key.** Supabase exposes two API keys per project: `anon` (subject to Row Level Security) and `service_role` (bypasses RLS). RCA tools that need to query across tenants or read `auth.users` and `auth.sessions` will need service-role; tools that only need a tenant's own data should be on anon. The integration config I see only carries one `api_key`. If service-role is used by default, that's a major security surface (anyone with read access to OpenSRE sees every Supabase project's data); if anon is used, Storage and Auth RCA will silently return empty for most schemas. Splitting into `api_key` (default anon) and an optional `service_role_key` (gated, used only for explicit cross-tenant tools) is the cleaner shape.

**Pooler vs direct connection.** Supabase Postgres has three endpoints per project: direct (port 5432, ~100 conn limit), session pooler (5432 with prepared statement quirks), and transaction pooler (6543, recommended for short-lived workloads). The verification probe and any continuous-running RCA tool that opens a connection per investigation should be on the transaction pooler (6543) by default to avoid hitting the connection cap on busy projects. Worth either documenting which endpoint the integration assumes or making it configurable.

**Auth schema noise.** Supabase ships its own `auth` schema (users, sessions, refresh_tokens, identities, mfa_*, sso_*, etc.) plus `storage` schema, plus `graphql` and `pgsodium` and `vault` extensions. None of these are typically what an RCA agent should investigate when the alert is about user-facing tables. The PostgreSQL slow-queries / current-queries / table-stats tools probably want a default exclusion list for these schemas (or a denylist toggle in the integration config) so the agent doesn't burn context on irrelevant catalog tables.

**Storage RCA edge case.** Supabase Storage maps to S3 but uses a project-scoped bucket prefix (`<project-ref>.storage.supabase.co`). If the Storage RCA tool was modelled on the existing AWS S3 tool, the bucket-listing path may not match. Worth a manual smoke against a real Supabase project before merge if that hasn't happened already.

**Already covered upstream.** Greptile's existing thread (if any, didn't scroll fully) is the right shape, no need to pile on.

The shape of this is good and there's clearly a sizeable Supabase userbase that will care. The auth-key separation and pooler defaults are the two items I'd want resolved before this is exposed to anything beyond closed test.

dekunlab:
@muddlebee Hey , on it !

dekunlab:
@WatchTree-19  thanks for reviewing , working on it right now

dekunlab:
@muddlebee @WatchTree-19  @yashksaini-coder 

Thanks for the detailed review guys ,  I’ve gone through the blockers and agree with the priority order. I’m tied up for a bit but will work through the security fix and integration wiring shortly.

muddlebee:
no hurries, this is shaping up good. good work @laxlab :+1:

dekunlab:
@greptile review

rrajan94:
@laxlab CI is failing, please check.

yashksaini-coder:
@greptile review

yashksaini-coder (commented):
Image

@laxlab these 5 file will be formatted on running the `make format-check` and it also seems to be the problem in your failing ci job. 

Kindly run the format-check and commit the changes to rerun the ci jobs. Also try to complete all the copilot and greptile reviewed comments to ensure a perfect 5/5 score as well.

muddlebee:
Hey, please ensure to follow the contribution guidelines. 

- PR template missing, restore back
- e2e demo missing, please add a video demo of full integration pipeline

dekunlab:
@greptileai

dekunlab:
Hi @muddlebee @VaibhavUpreti — I've addressed all the review feedback including the formatting fix, store-based credential resolution, and added the e2e demo recording. Could you approve the CI workflow run so the checks can complete? Thanks!

muddlebee:
@laxlab we need live e2e demo and not just running tests, how does it help?

muddlebee:
@laxlab pushing some commtis and fixes

muddlebee:
## Live local verification (Supabase CLI)

We validated the integration against a **real** local Supabase stack (not mocks):

1. **CLI / stack:** `npx supabase@latest init` then `npx supabase start` from a project directory.
   - **Docker note:** Initially `supabase start` failed when the project lived under `/tmp` because Docker Desktop did not share that path for volume mounts. Moving the project under the home directory (e.g. `~/opensre-supabase-demo`) fixed Studio/container bind mounts.
2. **Credentials:** `npx supabase status -o env` → set `SUPABASE_URL` to `API_URL` (e.g. `http://127.0.0.1:54321`) and `SUPABASE_SERVICE_KEY` to `SERVICE_ROLE_KEY`.
3. **OpenSRE checks:** From this repo, called `resolve_supabase_config`, `get_service_health`, and `get_storage_buckets`:
   - PostgREST, Auth (`/auth/v1/health`), and Storage (`/storage/v1/health`) reported healthy.
   - Storage buckets call succeeded; empty bucket list on a fresh project (expected).

**Automated suite:** `tests/integrations/test_supabase.py` (38 tests) exercises the same code paths with mocked HTTP, including v2 store credential resolution, origin checks, and bucket truncation metadata.

Latest branch push includes merge of current `main` plus review fixes (store v2 resolution, CodeQL logging, accurate bucket counts, HTTP client under `app/services/supabase/`).

muddlebee:
@greptile review

muddlebee:
@laxlab pushed this to the final lap. nice work but needs more fine-tuned work for PR mergeability :)

and congrats

dekunlab:
@muddlebee Thank you so much for the thorough review and patience throughout the process  really appreciated the detailed feedback, it made the integration genuinely better. Looking forward to contributing more! 🙏

PR #1658, opened 2026-05-08
fix: prevent REPL prompt session signature crash in interactive shell

7vignesh:
### Describe the Changes

This PR fixes a runtime crash in the interactive shell caused by a prompt session function signature mismatch.

**Changes included:**

- Updated the REPL loop to pass the current session into prompt session construction.
- Made the prompt session builder accept an optional `session` argument for compatibility.
- Added a regression test to ensure prompt session creation works when a session is passed.

### Why This Works

- The Sentry issue reports a missing required argument for the prompt session builder.
- The loop now provides the session at call time, aligning with the updated signature.
- The builder remains backward compatible by accepting `session` as optional, reducing risk across mixed runtime/module states.

## Validation evidence:

**Focused tests:**

```bash
pytest tests/cli/interactive_shell/test_loop.py -q
```

Result: ✅ 45 passed

**Lint:**

```bash
ruff check app/ tests/
```

Result: ✅ passed

**Format check:**

```bash
ruff format --check app/ tests/
```

Result: ✅ passed

**Typecheck:**

```bash
mypy app/cli/interactive_shell/
```

Result: ✅ passed

PR #1662, opened 2026-05-08
fix(cli): accept optional session param in _build_prompt_session for compat

### LINKED ISSUE #1656: TypeError: _build_prompt_session() missing 1 required positional argument: 'session'
Sentry Issue: [PYTHON-17](https://tracer-30.sentry.io/issues/7467658622/)

```
TypeError: _build_prompt_session() missing 1 required positional argument: 'session'
(8 additional frame(s) were not displayed)
...
  File "app/cli/__main__.py", line 211, in main
    cli(
  File "app/cli/__main__.py", line 164, in cli
    raise SystemExit(run_repl(config=config))
  File "app/cli/interactive_shell/loop.py", line 250, in run_repl
    return asyncio.run(_repl_main(initial_input=initial_input, _config=cfg))
  File "app/cli/interactive_shell/loop.py", line 185, in _repl_main
    prompt = _build_prompt_session()
```

davincios:
$(cat <<'EOF'
## Summary

- Some deployed builds call `_build_prompt_session(session)` with a positional argument; without that parameter in the signature the call raises `TypeError: _build_prompt_session() missing 1 required positional argument: 'session'`
- Fix: add `_session: ReplSession | None = None` as an optional parameter — the `_` prefix follows Python / ruff convention for intentionally-unused arguments (silences `ARG001`) and removes the need for an explicit discard statement

## Test plan

- [x] `make lint` — clean
- [x] `make format-check` — clean
- [x] `make typecheck` — 520 files, no issues
- [x] `uv run pytest tests/cli/interactive_shell/test_loop.py -q` — 44 passed

Closes #1656

https://claude.ai/code/session_018rsyVPDNXVCrEwPCaCAdo7
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_018rsyVPDNXVCrEwPCaCAdo7)_

### DIFF
--- a/app/cli/interactive_shell/prompt_surface.py
+++ b/app/cli/interactive_shell/prompt_surface.py
@@ -300,7 +300,7 @@ def _build_prompt_style() -> Style:
     )
 
 
-def _build_prompt_session() -> PromptSession[str]:
+def _build_prompt_session(_session: ReplSession | None = None) -> PromptSession[str]:
     return _install_prompt_frame(
         PromptSession(
             completer=ShellCompleter(),


WatchTree-19:
Quick observation, not a blocker:

#1658 ("fix: prevent REPL prompt session signature crash in interactive shell" by @7vignesh) appears to be addressing the exact same TypeError, opened ~6 minutes apart from this one. Worth coordinating before merge so one of them lands and the other rebases or closes, otherwise whichever lands second hits a no-op merge.

On the fix itself: the underscore-prefix convention is correct here, but if the deployed callers passing `session` positionally expected it to be wired up (the issue's framing leans that way), the placeholder fix silently absorbs the data instead of using it. If the eventual intent is to persist `prompt.history` from the passed-in `ReplSession`, a follow-up that actually consumes `_session` (and renames it back to `session`) would close the loop. Reasonable to ship the unblocking fix first and track the wire-up separately. Just flagging.

davincios:
Thanks for flagging! Both points addressed:

**Coordination with #1658**: Already handled — I left a comment on #1658 explaining the situation (the `_ = session` pattern it used triggers ruff's `ARG001`) and it's being superseded by this PR.

**Wire-up concern**: `_session` genuinely doesn't need to be consumed inside the function. The current design already wires the history correctly: `_build_prompt_session` loads history from disk via `load_prompt_history()`, and immediately after the call in `loop.py` the caller does `session.prompt_history_backend = prompt.history` — so the `ReplSession` gets the history handle via that assignment, not via what's passed into this function. The parameter is a pure compatibility shim for deployed callers that pass `session` positionally; there's no data loss.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1664, opened 2026-05-08
fix(cli): Performance and usability fix for long streamed responses. Long streamed responses in the CLI visibly stalled

### LINKED ISSUE #1649: [BUG] address issues found in #1345 streaming implementation: re-parse on long streams
### Summary

Follow-up to #1263 (PR #1345). This issue tracks the bugs and gaps
discovered after the initial streaming PR went up for review.

### Problems

- [x] Issue : O(n²) re-parse on long streams

#### Current code in streaming.py and renderer.py:
```
for chunk in chunks:
    buffer.append(chunk)
    live.update(Markdown("".join(buffer)))   # ← re-parses entire buffer
```

Every chunk causes:

"".join(buffer) — O(n) where n = total tokens so far
Markdown(...) — parses the entire accumulated text as Markdown
live.update(...) — Rich computes layout for the whole document
For a 10,000-token response with ~10,000 chunks, that's ~10⁸ operations total. At Anthropic's ~50–200 chunks/sec, the Markdown parser becomes the bottleneck. The terminal isn't "stuck" — it's churning. Each individual chunk is doing more parsing work than the chunk before it.

Rich.Live's refresh_per_second=20 only caps the visible repaint rate. It does NOT cap how often we call live.update(). Every call still parses Markdown.

Batched providers:
For providers that emit one large delta or fall back to yield-once (Bedrock, CLI-backed, some OpenAI-compat), invoke_stream yields a single big chunk. Our buffer-and-render loop fires once with the full response. UX is: spinner → text dump → done. Not progressive but also not slow — single render.

@VaibhavUpreti  recommended "true incremental" likely refers to the yield-once fallbacks

### Context

- Parent issue: #1263
- Original PR: #1345

### Acceptance criteria

- [x] All checkboxes above resolved
- [x] Tests added that catch the original error
- [x] `make lint && make format-check && make typecheck && make test-cov`
      green

**Expected:**
Long streamed responses (e.g. 10,000-word essay) render at the same
smooth rate as short ones — the terminal stays responsive throughout.

**Actual:**
Long streams progressively slow down and appear stuck. After the first
few hundred tokens, each new chunk visibly takes longer to render than
the one before it, with multi-second pauses on responses past ~1k tokens.

### Steps to reproduce

**Reproduction:**
1. `opensre`
2. Type: `What is opensre? Explain in 10,000 words.`
3. Observe: streaming starts smoothly, becomes progressively slower
   past ~1k tokens, looks stuck for long stretches near the end.

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

_No response_

### Additional context

_No response_

YauhenBichel:
Fixes #1649 

Performance and usability fix for long streamed responses. Long streamed responses in the CLI visibly stalled. A 10k-token response would freeze for several seconds at a time after the first ~1k tokens

- Perfomance and usability fix: `Markdown(buffer)` re-parse is throttled to once per refresh window. Without this, every incoming token triggered a full re-parse of the accumulated buffer — overall O(n²) work — which visibly stalled long streams past ~2k tokens. Final flush before `Live.stop()` ensures trailing-window chunks always render.
- Refactoring (applied separation of concerns): extracted `_DiagnoseStreamRenderer` so `StreamRenderer` keeps a single responsibility (event dispatch + node lifecycle + final report); diagnose-specific state lives in one focused class.

https://github.com/user-attachments/assets/c9421ae2-107b-477f-8a2e-c9a9c5985dd1

---

The streaming renderer's per-token work was O(n) on the accumulated buffer (rebuild the string + parse it as Markdown + lay it out). Done n times across a stream of n chunks, that's O(n²) — fine for short responses, brutal for anything past a few thousand tokens.

The throttle gate is one `time.monotonic()` comparison: if we already re-rendered within the current refresh window, skip. The chunk still gets appended to the buffer (no data loss), but the expensive re-parse waits for the next window. With a 100ms window at 10fps, that bounds parse work to ~10 calls per second of streaming — constant rate regardless of how fast tokens arrive.

The crucial detail is the **final flush**. Without one, chunks that arrive in the last throttle window before the stream ends — or chunks already buffered when an exception interrupts the loop — would never render. Wrapping the chunk loop in an inner `try/finally` inside the Live context manager guarantees a final `live.update(Markdown(buffer))` runs on both the happy and error paths, before Live closes.

**Performance characteristics**

| Metric | Before | After |
|---|---|---|
| Markdown parse work per stream | O(n²) — n chunks × O(n) per parse | O(refresh-rate × duration) — capped at 10/sec |
| Long-stream UX (10k+ tokens) | Visible freezes of several seconds; rendering rate degrades as buffer grows | Continuous progressive rendering; no perceptible degradation |
| Short-stream UX (< 500 tokens) | Smooth | Smooth (unchanged) |
| CPU during long generation | One core pegged at ~100% (parser saturation) | Typically < 30% — parse work is now bounded |
| Token loss on mid-stream exception | Possible — trailing-window chunks rendered per-chunk; without final flush, the last buffer state could be missed | None — final flush inside `try/finally` always runs before `Live.stop()` |

**Code path — the streaming hot loop**

| | Before | After |
|---|---|---|
| Per-chunk action | `live.update(Markdown("".join(buffer)))` runs unconditionally | `live.update(Markdown("".join(buffer)))` gated by `time.monotonic() - last_render >= _LIVE_RENDER_INTERVAL_S` |
| Final flush | None — relied on the last per-chunk update | Explicit unconditional flush before `Live.stop()` so trailing chunks always land |
| Failure-path render | Whatever the last per-chunk update happened to capture (could be stale) | Inner `try/finally` flushes latest buffer state regardless of success or exception |

**Code structure — separation of concerns**

| | Before | After |
|---|---|---|
| `StreamRenderer.__init__` | 5 diagnose-state fields (`_diagnose_buffer`, `_diagnose_live`, `_diagnose_started`, `_diagnose_last_render`, `_diagnose_console`) | 1 helper instance (`self._diagnose = _DiagnoseStreamRenderer()`) |
| Diagnose lifecycle | 3 long methods on `StreamRenderer` (~120 lines combined) | 3 methods on `_DiagnoseStreamRenderer` + 2 thin orchestrators (`_begin_diagnose`, `_end_diagnose`) on `StreamRenderer` |
| `_handle_events_mode` diagnose branch | ~15 lines of inline state manipulation | 4 lines of delegation: `self._diagnose.start()` / `.append_chunk()` / `.finish()` |
| Adding a second special-streaming node | Another long branch in `_handle_events_mode` | Another helper instance + 4-line dispatch entry |

---

---


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

WatchTree-19 (commented):
Drive-by review, take or leave.

**Streaming-buffer cardinality.** "Slow long streams" usually means one of three things in agentic LLM code: (a) the streaming chunks are arriving fast but the receiver is doing per-chunk synchronous work that blocks the event loop, (b) chunks are arriving slowly because of throttling on the model side, (c) chunks are arriving at normal speed but downstream rendering (Rich, prompt-toolkit, terminal flush) is the bottleneck. Knowing which one this PR addresses is important because the fix shape is different in each case. The PR title says "fix slow long streams" but the cure depends on the diagnosis. Worth a one-paragraph summary in the PR body of which symptom + which root cause.

**Cancellation behaviour.** Long streams interact awkwardly with Ctrl+C and prompt cancellation. If the user cancels mid-stream, the streaming consumer should: (1) stop calling `.next()` or equivalent on the iterator, (2) close the underlying HTTP connection so the server stops sending tokens (and so the user is not billed for tokens after cancel), (3) leave the partial output in a recoverable place rather than silently discarded. Worth confirming the fix preserves these.

**Backpressure.** If the buffering layer this PR touches is unbounded, a particularly verbose model response can OOM the process on a low-memory host. The standard pattern is bounded `asyncio.Queue` with size proportional to expected max-tokens, plus a backpressure signal back to the consumer. Without backpressure, "slow long streams" gets fixed at the cost of "OOM long streams" later.

**Tests.** Stream-handling logic is hard to unit-test deterministically. Two patterns that work: (a) replay-fixture tests where a recorded LLM response stream is fed through the consumer at controlled speeds, (b) async unit tests with `pytest-asyncio` that exercise the cancellation and backpressure paths explicitly. If this PR adds either, it'll prevent the next regression on this surface.

Slow streams in LLM-driven CLI work are a real long-tail bug source. Good to see this getting attention.

YauhenBichel:
@greptile review

VaibhavUpreti (approved):
awesome @YauhenBichel !

PR #1666, opened 2026-05-08
fix(cli): disable Rich legacy Windows renderer to prevent UnicodeEncodeError

### LINKED ISSUE #1663: UnicodeEncodeError: 'charmap' codec can't encode characters in position 0-2: character maps to <undefined>
Sentry Issue: [PYTHON-19](https://tracer-30.sentry.io/issues/7467759440/)

```
UnicodeEncodeError: 'charmap' codec can't encode characters in position 0-2: character maps to <undefined>
(18 additional frame(s) were not displayed)
...
  File "rich\console.py", line 2091, in _write_buffer
  File "rich\_windows_renderer.py", line 17, in legacy_windows_render
  File "rich\_win32_console.py", line 441, in write_styled
  File "rich\_win32_console.py", line 402, in write_text
  File "encodings\cp1252.py", line 19, in encode
```

davincios:
## Summary

- On Windows, Rich's legacy console renderer (`_windows_renderer.py`) encodes text through the system codepage (e.g. CP1252). Characters outside that codepage raise `UnicodeEncodeError: 'charmap' codec can't encode characters`.
- Add `legacy_windows=False` to the two `Console` instances that already set `force_terminal=True, color_system="truecolor"`, so Rich always takes the ANSI path instead.

## Root cause

Rich auto-detects on Windows whether to use its legacy `WriteConsole` API path or ANSI escape sequences. The `legacy_windows` flag defaults to `None` (auto), which on a Windows legacy terminal activates `_windows_renderer.py → _win32_console.py → encodings/cp1252`. Characters like emoji or non-Latin Unicode crash there.

The affected consoles (`loop.py` and `wizard/flow.py`) already request `color_system="truecolor"`, which implies an ANSI-capable terminal. Setting `legacy_windows=False` is therefore correct and consistent with the existing intent.

## Files changed

- `app/cli/interactive_shell/loop.py` — REPL main loop console
- `app/cli/wizard/flow.py` — onboarding wizard console

## Test plan

- [ ] On Windows, run `opensre` and send a response containing non-Latin Unicode (e.g. emoji or CJK characters); confirm no `UnicodeEncodeError`.
- [ ] On macOS/Linux, confirm normal terminal output is unchanged.

Closes #1663

https://claude.ai/code/session_01JDwPRyFyRn8FywHbPBwXMm

---
_Generated by [Claude Code](https://claude.ai/code/session_01JDwPRyFyRn8FywHbPBwXMm)_


WatchTree-19 (commented):
Drive-by from someone who's been on Windows for the recent eval-bench work, this fix lands at exactly the right surface.

**Legacy Windows renderer scope.** Rich's `legacy_windows=True` path is the one that drops to ASCII fallback for terminals that don't support full ANSI / Unicode. The UnicodeEncodeError was almost certainly a cp1252-encoded stdout choking on em dashes / arrows / box-drawing characters. Disabling `legacy_windows` and forcing the modern renderer path is the right move because:
- Windows Terminal (default on Win11 + a free Win10 install) handles full Unicode + ANSI
- conhost on Win10 has UTF-8 support behind `chcp 65001` since 1903 (5 years ago)
- the legacy path was for cmd.exe-on-XP-era which is no longer a realistic target

**Fallback for the actual cmd.exe holdouts.** The one risk: someone running on a fresh Win10 image with stock cmd.exe at code page 850 will still hit UnicodeEncodeError on the modern renderer. Worth confirming Rich's modern path silently degrades to ASCII when the terminal can't encode (Rich does this internally via `safe_box=True` fallback) rather than raising. If not, an explicit `force_terminal=True, safe_box=True` pair on Console init is the belt + braces.

**PYTHONIOENCODING signal.** Setting `PYTHONIOENCODING=utf-8` in the install scripts (or documenting it as a Windows requirement in quickstart) prevents this whole class of bug at the Python layer rather than the Rich layer. Worth the one-line addition if it's not there already, especially since this also covers downstream tools that aren't Rich.

**Test coverage.** Hard-to-CI on Linux runners but the failure mode is real for Windows users. If there's a Windows runner in CI Labels (saw the windows quality / windows test lanes earlier today), gating a smoke test there for "import opensre + render banner without raising" closes the regression vector. If those Windows lanes aren't running on every PR yet, this is the right surface to wire them in.

Windows install ergonomics is the kind of thing that costs disproportionate user trust when it breaks. Good call to fix at this layer rather than chasing every emoji escape downstream.

davincios:
@greptile review

PR #1674, opened 2026-05-08
feat(synthetic/rds_postgres): wire --axis2 CLI flag through to _print…

### LINKED ISSUE #1672: [IMPROVEMENT] wire --axis2 CLI flag through to _print_gap_report in run_suite.py
### Current state

`tests/synthetic/rds_postgres/run_suite.py` exposes a `--axis2` argument via `parse_args` and an aggregator function `_print_gap_report` (around line 455) that prints overall and per-difficulty-level pass rates plus the gap. The flag and the aggregator both exist, but `run_suite()` (around lines 494-522) never references `args.axis2` and never calls `_print_gap_report`.

Net effect today: a user who runs `python -m tests.synthetic.rds_postgres.run_suite --axis2` (the command that #1587 originally documented) gets the normal per-scenario PASS/FAIL printout with no gap numbers and no error. The flag is parsed and silently ignored.

Caught during review of #1587 (docs PR) by @yashksaini-coder. Docs in #1587 have been softened to remove the misleading invocation, but the underlying CLI gap is still open.

### Desired state

`run_suite()` should branch on `args.axis2`:

- when `--axis2` is set, load both axis 1 and axis 2 result sets and call `_print_gap_report(axis1_results, axis2_results)` instead of (or in addition to) the per-scenario PASS/FAIL printout
- when not set, current behaviour is preserved

Implementation sketch:

```python
def run_suite(...):
    args = parse_args()
    if args.axis2:
        axis1_results = ...   # full mock backend (FixtureGrafanaBackend)
        axis2_results = ...   # selective backend (SelectiveGrafanaBackend)
        _print_gap_report(axis1_results, axis2_results)
        return
    # existing per-scenario flow
    ...
```

Once this is wired, the docs in #1587 can re-add the canonical `python -m tests.synthetic.rds_postgres.run_suite --axis2` line, and the AXIS-GAP marker block in `docs/benchmarks/README.md` can be auto-populated by a CI workflow that runs both axes on every push to `main`.

Happy to take this — small, self-contained, complements #1587 (docs) and #1580 (TP/FP/TN/FN counts that the gap report can include). Will be a ~30-line PR.

WatchTree-19:
Closes #1672. Addresses the gap caught by @yashksaini-coder during review of #1587: `--axis2` was parsed by `parse_args` but never read by `run_suite`, so the documented invocation `python -m tests.synthetic.rds_postgres.run_suite --axis2` silently produced normal per-scenario output with no gap report.

## What this changes

In `tests/synthetic/rds_postgres/run_suite.py`:

- new import for `SelectiveGrafanaBackend` (already exists at `tests/synthetic/mock_grafana_backend/selective_backend.py`)
- `run_suite()` now branches on `args.axis2`. When set, every fixture is run twice: once with the full mock backend (`FixtureGrafanaBackend`, axis 1) and once with the selective backend (`SelectiveGrafanaBackend`, axis 2). Both result sets are then passed to `_print_gap_report` which already exists and prints overall + per-difficulty pass-rates plus the gap.
- when `--axis2` is not set, behaviour is unchanged

Net diff: ~20 lines added, 0 modified outside the dispatch branch.

## What this does not change

- the per-scenario PASS/FAIL flow when `--axis2` is not set
- the JSON output path
- the existing `_print_gap_report` aggregator (used as-is)
- any test fixtures or scoring logic

## Local validation

```
$ python3 -c "import ast; ast.parse(open('tests/synthetic/rds_postgres/run_suite.py').read())"
# clean

$ ruff check tests/synthetic/rds_postgres/run_suite.py
All checks passed!
```

The full pytest run requires the opensre dependency tree (langgraph, opentelemetry, etc.) so the end-to-end smoke against a real fixture happens in CI on developer machines that have it.

## Why a docs-only fix on #1587 wasn't enough

#1587 originally documented `python -m tests.synthetic.rds_postgres.run_suite --axis2` as the canonical command for the gap report. After @yashksaini-coder caught that the flag was a no-op, the docs in #1587 were softened (PR #1587 is now docs-only and explicitly notes the wiring is tracked separately). This PR is that separately-tracked wiring. Once this lands, #1587's docs can be updated in a small follow-up to re-add the canonical invocation.

## Mergeability

Branched off `upstream/main`. Independently mergeable from #1580 (TP/FP/TN/FN counts) and #1420 (memory_mode field). When all three land, `_print_gap_report` could optionally emit the new classification metrics too, but that is its own scope.

cc @VaibhavUpreti @muddlebee @yashksaini-coder

---

- [x] Yes
- [ ] No

- [ ] No

- [x] Yes
- [ ] No

### DIFF
--- a/tests/synthetic/rds_postgres/run_suite.py
+++ b/tests/synthetic/rds_postgres/run_suite.py
@@ -34,6 +34,7 @@ from rich.table import Table
 
 from tests.synthetic.mock_aws_backend import FixtureAWSBackend
 from tests.synthetic.mock_grafana_backend.backend import FixtureGrafanaBackend
+from tests.synthetic.mock_grafana_backend.selective_backend import SelectiveGrafanaBackend
 from tests.synthetic.rds_postgres.observations import (
     TrajectoryPolicy,
     TrajectoryPolicyResult,
@@ -43,6 +44,7 @@ from tests.synthetic.rds_postgres.observations import (
     render_report_to_console,
     write_observation,
 )
+from tests.synthetic.rds_postgres.reporting import print_gap_report
 from tests.synthetic.rds_postgres.runner_api import (
     LevelRunConfig,
     LevelRunResult,
@@ -721,9 +723,63 @@ def run_synthetic_suite(config: SuiteRunConfig) -> SuiteRunResult:
     )
 
 
+def _run_axis2_suite(
+    fixtures: list[ScenarioFixture],
+    *,
+    output_json: bool,
+) -> list[ScenarioScore]:
+    """Run every fixture twice (axis 1 and axis 2) and emit the gap report.
+
+    Axis 1 uses ``FixtureGrafanaBackend`` (full mock data, the same backend the
+    default suite uses with ``--mock-grafana``). Axis 2 uses
+    ``SelectiveGrafanaBackend`` (query-aware adversarial mock). The combined
+    result list is returned so :func:`main`'s exit code reflects failures on
+    either axis — a fully-failing axis 2 run still surfaces as non-zero.
+    """
+    axis1_results: list[ScenarioScore] = []
+    axis2_results: list[ScenarioScore] = []
+    for fixture in fixtures:
+        _, score1 = run_scenario(fixture, use_mock_grafana=True)
+        axis1_results.append(score1)
+        _, score2 = run_scenario(
+            fixture,
+            use_mock_grafana=False,
+            grafana_backend=SelectiveGrafanaBackend(fixture),
+        )
+        axis2_results.append(score2)
+
+    if output_json:
+        print(
+            json.dumps(
+                {
+                    "axis1": [asdict(r) for r in axis1_results],
+                    "axis2": [asdict(r) for r in axis2_results],
+                },
+                indent=2,
+            )
+        )
+    else:
+        print_gap_report(axis1_results, axis2_results, fixtures)
+
+    return axis1_results + axis2_results
+
+
 def run_suite(argv: list[str] | None = None) -> list[ScenarioScore]:
     args = parse_args(argv)
     config = _build_run_config(args)
+
+    # --axis2 short-circuits the default per-level orchestration: every selected
+    # fixture is run twice (FixtureGrafanaBackend then SelectiveGrafanaBackend)
+    # and the cross-axis gap is printed via ``print_gap_report``. This is the
+    # canonical command documented in tests/synthetic/rds_postgres/README.md.
+    if args.axis2:
+        all_fixtures = load_all_scenarios(SUITE_DIR)
+        try:
+            selected_fixtures = select_fixtures(all_fixtures, config)
+        except ValueError as exc:
+            raise SystemExit(str(exc)) from exc
+        return _run_axis2_suite(selected_fixtures, output_json=bool(args.json))
+
     suite_result = run_synthetic_suite(config)
     results = list(suite_result.scores)
     canonical_payloads = dict(suite_result.canonical_payloads)
--- a/tests/synthetic/rds_postgres/test_run_suite_axis2_cli.py
+++ b/tests/synthetic/rds_postgres/test_run_suite_axis2_cli.py
@@ -1,0 +1,203 @@
+"""Unit tests for the ``--axis2`` CLI dispatch in ``run_suite``.
+
+Pins the wiring fix for #1672: ``--axis2`` was previously parsed by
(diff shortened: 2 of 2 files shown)


cerencamkiran:
Good work wiring the previously no-op "--axis2" flag.

A few blockers before this can be merged though:

1. "_print_gap_report" is currently called with two args, but its signature needs "all_fixtures" as the third arg, so "--axis2" will crash immediately.
2. Returning only "axis1_results" means the exit code ignores Axis 2 failures. I think the return should include Axis 2 results as well, otherwise CI can pass even if the selective/adversarial run fails.
3. "--axis2 --json" currently ignores JSON output. Either support a structured JSON shape for both axes, or fail fast if "--json" and "--axis2" are mutually exclusive.

Once those are fixed and Greptile/CodeQL are green, this should be much cleaner.

WatchTree-19:
@yashksaini-coder @cerencamkiran thanks, all three issues addressed in the new commit.

- `_print_gap_report` now receives `fixtures` as the third positional argument. The TypeError on every `--axis2` invocation is resolved and the gap report actually prints.
- `--json` is now honoured under `--axis2`. Output shape is `{"axis1": [...], "axis2": [...]}` so both result sets are addressable downstream rather than the JSON path being silently dropped.
- Return value is now `axis1_results + axis2_results` so `main()`'s exit-code logic reflects failures on either axis. A fully-failing axis2 run will exit 1 rather than 0.

Ran `python3 -c "import ast; ast.parse(open('tests/synthetic/rds_postgres/run_suite.py').read())"` clean, plus `ruff check` clean. The end-to-end smoke against real fixtures runs in CI.

Ready for re-review.

cerencamkiran:
@greptileai review

rrajan94:
@WatchTree-19 CI is failing, please check.

WatchTree-19:
Note on the `test (ubuntu-latest)` failure: this PR's only change is in `tests/synthetic/rds_postgres/run_suite.py` (adding a `SelectiveGrafanaBackend` import + branching `run_suite()` on `args.axis2`). The failing CI step runs `pytest --ignore=tests/synthetic -m "not synthetic"`, which excludes the path my changes touch.

Logically this PR cannot have broken a non-synthetic test. The failure is either a flaky test or a pre-existing main breakage. A re-run of CI would likely clear it. Happy to confirm by rebasing onto latest `main` if that's preferred, or if a maintainer can hit re-run on the failing job.

Quality + CodeQL both green. No conflicts with base.

davincios:
Rebased and refreshed the PR to land on top of current `main`. Two things were happening:

**1. The "test (ubuntu-latest)" failure was not flaky — it was stale tests.**
The two failing renderer tests (`tests/remote/test_renderer.py::TestStreamRendererDiagnoseStreaming::test_other_nodes_bypass_diagnose_streaming` and `test_diagnose_live_closed_on_mid_stream_exception`) referenced a `_diagnose_live` attribute that was renamed to `_diagnose._live` during the diagnose-renderer refactor (#1692). `main` already carries the fix in `1ce4d5a fix(remote): align StreamRenderer tests with diagnose refactor`, so rebasing onto current `main` makes those tests green here too.

**2. The PR's `run_suite.py` diff no longer applied cleanly.**
After the synthetic-suite refactor (#1786, #1798) the file went from ~515 lines (single `run_suite()` body) to ~754 lines split across `run_suite.py`, `runner_api.py`, `observations.py`, `reporting.py`. The original 35-line patch couldn't merge.

The wiring intent is preserved and re-implemented for the new architecture:

- `--axis2` now short-circuits the default per-level orchestration in `run_suite()` and runs every selected fixture twice — once with `FixtureGrafanaBackend` (axis 1), once with `SelectiveGrafanaBackend` (axis 2) — then calls `print_gap_report` from `reporting.py` (renamed from `_print_gap_report`).
- `--axis2 --json` emits `{"axis1": [...], "axis2": [...]}` so both axes stay addressable downstream.
- Combined return (`axis1_results + axis2_results`) so `main()`'s exit code reflects failures on either axis — a fully-failing axis 2 run still exits 1.

Also added `tests/synthetic/rds_postgres/test_run_suite_axis2_cli.py` with three unit tests pinning the wiring (each fixture runs twice with the right backend, JSON shape is structured, axis-2 failures propagate through `main()`'s exit code), so this regression is locked in.

Local checks before push: `ruff check`, `ruff format --check`, `mypy`, plus the full `tests/remote/test_renderer.py` (53 passed) and `tests/synthetic/rds_postgres/test_run_suite_*` (20 passed). CI is green.

Merging.

PR #1683, opened 2026-05-08
Fix LangGraph config adapter typing

### LINKED ISSUE #1681: [BUG] LangGraph emits repeated RunnableConfig warnings when building pipeline graph
### Summary

Building the pipeline graph emits repeated LangGraph warnings because wrapped node functions expose `config: Any` instead of `RunnableConfig | None`.

**Expected:** Building the graph should not emit warnings for every wrapped node.

**Actual:** Graph construction emits repeated warnings like:

```text
app/pipeline/graph.py:57: UserWarning: The 'config' parameter should be typed as 'RunnableConfig' or 'RunnableConfig | None', not 'typing.Any'.
  graph.add_node("inject_auth", _accept_langgraph_config(inject_auth_node))
app/pipeline/graph.py:60: UserWarning: The 'config' parameter should be typed as 'RunnableConfig' or 'RunnableConfig | None', not 'typing.Any'.
  graph.add_node("chat_agent", _accept_langgraph_config(chat_agent_node))
```

### Steps to reproduce

1. From the repo root, run the minimal graph-construction repro:

```powershell
uv run python -W default -c "from app.pipeline.graph import build_graph; build_graph()"
```

2. The same warning also appears during an investigation
 in the interactive shell:

```text
/investigate tests/e2e/kubernetes/fixtures/datadog_k8s_alert.json
```

Yes

### How often does it occur?

Every time

### Operating system

Windows

### Logs and error output

_No response_

### Additional context

_No response_

gbsierra:
Fixes #1681 

- Typed the LangGraph boundary adapter's `config` parameter as `RunnableConfig | None`, which matches LangGraph's expected node signature and removes the repeated config annotation warnings during graph construction.
- Preserved OpenSRE's internal `NodeConfig` abstraction by casting only inside `_accept_langgraph_config()`, so core node signatures do not need to import LangGraph/LangChain runtime types.
- Narrowed the existing `RunnableConfig` import guard so it still prevents framework types from leaking into core nodes and runners, while allowing the graph adapter to use the framework boundary type.
- Added regression coverage that builds the graph under warning capture and verifies the LangGraph config annotation warning is no longer emitted.

Focused tests:

```powershell
.\.venv\Scripts\python.exe -m pytest tests\pipeline\test_graph_config_adapter.py tests\types\test_config.py
```

Result:

```text
4 passed
```

The repeated warnings like this are no longer emitted:

```text
UserWarning: The 'config' parameter should be typed as 'RunnableConfig' or 'RunnableConfig | None', not 'typing.Any'.
```

Note: the command still emits a separate pre-existing LangGraph deprecation warning from `app/pipeline/routing.py` about importing `Send` from `langgraph.constants`; that is unrelated to this fix.

### DIFF
--- a/app/pipeline/graph.py
+++ b/app/pipeline/graph.py
@@ -3,6 +3,7 @@
 from collections.abc import Callable
 from typing import Any, cast
 
+from langchain_core.runnables import RunnableConfig
 from langgraph.graph import END, StateGraph
 from langgraph.graph.state import CompiledStateGraph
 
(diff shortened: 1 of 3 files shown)


muddlebee:
already fixed in main. thank you for reporting

PR #1704, opened 2026-05-08
Fix doctor LLM key check to use keyring-aware resolver

### LINKED ISSUE #1686: [BUG] /doctor reports missing LLM API key when key is available in keyring
### Summary

`/doctor` checks only environment variables for the active LLM provider API key, so it warns on a valid onboarded setup where the key is stored in the system keyring.

**Expected:** `/doctor` should use the same credential resolution path as onboarding/runtime and recognize API keys stored in the system keyring.

**Actual:** `/doctor` reports the active provider key as missing:

```text
llm_provider │ warn │ provider=gemini, but GEMINI_API_KEY is not set
```

But runtime credential resolution succeeds:

```powershell
python -c "import os; print('env:', bool(os.getenv('GEMINI_API_KEY'))); from app.llm_credentials import has_llm_api_key; print('runtime:', has_llm_api_key('GEMINI_API_KEY'))"
```

```text
env: False
runtime: True
```

This appears inconsistent with onboarding, which stores provider API keys in the system keyring and writes only non-secret provider settings to `.env`.

### Steps to reproduce

1. Configure an LLM provider through `opensre onboard`, using a provider that stores an API key such as Gemini.

2. Confirm `.env` contains the provider/model settings but not the API key, for example:

```text
LLM_PROVIDER=gemini
GEMINI_REASONING_MODEL=<redacted>
GEMINI_MODEL=<redacted>
```

3. Confirm runtime credential resolution can find the key:

```powershell
python -c "import os; print('env:', bool(os.getenv('GEMINI_API_KEY'))); from app.llm_credentials import has_llm_api_key; print('runtime:', has_llm_api_key('GEMINI_API_KEY'))"
```

Expected output for this setup:

```text
env: False
runtime: True
```

4. Run:

```text
/doctor
```

5. Observe that `/doctor` reports:

```text
provider=gemini, but GEMINI_API_KEY is not set
```

Yes

### How often does it occur?

Every time

### Operating system

Windows

### Logs and error output

_No response_

### Additional context

Note: If `/doctor` is intentionally meant to validate only environment variables, then the message may need to distinguish “not set in env” from “not available to OpenSRE runtime.” Right now it reads like the LLM provider is not configured, even though runtime resolution succeeds via keyring.

gbsierra:
Fixes #1686

Updated `/doctor` so hosted LLM API-key checks use the existing keyring-aware `has_llm_api_key(...)` helper instead of checking only environment variables.

This makes `/doctor` recognize valid onboarded setups where the provider API key is stored in the system keyring. Non-secret env checks such as `AWS_DEFAULT_REGION` and `OLLAMA_HOST` remain env-only, and CLI provider checks are unchanged.

Before:
Before
After:
After

added tests:
```text
uv run pytest tests/cli/test_doctor.py
```


cerencamkiran:
Small non-blocking thought: right now the provider requirements are split between api_key_vars and env_vars in /doctor. If we add more providers later, there’s a small risk the runtime credential rules and doctor checks drift again.

Might be worth eventually centralising provider credential metadata in a shared helper/source-of-truth, but I think the focused fix here is good 👍

gbsierra:
> Small non-blocking thought: right now the provider requirements are split between api_key_vars and env_vars in /doctor. If we add more providers later, there’s a small risk the runtime credential rules and doctor checks drift again.
> 
> Might be worth eventually centralising provider credential metadata in a shared helper/source-of-truth, but I think the focused fix here is good 👍

yes i mentioned this here https://github.com/Tracer-Cloud/opensre/issues/1686#issuecomment-4408373205
we decided to do the minimal fix for now

VibhorGautam (commented):
clean separation of api_key_vars vs env_vars - secrets go through keyring, infrastructure config stays env-only. the _raise_if_called pattern in the bedrock test is a nice guardrail to make sure that boundary doesnt get accidentally crossed

the updated error message "not available in env or keyring" is a good UX improvement, gives the user actionable info about where to look

one thing i was curious about - if a new provider gets added to api_key_vars but not env_vars (or vice versa), is there anything that catches that? might be worth a quick test or assertion that the two dicts cover all expected providers

rrajan94:
@gbsierra please resolve the conflicts

gbsierra:
@greptile review

PR #1708, opened 2026-05-08
fix(catalog): log warning when integration validation fails

### LINKED ISSUE #1554: [BUG] Silent swallowing of exceptions during integration classification in app/integrations/catalog.py hides validation errors
### Summary

Currently, when classifying integrations in app/integrations/catalog.py via _classify_service_instance, the codebase relies on Pydantic models (e.g., GrafanaIntegrationConfig, AWSIntegrationConfig, etc.) to validate credentials.

However, all of these validations are wrapped in bare except Exception: blocks that return None, None without logging the exception.

```# Example from app/integrations/catalog.py (Lines ~200-209)
try:
    grafana_config = GrafanaIntegrationConfig.model_validate(
        {
            "endpoint": credentials.get("endpoint", ""),
            "api_key": credentials.get("api_key", ""),
            "integration_id": record_id,
        }
    )
except Exception:
    return None, None
```
Why this is an issue: If a user configures an integration incorrectly (e.g., provides a malformed URL, or a required field fails Pydantic's strict validation), the ValidationError is completely swallowed. The integration is silently dropped, leaving no logs, warnings, or traces for the user or the SRE team to understand why the integration failed to load.

Proposed Solution:

Replace the bare except Exception: with except pydantic.ValidationError as e: (or except Exception as e: if broader coverage is strictly needed).
Use the logger initialized at the top of the file to emit a warning when a configuration fails to validate:
```except Exception as e:
    logger.warning("Failed to validate %s integration %s: %s", key, record_id, e)
    return None, None
```

This will vastly improve the debuggability of integration setups without breaking the current flow.

**Expected**
When an integration configuration is loaded from the environment or data store and contains invalid credentials (e.g., a malformed URL, missing required fields, or incorrect types), the system should log a warning or error message indicating which integration failed validation and why (e.g., logging the pydantic.ValidationError). This allows users or operators to quickly identify and fix the misconfiguration.

**Actual**
When a configuration is invalid, the except Exception: block blindly catches the ValidationError and returns None, None. The integration is silently skipped and omitted from the active integrations list. No warnings, errors, or traceback logs are printed to standard output or log files. The user is left completely blind as to why their integration isn't registering.

### Steps to reproduce

Configure an integration with intentionally invalid data or missing required fields. For example, configure a Datadog or Grafana integration where an integer is expected but an invalid string is provided, or provide a malformed URL.
Start the OpenSRE agent or run an integration-related CLI command (e.g., opensre integrations list or opensre integrations verify).
Check the console output and logs.
Result: The malformed integration does not appear in the active list, and there are absolutely zero warnings or error messages in the logs indicating that the configuration failed to parse or validate.

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

_No response_

### Additional context

_No response_

aayushprsingh:
## Summary

Fixes silent exception swallowing in `_classify_service_instance()` in `app/integrations/_catalog_impl.py`.

**Problem:** All 29 service classifiers (`grafana`, `aws`, `datadog`, etc.) wrapped their `Pydantic` model validation in bare `except Exception:` blocks that returned `None, None` with no logging. When a user misconfigured an integration — e.g. malformed URL, wrong field type, missing required field — the `ValidationError` was silently dropped. The integration would simply be absent from the active list with no trace in any log.

**Fix:** Replace each bare `except Exception:` with:

```python
except Exception as exc:
    logger.warning("Failed to validate %s integration %s: %s", key, record_id, exc)
    return None, None
```

This gives operators visibility into misconfigured integrations without changing the runtime behaviour (integration is still skipped, just now with a warning).

## Changes

- `app/integrations/_catalog_impl.py`: +58/-29 lines
  - 29 `except Exception:` blocks in `_classify_service_instance()` now log a warning with the integration key, `record_id`, and exception message before returning `None, None`

## Testing

```bash
# Should log a warning when an integration has invalid config
python -c "
from app.integrations._catalog_impl import _classify_service_instance
import logging
logging.basicConfig(level='WARNING')
_classify_service_instance('grafana', {'endpoint': 'not-a-url', 'api_key': 12345}, record_id='test-1')
"  # expected: WARNING - Failed to validate grafana integration test-1: ...
```

Fixes: #1554

### DIFF
--- a/app/integrations/_catalog_impl.py
+++ b/app/integrations/_catalog_impl.py
@@ -162,7 +162,8 @@ def _classify_service_instance(
                     "integration_id": record_id,
                 }
             )
-        except Exception:
+        except Exception as exc:
+            logger.warning("Failed to validate %s integration %s: %s", key, record_id, exc)
             return None, None
         if not grafana_config.endpoint:
             return None, None
@@ -194,7 +195,8 @@ def _classify_service_instance(
                 AWSIntegrationConfig.model_validate(raw_config).model_dump(exclude_none=True),
                 "aws",
             )
-        except Exception:
+        except Exception as exc:
+            logger.warning("Failed to validate %s integration %s: %s", key, record_id, exc)
             return None, None
 
     if key == "datadog":
@@ -207,7 +209,8 @@ def _classify_service_instance(
                     "integration_id": record_id,
                 }
             )
-        except Exception:
+        except Exception as exc:
+            logger.warning("Failed to validate %s integration %s: %s", key, record_id, exc)
             return None, None
         if datadog_config.api_key and datadog_config.app_key:
             return datadog_config.model_dump(), "datadog"
@@ -223,7 +226,8 @@ def _classify_service_instance(
                     "integration_id": record_id,
                 }
             )
-        except Exception:
+        except Exception as exc:
+            logger.warning("Failed to validate %s integration %s: %s", key, record_id, exc)
             return None, None
         if honeycomb_config.api_key:
             return honeycomb_config.model_dump(), "honeycomb"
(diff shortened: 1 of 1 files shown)


aayushprsingh:
@greptile review

PR #1734, opened 2026-05-09
fix(cli): validate slash-command args before policy gate (#1712)

### LINKED ISSUE #1712: [BUG] /investigate prompts before usage check
### Summary

`/investigate` asks for confirmation before it checks required args.

**Expected:** Missing args should print usage and stop.

**Actual:** The REPL shows `Proceed? [y/N]`, then prints `usage: /investigate <file>` after confirm.

### Steps to reproduce

1. Start `opensre`.
2. Run `/investigate` with no file.
3. Confirm the prompt.
4. Notice the usage message appears after the confirm prompt.

Image

Yes

### How often does it occur?

Every time

### Operating system

Windows

### Logs and error output

_No response_

### Additional context

This is a small fix, but it points to a bigger issue that command handling is too brittle and not abstract enough.

New commands should not need special-case checks added later.

Validation should live with the command definition, not in scattered policy branches.

In short, the current flow is "policy first, usage later."
The better flow is "validate first, policy second, handler last."

Mohdtalibakhtar:
Fixes #1712

Adds an optional `validate_args` field to `SlashCommand` and wires it into `dispatch_slash` so per-command arg validation runs before the policy gate, not inside the handler. With this in place, typing `/investigate` (or `/save`, or `/cancel`) with no args prints the usage line immediately and never triggers the `Proceed? [y/N]` confirmation prompt.

The same bug shape exists across three ELEVATED-tier commands, not just the one named in the issue: `/investigate` (`investigation.py`), `/save` (`investigation.py`), and `/cancel` (`tasks_cmds.py`) all carry an `if not args: print usage` block inside the handler that runs after the policy prompt. Fixing only `/investigate` would leave `/save` and `/cancel` broken in the same way, so the PR migrates all three together. SAFE-tier commands (`/template`, `/last`, `/list`, etc.) are untouched because they never prompt and their usage messages already print immediately.

Screenshot 2026-05-09 at 3 00 52 PM

---

The validator is a `Callable[[list[str]], str | None]` declared on `SlashCommand`. It returns `None` for valid args, or a fully-rendered usage string (Rich markup included) for invalid ones. The dispatcher prints the string, records the slash entry as `ok=False`, and short-circuits before the policy tier resolution runs. This keeps the existing handler bodies focused on the happy path and addresses the reporter's point that validation should live with the command definition rather than in scattered handler branches.

The validator runs even when `policy_precleared` is set, because pre-clearance only attests that policy was already approved by the caller, not that args are correct. Invalid input should still fail fast on that path. Trust mode also does not bypass the validator, because trust mode is about confirmations, not correctness. A regression test pins this.

Considered and rejected: a surgical patch in `dispatch_slash` that hardcodes command names like `if name in {"/investigate", "/save", "/cancel"} and not args: ...`. That couples the dispatcher to specific handlers and ignores the reporter's design framing. A full refactor of every command into declarative validators was also rejected as out of scope for a bug fix; the new field is opt-in so existing commands are unchanged.

The /cancel validator preserves its [ERROR] red usage styling rather than standardising to the [DIM] grey used by /investigate and /save. This is not the PR to standardise styling.

Tests live in `tests/cli/interactive_shell/test_commands.py` under two new classes. `TestPrePolicyValidation` parametrizes the ordering invariant across all three commands (with `is_tty=True` and a recording `confirm_fn`, missing args print usage and never call `confirm_fn`), plus the trust-mode-still-validates case and a regression test that valid args still fire the policy prompt. `TestSlashValidatorFunctions` adds direct unit tests on each validator. The existing `test_missing_arg_prints_usage` cases continue to pass because they exercise the same string output, just via the new code path.

---

### DIFF
--- a/app/cli/interactive_shell/command_registry/__init__.py
+++ b/app/cli/interactive_shell/command_registry/__init__.py
@@ -111,6 +111,12 @@ def dispatch_slash(
         console.print()
         console.print(f"[{ERROR}]unknown command:[/] {escape(name)}  (type [bold]/help[/bold])")
         return True
+    if cmd.validate_args is not None:
+        validation_error = cmd.validate_args(args)
+        if validation_error is not None:
+            console.print(validation_error)
+            session.record("slash", stripped, ok=False)
+            return True
     if policy_precleared:
         session.record("slash", stripped, ok=True)
         return cmd.handler(session, console, args)
--- a/app/cli/interactive_shell/command_registry/investigation.py
+++ b/app/cli/interactive_shell/command_registry/investigation.py
@@ -72,13 +72,20 @@ def _cmd_template(session: ReplSession, console: Console, args: list[str]) -> bo
     return True
 
 
+def _validate_investigate_args(args: list[str]) -> str | None:
+    if not args:
+        return f"[{DIM}]usage:[/] /investigate <file>"
+    return None
+
+
+def _validate_save_args(args: list[str]) -> str | None:
+    if not args:
+        return f"[{DIM}]usage:[/] /save <path>  (e.g. /save report.md or /save out.json)"
+    return None
+
+
 def _cmd_investigate_file(session: ReplSession, console: Console, args: list[str]) -> bool:
     from app.cli.investigation import run_investigation_for_session
-
-    if not args:
-        console.print(f"[{DIM}]usage:[/] /investigate <file>")
-        session.mark_latest(ok=False, kind="slash")
-        return True
 
     path = Path(args[0])
     if not path.exists():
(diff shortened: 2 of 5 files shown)


Mohdtalibakhtar:
Pushed `4c3c16e3` removing the misleading `session.last_state` preset and the comment around it. Agree the line was a no-op for the test's purpose under the new `validate_args`-runs-before-handler ordering — the validator catches missing args regardless of `last_state`, so presetting it was leftover thinking from the old in-handler check. All 5 `TestPrePolicyValidation` cases still pass; full gate green (`6134 passed, 6 skipped`).

@greptile review

Devesh36:
LGTM 👍

Devesh36:
one more review from @yashksaini and then i'll merge this

PR #1737, opened 2026-05-09
feat(agents): /agents trace <pid> live tail view (#1493)

### LINKED ISSUE #1493: `/agents trace <pid>` live tail view

```
> /agents

> /agents trace 8421

> /agents kill 7702
```

|---|---|

---

## This issue: `/agents trace <pid>` live tail view

**Phase:** 1 — Golden signals MVP
**Effort:** M (~2-3 days)
**Depends on:** #1490

### Why we need this

When a user sees an agent acting suspiciously in the `/agents` dashboard — stuck, looping, error rate spiking — they want to drop into a live tail of its tool-call stream **without leaving the `opensre` shell**. This is the equivalent of `kubectl logs -f` but inline in the same REPL, and it's the diagnostic surface users will reach for first when something looks wrong.

### What ships in this issue

**Not a good-first issue** — requires PTY/lsof handling and platform-specific code. Open a Rich `Live` view streaming the agent's stdout buffer inline in the REPL. For OpenSRE-spawned agents, read the captured PTY; for externally-attached PIDs, document the limitation and best-effort tail `/proc/<pid>/fd/1` on Linux.

Inside the `opensre` shell, typing `/agents trace 8421` opens a live, scrolling view of claude-code's stdout in the same terminal until the user hits Ctrl-C, at which point the prompt returns.

### Files to touch

- `app/agents/tail.py` (new) — `TailBuffer` ring buffer + `attach(pid)` async iterator
- `app/cli/interactive_shell/command_registry/agents.py` — `trace` subcommand using Rich `Live`
- `docs/agents.mdx` (new) — document attach limits per platform

### Acceptance

- [ ] `/agents trace 8421` shows live output until Ctrl-C
- [ ] Ring buffer caps memory at 4 MB per agent
- [ ] macOS/Linux behavior documented; Windows surfaced as not supported

### Where to start

Rich `Live` docs: https://rich.readthedocs.io/en/stable/live.html — and look at how `app/cli/commands/general.py`'s `health --watch` flag uses Rich Live.

---

kespineira:
Fixes #1493

Adds the `/agents trace <pid>` subcommand to the OpenSRE interactive shell — a `kubectl logs -f`-style live tail of a registered local AI agent's stdout, inline in the same REPL. This is the diagnostic surface a user reaches for first when an agent in the `/agents` dashboard looks stuck, looping, or noisy.

Only **regular files** backing fd 1 are tail-able. TTY/PTY/pipe/socket/anon_inode/`/dev/null` are rejected at attach time with a precise reason — tailing those would compete with the legitimate consumer for bytes and produce corrupted output. PTY interception for OpenSRE-spawned agents is left to a future change once a spawn path lands.

| Platform | Resolver | Supported |
| --- | --- | --- |
| Linux | `os.readlink("/proc/<pid>/fd/1")` | regular files |
| macOS (best-effort) | `lsof -F ftn -p <pid>` | regular files |
| Windows | — | not supported |

Memory is bounded at every layer: a 4 MiB ring buffer (acceptance criterion), a maxsize-128 queue with drop-oldest on producer overflow, and a 64 KiB on-screen render slice so a burst writer can't peg a core through Rich. A single Ctrl+C returns to the prompt — deliberately different from `streaming.py`'s double-press pattern (LLM streaming wants to ignore stray keypresses; logs-style views want to stop on the first one).

Files touched:

- `app/agents/tail.py` (new) — `TailBuffer`, `AttachSession`, resolvers, `attach()` factory with eager validation
- `app/cli/interactive_shell/command_registry/agents.py` — `_cmd_agents_trace`, `_render_live_tail`, dispatch + completion
- `tests/agents/test_tail.py` (new, 49 tests)
- `tests/cli/interactive_shell/test_agents_commands.py` (extended, +9 tests)
- `docs/agents.mdx` (new) + `docs/docs.json` (Mintlify nav)

Captured by running `dispatch_slash` with a stubbed `attach()` (no real REPL needed):

**Happy path** (`/agents trace 8421` against an agent whose stdout goes to a regular file):

```
trace pid 8421  Ctrl+C to stop
2026-05-09T11:42:00Z investigating mongodb cluster
2026-05-09T11:42:01Z found connection-pool exhaustion
2026-05-09T11:42:02Z proposing rolling restart
· trace ended
```

**Rejection paths** (precise reasons, no traceback, REPL keeps running):

```
cannot trace pid 8421: stdout is on a terminal; live tail not supported
cannot trace pid 8421: stdout is a pipe; live tail not supported
cannot trace pid 99999: no such pid 99999
cannot trace pid 0: invalid pid 0 (must be positive)
```

A live in-REPL recording requires a TTY, but the smoke commands in `docs/agents.mdx` (writing to `/tmp/agent.log` from one terminal and tracing from another) reproduce all four rejection branches plus the happy path on Linux/macOS.

---

The problem: when an agent in `/agents` looks suspicious, the user wants the inline equivalent of `kubectl logs -f` without leaving the REPL. The naive approach — "tail `/proc/<pid>/fd/1` whatever it is" — competes with the agent's real consumer when fd 1 is a TTY/pipe/socket and produces interleaved garbage. So the design separates *resolution* from *tailing*: `_resolve_target` only returns a target when fd 1 is a regular file (Linux via `readlink`, macOS via `lsof -F ftn`); everything else lands in `AttachUnsupported` with a precise, user-facing reason.

Alternatives I considered and rejected:

- **`select()` + `read()` on `/proc/<pid>/fd/1`** — `select()` always reports regular files as readable, busy-loops at EOF.
- **`subprocess` to wrap `tail -f`** — defeats the in-REPL live view, fights process-tree teardown on Ctrl+C.
- **Generator function for `attach()`** — defers `AttachUnsupported` to first iteration, hides errors from the call site. Made `attach()` a regular factory returning an `AttachSession` so validation lands eagerly.
- **Single unbounded buffer** — fails the acceptance memory bound under burst writers. Each layer (queue, ring buffer, render slice) is bounded with drop-oldest semantics so the live view stays fresh.

Key components:

- **`TailBuffer`** caps total bytes and drops *whole* chunks from the head, never splitting a chunk — `decoded()` therefore never breaks a multi-byte UTF-8 codepoint mid-byte.
- **`AttachSession`** is the producer/consumer hand-off: opens the fd, seeks to EOF, runs a daemon reader thread that writes into a bounded queue, and exposes itself as both an iterator and a context manager so resource cleanup is explicit.
- **`_render_live_tail`** wraps `Rich.Live` + `prompt_toolkit.patch_stdout`, swallows a single `KeyboardInterrupt` (kubectl-logs-style), and renders only the most recent 64 KiB of the buffer so burst writers don't make Rich reparse 4 MiB at 10 fps.
- **Logrotate semantics** — the reader's only liveness condition is `pid_exists(pid)`; the pathname is *not* checked, so a `mv agent.log agent.log.1` mid-trace doesn't terminate the view (the open fd still points to the live inode, same as `tail -f`).

Edge cases I deliberately tested or pinned:

- non-positive pids (`0`, `-1`) — rejected before `psutil.pid_exists` is touched (it raises `PermissionError(0)` on macOS, which the slash handler doesn't catch)
- fd 1 is a TTY/pipe/socket/anon_inode/`/dev/null` — each gets a dedicated, distinct error message (a generic "not supported" is unhelpful when debugging)
- file truncated below the read offset — current behavior is "silently miss subsequent writes" (documented as a limitation)
- file renamed mid-trace (logrotate) — reader follows the inode, does not exit
- producer dies during tail — reader exits within one poll interval
- single Ctrl+C — swallowed inside the handler; the REPL prompt comes back, no traceback
- concurrent close + read — fd close is deferred until the reader thread has actually exited, so we never close from under an in-flight `read()`

---

---


kespineira:
@greptile review

kespineira:
Hey! Ready to review @cerencamkiran

cerencamkiran:
Good work @kespineira 👍

Small non-blocking thought: the drop-oldest "get_nowait()" → "put_nowait()" pattern here technically relies on the current “single producer thread per session” invariant.

That seems fine for the current design, but it might be worth leaving a short comment near these queue paths so a future refactor doesn’t accidentally introduce another producer and turn this into a real race window.

(app/agents/tail.py: lines 327–330 in "_reader_loop()" and 339–342 in "_publish()")

kespineira:
Good point @cerencamkiran 👍 Documented the invariant at both queue sites so a future second producer doesn't trip on it

Devesh36:
Looks good , wait till @yashksaini reviews this

rrajan94:
Minor nit before merging: `_reject(reason)` in `tail.py:65` returns an exception instead of raising it — callers do `raise _reject(...)`. Easy to miss the `raise` in a future edit and silently swallow the error. Consider inlining as `raise AttachUnsupported(reason)` at each call site.

Everything else looks great 👍

WatchTree-19 (commented):
Drive-by from someone in the bench-side lane. Live tail for /agents is the right shape, take or leave.

**Tail behaviour under disconnect.** Live-tail UIs in TUI tools have three failure modes worth nailing down: (a) the agent process dies mid-stream, (b) the user's terminal is resized or backgrounded, (c) the underlying log source rotates or truncates. Worth confirming the tail handles all three rather than just stalling silently. The kubectl-style "lost connection, retrying..." inline message is the standard pattern.

**Backpressure on noisy agents.** If an agent is emitting hundreds of lines per second (very possible for a tight loop or a stuck retry), the tail consumer needs a bounded ring-buffer or rate cap. Otherwise the terminal becomes useless and the user can't even Ctrl+C cleanly because the render loop is saturated. Worth a quick check on what happens at, say, 1000 lines/sec.

**PID resolution.** `<pid>` is brittle if the user runs `/agents trace` and the agent has already restarted (PID changed). Worth supporting `/agents trace <agent-id>` (the stable identifier from `/agents list`) as the primary form, with PID as a convenience alias. Otherwise users with crash-loop agents will keep typing the wrong PID.

**Termination semantics.** When the user Ctrl+C's out of the tail, does the agent process keep running? It should — tail is read-only — but worth a sanity check that the cleanup path doesn't accidentally signal the agent. Same applies if the tail is closed via the parent shell (SIGHUP).

**Already covered upstream.** Existing reviewer thread looks like the right shape, no need to pile on.

The /agents surface is becoming the operator UX for local agents — well worth getting these primitives right.

kespineira:
@WatchTree-19 thanks for the look 👍

**(1)** Producer-death and rotation are covered (`test_reader_exits_when_pid_dies`, `test_reader_follows_inode_through_rename`); truncate-below-offset is a documented limit; resize/SIGTSTP are fine because the reader is daemon + drop-oldest keeps memory bounded across suspension. Agreed an inline `· process exited` note before the trailer reads better — adding in a follow-up commit.

**(2)** Memory bounds and Ctrl+C responsiveness are in place (4 MiB `TailBuffer` + 128-slot drop-oldest queue; main thread sits in `queue.get(timeout=0.1)`). What's *not* capped is `live.update()` itself — under a 1k-line/sec stream we'd reparse the 64 KiB slice ~100× more than Rich actually paints. Time-based throttle in a follow-up commit.

**(3)** Fair UX concern, but `/agents kill <pid>` already shipped with PID semantics — making `trace` alone accept agent-id would split the surface. Filing a follow-up to unify `<pid>|<name>` across kill/trace/(future) inspect together rather than diverge here.

**(4)** Read-only by construction — the reader holds an fd to the stdout file, never imports `lifecycle.terminate`, never signals the pid. `with sess:` cleanup only sets the stop event, joins the reader, closes our fd. SIGHUP kills daemon threads (ours); the agent is a separate process and unaffected.

Will link the follow-up commits and issue back here.

yashksaini-coder (approved):
every reviewer concern (Greptile P2, 8 CodeQL findings, WatchTree-19's drive-by) has been resolved by traceable commits, CI is fully green, and the test coverage matches the documented edge cases 

LGTM @cerencamkiran @kespineira @WatchTree-19 @rrajan94

kespineira:
@WatchTree-19 follow-up landed:

- **(1a)** inline `· process exited` trailer when the producer dies → 017e6d38
- **(2)** time-based throttle on `live.update()` matched to `_TRACE_REFRESH_PER_SECOND` → d0e8358a
- **(3)** `<pid>|<name>` unification across `kill`/`trace`/(future) `inspect` → tracked in #1749

(4) confirmed read-only by construction in the previous reply, no code change needed.

rrajan94:
Hey @kespineira, please fix the conflicts and a minor thing:
`test_module_thread_cleanup_under_pytest` ordering is implicit — safe now, but fragile if `pytest-randomly` is ever added.

muddlebee:
@greptile review

muddlebee:
fixed the merge conflicts

app/agents/tail.py
```diff
@@ -0,0 +1,444 @@
...
+                    continue
+                # EOF for now: the *only* liveness condition is the
+                # producer pid. We deliberately do NOT check
+                # ``self.target.path.exists()`` — once the fd is open we
+                # follow the inode through rename/unlink (logrotate
+                # semantics), the same way ``tail -f`` does. If the
+                # path-check were here, a logrotate move would silently
+                # end the trace mid-incident even though the agent is
+                # still writing to the original inode.
+                if not pid_exists(self.target.pid):
```
muddlebee, line 324: When read() hits EOF while pid_exists is still true we only sleep and poll. A process that simply stops writing leaves the trace visually frozen with no hint that it is still idle. Call that out in docs or add periodic idle/reconnect feedback like common log tail UIs.

app/cli/interactive_shell/command_registry/agents.py
```diff
@@ -53,8 +59,52 @@
...
+# writers. A few screens of context is plenty for "what is the agent
+# doing right now"; the full tail is still in ``sess.buffer`` for any
+# future drill-down view.
+_TRACE_RENDER_TAIL_BYTES = 64 * 1024
+
+
+def _render_trace_snapshot(live: Live, sess: AttachSession) -> None:
+    """Pull the bounded snapshot, decode on a UTF-8 boundary, hand to ``Live``."""
+    snapshot = _slice_to_utf8_boundary(sess.buffer.snapshot(), _TRACE_RENDER_TAIL_BYTES)
+    live.update(Text.from_ansi(snapshot.decode("utf-8", errors="replace")))
```
muddlebee, line 86: Traced stdout is fed through Rich Text.from_ansi. Malicious or buggy ANSI sequences can manipulate the viewer terminal the same way unfiltered kubectl logs can. Extend the trust/limitations section or constrain interpretation if hostile agents matter.

muddlebee: added in docs

PR #1743, opened 2026-05-09
feat(error-handling): capture action-execution thread/future failures

### LINKED ISSUE #1466: [error-handling] Capture action-execution thread/future failures (no-more-silent retries)
## Problem

`app/nodes/investigate/execution/execute_actions.py` runs investigation actions in a thread pool and folds every failure into a result struct without logging or Sentry capture:

```82:88:app/nodes/investigate/execution/execute_actions.py
        except Exception as e:
            last_error = e
            if attempt < max_attempts - 1 and _is_transient_error(e):
                ...
                continue
            break
```

```170:179:app/nodes/investigate/execution/execute_actions.py
            except Exception as e:
                results[action_name] = ActionExecutionResult(
                    action_name=action_name,
                    success=False,
                    data={},
                    error=f"Execution failed: {type(e).__name__}: {str(e)}",
                )
```

Outcomes:

- A buggy tool that always raises consumes its retry budget silently, returns `success=False`, and the agent moves on. Sentry sees nothing.
- A pool-level `Exception` (e.g. cancelled future, broken executor) becomes a terse `error` string in the result dict.

This is one of the highest-value places to instrument, because it's the actual hot loop where the agent invokes external systems.

## Proposal

1. **Instrument the retry loop** at L82–88:
   - On each attempted-but-failed iteration, emit a Sentry breadcrumb with `event=action_attempt_failed`, `action_name`, `attempt`, `transient=True/False`.
   - On final failure (loop exits with `last_error`), call `report_exception(last_error, severity="error", tags={...})`.
2. **Instrument the future-collection** at L170–179:
   - Call `report_exception(e, severity="error", tags={"surface": "node", "component": "execute_actions", "action_name": action_name})` before storing the result.
3. **Differentiate transient errors**: `_is_transient_error` already classifies; reuse it to set `severity=warning` for transient (final-attempt) failures, `error` otherwise.
4. **Tests**:
   - Force a tool to raise once and succeed on retry: assert Sentry sees a breadcrumb but not an event.
   - Force a tool to raise on every retry: assert exactly one Sentry event with `attempt` set to `max_attempts`.
   - Force the future to raise: assert Sentry capture once with `action_name`.

## Acceptance criteria

- [ ] Retry loop adds Sentry breadcrumbs for non-final attempts.
- [ ] Final retry exhaustion calls `report_exception` exactly once with the right severity and tags.
- [ ] Future-collection failures call `report_exception` exactly once.
- [ ] Existing tests still pass; new tests cover the three branches above.

## Depends on

#1454

VibhorGautam:
Fixes #1466

adds sentry instrumentation to the retry loop and future handler in `execute_actions.py`

three things:
1. **breadcrumbs on transient retries** - when an action hits a transient error and retries, a sentry breadcrumb gets logged with action name, attempt number, and error so the retry history shows up if it eventually dies
2. **report_exception on final failures** - once retries run out, calls `report_exception` with severity="warning" for transient exhaustion vs severity="error" for non-transient
3. **report_exception on future exceptions** - the ThreadPoolExecutor future handler now captures exceptions that escape the executor

uses `report_exception` from `app/utils/errors.py` (from #1620) to stay consistent with how the rest of the error-handling layer works

```
tests/nodes/investigate/test_execute_actions.py::TestRetryBreadcrumbOnTransientRecovery::test_breadcrumb_emitted_no_sentry_event PASSED
tests/nodes/investigate/test_execute_actions.py::TestRetryExhaustionReportsToSentry::test_transient_failure_reports_warning PASSED
tests/nodes/investigate/test_execute_actions.py::TestRetryExhaustionReportsToSentry::test_nontransient_failure_reports_error PASSED
tests/nodes/investigate/test_execute_actions.py::TestFutureFailureReportsToSentry::test_future_exception_captured PASSED

4 passed in 1.12s
```

---

the retry loop already catches exceptions and decides transient vs not so breadcrumbs go right before the `continue` and the final report goes after the loop exits when it failed - severity is warning for transient exhaustion (network hiccup that lasted too long) vs error for non-transient (actually broken). future handler just wraps the existing except block

---

---

### DIFF
--- a/app/nodes/investigate/execution/execute_actions.py
+++ b/app/nodes/investigate/execution/execute_actions.py
@@ -1,12 +1,17 @@
 """Investigation action execution."""
 
+import logging
 import time
 from collections.abc import Iterable
 from concurrent.futures import ThreadPoolExecutor, as_completed
+from contextlib import suppress
 from dataclasses import dataclass
 from typing import Any
 
 from app.tools.registered_tool import RegisteredTool as InvestigationAction
+from app.utils.errors import report_exception
+
+logger = logging.getLogger(__name__)
 
 
 @dataclass
@@ -43,8 +48,11 @@ def _execute_with_retry(
 ) -> ActionExecutionResult:
     """Execute action with exponential backoff retry for transient failures."""
     last_error = None
+    last_transient = False
+    attempts_made = 0
 
     for attempt in range(max_attempts):
+        attempts_made = attempt + 1
         try:
             kwargs = action.extract_params(available_sources)
             data = action.run(**kwargs)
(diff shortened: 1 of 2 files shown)


VibhorGautam:
@greptile review

VibhorGautam:
@greptile review

WatchTree-19 (commented):
Drive-by from someone in the bench-side lane, take or leave.

**Future-failure capture: which exception types.** `concurrent.futures.Future.exception()` returns the captured exception or None. The dangerous case is when the worker thread itself dies before the Future is set (segfault, OS-level signal, hard kill) — `exception()` returns None and the Future is stuck pending. If the capture relies only on `add_done_callback`, those silent-kill cases never trigger the callback. Worth a watchdog timeout on each Future so a never-resolving one surfaces too, not just exceptioned ones.

**Re-raise semantics.** When a Future failure surfaces in the supervisor, you have two choices: re-raise on the calling thread (preserves stack but kills the supervisor), or capture-and-log-then-continue (supervisor survives but loses the original traceback unless you also `logger.exception()`). The PR description should probably name which behaviour was chosen and why, since downstream callers need to know whether to expect raised exceptions or just log entries.

**Sentry capture coupling.** Action-execution failures are exactly what Sentry wants in production (most actionable error category). If `init_sentry()` is in scope here, the capture path should call `sentry_sdk.capture_exception(exc)` at the same point as the local logging, with `extras` set to the action name + arguments. Otherwise the local log is the only record. Adjacent to shoaib050326's #1619 Sentry coverage work.

**Cancellation vs failure.** `Future.cancelled()` is distinct from `Future.exception()`. A user-initiated cancel shouldn't fire the failure-capture path or the operator gets noise. Worth confirming the capture predicate is `exception is not None and not cancelled`.

**Already covered upstream.** Existing thread + greptile look like the right shape, no need to pile on.

Action-execution failure handling is one of those things that's invisible until production blows up. Good direction to make it explicit.

VibhorGautam:
thanks for the drive-by! watchdog timeout and cancellation handling are solid ideas but feel like separate issues from what #1466 asks for so keeping this scoped to the three capture paths

re sentry coupling - report_exception already wraps capture_exception with extras internally so that path is covered, just going through the shared helper

PR #1794, opened 2026-05-10
refactor(cli): LLM-based intent routing + reorganize interactive_shell into subpackages

davincios:
## Summary

This PR reorganizes `app/cli/interactive_shell/` from a flat module layout into nine purpose-based subpackages and adds an LLM-based intent router for the shell's routing layer. Each new subpackage has one clear responsibility so imports become self-documenting:

| Subpackage | Purpose |
| --- | --- |
| `chat/` | User-facing chat agent + inline help |
| `config/` | REPL config, env loaders, tool catalog |
| `history/` | Persistent history storage + redaction policy |
| `intent/` | Intent parsing, terminal intent, interaction models |
| `orchestration/` | Action planner / executor, execution policy + tier |
| `prompting/` | Follow-up generation, prompt rules, prompt surface |
| `references/` | Docs / AGENTS.md / graph-pipeline reference + chunker |
| `routing/` | Rule-based + LLM-based intent classification |
| `runtime/` | Hot reload, REPL session, tasks |
| `shell/` | Shell execution + policy |
| `ui/` | Rendering, streaming, theme, banner, choice menu, agents view |

Also bundled, since they exercise the same surface:

- **Streaming UX fix** — assistant responses no longer staircase into terminal scrollback when rendered with `rich.Live`; the live preview is now transient and a single static `Markdown` is printed when the stream finishes.
- **Hot-reload safety** — `_run_throttled_markdown_loop` accepts both `preview` and the legacy `set_view` kwarg via a compat shim, so stale callers after a partial hot reload do not raise `TypeError`.
- **RCA prompt + planner hardening** — for RDS resource-threshold alerts the planner now prohibits `query_grafana_traces` (was a soft hint), and the diagnosis directives disambiguate `connection_exhaustion` from `dual_resource_exhaustion` when `Client:ClientRead` waits dominate.
- **Two callsite bugs exposed by the move are fixed**:
  - `config/tool_catalog.py`: `_REPO_ROOT` now uses `parents[4]` to account for the extra directory level, so tool source paths render as `app/tools/registry.py` instead of `tools/registry.py`.
  - `test_history_commands._redirect_history_path` also patches `history.storage.prompt_history_path`, since `clear_persisted_history` now resolves the name in its own module.

## Test plan

Run locally before merging:

- [x] `make lint` — clean
- [x] `make format-check` — clean
- [x] `make typecheck` — 566 source files, no issues
- [x] `make test-cov` — 6358 passed, 5 skipped
- [x] `pytest tests/cli/interactive_shell/` — 733 passed
- [ ] Manual: open the interactive shell, send a short message, confirm the streamed response no longer duplicates into scrollback
- [ ] Manual: `/list tools` shows `app/tools/...` source paths (not `tools/...`)
- [ ] Manual: `/history clear` truncates `~/.opensre/interactive_history`

@greptileai please review.

Made with [Cursor](https://cursor.com)


davincios:
@greptile review

davincios:
@greptile review

Addressed in de354b48:
- **P1 (action_planner repo-root depth)** — bumped `parents[3]` → `parents[4]` in `orchestration/action_planner.py` with a counting NOTE comment so future moves catch the issue.
- **CodeQL mixed imports in `test_action_planner.py`** — dropped the `import ... as action_planner_module` alias and added `_list_rds_postgres_scenarios` to the named-import list; `cache_clear()` fixture still works.
- **CodeQL mixed imports in `test_agent_actions.py`** — kept the module alias (~10 monkeypatch sites need the module object) and removed the bare `from ... import (...)`; qualified all 49 call sites with `agent_actions.`.

Quality gates: lint / format-check / typecheck / 6358 unit tests all green.

davincios:
@greptile review

29adfd86 closes the security observation:
- `_sanitise_text` in both `llm_intent_classifier.py` and `llm_synthetic_scenario_resolver.py` now collapses any 3+ run of `<` or `>` chars into a single space, neutralising the `<<<`/`>>>` delimiter-escape vector.
- Docstrings tightened to scope the claim accurately (no more overclaiming).
- New regression test `test_delimiter_escape_sequences_neutralised_before_prompt` asserts the assembled prompt only ever contains exactly one `<<<` and one `>>>` token from the template — extra copies would mean user input escaped.

davincios:
@greptile review

davincios:
@greptile review

61a34d22 fixes the third instance of the same parents[N] off-by-one:
- `HotReloadCoordinator` default `watch_root` was `parents[3] / "app"` which after the `runtime/` move resolved to `<repo>/app/app/` (nonexistent). `_scan()` returned `{}` and hot reload silently never fired in production.
- Extracted into a `_default_watch_root()` helper using `parents[4]` with an explicit counting NOTE so a future move catches it.
- Added `test_default_watch_root_points_at_app_directory` that asserts the implicit `HotReloadCoordinator()` picks a watch_root that exists on disk; without the fix the test fails.

davincios:
@greptile review

Addressed in 61a34d22 and 0b7ffa1f:

**P1: hot_reload watch_root depth** — fixed in **61a34d22** (the commit Greptile hasn't seen yet; previous review tail commit was d873e38b). Extracted into a `_default_watch_root()` helper using `parents[4]` with an explicit counting comment, and added `test_default_watch_root_points_at_app_directory` that fails without the fix.

**P1: brace-escape gap in `_sanitise_text` — false positive.** Verified empirically: `str.format()` does not recursively interpret substituted values, so user text containing `{...}` tokens passes through as literals and never raises `KeyError`:

```python
>>> "<<<{text}>>>".format(text="see {alert_id}")
'<<<see {alert_id}>>>'   # literal; no KeyError, no double-substitution
```

I exercised the seven concrete inputs in the review (`{alert_id}` interpolations, deploy templates, JSON alert payloads, shell globs with `{1,2,3}`, double-braces, bare `{}`, chained `{a}{b}{c}`) — every one rendered without exception and the LLM client was invoked normally.

Rather than reach for an "escape `{`/`}` in the sanitiser" fix that would actively corrupt user input (`ls *.py {1,2,3}` would become `ls *.py {{1,2,3}}` in the prompt the model sees, which is a visible regression in routing accuracy), commit **0b7ffa1f** pins the correct behaviour with `test_braces_in_user_input_pass_through_without_keyerror`. That test fails the moment any future change tries to escape braces in the sanitiser.

If you still see a path where this raises `KeyError`, please point me at the exact line and input — happy to revisit.

Issue #1679, opened 2026-05-08
[BUG] Persistent input frame so users can type/cancel while streaming


YauhenBichel:
### Problem statement

Related to #1345 

Issue:

REPL loop:

```
async def _run_one_turn(prompt, session, console) -> bool:
    text = await prompt.prompt_async(...)        # ← prompt UI runs HERE
    # text submitted; prompt_toolkit's UI exits
    ...
    answer_cli_agent(text, session, console)     # ← streaming runs HERE; UI is gone
    return True
```

When the model is streaming a response, the input frame should remain rendered at the bottom of the terminal (like Claude Code, ChatGPT desktop, etc.). 

Specifically:

- The user can begin typing the next prompt without waiting for the current stream to finish.
- The user can press Esc or Ctrl+C to cancel the running stream without leaving the REPL.
- Streamed output flows into a scrollable area above the input frame, not into the area where the input frame would have been.

###  Expected as in Claude CLI:

While a stream is in progress, the input frame stays pinned at the bottom; the user can type the next prompt or press a key to cancel. Output flows in the area between the previous prompt and the input frame.

What Claude CLI does? - Claude CLI uses a single long-running prompt_toolkit Application with a multi-area layout:

<img width="536" height="200" alt="Image" src="https://github.com/user-attachments/assets/d8688047-439f-4ca3-b6bb-c7864d2715c7" />

Both the streaming output and the input frame run concurrently in the same async event loop. Pressing Enter doesn't exit the Application — it submits the current input as a message, schedules an LLM stream as a Task, and the Application keeps running. Tokens flow into the output area; the user can type, scroll, or hit Esc to cancel.

@VaibhavUpreti: "Like in claude when you start typing and you press enter" and shared screenshots by @VaibhavUpreti 

<img width="3454" height="842" alt="Image" src="https://github.com/user-attachments/assets/b3d0e120-a314-4ad8-b281-f2187ab2fce1" />

<img width="3456" height="1324" alt="Image" src="https://github.com/user-attachments/assets/6d187916-7224-40d5-98ef-5d5440d5b37b" />

### Actual behavior

The input frame disappears the moment the previous prompt is submitted. Stream output renders where the input frame used to be. The frame only re-appears when the stream completes.

@VaibhavUpreti: "we see that area only when the prompt has ended streaming the response"

[...]

PR #1797, opened 2026-05-11
Use Streamdown for interactive shell Markdown streaming

### LINKED ISSUE #1782: [BUG] Interactive terminal repeats streamed assistant output
### Summary

Assistant responses in the interactive terminal can appear duplicated while streaming.

**Expected:** The assistant response streams once and remains readable.

**Actual:** Parts of the response are printed again as the answer streams, making it look like the assistant is repeating itself.

### Steps to reproduce

1. Open the interactive terminal.

2. Send a few prompts that produce longer assistant responses.

3. Watch the response while it streams.

4. Notice that earlier parts of the same response can appear again before the answer finishes.

Yes

### How often does it occur?

Intermittent

### Operating system

Windows

### Logs and error output

_No response_

### Additional context

Image

**Current Root Cause**
The current renderer streams by repeatedly:

Markdown("".join(buffer))
So every update is the whole answer so far.

That requires Rich.Live to erase the previous rendered Markdown perfectly. With long wrapped output, Windows, prompt-toolkit, and vertical_overflow="visible", old frames can remain. That creates the repeated-output bug.

gbsierra:
Fixes #1782

Fixes duplicated assistant output in the interactive shell.
The old streaming renderer repeatedly re-rendered the full Markdown response (in a Rich Live region). 

This PR changes the interactive shell to use `streamdown` for real terminal Markdown streaming. It keeps the existing `stream_to_console(...)` API, so chat/help/follow-up callers do not need to know about the renderer.

Design decisions:
- Keep streaming enabled.
- Keep rendered Markdown output.
- Use `streamdown` only for real TTY output.
- Keep Rich Markdown as the safe fallback for tests, pipes, and captured output.
- Keep JSON action-plan suppression and Ctrl+C behavior unchanged.
- Keep renderer-specific code isolated in `streaming_markdown.py`.

Added the `streamdown` dependency because this bug comes from treating Markdown like a full-screen redraw. The safer fix is to use a renderer built for incremental Markdown output instead of trying to make terminal clearing more reliable.

**Before**:
image
**After**:
image

---


gbsierra:
@greptile review

gbsierra:
@greptile review

gbsierra:
@greptile review

gbsierra:
@greptile review

YauhenBichel:
Hi @gbsierra  #1782's root cause as described (repeated Markdown(buffer) + Live erase failure) is structurally fixed by #1679 — rich.Live is no longer in the streaming path at all. Could you verify the duplication still reproduces on top of #1679 before merging #1797? Could you please check in main branch if it is still reproducible

gbsierra:
> Hi @gbsierra #1782's root cause as described (repeated Markdown(buffer) + Live erase failure) is structurally fixed by #1679 — rich.Live is no longer in the streaming path at all. Could you verify the duplication still reproduces on top of #1679 before merging #1797? Could you please check in main branch if it is still reproducible

Seems fixed. I can’t reproduce the duplicate output on main.

but im not sure it handles every Markdown edge case though. It still uses a paragraph-based heuristic, not a full streaming Markdown parser. I think the tradeoff is between keeping the current simpler/no new dependency fix versus taking on Streamdown for better streaming Markdown handling.

YauhenBichel:
> > Hi @gbsierra #1782's root cause as described (repeated Markdown(buffer) + Live erase failure) is structurally fixed by #1679 — rich.Live is no longer in the streaming path at all. Could you verify the duplication still reproduces on top of #1679 before merging #1797? Could you please check in main branch if it is still reproducible
> 
> Seems fixed. I can’t reproduce the duplicate output on main.
> 
> but im not sure it handles every Markdown edge case though. It still uses a paragraph-based heuristic, not a full streaming Markdown parser. I think the tradeoff is between keeping the current simpler/no new dependency fix versus taking on Streamdown for better streaming Markdown handling.

Thank you @gbsierra 
Streamdown would be a perfect fit for the parts we render all at once (not streamed): the final RCA report, Slack/PagerDuty message previews, and the daily-update summary. Those don't run inside the interactive shell, so Streamdown gets to do what it's best at without any REPL conflicts

muddlebee:
closing as per comments here https://github.com/Tracer-Cloud/opensre/pull/1797#issuecomment-4421943586

@gbsierra feel free to open a new issue post the latest fixes by team

PR #1827, opened 2026-05-11
Feature/sentry logging integration

### LINKED ISSUE #1455: [error-handling] Wire Sentry LoggingIntegration so logger.exception automatically reports
## Problem

The audit found dozens of operational error paths that already call `logger.exception(...)` or `logger.warning(..., exc_info=True)` but **never** call `app.utils.sentry_sdk.capture_exception`. Examples:

- `app/pipeline/routing.py` ~L75–98 — `should_continue_investigation` defaults to `publish` on any exception with `logger.exception` only.
- `app/nodes/investigate/merge.py` ~L27–37 — OpenSRE telemetry load failure logged with `logger.exception` only.
- `app/remote/vercel_poller.py` ~L839–854 — per-candidate handler and outer poll loop use `logger.exception(...)`, never `capture_exception`.

Even after we migrate the worst offenders to the new helper from #1454, we'll still want a safety net for any remaining `logger.exception` call so nothing slips through.

## Proposal

Wire Sentry's [`LoggingIntegration`](https://docs.sentry.io/platforms/python/integrations/logging/) into `_init_sentry_once` in `app/utils/sentry_sdk.py`:

```python
from sentry_sdk.integrations.logging import LoggingIntegration

sentry_sdk.init(
    ...,
    integrations=[
        LoggingIntegration(
            level=logging.INFO,        # breadcrumbs
            event_level=logging.ERROR, # send as Sentry events
        ),
    ],
)
```

This forwards:

- All `logger.error(...)` and `logger.exception(...)` calls anywhere in `app/` to Sentry as events (subject to `SENTRY_ERROR_SAMPLE_RATE`).
- Lower-severity records as breadcrumbs that get attached to subsequent events.

## Scope

- Update `app/utils/sentry_sdk._init_sentry_once` to register `LoggingIntegration`.
- Add an env switch `OPENSRE_SENTRY_LOGGING_DISABLED=1` so users can opt out (mirrors the pattern of the other Sentry env switches).
- Audit a handful of high-volume `logger.warning(..., exc_info=True)` sites and demote them to `logger.warning(...)` (without `exc_info`) where the warning is intentional — we don't want to spam Sentry on every routinely-flaky integration probe.
- Update `tests/test_sentry_init.py` to assert the integration is registered.
- Document the new env var alongside the existing `OPENSRE_SENTRY_DSN` / `OPENSRE_SENTRY_DISABLED` knobs in `app/utils/sentry_sdk.py` docstring and the appropriate doc page.

## Acceptance criteria

- [ ] `LoggingIntegration` is registered when Sentry is enabled; unit tests assert this.
- [ ] `OPENSRE_SENTRY_LOGGING_DISABLED=1` disables the integration without affecting exception capture from `capture_exception`.
- [ ] No regression on `make test-cov`.
- [ ] Ad-hoc verification: triggering a `logger.exception("test")` in a CLI command results in a Sentry event with the correct release/environment tags.

## Depends on

#1454 (foundation)

CoderHariswar:
Fixes #1455 

Implemented Sentry LoggingIntegration support in app/utils/sentry_sdk.py to automatically forward logger.error(...) and logger.exception(...) events to Sentry.

Changes made:

- Added LoggingIntegration registration inside _build_sentry_integrations()
- Added OPENSRE_SENTRY_LOGGING_DISABLED=1 environment flag to disable automatic logging integration without affecting explicit capture_exception(...)
- Updated init_sentry() docstring/comments for the new env flag
- Added unit test coverage in tests/test_sentry_init.py for:
- LoggingIntegration registration when enabled
- LoggingIntegration exclusion when OPENSRE_SENTRY_LOGGING_DISABLED=1
- Verified the implementation scope against the issue acceptance criteria

https://github.com/user-attachments/assets/3c096c16-c92d-43ee-9070-2aa33c2dff0b

test1
test2

---

This PR adds a centralized safety-net mechanism for operational errors already logged using logger.error(...) and logger.exception(...) but not explicitly captured through capture_exception(...).

I used Sentry’s built-in LoggingIntegration because it automatically forwards error-level log records to Sentry while also attaching lower-severity logs as breadcrumbs. This avoids manually adding capture_exception(...) calls across multiple files and keeps the implementation centralized.

The integration registration is handled inside _build_sentry_integrations() to preserve the existing lazy-loading behavior for Sentry integrations.

I also added the OPENSRE_SENTRY_LOGGING_DISABLED=1 environment switch so users can disable automatic logger forwarding without disabling explicit exception capture behavior.

The added tests validate both:

- LoggingIntegration registration when enabled
- LoggingIntegration exclusion when disabled via environment variable

---

---

### DIFF
--- a/app/utils/sentry_sdk.py
+++ b/app/utils/sentry_sdk.py
@@ -67,7 +67,7 @@ def _is_sentry_disabled() -> bool:
 
 
 def _is_sentry_logging_disabled() -> bool:
-    """Checking Sentry Disable login"""
+    """Return True when automatic Sentry logging integration is disabled."""
     return os.getenv("OPENSRE_SENTRY_LOGGING_DISABLED", "0") == "1"
 
 
--- a/app/utils/sentry_sdk.py
+++ b/app/utils/sentry_sdk.py
@@ -66,6 +66,11 @@ def _is_sentry_disabled() -> bool:
     )
 
 
+def _is_sentry_logging_disabled() -> bool:
+    """Checking Sentry Disable login"""
+    return os.getenv("OPENSRE_SENTRY_LOGGING_DISABLED", "0") == "1"
+
+
 def _sample_rate_from_env(env_var: str, default: float) -> float:
     try:
         sample_rate = float(os.getenv(env_var, str(default)))
@@ -265,11 +270,20 @@ def _build_sentry_integrations() -> list[Any]:
     from sentry_sdk.integrations.httpx import HttpxIntegration
     from sentry_sdk.integrations.logging import LoggingIntegration
 
-    return [
-        LoggingIntegration(level=logging.INFO, event_level=logging.ERROR),
+    integrations: list[Any] = [
         AsyncioIntegration(),
         HttpxIntegration(),
     ]
+
+    if not _is_sentry_logging_disabled():
+        integrations.append(
+            LoggingIntegration(
+                level=logging.INFO,
+                event_level=logging.ERROR,
+            )
+        )
+
+    return integrations
 
 
 @contextmanager
@@ -375,6 +389,8 @@ def init_sentry(entrypoint: str | None = None) -> None:
     env var, then the bundled constant. Set ``OPENSRE_NO_TELEMETRY=1`` or
     ``DO_NOT_TRACK=1`` to disable both Sentry and PostHog product analytics.
     ``OPENSRE_SENTRY_DISABLED=1`` disables Sentry only;
+    ``OPENSRE_SENTRY_LOGGING_DISABLED=1`` disables automatic
+        logger forwarding while preserving explicit ``capture_exception`` calls.
     ``OPENSRE_ANALYTICS_DISABLED=1`` disables PostHog only.
 
     ``entrypoint`` identifies the calling surface (``cli``, ``webapp``,
--- a/app/utils/sentry_sdk.py
+++ b/app/utils/sentry_sdk.py
@@ -389,8 +389,8 @@ def init_sentry(entrypoint: str | None = None) -> None:
     env var, then the bundled constant. Set ``OPENSRE_NO_TELEMETRY=1`` or
     ``DO_NOT_TRACK=1`` to disable both Sentry and PostHog product analytics.
     ``OPENSRE_SENTRY_DISABLED=1`` disables Sentry only;
-    ``OPENSRE_SENTRY_LOGGING_DISABLED=1`` disables automatic logger forwarding
-    while preserving explicit ``capture_exception`` calls.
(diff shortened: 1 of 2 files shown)


CoderHariswar:
@greptile review

WatchTree-19:
drive-by from someone in the eval/observability lane, take or leave.

**The test gap greptile flagged is the merge blocker, and the shape of the fix matters.** patching `_build_sentry_integrations` away in the enabled-case test means the new conditional path is never exercised, so a regression in `_is_sentry_logging_disabled` or the append branch would land green. the right fix is two new tests that call `_build_sentry_integrations` directly (not via `_init_sentry_once`): one with the env var unset asserting `LoggingIntegration` is in the returned list, one with `OPENSRE_SENTRY_LOGGING_DISABLED=1` asserting it is not. `monkeypatch.setenv` / `monkeypatch.delenv` is the cleanest way. that closes greptile's confidence-3 concern.

**`OPENSRE_SENTRY_LOGGING_DISABLED=1` env truthy check is brittle.** if `_is_sentry_logging_disabled` only matches the exact string `"1"`, then `true`, `yes`, `on`, `TRUE` all silently pass through as not-disabled. either match `{"1","true","yes","on"}` case-insensitively (common pattern), or document that exact string `1` is the only accepted value. matters because every other `OPENSRE_*` env flag in the repo should converge on one convention, and a future contributor flipping `true` will assume it works.

**LoggingIntegration default levels.** `sentry_sdk.integrations.logging.LoggingIntegration()` defaults to `level=INFO` (breadcrumbs) + `event_level=ERROR` (sent as events). worth confirming that's what #1455 wanted. if the team is on the "operational errors only" framing in the PR description, `event_level=ERROR` is right. if there's any expectation that `logger.warning` events show up as Sentry events, this needs adjusting. one line of config or one inline comment locks the contract.

**Coordination with #1619 (sentry fingerprinting) and #1743 (sentry instrumentation for action execution).** all three PRs touch the same surface. once `LoggingIntegration` lands, every `logger.exception` from a tool execution path that #1743 also captures via `capture_exception` will be duplicated in Sentry (one as a logged event, one as a captured exception). the dedup is handled by Sentry's event_id matching most of the time, but worth flagging on #1743's thread so the two don't double-report. similarly #1619's fingerprint rules apply differently to logging-integration events vs captured exceptions.

**CI quality is failing.** same blocker on the merge. running `make lint` and `make format` locally should clear it. quality + Windows are the only failing checks otherwise.

small surface, the LoggingIntegration approach is the right call for centralisation (matches the description's "no per-file capture_exception sprinkle" framing). the test rework + env-truthy widening are the two items i'd want closed before merge.

muddlebee:
closing the PR, not planned for now

PR #1830, opened 2026-05-11
synthetic: add 021 RDS CPU near-critical from API surge with no correlation

hamzzaaamalik:
Adds a new Level 4 test scenario based on a real customer pattern:
the database CPU climbs to 92% because of a 5x surge in API requests,
not because of anything wrong inside the database itself.

The tricky part this tests:

The API servers do not show high CPU during the surge because they
are mostly waiting on the database to respond. The only signals that
go up on the API side are network traffic and request count, not CPU.
So the agent cannot just look at "where is CPU high" and find the
answer, since the only place CPU is high is the database. It has to
look at non-CPU metrics on the API side to find the actual cause.

Existing similar scenarios are easier:
- 015 has the web tier CPU also climbing, so the correlation is obvious
- 020 tests the same idea on the Kubernetes side

021 is the version where the agent cannot rely on CPU shape alone.

The scenario also includes a few wrong-answer traps:
- DB connections climb a bit (the easy wrong answer is "connection
  exhaustion") but the connection pool absorbs the surge
- A parameter group change happened six hours before the alert (red
  herring, unrelated)
- One query dominates the database load but the query itself is fine,
  it just runs 5x more often than usual

The agent has to figure out that the load is volume not cost, and
that the cause is upstream.

Validated end to end with a real LLM run. Test passes in one attempt.
Tuned a few details during testing:
- removed a forbidden keyword that was being matched in negated form
  (same trap PR 942 caught)
- accepted a couple of category labels the agent naturally produces
- swapped one keyword for a more reliable one the agent uses
  consistently
- updated the expected tool sequence to match what the agent
  actually does


muddlebee:
image

needs to be fixed

WatchTree-19:
nice work @hamzzaaamalik. 021 + 022 hit exactly the gap on `tests/synthetic/rds_postgres/` where every existing level-4 was DB-internal. the no-trace-ID time-correlation shape is the right hard case.

couple of observations.

one thing not blocking. `optimal_trajectory` puts `query_grafana_metrics` and `query_grafana_logs` *after* `ec2_instances_by_tag` and `get_elb_target_health`. but the rds CPU + ALB RequestCount + EC2 NetworkOut comparison is the diagnostic core of the scenario, and an agent reaching for the cloudwatch metrics last would have already committed to a hypothesis from the topology lookups. 022 (your sibling PR) has the inverse order with grafana_metrics/logs sitting before ec2/elb, which reads more naturally for a load-spike attribution. probably worth aligning the two scenarios on one order so the trajectory-shape grading doesn't accidentally reward different reasoning styles for what is the same failure_mode (`application_load_spike`).

second, `forbidden_categories` includes bare `cpu_saturation`, and `equivalent_root_cause_categories` includes `cpu_saturation_workload_burst` and `cpu_saturation_bad_query`. the scorer in `scoring.py` does exact-string match (`actual_category in answer_key.forbidden_categories`) so the substring overlap is safe today, but it's a footgun if anyone ever loosens that to startswith/contains. worth a one-line comment in the answer.yml saying the bare `cpu_saturation` form is the symptom-only mis-attribution being ruled out, and the `cpu_saturation_*` equivalents are the structured forms that *are* acceptable.

third, smaller. `aws_rds_events.json` is in `available_evidence` but not in `required_evidence_sources`, which i think is the intent (it's the red-herring source for the param-group-change adversarial signal). worth saying that explicitly somewhere - either in the QA_VALIDATION.md or as a comment in the scenario.yml - so a future maintainer reading the scenario doesn't think it's an oversight.

scenario quality is high and the model_response in answer.yml is the cleanest causal-chain narration i've seen in the level-4 set. lgtm once the trajectory order is sorted with 022.

cerencamkiran:
Both 021 and 022 look really solid now and I don’t think fixture correctness is the issue anymore.

My only concern is consistency between them. They’re testing almost the same RCA pattern, but the investigation flow is still a bit different:

- one leans more on topology first,
- the other on metrics/log correlation first.

Small thing, but if these scenarios get used later for regression tracking or provider comparisons, I think sibling scenarios should reward roughly the same reasoning flow.

021 also still has a tiny mismatch: if "mean_call_duration_ms" is not actually visible to the agent, the expected response probably shouldn’t describe it like directly observed evidence.

But overall really nice work on both.

WatchTree-19:
thanks @cerencamkiran. agree on sibling consistency, the regression-tracking value drops fast when two flows for the same failure family diverge.

on the field-visibility: calls_per_sec is fine, it's in TopSQL and on every 020-series fixture. mean_call_duration_ms is real schema drift though, not in TopSQL and not in any other scenario. happy to file the schema follow-up if hamza wants to keep the field in the response narration.

lgtm once trajectory order converges with 022.

PR #1832, opened 2026-05-11
synthetic: add scenario 022 multi-tenant noisy neighbor on shared RDS

hamzzaaamalik:
Adds a new Level 4 RDS test scenario covering the multi-tenant
"noisy neighbor" pattern. Three customers (acme-corp, globex,
initech) share payments-multi-prod. One tenant's API tier surges 8x
and consumes ~78% of the database load. The other two tenants stay
at baseline. RDS CPU climbs to 91%.

The agent has to identify WHICH tenant is responsible, not just
which query is dominating.

What makes this different from existing scenarios:

015 and 021 both test "find the upstream cause" but with one shared
API tier. 022 is the first scenario where the same dominant SQL
pattern is used by every tenant, so the agent has to read PI
top_users and per-target-group ALB request counts to attribute the
load to one specific customer.

This is the first synthetic scenario in the suite where PI top_users
is the load-bearing piece of evidence. It is also the first
multi-tenant scenario.

Wrong-answer traps included:
- The dominating query (orders+order_items JOIN by customer_id) has
  a normal plan and per-call duration. Calling it a "slow query" or
  blaming a schema regression is wrong.
- DB connections climb modestly (78 to 139). cpu_saturation and
  connection_exhaustion are tempting but wrong.
- Without reading PI top_users or per-tenant ALB metrics, the agent
  has no way to identify which tenant is responsible.

Validated end to end with a real LLM run. Single attempt, all gates
clean. Applied lessons from PR #1830 (Greptile P1s) up front:
- forbidden_keywords list does not substring-overlap with any
  equivalent root_cause_category
- optimal_trajectory includes describe_rds_events (matches scenario
  015's pattern)
- pre-commit self-audit verified all keyword/category cross-checks


muddlebee:
image

check https://github.com/Tracer-Cloud/opensre/pull/1832#issuecomment-4422120669

WatchTree-19:
nice work @hamzzaaamalik. the top_users + top_hosts attribution mechanism in answer.yml is a clean way to make the noisy-neighbor pattern solvable without inventing new evidence shapes. the inline comment about PI not being in `optimal_trajectory` because there's no `get_performance_insights` tool in the registry is exactly the kind of context a future maintainer needs and i'd not seen called out before.

one thing not blocking. the topology block has the right escape hatch for the single-string `target_group_arn` field (your comment about it being inherited from 015's shape and the agent discovering the other two tenant TGs via `get_elb_target_health`). the comment helps but the schema gap is real - any future multi-target-group scenario will hit the same workaround. worth filing a separate small issue to grow `target_group_arn` to a list field in the topology schema. happy to file that one if it's useful, it's the kind of follow-up that gets lost otherwise.

pair-wise nit. 021 has the opposite `optimal_trajectory` ordering (grafana_metrics/logs after ec2/elb). same failure_mode, same difficulty, sibling PR. left a longer note on 1830 about aligning - flagging here so both PRs converge on whichever order you prefer.

lgtm modulo the alignment with 1830.

WatchTree-19:
hey @hamzzaaamalik just filed the follow-up on the topology.target_group_arn schema gap from my #1832 review as #2100. happy to loop you in if needed.

PR #1843, opened 2026-05-11
fix(interactive-shell): refuse interactive wizards (onboard, integratons setup) with clear handoff message

### LINKED ISSUE #1841: [BUG] Run 'opensre onboard': wizard renders incorrectly when triggered from inside the interactive shell. Options are not clickable. Input keeps focus
### Expected
Radio widget renders in place, arrow keys update the highlighted option, Enter selects it.
Run 'opensre onboard': wizard renders incorrectly when triggered from inside the interactive shell. Options are not clickable. Input keeps focus 

### Actual
Each redraw of the radio widget is appended below the previous one — the screen fills with stacked copies of `Quickstart` / `Advanced` rows. Selection is impossible.

### Likely cause
`questionary` (used by the wizard) embeds its own prompt_toolkit Application. The interactive shell already has a prompt_toolkit Application running under `patch_stdout` (the persistent input frame from #1679). Two prompt_toolkit apps cannot cleanly share the same terminal.

### Additional observation
Every output line carries a `d76b524a stdout` prefix, suggesting the remote-stream renderer (#1415) is also wrapping local-onboard output where it shouldn't.

### Workaround
Run `opensre onboard` in a separate terminal before launching the interactive shell.

### Screenshot

Image

### Expected
Radio widget renders in place, arrow keys update the highlighted option, Enter selects it.

### Actual
Each redraw of the radio widget is appended below the previous one — the screen fills with stacked copies of `Quickstart` / `Advanced` rows. Selection is impossible.

### Steps to reproduce

1. Fresh install: `pip install opensre` (or `uv sync`)
2. Run `uv run opensre` (or `opensre`)
3. Shell auto-launches onboarding wizard
4. Run 'opensre onboard'
5. Try to interact with the "How do you want to get started?" radio prompt

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

_No response_

### Additional context

_No response_

YauhenBichel:
*opensre onboard* and *opensre integrations setup* are full-TTY interactive wizards driven by questionary. When invoked from inside the persistent REPL (interactive CLI) — either via command as /onboard or via natural-language intent that the router classifies as an onboard action — the focus is in input textbox.

Piped path (background task): the wizard writes cursor-control escapes to stdout, but stdout is a pipe in this case. The line-relay strips ANSI escapes and prefixes every line with <task_id> stdout │. Each redraw of the radio widget arrives as a new plain-text line, so the screen fills with stacked copies of Quickstart / Advanced rows. Selection is impossible.
Inherited-stdio path (/onboard): both apps write cursor escapes to the same TTY. Rendering corrupts in different ways depending on terminal.

Fixes #1841

- app/cli/interactive_shell/orchestration/action_executor.py: new _INTERACTIVE_OPENSRE_SUBCOMMANDS constant + guard in run_opensre_cli_command that catches onboard and integrations setup.
- app/cli/interactive_shell/command_registry/cli_parity.py: _cmd_onboard no longer spawns the subprocess — prints the same message and preserves user-supplied args (e.g. /onboard local_llm → suggests opensre onboard local_llm).
Tests for both refusal paths, plus a regression guard that integrations list still works normally (the guard is scoped to setup only).

Before fix:
Screenshot 2026-05-11 at 8 22 26 PM

After fix:

https://github.com/user-attachments/assets/0837c9ba-4257-4080-b2d8-7b4ab9a4d440

---

- What problem does your code solve? 
Some commands do not work in interactive CLI. They need CLI without input

So, refuse interactive subcommands at the dispatcher with a clear handoff message pointing at the correct invocation.

---

---


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

VibhorGautam (commented):
clean design - the data-driven set for wizard detection means adding a new interactive command is a one-line change. the case-insensitive lookup and two-token matching for integrations setup handle input variations well

one thing i noticed - the wizard refusal returns True (command handled) while _OPENSRE_BLOCKED_SUBCOMMANDS returns False (command failed). thats the right call since the wizard exists but cant run in this context, vs a blocked command that shouldnt run at all. might be worth a brief inline note on the return True to surface that intent for the next contributor who adds a command restriction

the anti-regression test for integrations list not being blocked is a nice guard against over-broad refusal

WatchTree-19:
drive-by from someone in the cli/ux lane, take or leave.

**Pairs well with #1785.** that PR keeps the input frame editable WHILE a response streams; this PR refuses interactive wizards from spawning a competing prompt-toolkit app inside the same REPL. between them the full surface (in-REPL streaming + subprocess-wizard handoff) is consistent. worth cross-linking in the PR body so the next reviewer sees the two as one design rather than two unrelated fixes.

**`opensre onboard local_llm` is an edge case worth thinking about.** the case-insensitive two-token matching on `integrations setup` is the right shape for the multi-token wizard, but `onboard` matches as a single token regardless of subcommand. `onboard local_llm` (the Zero-config-via-Ollama path that has no questionary prompts, just runs) will currently hit the same refusal as `onboard` interactive. that's almost certainly the right call (the message points at the correct invocation either way), but worth either explicitly testing `onboard local_llm` in the refusal path so the behaviour is documented, or special-casing `onboard local_llm` to fall through since it has no full-TTY ask. flagging because i just filed #1876 proposing to promote `onboard local_llm` as the headline quickstart path, so the in-REPL UX for that specific subcommand becomes a real footgun if the handoff message implies "you need a fresh terminal" when actually the subcommand would just work.

**Handoff message quality is the user-facing surface.** can't read the exact message text from the diff summary alone. if it says something like "this command needs to run in a standalone terminal — exit this shell and run `opensre onboard`" then perfect. if it just says "blocked: interactive wizard" without naming the recovery, that's a worse-UX state than the original bug because the user doesn't know what to do next. worth confirming.

**`return True` vs blocked-subcommand `False` distinction.** +1 to VibhorGautam's comment: the inline note is worth it. the next contributor adding a new restriction will read this code as the template, so the semantic difference between "exists but not here" vs "blocked entirely" should be commented at the return site, not just in the docstring.

clean design overall, the data-driven set is the right shape and the parametrised refusal tests lock the contract. the local_llm sub-path and the handoff-message wording are the two items worth confirming before merge.

YauhenBichel:
@VibhorGautam @WatchTree-19 thanks both.

I added a comment

// True = wizard exists and was handed off; ``_OPENSRE_BLOCKED_SUBCOMMANDS`` returns False for "shouldn't run at all".

// True = wizard exists and was handed off; the ``_OPENSRE_BLOCKED_SUBCOMMANDS`` branch above returns False for "shouldn't run at all".

YauhenBichel:
@greptile review

muddlebee:
LGTM, nice catch :+1:

PR #1848, opened 2026-05-11
fix(cli): /cancel stops the spinner when worker is parked on a confirmation

davincios:
## Summary

Cancelling a dispatch (via `Esc`, `/cancel`, `/stop`, or `/abort`) now actually stops the agent's worker thread — not just the visible spinner.

### The bug

You typed `/cancel` while parked on a `Proceed? [y/N]` confirmation. After the first commit on this branch, the spinner stopped. But you correctly noticed the agent was still working on the task — `pondering…` reappeared with a higher elapsed counter. **Esc had the exact same bug.**

### Root cause

When the cancel signal fired, `_route_confirm_through_prompt` returned the empty string `""` to its caller. `execution_policy.execution_allowed` then evaluated:

```python
if answer not in {"", "y", "yes"}:
```

…and `""` IS in that set (because the upstream prompt is `[Y/n]` with capital-Y default), so the worker treated cancellation as a confirmation and ran the action. The `Task.cancel()` on the asyncio side stopped the visible spinner; the worker thread kept executing in the background, output and all.

### The fix

1. **`DispatchCancelled` exception.** `_route_confirm_through_prompt` now raises it instead of returning `""` when the cancel event fires (or when the confirm event was published without a delivered answer — that's the `cancel_current_dispatch` fingerprint). The exception propagates out of `execution_allowed` and the surrounding action loop, so the in-flight action never runs.
2. **`_run_one_dispatch` catches `DispatchCancelled`** and prints the standard `· interrupted` line, matching the `Esc`-on-streaming UX.
3. **`execute_cli_actions` checks `console.cancel_requested`** at the top of each action iteration, so multi-action plans halt after the current action when the user cancels mid-plan.
4. **Prompt-loop intercept** (kept from the first commit): bare `/cancel` / `/stop` / `/abort` while a dispatch is running routes through `state.cancel_current_dispatch()` instead of being queued behind the parked dispatch. `/cancel <task_id>` (with arguments) is intentionally unchanged — that targeted background-task cancel still flows through normal slash dispatch.

### Reproduction

1. Trigger any action plan that asks for `Proceed? [y/N]` confirmation (e.g. `opensre deploy ec2 --help`).
2. Without answering, type `/cancel` (or hit `Esc`).
3. **Before this PR:** spinner stops with `· interrupted`, but the worker thread silently runs the action anyway. With `/cancel`, the spinner doesn't even stop because the slash got queued behind the parked dispatch.
4. **After this PR:** worker raises `DispatchCancelled`, the action never runs, the dispatch ends cleanly, and the prompt becomes responsive immediately.

## Changes

- `app/cli/interactive_shell/loop.py`
  - New `DispatchCancelled(Exception)`.
  - `_route_confirm_through_prompt` raises `DispatchCancelled` on cancel (replaces the `""` return).
  - `_run_one_dispatch` catches `DispatchCancelled` alongside `asyncio.CancelledError`.
  - `_looks_like_cancel_request()` recognizer + intercept of bare `/cancel` / `/stop` / `/abort` in the prompt loop.
- `app/cli/interactive_shell/orchestration/agent_actions.py`
  - Top-of-loop `console.cancel_requested` check in `execute_cli_actions` so multi-action plans break after the current action.
- `tests/cli/interactive_shell/test_loop.py`
  - `TestLooksLikeCancelRequest` — recognizer coverage.
  - Updated three `TestRouteConfirmThroughPrompt` cancel-path tests for the new raise-based contract.
  - New `TestExecutionAllowedRespectsDispatchCancelled` pins the end-to-end contract AND documents the original footgun (an empty-string return IS silently confirmed as YES, which is exactly why the cancel path must raise).
- `tests/cli/interactive_shell/orchestration/test_agent_actions.py`
  - `test_execute_cli_actions_skips_remaining_actions_when_cancelled` pins the multi-action between-iterations break.

## Test plan

- [x] `make lint` — passes
- [x] `make format-check` — passes
- [x] `make typecheck` — passes (576 source files, no issues)
- [x] `uv run pytest tests/cli/` — 1371 passed

### DIFF
--- a/app/cli/interactive_shell/loop.py
+++ b/app/cli/interactive_shell/loop.py
@@ -154,6 +154,26 @@ def _looks_like_cancel_request(text: str | None) -> bool:
     return (text or "").strip().lower() in _CANCEL_REQUEST_TOKENS
 
 
+class DispatchCancelled(Exception):
+    """Raised in the dispatch worker thread when the user interrupts
+    (Esc / ``/cancel`` / ``/stop`` / ``/abort``) and the worker is
+    parked on a ``Proceed? [y/N]`` confirmation prompt.
+
+    Closes a real footgun in the cancel path: ``execution_policy``
+    treats an empty answer as **YES** (the upstream prompt is
+    ``[Y/n]``, capital Y), so if the cancel handler returned ``""``
+    the worker would happily run the action it was supposed to
+    interrupt — only the spinner would stop, the agent would keep
+    going. Raising propagates out of ``execution_allowed`` and the
+    surrounding action loop, so the in-flight action never runs and
+    any remaining actions in the same plan are skipped.
+
+    Caught by :func:`_run_one_dispatch` so the user just sees the
+    standard ``· interrupted`` line and the prompt becomes responsive
+    again, identical to a clean Esc on a streaming response.
+    """
+
+
 def _looks_like_correction(text: str) -> bool:
     """True when text begins with a short correction cue (intervention signal)."""
     stripped = text.lstrip()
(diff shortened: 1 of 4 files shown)


muddlebee:
@greptile review

PR #1849, opened 2026-05-12
fix(remote): report swallowed runtime fallback errors

### LINKED ISSUE #1471: [error-handling] Add Sentry capture to remote runtime helpers (Vercel poller, IMDS, system metrics, Slack)
## Problem

The remote runtime helpers under `app/remote/` have several silent failure paths despite `app/remote/server.py` itself being well-instrumented at the top level.

### `app/remote/vercel_poller.py`

```python
# L314–316 — config resolve
except Exception:
    logger.warning(...)
    return None
# L839–844 — per-candidate handler
except Exception:
    logger.exception(...)
# L853–854 — outer poll loop
except Exception:
    logger.exception(...)  # continues loop
```

`logger.exception` writes locally — never to Sentry. The poller is a long-running async task; if it silently fails to handle a Vercel candidate, we lose investigations.

### `app/remote/server.py` — IMDS helpers

```python
# L600–601, 610–611 — IMDS metadata fetch
except <narrow exc>:
    return None
```

If we're running on EC2 and IMDS fails, we transparently degrade. That's fine — but no telemetry means we can't dashboard "% of requests where IMDS failed".

### `app/remote/server.py:_check_llm_connectivity` — L634–638

`# noqa: BLE001` returns a "failed" check object. Bedrock / OpenAI errors during health checks are **not** sent to Sentry, only embedded in `/health/deep` JSON.

### `app/remote/system_metrics.py`

```python
# L93–94, L182–183, L257–258
except Exception:  # noqa: BLE001
    return None
```

Each broad-except drops a metric section silently.

### `app/remote/slack_context.py`

```python
# L68–70
except Exception:
    logger.warning(...)
    return {"error": ...}
```

No Sentry capture.

### `app/remote/client.py`

L157–158, L186–187, L204–205, L271–272 — `# noqa: BLE001` → fallback returns. Health, version, deep checks, and preflight all silently degrade.

## Proposal

1. **Vercel poller**: replace each `logger.exception(...)` and `logger.warning(...) + return None` with `report_exception(..., severity="error", tags={"surface": "remote_server", "component": "vercel_poller", "event": <event>, "candidate_id": <when known>})`. The two long-running loops should also wrap themselves in a top-level `report_and_swallow` so an unhandled error in *new* code can't kill the task silently.
2. **IMDS / system_metrics**: capture at `severity="info"` (sampled) so high-cardinality but useful "we couldn't fetch X" trends end up in Sentry without flooding.
3. **`_check_llm_connectivity`**: capture at `severity="warning"` because LLM connectivity issues are operational. Health check still returns the same structured payload.
4. **`slack_context`**: capture at `severity="warning"`; Slack hook failures should be visible to ops.
5. **`remote/client.py`**: capture at `severity="warning"` for preflight / health / version probes; preserve the fallback behaviour.

Add tests under `tests/remote/` that, for each call site, force the underlying call to raise and assert capture happens with the right tags.

## Acceptance criteria

- [ ] All listed sites in `app/remote/` route exceptions through `report_exception`.
- [ ] The Vercel poller is wrapped so a new bug cannot kill the task silently.
- [ ] Sentry events include `surface=remote_server` and a precise `component`/`event` tag.
- [ ] Tests cover each migrated site.

## Depends on

#1454

gbsierra:
Fixes #1471

Added Sentry reporting for remote runtime fallback paths that were previously only logged or silently ignored.

This covers Vercel polling, IMDS metadata lookup, Bedrock health checks, system metrics, Slack thread fetching, and remote client health/preflight probes. The existing fallback behavior is preserved; failures still degrade gracefully, but now they are reported with consistent remote tags.

Acceptance criteria:
All listed sites in app/remote/ route exceptions through report_exception.
The Vercel poller is wrapped so a new bug cannot kill the task silently.
Sentry events include surface=remote_server and a precise component/event tag.
Tests cover each migrated site.
make lint && make typecheck && make test-cov pass.
( make lint and make typecheck pass.
 make test-cov attempted; unrelated Windows socket.AF_UNIX failures in tests/agents/test_bus.py. Targeted remote tests pass.)


gbsierra:
@greptile review

gbsierra:
@greptile review

gbsierra:
@greptile review

gbsierra:
@greptile review

VibhorGautam (commented):
clean approach - the report_remote_exception wrapper that bakes in surface: remote_server removes repetition across the module, and the per-process dedup in _report_remote_once is smart for noisy fallback paths like IMDS where the same error fires every health cycle

one thought on the dedup sets - theres no reset path, so if a transient failure resolves and later recurs (different root cause, same event name), the second occurrence gets silently swallowed. might be worth either a TTL-based expiry or clearing the relevant key on a successful probe so recurring failures can surface again

test coverage is solid - test_imds_get_reports_failure_once calling twice and asserting called_once() is exactly the right way to pin the dedup behavior

gbsierra:
> one thought on the dedup sets - theres no reset path, so if a transient failure resolves and later recurs (different root cause, same event name), the second occurrence gets silently swallowed. might be worth either a TTL-based expiry or clearing the relevant key on a successful probe so recurring failures can surface again

Good point. I’m leaning toward clearing the dedupe key after a successful metadata check, so we report once per failure episode instead of once per process. That seems cleaner than a timer here because a successful check tells us the issue recovered.

Do you agree, or is there a better fix for this?

WatchTree-19:
drive-by from someone in the observability / audit lane, take or leave.

**Clear-on-success is the right call, agree with gbsierra's reply to VibhorGautam.** TTL gives you "report once per N-second window" which is arbitrary; clear-on-success gives you "report once per failure episode," which is the actual signal an operator wants. concrete shape:

```python
def _report_remote_once(event: str, ...):
    if event in _REPORTED_REMOTE_EVENTS:
        return
    _REPORTED_REMOTE_EVENTS.add(event)
    report_remote_exception(event, ...)

def _mark_remote_recovered(event: str):
    _REPORTED_REMOTE_EVENTS.discard(event)
```

then call `_mark_remote_recovered` at every success path that matches a dedup key. yes it adds one line per success site, but each pair is local to the surface (IMDS, Bedrock health, metric collector) so review burden is per-surface, not cross-cutting.

**Coordination with the other Sentry PRs in flight.** #1619 (fingerprint rules + tool exception capture), #1743 (action-execution thread/future failures), #1827 (LoggingIntegration), and this PR all touch Sentry instrumentation. each surface has its own dedup / capture story. once they all land, the fingerprint-rule layer in #1619 becomes the canonical place to deduplicate at the Sentry side, and the per-module process-local dedup sets here become belt-and-braces. worth a tracking note in #1619's body so the maintainer can decide whether to keep both layers or simplify once everything's in.

**Process-level dedup semantics for the CLI path vs the server path.** the dedup sets are module-globals, so per-process. for the CLI (`opensre investigate` invocations), every fresh invocation gets a clean dedup set, which is correct: each user session = one report. for `app/remote/server.py` running as a long-lived process, the same dedup persists across many client requests, which is the noisy-fallback scenario this is designed for. fine for both, just worth a one-line comment at the module-global declaration noting that the lifetime is "process lifetime" and that the long-lived case is the intended one.

**Cross-link to #1485 (hash-chained audit log).** the events being reported here (Vercel poll failure, IMDS metadata lookup miss, Bedrock health probe fail, Slack thread fetch fail) are exactly the cross-system actions the audit-chain wants to record for investigation timelines. once #1485 lands, each call to `report_remote_exception` is a natural hook point for emitting an audit entry alongside the Sentry event. not in scope here, just flagging so the lane stays coordinated.

clean change overall. test coverage on every site is the right shape (`called_once` + `called_once` after reset = the dedup contract is locked). clear-on-success + the cross-PR coordination note are the only two items i'd want closed before merge.

gbsierra:
@greptile review

gbsierra:
> . clear-on-success

I updated this so a successful check resets the “already reported” marker. That means we still avoid repeated reports while the same problem is ongoing, but we will report it again if it recovers and later fails again. I also added a short comment and tests for that behavior.

muddlebee:
@gbsierra need to fix conflicts otehrwise good to go

Issue #1469, opened 2026-05-06
[error-handling] Add Sentry capture to integration verification adapters


davincios:
## Problem

`app/integrations/_verification_adapters.py` is the single home for "is this integration healthy?" probes used by `make verify-integrations`. Every probe collapses any exception into a string result without logging or Sentry capture:

| Lines | Site |
|------:|------|
| 107–113 | Probe verifier `build_config` failure |
| 167–176 | Grafana datasource discovery (`# noqa: BLE001`) |
| 207–208 | AWS STS check |
| 249–250 | Slack webhook POST |
| 262–263 | Tracer JWT decode |
| 274–275 | Tracer API list integrations |
| 288–289 | Discord `discord.py` ImportError (acceptable to keep silent) |
| 302–306 | Discord `# noqa: BLE001` |
| 319–320 | Telegram `getMe` |

When verification reports "failed: connection refused", that's surfaced to the operator. Good. But:

1. The full traceback never reaches Sentry, so we can't dashboard verification failure rates across all installs.
2. When the probe itself bugs out (our code, not vendor), it looks identical to "vendor is down".

## Proposal

1. Wrap every probe's body in:

```python
try:
    return _probe_<vendor>(...)
except Exception as exc:
    report_exception(exc, logger=logger, severity="warning",
                     tags={"surface": "integration", "event": "verify_failed",
                           "integration": vendor})
    return result(vendor, source, "failed", str(exc))
```

2. **Distinguish vendor failures from probe bugs**: catch known vendor exceptions (`httpx.HTTPStatusError`, `botocore.ClientError`, `slack_sdk.errors.SlackApiError`, etc.) and tag `event=vendor_failure` with `severity=info` (we still want a Sentry record, sampled, but at low volume). Anything else → `severity=error`.

3. **Discord ImportError** (L288–289) is genuinely OK to drop — `discord.py` is an optional extra. Keep the existing return but add a one-shot `set_tag("integrations.discord.import_failed", True)` so we know it happened.

4. **Add tests** in `tests/integrations/test_verify.py` that, for each adapter, force the probe to raise and assert Sentry sees a single event with the documented tags.

## Acceptance criteria

- [ ] Every adapter in `_verification_adapters.py` reports exceptions to Sentry, classifying vendor failures vs probe bugs.
- [ ] Discord ImportError remains acceptable but is tagged.
- [ ] Tests cover the failure path of each adapter.
[...]

PR #1922, opened 2026-05-13
fix(services): standardize error telemetry across service clients

### LINKED ISSUE #1458: [error-handling] Standardize service-client error telemetry across app/services/*/client.py
## Problem

Almost every service client in `app/services/*/client.py` follows the same pattern:

```python
try:
    response = httpx.post(...)
    response.raise_for_status()
    ...
except Exception as exc:  # noqa: BLE001
    return {"success": False, "error": str(exc)}
```

There is **no logger call**, **no Sentry capture**, and **no integration tag**. Production has zero visibility into how often vendor APIs are failing the agent.

### Concrete sites

| File | Lines | Pattern |
|------|------:|---------|
| `app/services/jira/client.py` | ~84–86, 110–112, several methods | `except Exception` → warning + `{success: False, error}` |
| `app/services/notion/client.py` | ~102–104, 126–128 | Same |
| `app/services/datadog/client.py` | ~120–127 | Same |
| `app/services/elasticsearch/client.py` | ~117–121, 148–152 | Same |
| `app/services/prefect/client.py` | several | Same |
| `app/services/opsgenie/client.py` | several | Same |
| `app/services/alertmanager/client.py` | several | Same |
| `app/services/vercel/client.py` | several endpoints | Same |
| `app/services/honeycomb/client.py` | 90–91, 113–114, 142–143, 160–161 | `# noqa: BLE001` → error dict, no logging |
| `app/services/coralogix/client.py` | 220–221 | `# noqa: BLE001` → error dict, no logging |
| `app/services/splunk/client.py` | 108–109, 173–174 | `# noqa: BLE001` → error dict, no logging |
| `app/services/victoria_logs/client.py` | 150–152 | `# noqa: BLE001` → warning + error dict |
| `app/services/argocd/client.py` | 221–223, 254–260, 304–310 | `# noqa: BLE001` → warning + `_error_result` |
| `app/services/aws_sdk_client.py` | 257–265 | `except Exception` → `{success: False, error}`, **no log** |
| `app/services/lambda_client.py` | 147–148 | broad catch surfaces in `extract_error` only |

## Proposal

Define a shared mix-in / decorator in `app/services/_base.py` (new) that wraps a service-client method and:

1. Logs at `warning` for vendor 4xx, `error` for vendor 5xx and unexpected exceptions.
2. Calls `app.utils.errors.report_exception` with tags `surface=service_client`, `component=<module>`, `integration=<vendor>`, `method=<name>`.
3. Distinguishes `httpx.HTTPStatusError` (vendor problem, expected) from any other `Exception` (almost certainly our bug — should be reported at `error` level).
4. Preserves the existing `{"success": False, "error": str(exc)}` return shape for backward compatibility.

Example:

```python
from app.utils.errors import report_exception

class ServiceClient:
    integration: str  # "jira", "notion", ...

    def _capture_method_error(
        self, exc: BaseException, *, method: str, severity: str = "error",
    ) -> dict[str, Any]:
        report_exception(
            exc,
            logger=self._logger,
            message=f"[{self.integration}] {method} failed",
            severity=severity,
            tags={
                "surface": "service_client",
                "component": type(self).__module__,
                "integration": self.integration,
                "method": method,
            },
        )
        return {"success": False, "error": str(exc)}
```

Then convert each `except Exception: return {"success": False, "error": str(exc)}` to:

```python
except httpx.HTTPStatusError as exc:
    return self._capture_method_error(exc, method="get_issue", severity="warning")
except Exception as exc:
    return self._capture_method_error(exc, method="get_issue")
```

## Out of scope

- The `make_*_client` factory pattern is handled in #1459.
- NDJSON / streaming JSON parsers are handled in #1460.
- Grafana service has its own quirks (datasource discovery + tempo + loki + mimir) and gets its own issue (#1461).
- AWS clients (`aws_sdk_client.py`, `lambda_client.py`) get their own issue (#1462) because they have additional HTTP/JSON concerns.

## Acceptance criteria

- [ ] All listed clients are migrated to the shared error-capture helper.
- [ ] Every previously-silent broad-except path now produces a Sentry event with `integration` and `method` tags.
- [ ] Unit tests assert that a forced exception in each client method calls `capture_exception` exactly once.
- [ ] No regression in existing happy-path tests.

## Depends on

#1454 (foundation)

VibhorGautam:
Fixes #1458

adds a shared `capture_service_error()` helper in `app/services/_error_helpers.py` that wraps `report_exception` with `surface=service_client` tags. severity logic: 5xx (vendor-side, e.g. 502/503) gets `warning`, everything else (4xx auth failures, timeouts, connection errors) gets `error`. this way a bad API key or expired token hits Sentry as an error, not a silent warning.

migrates 13 of 15 service clients listed in the issue from ad-hoc `logger.warning()` calls to the shared helper:
- jira, notion, datadog (sync + async), honeycomb, splunk, coralogix, elasticsearch, prefect, opsgenie, alertmanager, vercel, victoria_logs, argocd

**intentionally out-of-scope:** `aws_sdk_client.py` and `lambda_client.py` use boto3/botocore exceptions (ClientError, NoCredentialsError, ParamValidationError) which are a different error hierarchy from httpx. migrating those warrants a dedicated PR with boto-aware severity logic rather than shoehorning them into the httpx-based helper.

the helper accepts an optional `extras` dict so callers can forward request-specific context that was previously baked into ad-hoc log messages:
- opsgenie: `alert_id`
- prefect: `flow_run_id`, `work_pool_name`
- vercel: `deployment_id`, `project`, `project_id`, `state`
- jira: `issue_key`
- datadog/elasticsearch: `query`, `time_range_minutes`
- victoria_logs: `query`
- argocd: `application_name`

the `method` key in extras is protected and cannot be overwritten by caller-supplied extras (test included).

return shapes (`{"success": False, "error": ...}`) are unchanged across all clients so callers are unaffected.

vercel `get_runtime_logs` only reports telemetry after retry exhaustion, not on each intermediate attempt, to avoid false failure noise when retries eventually succeed.

also renames exception variables from `e` to `exc` for consistency with the rest of the codebase.

---

## Acceptance Criteria

| Criterion | Status |
|-----------|--------|
| Shared helper that logs + calls `report_exception` with `surface=service_client` tags | done - `capture_service_error()` in `_error_helpers.py` |
| Severity split: 5xx vendor = warning, other = error | done - `_is_server_error()` checks `status_code >= 500` |
| Every client in the issue migrated or explicitly scoped out | 13/15 migrated, 2 boto3 clients scoped out with rationale |
| Per-request context preserved in Sentry extras | done - extras forwarded from every client that had context in old logs |
| Reserved `method` key protected from caller overwrite | done - test `test_caller_extras_cannot_overwrite_method` |
| Existing return shapes unchanged | done - `{"success": False, "error": ...}` preserved everywhere |
| `make lint && make typecheck && make test-cov` pass | done - 11 helper tests pass, 6815/6837 suite passes (22 pre-existing bus.py failures) |

## Demo Video

https://github.com/user-attachments/assets/4195453b-bfea-49b1-ae75-05aa9d2bb546

---

I designed the approach and wrote the helper based on looking at how `report_exception` works in `app/utils/errors.py` and what each client was doing differently. used copilot for some of the repetitive migration edits across the 13 clients but reviewed and adjusted each one manually since the except block shapes varied (some had retry loops, some had special 404 handling, some had context in the log messages that needed to be preserved as extras).

the core decision was standalone helper vs base class method. most clients dont share a base class so a function was simpler and avoids inheritance changes. severity split came from the issue description itself - HTTP 5xx means the vendor is having problems (warning, not our fault), while 4xx (bad creds, missing scopes) and connection/timeout errors are things we need to fix on our end (error).

the `extras` parameter was added after review feedback flagged that some clients had per-request context (query strings, alert IDs, deployment IDs, etc.) in their old log messages that got lost in the initial migration. the optional dict lets callers forward that context without changing the helper's core signature. the `method` key is set last so caller-supplied extras can't accidentally overwrite it.

for argocd specifically, the existing `_error_result` return-shape helper is kept intact - `capture_service_error` is called before it for telemetry, preserving the redaction logic argocd already had.

for vercel, the retry loop was tricky because the original code reported on every failed attempt. moved the telemetry call to after exhaustion so sentry doesnt get spammed with false failures when the next retry succeeds.

aws_sdk_client and lambda_client were intentionally left out because they use boto3 exceptions (ClientError, NoCredentialsError, ParamValidationError) which dont fit the httpx severity model. those should get a boto-aware helper in a follow-up.

---

---

### DIFF
--- a/app/services/victoria_logs/client.py
+++ b/app/services/victoria_logs/client.py
@@ -144,7 +144,7 @@ class VictoriaLogsClient:
                 exc,
                 logger=logger,
                 integration="victoria_logs",
-                method="search_logs",
+                method="query_logs",
                 extras={"query": query},
             )
             return {
@@ -156,7 +156,7 @@ class VictoriaLogsClient:
                 exc,
                 logger=logger,
                 integration="victoria_logs",
-                method="search_logs",
+                method="query_logs",
                 extras={"query": query},
             )
             return {"success": False, "error": str(exc)}
--- a/app/services/victoria_logs/client.py
+++ b/app/services/victoria_logs/client.py
@@ -20,6 +20,7 @@ import httpx
 from pydantic import field_validator
 
 from app.integrations.probes import ProbeResult
+from app.services._error_helpers import capture_service_error
 from app.services._streaming import StreamingParseStats
 from app.strict_config import StrictConfigModel
 
@@ -138,18 +139,27 @@ class VictoriaLogsClient:
             resp.raise_for_status()
             rows = _parse_ndjson(resp.text, limit=limit)
             return {"success": True, "rows": rows, "total": len(rows)}
-        except httpx.HTTPStatusError as e:
-            logger.warning(
-                "[victoria_logs] Query HTTP failure status=%s",
-                e.response.status_code,
+        except httpx.HTTPStatusError as exc:
+            capture_service_error(
+                exc,
+                logger=logger,
+                integration="victoria_logs",
+                method="search_logs",
+                extras={"query": query},
             )
             return {
                 "success": False,
-                "error": f"HTTP {e.response.status_code}: {e.response.text[:200]}",
+                "error": f"HTTP {exc.response.status_code}: {exc.response.text[:200]}",
             }
-        except Exception as e:
-            logger.warning("[victoria_logs] Query error: %s", e)
-            return {"success": False, "error": str(e)}
+        except Exception as exc:
+            capture_service_error(
+                exc,
+                logger=logger,
+                integration="victoria_logs",
+                method="search_logs",
+                extras={"query": query},
+            )
+            return {"success": False, "error": str(exc)}
 
 
 def _parse_ndjson(text: str, *, limit: int) -> list[dict[str, Any]]:
--- a/app/services/_error_helpers.py
+++ b/app/services/_error_helpers.py
@@ -1,0 +1,27 @@
+"""Shared error-telemetry helper for service-client except blocks."""
+
+from __future__ import annotations
+
+import logging
+
+import httpx
+
+from app.utils.errors import report_exception
+
+
+def capture_service_error(
+    exc: BaseException,
+    *,
+    logger: logging.Logger,
+    integration: str,
+    method: str,
+) -> None:
+    severity = "warning" if isinstance(exc, httpx.HTTPStatusError) else "error"
+    report_exception(
+        exc,
+        logger=logger,
+        message=f"[{integration}] {method} failed",
+        severity=severity,
+        tags={"surface": "service_client", "integration": integration},
+        extras={"method": method},
+    )
--- a/app/services/_error_helpers.py
+++ b/app/services/_error_helpers.py
@@ -10,8 +10,12 @@ import httpx
 from app.utils.errors import report_exception
 
 
-def _is_server_error(exc: BaseException) -> bool:
-    return isinstance(exc, httpx.HTTPStatusError) and exc.response.status_code >= 500
+def _is_transient_vendor_error(exc: BaseException) -> bool:
+    if not isinstance(exc, httpx.HTTPStatusError):
+        return False
+    sc = exc.response.status_code
+    # 429 = vendor rate-limit (transient throttling, not a config error)
+    return sc == 429 or sc >= 500
 
 
 def capture_service_error(
@@ -22,8 +26,9 @@ def capture_service_error(
     method: str,
     extras: dict[str, Any] | None = None,
 ) -> None:
-    severity = "warning" if _is_server_error(exc) else "error"
+    severity = "warning" if _is_transient_vendor_error(exc) else "error"
     merged_extras: dict[str, Any] = dict(extras) if extras else {}
+    merged_extras.pop("surface", None)
     merged_extras["method"] = method
     report_exception(
         exc,
--- a/app/services/_error_helpers.py
+++ b/app/services/_error_helpers.py
@@ -23,9 +23,8 @@ def capture_service_error(
     extras: dict[str, Any] | None = None,
 ) -> None:
     severity = "warning" if _is_server_error(exc) else "error"
-    merged_extras: dict[str, Any] = {"method": method}
-    if extras:
-        merged_extras.update(extras)
+    merged_extras: dict[str, Any] = dict(extras) if extras else {}
+    merged_extras["method"] = method
     report_exception(
         exc,
         logger=logger,
--- a/app/services/_error_helpers.py
+++ b/app/services/_error_helpers.py
@@ -3,10 +3,15 @@
 from __future__ import annotations
 
 import logging
+from typing import Any
 
 import httpx
 
 from app.utils.errors import report_exception
+
+
+def _is_server_error(exc: BaseException) -> bool:
+    return isinstance(exc, httpx.HTTPStatusError) and exc.response.status_code >= 500
 
 
 def capture_service_error(
@@ -15,13 +20,17 @@ def capture_service_error(
     logger: logging.Logger,
     integration: str,
     method: str,
+    extras: dict[str, Any] | None = None,
 ) -> None:
-    severity = "warning" if isinstance(exc, httpx.HTTPStatusError) else "error"
+    severity = "warning" if _is_server_error(exc) else "error"
+    merged_extras: dict[str, Any] = {"method": method}
+    if extras:
+        merged_extras.update(extras)
     report_exception(
         exc,
         logger=logger,
         message=f"[{integration}] {method} failed",
         severity=severity,
         tags={"surface": "service_client", "integration": integration},
-        extras={"method": method},
+        extras=merged_extras,
     )
--- a/app/services/alertmanager/client.py
+++ b/app/services/alertmanager/client.py
@@ -18,6 +18,7 @@ import httpx
 
 from app.integrations.config_models import AlertmanagerIntegrationConfig
 from app.integrations.probes import ProbeResult
+from app.services._error_helpers import capture_service_error
 
 logger = logging.getLogger(__name__)
 
@@ -87,18 +88,19 @@ class AlertmanagerClient:
             resp.raise_for_status()
             data = resp.json()
             return {"success": True, "status": data}
-        except httpx.HTTPStatusError as e:
-            logger.warning(
-                "[alertmanager] Status check HTTP failure status=%s",
-                e.response.status_code,
+        except httpx.HTTPStatusError as exc:
+            capture_service_error(
+                exc, logger=logger, integration="alertmanager", method="get_status"
             )
             return {
                 "success": False,
-                "error": f"HTTP {e.response.status_code}: {e.response.text[:200]}",
+                "error": f"HTTP {exc.response.status_code}: {exc.response.text[:200]}",
             }
-        except Exception as e:
-            logger.warning("[alertmanager] Status check error: %s", e)
-            return {"success": False, "error": str(e)}
+        except Exception as exc:
+            capture_service_error(
+                exc, logger=logger, integration="alertmanager", method="get_status"
+            )
+            return {"success": False, "error": str(exc)}
 
     def list_alerts(
         self,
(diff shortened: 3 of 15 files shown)


VibhorGautam:
@greptile review

cerencamkiran:
Looks good overall @VibhorGautam the shared helper direction, severity split, demo video, and validation flow all line up pretty well with #1458 👍

A couple small things I’d still like to see before merge just to fully close out the acceptance criteria / edge-cases cleanly:

* Please confirm every client listed in the issue is either migrated or explicitly called out as intentionally out-of-scope for this PR.
* The `merged_extras.update(extras)` path in `_error_helpers.py` can currently overwrite reserved keys like `method`; worth locking that down before merge.
* Some integrations still seem to drop useful request context that existed in previous logs (`alert_id`, `deployment_id`, `flow_run_id`, etc.). Would be good to preserve those where possible for debugging consistency.
* Since the issue explicitly calls out telemetry coverage expectations, adding a short note in the PR description summarizing how the acceptance criteria are satisfied would help future readers too.

VibhorGautam:
thanks for the thorough review @cerencamkiran, pushed fixes for all 4 points:

**1. client coverage** - migrated victoria_logs and argocd (now 13/15). aws_sdk_client and lambda_client are intentionally out-of-scope because they use boto3/botocore exceptions (ClientError, NoCredentialsError, ParamValidationError) which dont map to the httpx severity model. those should get a dedicated boto-aware helper in a follow-up. called this out in the PR description now.

**2. reserved key protection** - flipped the merge order so `method` is set last and cant be overwritten by caller extras. added `test_caller_extras_cannot_overwrite_method` to cover it.

**3. request context restored** - went through every client and added back the per-request IDs that were in the old log lines:
- opsgenie: `alert_id` on get_alert, get_alert_logs, add_note
- prefect: `flow_run_id` on get_flow_run_logs, `work_pool_name` on get_workers
- vercel: `deployment_id` on get_deployment/events/runtime_logs, `project` on get_project, `project_id`+`state` on list_deployments
- jira: `issue_key` on update_issue, add_comment, get_issue
- victoria_logs: `query` on search_logs
- argocd: `application_name` on get_application_summary and get_application_diff

**4. acceptance criteria summary** - added a table to the PR description mapping each criterion to where its satisfied.

all green: `make lint && make typecheck && make test-cov` (11 helper tests, 6815 suite pass)

VibhorGautam:
@greptile review

WatchTree-19:
drive-by from someone in the observability / sentry-instrumentation lane, take or leave.

**The aws/lambda exclusion rationale is fair but the scope-out leaves a coverage gap worth tracking.** boto3's `ClientError` does expose status via `exc.response["ResponseMetadata"]["HTTPStatusCode"]` and an error code via `exc.response["Error"]["Code"]`. so the same 5xx-vs-4xx severity split actually does map onto boto3 cleanly, the work is just translating ClientError into a status_code at the helper entry. `NoCredentialsError` and `ParamValidationError` are operator-actionable so they map to `error`. concrete follow-up shape: `capture_boto_error(exc, *, service, method, extras=None)` that calls `capture_service_error` under the hood after pulling status from response metadata. worth filing as a follow-up issue alongside merge so the gap stays visible.

**429 rate-limit edge case is worth a paragraph in the severity rationale.** `status_code >= 500` puts 429 (technically 4xx) into `error` severity. but 429 means the vendor is saying "back off, transient capacity issue," not "your config is broken." in practice operator-action on 429 is "wait" not "fix something." would lean toward bumping 429 (and 503 / 504 which you already get for free as >=500) into the warning bucket for consistency. one-line tweak: `if status_code == 429 or status_code >= 500`. low priority since 429s should be retried by the client anyway, but worth a comment explaining the choice either way so the next reviewer doesn't second-guess it.

**Cross-cluster coordination with #1827, #1849, #1619, #1743.** all four PRs (plus this one) form a coherent Sentry-instrumentation cluster:
- #1827 (CoderHariswar) — LoggingIntegration as the safety net
- #1849 (gbsierra) — `surface=remote_server` tag for runtime fallback events  
- this PR — `surface=service_client` tag for outbound API calls
- #1619 (in flight) — fingerprint rules that will pivot on these surface tags
- #1743 — action-execution thread/future failures

once #1619's fingerprint rules land, `surface=service_client` becomes the pivot key for grouping vendor-outage spikes vs operator-config issues at the Sentry side, replacing brittle message-pattern matching. worth a one-line note in `_error_helpers.py` flagging the surface tag's role in the cross-PR taxonomy so future implementers don't rename it inadvertently.

**`method` key protection is good. extend to `surface` while you're at it.** the same logic that protects `method` from caller overwrite should probably protect `surface` (currently set at the helper start but a caller passing `extras={"surface": "weird"}` would clobber it under `merged_extras.update(extras)`). one-line change: set both keys last, add a similar test.

**Cross-link to #1469** (integration verification adapters, cerencamkiran's referenced thread). that PR's adapters are exactly the kind of outbound-call surface this helper is built for. flagging so the two don't drift into parallel helper implementations.

**Vercel retry-exhaustion-only reporting is the right v1 call** for 5xx noise reduction. one nuance worth thinking about for v2: non-retryable errors (401 auth failure, 403 permission, 404 missing resource) on intermediate attempts probably shouldn't wait for exhaustion since they won't recover. the operator wants to know now. pattern: report-immediately on `status_code < 500` even mid-retry, defer telemetry only for `>= 500` until exhaustion. follow-up scope, not blocking this PR.

clean change overall, the scope decisions are well-justified, cerencamkiran's review already extracted the meaningful items, and VibhorGautam's responses are crisp. aws/lambda follow-up issue + the one-line severity comment + the surface-key protection are the only items i'd want to see before merge.

VibhorGautam:
great review @WatchTree-19, addressed the three items:

**1. surface key protection** - `merged_extras.pop("surface", None)` now strips it from caller extras before merge, same pattern as method. test `test_caller_extras_cannot_inject_surface` verifies it lands only in tags. good catch on the cross-PR taxonomy angle - once #1619's fingerprint rules pivot on `surface=service_client`, accidentally clobbering it from extras would be bad.

**2. 429 severity** - bumped to warning. renamed `_is_server_error` -> `_is_transient_vendor_error` since the function now covers `429 || >= 500`. the reasoning is exactly what you said: 429 = "back off, transient capacity" = vendor problem, operator action is "wait" not "fix config". test `test_429_rate_limit_uses_warning_severity` covers it.

**3. boto3 follow-up** - filed #1942 with the shape you suggested (`capture_boto_error` wrapping `capture_service_error` after pulling status from `response["ResponseMetadata"]["HTTPStatusCode"]`). covers both aws_sdk_client and lambda_client sites.

re: vercel retry-exhaustion-only for non-retryable errors - agreed that 401/403/404 mid-retry should report immediately since they wont recover. thats a v2 concern though, current implementation already handles the HTTPStatusError path separately from the retryable path so the plumbing is there.

re: cross-cluster coordination with #1827/#1849/#1619/#1743 - noted. the `surface` tag is the pivot key and its now protected from overwrite.

VibhorGautam:
@greptile review

app/services/victoria_logs/client.py
```diff
@@ -138,18 +139,27 @@ def query_logs(
...
-        except httpx.HTTPStatusError as e:
-            logger.warning(
-                "[victoria_logs] Query HTTP failure status=%s",
-                e.response.status_code,
+        except httpx.HTTPStatusError as exc:
+            capture_service_error(
+                exc,
+                logger=logger,
+                integration="victoria_logs",
+                method="search_logs",
```
Devesh36: should be "query_logs"

VibhorGautam: good catch, pushed the fix in ed6ed09d - both except blocks had search_logs instead of query_logs

VibhorGautam:
@greptile review

VibhorGautam:
rebased onto main after #2022's streaming NDJSON changes landed, and resolved the 4 import conflicts cleanly. also fixed the query_logs method tag from Devesh's catch

I believe the feedback from cerencamkiran and WatchTree-19 is covered now: reserved key protection, 429 severity, request context extras, and the acceptance criteria table. ready for another look whenever you get a chance

muddlebee:
@greptile review

muddlebee (approved):
LGTM

Davidson3556:
looks good to me

PR #1871, opened 2026-05-12
test: skip POSIX-specific tail tests on Windows to fix CI failures

Siddhazntx:
Fixes #1852 

#### Describe the changes you have made in this PR -
Added `@pytest.mark.skipif` decorators to skip POSIX-specific tail tests when running on Windows. This prevents `AttachUnsupported` (path parsing) and `WinError 32` (file-locking) errors during Windows CI runs, satisfying the host-neutral requirement mentioned in the issue.

### Demo/Screenshot for feature changes and bug fixes - 
```
======================================================= slowest 20 durations =======================================================
0.10s call     tests/agents/test_tail.py::TestAttachSession::test_does_not_replay_pre_attach_content
0.03s call     tests/agents/test_tail.py::TestAttachIntegration::test_attach_eagerly_opens_real_file
0.02s call     tests/agents/test_tail.py::TestAttachSession::test_reader_exits_when_pid_dies
0.02s call     tests/agents/test_tail.py::TestAttachSession::test_yields_chunks_written_after_attach
0.01s setup    tests/agents/test_tail.py::TestResolveMacosTarget::test_lsof_warnings_on_stderr_do_not_break_parse
0.01s setup    tests/agents/test_tail.py::TestResolveMacosTarget::test_regular_file
0.00s setup    tests/agents/test_tail.py::TestAttachSession::test_iteration_after_close_raises_stop_iteration
0.00s setup    tests/agents/test_tail.py::TestAttachSession::test_yields_chunks_written_after_attach
0.00s setup    tests/agents/test_tail.py::TestAttachSession::test_does_not_replay_pre_attach_content
0.00s setup    tests/agents/test_tail.py::TestAttachSession::test_close_is_idempotent
0.00s setup    tests/agents/test_tail.py::TestAttachSession::test_close_does_not_set_producer_exited
0.00s setup    tests/agents/test_tail.py::TestAttachIntegration::test_attach_eagerly_opens_real_file
0.00s setup    tests/agents/test_tail.py::TestAttachSession::test_reader_exits_when_pid_dies
0.00s setup    tests/agents/test_tail.py::TestAttachSession::test_publish_drops_oldest_when_queue_full
0.00s setup    tests/agents/test_tail.py::TestAttachSession::test_context_manager_closes_thread
0.00s call     tests/agents/test_tail.py::TestAttachSession::test_iteration_after_close_raises_stop_iteration
0.00s setup    tests/agents/test_tail.py::TestResolveMacosTarget::test_lsof_missing
[...]

Issue #1852, opened 2026-05-12
[BUG] Windows test-cov: agent tail tests assume POSIX paths


gbsierra:
### Summary

Agent tail tests fail on Windows because Linux/macOS tail logic assumes POSIX paths and file rename behavior.

### Expected vs actual behavior

**Expected:** POSIX-only tail tests should pass on POSIX and skip or be host-neutral on Windows.

**Actual:** `tests/agents/test_tail.py` fails on Windows with errors like:

`AttachUnsupported: stdout target 'C:\\...' is not a filesystem path`

and:

`PermissionError: [WinError 32] The process cannot access the file because it is being used by another process`

### Steps to reproduce

1. On Windows, run:

`make test-cov` or:
`uv run pytest -n auto -v --cov=app --cov-report=term-missing --ignore=tests/e2e/kubernetes_local_alert_simulation --ignore=tests/synthetic -m "not synthetic"`

2. See failures in `tests/agents/test_tail.py`.

### Can you reproduce it consistently?

Yes

### How often does it occur?

Every time

### Operating system

Windows

Siddhazntx:
Hey, I just opened PR #1871 to fix this.

To get the Windows CI pipeline unblocked quickly, I went with the `@pytest.mark.skipif` method since the issue mentioned 'skip or be host-neutral'. However, if you would prefer to actually make the tests host-neutral so they run on Windows, let me know! I'm happy to write the pathlib and file-locking fixes in a follow-up PR.

muddlebee:
@Siddhazntx needs more input from community

PR #1931, opened 2026-05-13
test(agents): skip AF_UNIX-only bus tests on platforms without Unix sockets (#1850)

### LINKED ISSUE #1850: [BUG] Windows test-cov: /agents bus fails without socket.AF_UNIX
### Summary

/agents bus tests fail on Windows because app.agents.bus assumes Unix sockets are available

**Expected:** Windows test runs should pass or skip Unix-socket-only bus tests clearly.

**Actual:** Several `tests/agents/test_bus.py` tests fail with:

`AttributeError: module 'socket' has no attribute 'AF_UNIX'`

### Steps to reproduce

1. On Windows, run:
make test-cov or the equivalent 
`uv run pytest -n auto -v --cov=app --cov-report=term-missing --ignore=tests/e2e/kubernetes_local_alert_simulation --ignore=tests/synthetic -m "not synthetic"`

2. See failures in `tests/agents/test_bus.py`, including `TestBusServerLifecycle`.

or
1. open interactive terminal
2. do /agents bus

Image

Yes

### How often does it occur?

Every time

### Operating system

Windows

### Logs and error output

_No response_

### Additional context

_No response_

Powlisher:
Fixes #1850 (test-cov half — the runtime half is left for the maintainer call @gbsierra raised, see below)

`tests/agents/test_bus.py` exercises the agent bus, which is built on `socket.AF_UNIX` (Unix-domain sockets). On platforms whose stdlib `socket` module doesn't expose `AF_UNIX` — notably Windows in CPython 3.12 — every `TestBusServer*` test fails at the first `socket.socket(socket.AF_UNIX, ...)` call with

```
AttributeError: module 'socket' has no attribute 'AF_UNIX'
```

This is exactly what blocks `make test-cov` (and the equivalent `uv run pytest -n auto -v --cov=app --cov-report=term-missing --ignore=tests/e2e/kubernetes_local_alert_simulation --ignore=tests/synthetic -m "not synthetic"`) on Windows, as documented in the issue.

This PR applies the same pattern PR #1871 used for the sibling `tests/agents/test_tail.py` failures (`#1852`): a single module-level `pytestmark = pytest.mark.skipif(not hasattr(socket, "AF_UNIX"))` so the whole module is collected-then-skipped on Windows and continues to run normally on Linux and macOS where `AF_UNIX` exists.

```python
pytestmark = pytest.mark.skipif(
    not hasattr(socket, "AF_UNIX"),
    reason=(
        "agent bus is built on AF_UNIX Unix-domain sockets; the platform's "
        "stdlib socket module does not expose AF_UNIX (e.g. Windows)"
    ),
)
```

Why `hasattr(socket, "AF_UNIX")` over `sys.platform == "win32"`:

- it matches the **exact** failure mode reported in the issue (the `AttributeError` is the trigger),
- it skips correctly on any platform/Python build that lacks `AF_UNIX`, not just Windows,
- it does **not** skip on future Windows Python builds that ship `AF_UNIX` (CPython has had partial work toward that — this guard stays right whichever way that lands).

#### Scope explicitly excluded

The runtime half of #1850 — `opensre` interactive shell `/agents bus` failing at runtime on Windows — is left untouched. As @gbsierra noted in the issue thread, that requires a maintainer call between two approaches:

1. a small bus transport abstraction that supports Unix sockets where available and local TCP elsewhere, or
2. moving the bus to local TCP on all platforms with a local auth token.

This PR doesn't preempt either choice; it only unblocks `test-cov` so contributors on Windows can iterate without the whole suite turning red on a known-environmental failure.

I couldn't run `tests/agents/test_bus.py` end-to-end locally because it imports `app.agents.bus`, which transitively pulls the full opensre stack (`httpx`, `anthropic`, `langgraph`, ...) and I didn't want to install the full dependency tree just to verify a one-line skip-marker.

Instead I verified the skip logic directly:

1. **On this Windows 11 box, `hasattr(socket, "AF_UNIX")` is `False`** — so the `skipif` condition `not hasattr(socket, "AF_UNIX")` is `True` and the marker triggers.

2. **Standalone repro** — a tiny test file mirroring the pattern, with the same skip marker and a body that would `AttributeError` on Windows:

   ```
   $ py -m pytest test_repro.py -v
   collected 2 items

   test_repro.py::test_unconditionally_fails_unless_skipped SKIPPED [50%]
   test_repro.py::TestBusServerLifecycleLookAlike::test_open_listener SKIPPED [100%]

   ============================== 2 skipped in 0.15s ==============================
   ```

   Both tests are skipped before they get to the AF_UNIX call, which is exactly the behaviour we need on Windows CI.

3. **AST sanity check** — `ast.parse(test_bus.py).pytestmark` is at line 17, before the `from app.agents import bus` line, so the marker is registered during module-level execution and applies to every test in the module.

4. **POSIX runs are untouched** — on Linux/macOS `hasattr(socket, "AF_UNIX")` is `True`, so `not hasattr(...)` is `False`, the skipif is inactive, and the existing test bodies run exactly as before. CI on `ubuntu-latest` already exercises this and should stay green.

---

1. **Per-test `@pytest.mark.skipif`** — the pattern PR #1871 uses for `test_tail.py`. Works but `test_bus.py` has ~20 tests across 6 test classes and the failure is uniform across all of them (any AF_UNIX socket creation explodes the same way), so per-test markers are pure noise.
2. **Module-level `pytestmark`** — one marker, applies to every test in the module. Chosen.
3. **A conftest fixture or platform-aware import shim** — too much machinery for a one-line, one-cause problem.

For the condition, I used `not hasattr(socket, "AF_UNIX")` rather than `sys.platform == "win32"` because it matches the exact attribute the test bodies reach for. A future Windows Python build that ships `AF_UNIX` would un-skip these tests automatically — and a hypothetical non-Windows build that lacks it would also be handled.

The marker is placed before the `from app.agents import bus` line so a future reader scanning the imports sees the platform gate immediately, and so that if `app.agents.bus` ever grows a module-level `AF_UNIX` reference (it doesn't today — all current uses are inside function bodies), the skip still works.

I left `_POSIX_FCNTL_AVAILABLE` (line 15) untouched. It's used by an existing `@pytest.mark.skipif` on a single fcntl-specific test (line 752), is independent of AF_UNIX, and removing it would be unrelated scope.

---

### DIFF
--- a/tests/agents/test_bus.py
+++ b/tests/agents/test_bus.py
@@ -14,6 +14,14 @@ import pytest
 
 _POSIX_FCNTL_AVAILABLE = importlib.util.find_spec("fcntl") is not None
 
+pytestmark = pytest.mark.skipif(
+    not hasattr(socket, "AF_UNIX"),
+    reason=(
+        "agent bus is built on AF_UNIX Unix-domain sockets; the platform's "
+        "stdlib socket module does not expose AF_UNIX (e.g. Windows)"
+    ),
+)
+
 from app.agents import bus as bus_module
 from app.agents.bus import (
     BUS_SCHEMA_VERSION,


yashksaini-coder:
@Powlisher Hi great work, can you add a screenshot or video demo as well please.

gbsierra:
> @Powlisher Hi great work, can you add a screenshot or video demo as well please. 

There is an unresolved discussion in #contribute the main team needs to discuss regarding supporting Windows before this is fully worked on

muddlebee:
closing this as per comments here https://github.com/Tracer-Cloud/opensre/pull/1931#issuecomment-4441235092

PR #1953, opened 2026-05-13
Add SQLite source store for interactive shell RAG

### LINKED ISSUE #1446: REPL: SQLite-backed chunk and vector store for future RAG (no embeddings yet)
## Background

Future work to make the OpenSRE interactive shell codebase-aware (RAG over `app/**`) needs a place to persist chunked source plus their embedding vectors. The store is well-bounded and self-contained: with a clean dataclass interface it has no compile-time dependency on the chunker, the embeddings client, or the retriever. Tests can use synthetic vectors (zeros / hashes), so this can ship as an independent first task.

## Scope

Add `app/cli/interactive_shell/source_store.py` exposing a small SQLite-backed store:

```python
DEFAULT_STORE_PATH = Path.home() / ".config" / "opensre" / "source_index.sqlite"

@dataclass(frozen=True)
class StoredChunk:
    relpath: str
    kind: str
    symbol: str
    start_line: int
    end_line: int
    content: str
    fingerprint: str   # caller-supplied per-file digest

class SourceStore:
    def __init__(self, path: Path = DEFAULT_STORE_PATH, *, vector_dim: int | None = None) -> None: ...
    @property
    def embedding_model(self) -> str | None: ...
    def set_meta(self, key: str, value: str) -> None: ...
    def get_meta(self, key: str) -> str | None: ...
    def file_fingerprints(self) -> dict[str, str]: ...        # relpath -> fingerprint
    def delete_file(self, relpath: str) -> int: ...           # returns rows deleted
    def upsert_chunks(self, chunks: list[StoredChunk], vectors: list[np.ndarray]) -> list[int]: ...
    def fetch_by_ids(self, ids: list[int]) -> list[StoredChunk]: ...
    def cosine_topk(self, query_vector: np.ndarray, k: int = 30) -> list[tuple[int, float]]: ...
    def stats(self) -> dict[str, Any]: ...                     # chunk_count, file_count, on_disk_bytes, embedding_model
```

### Schema

```sql
CREATE TABLE IF NOT EXISTS chunks (
  id           INTEGER PRIMARY KEY,
  relpath      TEXT NOT NULL,
  kind         TEXT NOT NULL,
  symbol       TEXT NOT NULL,
  start_line   INTEGER NOT NULL,
  end_line     INTEGER NOT NULL,
  content      TEXT NOT NULL,
  fingerprint  TEXT NOT NULL
);
CREATE INDEX IF NOT EXISTS idx_chunks_relpath ON chunks(relpath);
CREATE TABLE IF NOT EXISTS vectors (id INTEGER PRIMARY KEY, vector BLOB NOT NULL);
CREATE TABLE IF NOT EXISTS meta (key TEXT PRIMARY KEY, value TEXT NOT NULL);
```

Vectors are persisted as `numpy.float32` byte blobs (`np.ndarray.tobytes()`); load with `np.frombuffer(blob, dtype=np.float32)`. Cosine ranking can use plain numpy (`(M @ q) / (norms_M * norm_q)`) — no faiss/chroma dependency.

`set_meta` should be used by callers to record `embedding_model` and `index_version`; if the store is opened with a `vector_dim` that disagrees with the previously persisted dim (or model name disagrees), raise a clear `IncompatibleStoreError` so callers can prompt the user to rebuild.

The store should create parent directories on init and use a file-level lock or `BEGIN IMMEDIATE` to be safe against concurrent shells (one user, two terminals).

## Acceptance criteria

- [ ] New module `app/cli/interactive_shell/source_store.py` with the dataclass, the `SourceStore` class, and a custom `IncompatibleStoreError`
- [ ] Tests in `tests/cli/interactive_shell/test_source_store.py` (use a `tmp_path` SQLite file, no real `~/.config` writes) covering:
  - Round-trip upsert + fetch
  - `delete_file` removes both chunks and their vectors
  - `cosine_topk` returns expected order on a small hand-built fixture
  - `IncompatibleStoreError` on mismatched `vector_dim` or `embedding_model`
  - `stats()` reports correct counts after upsert/delete
  - `file_fingerprints()` returns deduplicated mapping
- [ ] Schema is created idempotently — opening an existing store is a no-op for migrations
- [ ] `make lint`, `make format-check`, `make typecheck`, `make test-cov` all pass

## References

- Reference for cache + fingerprint conventions in this folder: [`app/cli/interactive_shell/docs_reference.py`](app/cli/interactive_shell/docs_reference.py)

## Out of scope

- Walking `app/` to populate the store (callers will do that with the chunker from a separate issue)
- The retriever / prompt-grounding builder (separate issue)
- Schema migrations beyond initial create
- Multi-tenant / shared-volume locking

popsiclelmlm:
## Summary
- Add a SQLite-backed `SourceStore` and `StoredChunk` model for future interactive-shell source RAG
- Persist float32 vectors, metadata, vector dimension/model compatibility checks, file fingerprint lookup, delete/upsert, stats, and cosine top-k search
- Add focused tests covering round-trip persistence, deletion, ranking, compatibility errors, stats, fingerprints, and idempotent schema creation

Closes #1446

## Tests
- `uv run --extra dev python -m pytest tests/cli/interactive_shell/test_source_store.py -q`
- `uv run --extra dev python -m ruff check app/cli/interactive_shell/source_store.py tests/cli/interactive_shell/test_source_store.py`
- `uv run --extra dev python -m ruff format --check app/cli/interactive_shell/source_store.py tests/cli/interactive_shell/test_source_store.py`
- `uv run --extra dev python -m mypy app/cli/interactive_shell/source_store.py`


popsiclelmlm:
@greptile review

popsiclelmlm:
@greptile review

popsiclelmlm:
@greptile review

cerencamkiran:
Nice work @popsiclelmlm 

The main issue scope looks covered now, and the earlier Greptile concerns seem fixed too.

Only small thing I’d keep an eye on later is the mixed connection / transaction handling. Not blocking for this PR, but it may be cleaner to centralize it once the retriever/chunker parts are added.

popsiclelmlm:
Thanks, glad the main scope is covered. Agreed on the connection/transaction handling: I kept this PR focused on the source-store foundation, and I’d centralize that lifecycle before layering the retriever/chunker paths on top.

muddlebee:
need some sort of e2e demo

app/cli/interactive_shell/source_store.py
```diff
@@ -0,0 +1,398 @@
...
+
+import sqlite3
+from collections.abc import Iterable
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+import numpy as np
+
+DEFAULT_STORE_PATH = Path.home() / ".config" / "opensre" / "source_index.sqlite"
```
muddlebee, line 13: Use OPENSRE_HOME_DIR from app.constants (same as OPENSRE_HOME_DIR / source_index.sqlite) so the index lives with other OpenSRE data and the config root is not duplicated.

app/cli/interactive_shell/source_store.py
```diff
@@ -0,0 +1,398 @@
...
+            raise IncompatibleStoreError(
+                f"Store vector_dim is {self._vector_dim}, got {normalized.shape[0]}"
+            )
+        return normalized
+
+    def _resolve_upsert_vector_dim(self, vectors: list[np.ndarray]) -> int:
+        vector_dim = int(vectors[0].shape[0])
+        for vector in vectors:
+            if vector.shape[0] != vector_dim:
+                raise IncompatibleStoreError(
```
muddlebee, line 330: This branch is inconsistent vector lengths inside a single upsert batch. The error text reads like a store vs caller dim mismatch. Reword the message or raise ValueError here.

app/cli/interactive_shell/source_store.py
```diff
@@ -0,0 +1,398 @@
...
+                        SELECT id, relpath, kind, symbol, start_line, end_line, content, fingerprint
+                        FROM chunks
+                        WHERE id IN ({placeholders})
+                        """,
+                        batch,
+                    ).fetchall()
+                )
+
+        by_id = {int(row["id"]): _chunk_from_row(row) for row in rows}
+        return [by_id[chunk_id] for chunk_id in ids if chunk_id in by_id]
```
muddlebee, line 185: fetch_by_ids omits missing ids silently; the result length can be less than the input list. Document that contract or return or raise on missing ids.

app/cli/interactive_shell/source_store.py
```diff
@@ -0,0 +1,398 @@
...
+    def get_meta(self, key: str) -> str | None:
+        with self._connect() as conn:
+            row = conn.execute("SELECT value FROM meta WHERE key = ?", (key,)).fetchone()
+        return str(row["value"]) if row is not None else None
+
+    def file_fingerprints(self) -> dict[str, str]:
+        with self._connect() as conn:
+            rows = conn.execute(
+                """
+                SELECT relpath, MAX(fingerprint) AS fingerprint
```
muddlebee, line 90: MAX(fingerprint) is lexicographic when multiple chunks per relpath differ. Add a short docstring note or enforce one fingerprint per file at the indexer.

pyproject.toml
```diff
@@ -45,6 +45,7 @@ dependencies = [
     "questionary>=2.1.1",
     "prompt_toolkit>=3.0.0",
     "PyYAML>=6.0",
+    "numpy>=1.26.0",
```
muddlebee, line 48: numpy is added as a core dependency. Confirm the team wants the install weight for this store versus a lighter numerical path for small indexes.

app/cli/interactive_shell/source_store.py
```diff
@@ -0,0 +1,398 @@
...
+                        WHERE id IN ({placeholders})
+                        """,
+                        batch,
+                    ).fetchall()
+                )
+
+        by_id = {int(row["id"]): _chunk_from_row(row) for row in rows}
+        return [by_id[chunk_id] for chunk_id in ids if chunk_id in by_id]
+
+    def cosine_topk(
```
muddlebee, line 187: cosine_topk loads the full vectors table into memory; max_scan_rows caps scale. Add a brief class or method note that large corpora need a different index or ANN approach.

app/cli/interactive_shell/source_store.py
```diff
@@ -0,0 +1,398 @@
...
+
+class SourceStore:
+    """Persist source chunks and float32 embedding vectors in SQLite."""
+
+    def __init__(
+        self,
+        path: Path = DEFAULT_STORE_PATH,
+        *,
+        vector_dim: int | None = None,
+        embedding_model: str | None = None,
```
muddlebee, line 41: Treat one sqlite file as one embedding model plus one vector dim. Changing model should mean a new store path or a full reindex; wire EmbeddingsClient model_name and dim through to this constructor consistently when RAG lands.

app/cli/interactive_shell/source_store.py
```diff
@@ -0,0 +1,398 @@
...
+    content: str
+    fingerprint: str
+
+
+class SourceStore:
+    """Persist source chunks and float32 embedding vectors in SQLite."""
+
+    def __init__(
+        self,
+        path: Path = DEFAULT_STORE_PATH,
```
muddlebee, line 38: Optional follow-up: default path could incorporate a sanitized embedding_model so switching models naturally uses a separate DB without manual deletes.

pyproject.toml
muddlebee: I guess numpy works for now

popsiclelmlm:
Added a small end-to-end demo transcript for the SQLite source store lifecycle.

What it covers:
- creates a fresh temporary SQLite store
- upserts two source chunks with float32 vectors
- runs cosine search and fetches the top matching chunk
- reads file fingerprints
- deletes one file's chunks
- reopens the same DB path to verify persisted metadata and remaining rows

Validation command:

```bash
uv run python -m pytest tests/cli/interactive_shell/test_source_store.py -q
```

Result:

```text
13 passed in 0.53s
```

Demo transcript:

```text
inserted_ids [1, 2]
stats_after_upsert {'chunk_count': 2, 'file_count': 2, 'on_disk_bytes': 28672, 'embedding_model': 'demo-embeddings', 'vector_dim': 3}
top_match_for_alert_query StoredChunk(relpath='app/alerts.py', kind='py_func', symbol='send_alert', start_line=1, end_line=3, content="def send_alert(): return 'alert'", fingerprint='fp:app/alerts.py')
fingerprints {'app/alerts.py': 'fp:app/alerts.py', 'app/incidents.py': 'fp:app/incidents.py'}
deleted_alert_chunks 1
stats_after_delete {'chunk_count': 1, 'file_count': 1, 'on_disk_bytes': 28672, 'embedding_model': 'demo-embeddings', 'vector_dim': 3}
reopened_stats {'chunk_count': 1, 'file_count': 1, 'on_disk_bytes': 28672, 'embedding_model': 'demo-embeddings', 'vector_dim': 3}
```

PR #1905, opened 2026-05-12
feat(agents): wire psutil probe into REPL background sampler (#1490)

kasulani:
Fixes #1490

#### Describe the changes you have made in this PR 
- Added `app/agents/sampler.py`: a background `asyncio.Task` that probes every registered agent's PID every 5 seconds and stores the latest `ProcessSnapshot` in a module-level dict.
- Wired the sampler into `app/cli/interactive_shell/loop.py`: starts on REPL boot, cancels cleanly on exit.
- Updated `app/cli/interactive_shell/agents_view.py`: reads stored snapshots to populate the `cpu%`, `uptime`, and `status` columns (previously hardcoded as `-`).
- Added `tests/agents/test_sampler.py`: covers snapshot storage, None handling, per-PID fault isolation, and clean cancellation.
- Extended `tests/cli/interactive_shell/test_agents_view.py`: verifies the table renders live probe data when a snapshot exists.

### Demo/Screenshot for feature changes and bug fixes - 

https://github.com/user-attachments/assets/327ca96e-b6b2-4bad-9ee6-fc7c31ea2d8a

**Note for reviewers:**
During testing, I found that `cpu_percent(interval=0.1)` returns `0.0` for all processes when called inside `proc.oneshot()` in `probe.py`. Moving it outside `oneshot()` produces accurate readings. This is a pre-existing issue in #1489, not introduced by this PR. Happy to open a follow-up issue if desired.

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [x] No, I wrote all the code myself
- [ ] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [ ] I have reviewed every single line of the AI-generated code
- [ ] I can explain the purpose and logic of each function/component I added
- [ ] I have tested edge cases and understand how the code handles them
- [ ] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**

**What problem does this code solve?**

The `/agents` dashboard showed placeholder dashes for cpu%, uptime, and status columns. This PR connects the psutil probe (#1489) to the REPL via a background async task that samples registered agents every 5 seconds and feeds live data into the table.

**What alternatives were considered?**

- Probing on-demand (only when `/agents` is typed) — rejected because `probe()` blocks for 0.1s per PID, which would make the slash command feel sluggish with multiple agents.
[...]

PR #1960, opened 2026-05-13
fix(agents): move cpu_percent() outside oneshot() to return accurate values (#1950)

### LINKED ISSUE #1950: [BUG] cpu_percent() always returns 0.0 when called inside proc.oneshot()
## Summary

`probe()` in `app/agents/probe.py` calls `proc.cpu_percent(interval=0.1)` inside `proc.oneshot()`, which caches both internal readings from the same snapshot, making the delta always zero. The `/agents` dashboard always shows `0.0` CPU for every process regardless of actual load.

**Expected:** `cpu%` column in the `/agents` table shows actual CPU usage (e.g. `100.0` for a busy process like `yes > /dev/null`).

**Actual:** `cpu%` column always shows `0.0` for all processes.

## Steps to reproduce

```python
import psutil, os

proc = psutil.Process(os.getpid())

# Inside oneshot — always 0.0
with proc.oneshot():
    print(proc.cpu_percent(interval=0.1))  # 0.0

# Outside oneshot — correct value
print(proc.cpu_percent(interval=0.1))  # non-zero
```

Or via the REPL:
1. `opensre` (start interactive shell)
2. Register a CPU-heavy process (e.g. `yes > /dev/null &`)
3. Run `/agents`
4. Observe `cpu%` column shows `0.0`

Yes — every time.

## Operating system

macOS (also affects Linux — psutil caching behavior, not OS-specific)

## Additional context

- Discovered during #1490 implementation and demo
- Related: #1489 (probe module), #1490 (sampler wiring)
- The fix is to move `cpu_percent()` outside the `oneshot()` block while keeping other calls (`memory_info()`, `status()`, `create_time()`) inside for their caching benefit
- `psutil` docs confirm `oneshot()` caches process info; `cpu_percent(interval>0)` needs two distinct readings separated in time, which the cache defeats

kasulani:
Fixes #1950 

cpu_percent(interval>0) always returned 0.0 because it was called inside proc.oneshot(), which caches both internal readings from a single snapshot — making the delta always zero.

Moved proc.cpu_percent() outside the oneshot() block in app/agents/probe.py. All other calls (memory_info, status, create_time, num_fds, net_connections) remain inside oneshot() for their caching benefit. Added a regression test that spawns a CPU-burning subprocess and asserts non-zero CPU%.

This bug was surfaced during #1905 (wire probe into REPL), where the demo was recorded with cpu_percent() already moved outside oneshot() as a workaround. That PR's demo shows the fix working end-to-end in the agents view.

This PR formalizes the fix in isolation. Since probe() is an internal collector that requires the wiring in #1905 (not yet merged) to surface in the UI, here is the test output:
Screenshot 2026-05-13 at 22 35 39

---

Problem: psutil.Process.oneshot() caches all process attributes from a single syscall. cpu_percent(interval>0) needs two separate readings separated in time to compute a delta — inside oneshot, both readings return the same cached value, so the result is always 0.0.
Alternative: Could have called cpu_percent(interval=0) and managed baseline state externally in the sampler — rejected because it adds complexity to the consumer for no benefit.
Why this approach: Moving the single call outside the with block is the minimal fix. It preserves the oneshot optimization for all other fields while letting cpu_percent take real measurements.
Key change: One line moved in probe(), one regression test added with a busy-subprocess fixture.

---

---

### DIFF
--- a/app/agents/probe.py
+++ b/app/agents/probe.py
@@ -88,8 +88,9 @@ def probe(pid: int, *, cpu_interval: float = 0.1) -> ProcessSnapshot | None:
         return None
 
     try:
+        cpu = proc.cpu_percent(interval=cpu_interval)
+
         with proc.oneshot():
-            cpu = proc.cpu_percent(interval=cpu_interval)
             rss_mb = proc.memory_info().rss / _BYTES_PER_MIB
             num_fds = _safe_num_fds(proc)
             num_connections = _safe_num_connections(proc)
--- a/tests/agents/test_probe.py
+++ b/tests/agents/test_probe.py
@@ -4,16 +4,44 @@ from __future__ import annotations
 
 import os
 import pathlib
+import subprocess
 import sys
+import time
+from collections.abc import Iterator
 from datetime import UTC, datetime
 from unittest.mock import patch
 
 import psutil
+import pytest
 
 from app.agents.probe import ProcessSnapshot, probe
 
 _REPO_ROOT = pathlib.Path(__file__).resolve().parents[2]
 _PROBE_MODULE = _REPO_ROOT / "app" / "agents" / "probe.py"
+
+
+@pytest.fixture
+def busy_process() -> Iterator[int]:
+    """Spawn a subprocess burning CPU in a tight loop; yields its PID.
+    The warmup sleep ensures the process has accumulated enough CPU time for
+    psutil.cpu_percent(interval>0) to measure a non-zero delta.
+    """
+    proc = subprocess.Popen(
+        [sys.executable, "-c", "while True: pass"],
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+    )
+
+    time.sleep(0.1)
+
+    yield proc.pid
+
+    proc.terminate()
+    try:
+        proc.wait(timeout=5)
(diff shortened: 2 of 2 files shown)


kasulani:
@greptile review

kasulani:
@muddlebee / @yashksaini-coder please review this small PR

muddlebee:
@kasulani can you share a before/after demo please

kasulani:
> @kasulani can you share a before/after demo please

Hi @muddlebee,

Here is before moving the `proc.cpu_percent() outside the oneshot()`. All process show cpu utilization as `0.0`:

https://github.com/user-attachments/assets/a85075e9-6eb2-4360-a54a-f9f607d255f0

And here is after moving the `proc.cpu_percent() outside the oneshot()`. All process that have cpu utilization above `0.0` show none zero values

https://github.com/user-attachments/assets/b3a8fbe9-1288-42da-a8fc-1993d73fa7d2

Do you have any other further questions?

muddlebee:
@kasulani awesome work, feel free to tinker with `/agents` and report other issues, one visible issue I see is blank coming in 
image

which we need to make better

muddlebee:
@kasulani opened an issue here https://github.com/Tracer-Cloud/opensre/issues/1979

PR #1869, opened 2026-05-12
feat(error-handling): add Sentry capture to per-integration validators

kespineira:
Fixes #1470

#### Describe the changes you have made in this PR -

- Adds `app/integrations/_validation_helpers.py` with `report_validation_failure(exc, *, logger, integration, method, severity="warning", extras=None)` — a thin wrapper over the existing `app.utils.errors.report_exception` helper from #1454. It pins the tag schema (`surface=integration`, `integration=<vendor>`, `event=validation_failed`, `method=<func>`) so vendor-level validator failures are filterable in Sentry.
- Migrates every broad `except Exception` site listed in the issue (~50 sites across 20 files under `app/integrations/` plus `llm_cli/kimi.py`) to call the helper before returning the degraded `ValidationResult` / fallback dict.
- Operator-visible output is unchanged. Severity defaults to `warning` since these are vendor / config issues, not OpenSRE bugs.

**Special-shape sites**

- `airflow.get_recent_airflow_failures` — per-DAG-run capture with `dag_id` and `dag_run_id` as Sentry **extras**, not tags, so tag cardinality stays bounded.
- `daily_update.summarize_highlights` — capture fires before the `DAILY_UPDATE_REQUIRE_LLM` re-raise; the fallback path is unchanged.
- `mongodb.get_rs_status` and `postgresql.get_replication_status` — capture only on the genuine-error branch, not the intentional "not a replica" fallback.

**Deliberately not captured**

The nested `stmt_err` blocks in `mysql.get_replication_status` and `mariadb.get_replication_status` are MySQL <8.0.22 version-detection fallbacks: they `continue` on `ProgrammingError` and `raise` otherwise. The raise path propagates to the outer except, which **is** captured. Capturing the inner block would flood Sentry with normal version-detection events. The decision is documented in the test file's module docstring.

### Demo/Screenshot for feature changes and bug fixes -

Reliability-only refactor; operator-visible output is intentionally unchanged. Verified with `uv run opensre integrations verify` before and after — identical `missing` rows when no credentials are configured.

Validator-with-Sentry behavior is proven by the AST-parametrized test (72 cases, ~0.17s) which asserts every migrated `except Exception` block emits `report_validation_failure(..., integration=<vendor>, method=<func>)`:

```text
$ uv run pytest tests/integrations/test_validation_helpers.py tests/integrations/test_validator_sentry_capture.py -v
...
[...]

PR #1990, opened 2026-05-14
fix(services): stop silent make_*_client factory failures returning None (#1459)

### LINKED ISSUE #1459: [error-handling] Stop silent make_*_client factory failures returning None
## Problem

Several integration clients expose a `make_*_client(...)` factory that quietly swallows configuration / construction errors and returns `None`, making it impossible to distinguish "not configured" from "configured incorrectly" or "transient construction error":

```python
def make_argocd_client(config) -> ArgoCDClient | None:
    try:
        ...
    except Exception:
        return None
```

Concrete sites:

- `app/services/argocd/client.py` ~L527–539 (`make_argocd_client`)
- `app/services/jira/client.py` ~L269–270
- `app/services/alertmanager/client.py` ~L233–234
- `app/services/opsgenie/client.py` ~L236–237
- `app/services/prefect/client.py` ~L337–338
- `app/services/vercel/client.py` ~L547–548
- `app/services/victoria_logs/client.py` ~L184–187 (`make_victoria_logs_client`)

Callers commonly write:

```python
client = make_argocd_client(config)
if client is None:
    return  # treat as "not configured"
```

Today, both a missing `endpoint` env var and an exception inside `ArgoCDClient.__init__` produce the same `None`. The latter is silently wrong — we lose all telemetry on broken Argo CD configs.

## Proposal

Refactor each factory to:

1. Return `None` (or a typed `MissingConfig` sentinel) **only** when the integration is genuinely not configured — i.e., the required config fields are absent.
2. Let `ConfigError` / Pydantic `ValidationError` propagate when fields are present but invalid; the caller should report and skip.
3. Wrap unexpected exceptions in `report_exception` with `surface=service_client`, `integration=<vendor>`, `event=factory_failure`, then raise a typed `ServiceClientUnavailable` so the caller can decide (still skip, but now we have telemetry).

Concretely, introduce in `app/services/_base.py` (created in #1458):

```python
class ServiceClientUnavailable(RuntimeError):
    integration: str
    reason: str

    def __init__(self, integration: str, reason: str, original: BaseException | None = None):
        super().__init__(f"{integration}: {reason}")
        self.integration = integration
        self.reason = reason
        self.__cause__ = original
```

Then each factory:

```python
def make_argocd_client(config) -> ArgoCDClient | None:
    if not config or not config.endpoint:
        return None  # not configured — silent is correct
    try:
        return ArgoCDClient(config)
    except Exception as exc:
        report_exception(exc, logger=logger, message="argocd client construction failed",
                         tags={"surface": "service_client", "integration": "argocd", "event": "factory_failure"})
        raise ServiceClientUnavailable("argocd", "client construction failed", exc) from exc
```

Update callers (typically `resolve_integrations`, tools, and the verifier) to:

```python
try:
    client = make_argocd_client(config)
except ServiceClientUnavailable:
    client = None  # already reported to Sentry
```

## Acceptance criteria

- [ ] Every `make_*_client` factory listed above distinguishes "not configured" (`None`) from "broken" (`ServiceClientUnavailable`).
- [ ] All construction failures call `capture_exception` with `event=factory_failure` and the integration tag.
- [ ] All callers handle `ServiceClientUnavailable` cleanly (skip the integration with a structured warning, never crash the whole investigation).
- [ ] Unit tests cover the three branches (missing config, broken config, valid config) for each factory.

## Depends on

#1454, #1458

0xghost42:
Closes #1459 (stage 1 / 2 — see "Scope" below).

## Scope

This PR is **stage 1** of #1459 — telemetry only. It does **not** implement the typed `ServiceClientUnavailable` flow / caller updates listed in the acceptance criteria.

| Stage | What | Status |
|-------|------|--------|
| 1     | Route `make_*_client` construction-failure exceptions through `report_exception` with structured tags. Preserve the existing `None` caller contract verbatim. | **this PR** |
| 2     | Introduce `app/services/_base.ServiceClientUnavailable` (typed exception). Update the seven factories to raise it instead of returning `None`. Update every caller to catch it. | follow-up — gated on PR #1922 / issue #1458 landing the `_base.py` module |

The typed flow is deliberately deferred because `ServiceClientUnavailable` lives in `app/services/_base.py`, which is being added by @VibhorGautam in #1922 (currently OPEN, conflicting with main). Doing stage 2 now would either (a) ship a partial `_base.py` from this PR, cutting into #1922's scope, or (b) block the telemetry work behind #1922's rebase. Option (c) — staging the telemetry first — gets operators the Sentry signal immediately without forking the typed-exception design, and lets stage 2 graduate the seven factories + every caller in one atomic follow-up once #1922 lands. `_factory_telemetry.py` is intentionally a thin shim that folds into `_base.py` cleanly.

## What stage 1 does

Seven service-client factories previously collapsed both "integration not configured" (required field absent) and "integration broken" (constructor blew up — pydantic validator refused, malformed URL, bad type) into the same `None` return:

```python
def make_argocd_client(config) -> ArgoCDClient | None:
    if not config.endpoint:
        return None  # not configured
    try:
        return ArgoCDClient(config)
    except Exception:
        return None  # broken — but indistinguishable from "not configured"
```

Callers do:

```python
client = make_argocd_client(config)
if client is None:
    return  # treat as not configured
```

So a real config bug looked identical to a deliberately empty config, and the failure never reached Sentry.

## Affected factories

| File | Site | Vendor |
|------|------|--------|
| `app/services/argocd/client.py` | ~L527–539 | argocd |
| `app/services/jira/client.py` | ~L269–270 | jira |
| `app/services/alertmanager/client.py` | ~L233–234 | alertmanager |
| `app/services/opsgenie/client.py` | ~L236–237 | opsgenie |
| `app/services/prefect/client.py` | ~L337–338 | prefect |
| `app/services/vercel/client.py` | ~L547–548 | vercel |
| `app/services/victoria_logs/client.py` | ~L184–187 | victoria_logs |

Each `except Exception` inside the construction branch now routes through a shared helper `app/services/_factory_telemetry.report_factory_failure` before returning `None`. The pre-check "required field absent → return None" branches are unchanged — intentional empty config must stay silent, otherwise Sentry floods with every unconfigured integration on every dev box.

## Sentry tag set

Every captured exception carries:

| tag         | value                              |
|-------------|------------------------------------|
| `surface`   | `service_client`                   |
| `component` | `app.services.<integration>.client`|
| `integration` | `<vendor>`                       |
| `event`     | `factory_failure`                  |

Severity is `warning` — these failures are typically misconfigured credentials (typo in HTTPS URL, malformed bearer token, wrong region), not service bugs. They deserve a Sentry event so operators can distinguish "not configured" from "configured incorrectly", but they should not page on-call alongside real internal errors.

## Caller contract preserved (stage 1 only)

`make_*_client(...)` still returns `None` on both branches, so no downstream code needs to change in this PR. The typed-exception migration in stage 2 will lift every caller atomically.

## Tests (`tests/services/test_factory_silent_none_elimination.py`, 16 cases ~0.6s)

- Helper smoke test asserting the tag dict shape **and** that the original exception object is forwarded verbatim to `report_exception` (re-wrap regression guard, per Greptile P2).
- 7 parametrized cases over the 7 factories — force the `Client` constructor to raise, assert `None` is still returned *and* `report_factory_failure` was called with the right vendor tag, `event=factory_failure`, **and** the exact `RuntimeError` instance from the patched `__init__` (not a re-wrap).
- 8 parametrized negative cases — exercise every factory's "required field absent" pre-check branch and assert `report_factory_failure` is NOT called. Inverting this regression would flood Sentry on every dev box that hasn't configured a given integration.

Patching the `Client` constructor (rather than crafting bad credentials) proves the surrounding `try/except` routes the failure without coupling the test to validator strictness.

## Local verification

- `make lint`, `make format-check` clean.
- `tests/services/` clean (654 passed).
- Broader regression: 6574 passed, 6 skipped, 0 failed (excluding `synthetic/`/`e2e/`/`deployment/`/`chaos_engineering/` which need live infra/LLM access).


0xghost42:
@greptile review

cerencamkiran:
Nice @0xghost42

Before merge, I’d strongly suggest addressing the remaining acceptance-criteria gap explicitly in the PR description: this preserves the existing "None" contract and adds telemetry, but it does not yet implement the typed "ServiceClientUnavailable" flow / caller updates described in #1459.

Also please apply Greptile’s suggestion and assert that the original exception object is forwarded to "report_exception", so we don’t accidentally lose the real traceback in Sentry.

0xghost42:
thanks @cerencamkiran — addressed both points:

**1. Exception forwarding (Greptile P2).** Pushed `09fcf9bc`. The helper smoke test now asserts `mock_report.call_args.args[0] is exc`, and the per-factory parametrised test asserts the forwarded exception is the exact `RuntimeError` from the patched `__init__` (with the original message), so any factory that catches-and-re-raises instead of catches-and-forwards is caught by the suite. 16/16 in `tests/services/test_factory_silent_none_elimination.py`, full `tests/services/` 654/654 still green.

**2. Acceptance-criteria scope gap.** You're right that this PR doesn't cover the typed `ServiceClientUnavailable` flow from #1459's acceptance criteria — I've now made that explicit at the top of the PR description as a stage 1 / stage 2 split:

> **Stage 1 (this PR)** — telemetry routing only; preserve existing `None` caller contract.
> **Stage 2 (follow-up)** — introduce `app/services/_base.ServiceClientUnavailable`, raise it from the seven factories, lift every caller to the typed contract.

Stage 2 is gated on @VibhorGautam's PR #1922 (issue #1458) landing the `_base.py` module — currently OPEN, conflicting with main. Two paths I considered before splitting:

- **Ship `_base.py` partial from this PR.** Cuts into #1922's scope and forces them to rebase against my version. Felt rude.
- **Block telemetry behind #1922 rebase.** Operators stay blind to factory failures until #1922 settles, which has no firm ETA.

Splitting felt like the lowest-coordination cost — stage 1 gets the Sentry signal in front of operators today, and `_factory_telemetry.py` is intentionally a thin shim that folds into `_base.py` cleanly when stage 2 lands (the docstring on the module already calls this out).

Happy to either:

- **Land stage 1 now** and I'll open the stage 2 PR as a single atomic follow-up after #1922 merges (file a tracking issue too, linked from #1459).
- **Hold this** until #1922 lands and roll both stages into one PR — though that means waiting on someone else's rebase.

Let me know which shape you prefer. No strong opinion either way.

@greptile review

muddlebee:
@greptile review

cerencamkiran:
Thanks @0xghost42. Only remaining gap vs the full #1459 acceptance criteria is that the typed ServiceClientUnavailable flow + caller updates are not implemented yet.

Also would be nice to add a small demo/screenshot/log snippet showing the telemetry path firing end-to-end once.

0xghost42:
Thanks @cerencamkiran — agreed on the stage-2 gap, that's what the description split documents.

On the demo: ran the telemetry path end-to-end against the real argocd factory with a misconfigured `base_url` (typo'd scheme — exactly the historic "looks unavailable, was actually broken" case the issue calls out). Test harness patches only `report_exception` so we capture what would land in Sentry; everything else is the real production path.

**Reproducer**

```python
import logging
from unittest.mock import patch
import json

logging.basicConfig(level=logging.WARNING, format='%(levelname)s %(name)s: %(message)s')

captured = {}
def fake_report_exception(exc, *, logger, message, severity, tags):
    captured['exc_type'] = type(exc).__name__
    captured['exc_msg']  = str(exc)
    captured['message']  = message
    captured['severity'] = severity
    captured['tags']     = tags
    logger.warning('%s (exc=%s: %s)', message, type(exc).__name__, exc)

with patch('app.services._factory_telemetry.report_exception', side_effect=fake_report_exception):
    from app.services.argocd.client import make_argocd_client
    result = make_argocd_client('ftp://not-allowed', bearer_token='t0k')

print('caller contract:', repr(result))
print('Sentry payload:')
print(json.dumps(captured, indent=2))
```

**Output**

```
WARNING app.services.argocd.client: argocd client construction failed;
caller will treat as unavailable (exc=ValidationError: 1 validation error for
ArgoCDIntegrationConfig
base_url
  Value error, Argo CD base_url must use https:// unless targeting localhost/loopback.
    [type=value_error, input_value='ftp://not-allowed', input_type=str]
    For further information visit https://errors.pydantic.dev/2.13/v/value_error)

caller contract: None
Sentry payload:
{
  "exc_type": "ValidationError",
  "exc_msg":  "1 validation error for ArgoCDIntegrationConfig\nbase_url\n  Value error, Argo CD base_url must use https:// ...",
  "message":  "argocd client construction failed; caller will treat as unavailable",
  "severity": "warning",
  "tags": {
    "surface":     "service_client",
    "component":   "app.services.argocd.client",
    "integration": "argocd",
    "event":       "factory_failure"
  }
}
```

That's a real existing argocd-config validator firing on a typo'd URL scheme — pre-this-PR the operator gets `None` from the factory and silence in Sentry; now the same `None` is preserved for the caller and the typed `ValidationError` (with the validator's actual hint about `https://` / loopback) reaches Sentry under the standard `surface=service_client / event=factory_failure` tag set.

The same shape fires for every other vendor in the table — same tags, same `caller will treat as unavailable` message template, integration tag swaps to the vendor name. Test suite locks it down per-factory: `tests/services/test_factory_silent_none_elimination.py` calls each of the seven factories with a patched `__init__` that raises and asserts both (a) caller got `None` and (b) `report_exception` was called with the exception instance, the matching tags, and `severity='warning'` (15/15 in the file, full `tests/services/` 654/654 green).

Stage 2 (`ServiceClientUnavailable` + caller graduation) tracked as a follow-up — happy to file the tracking issue now so it's not lost. Let me know.

VibhorGautam (commented):
this shape makes sense to me - one helper, consistent tags, and callers still get the None behavior

one concern: putting every factory failure at warning might hide actual code bugs. missing credentials and typo'd URLs fit warning, but a TypeError from a validator or constructor bug probably deserves error since the operator cant fix it from config

the service-client helper from #1922 has a similar split, vendor throttles and 5xx stay warning while unexpected exceptions stay error

not blocking, but id rather not have real factory bugs buried in warning noise

could you file the stage 2 typed ServiceClientUnavailable follow-up as an issue before this merges so it doesnt disappear in the PR text

15 factory tests feels like the right granularity

Davidson3556:
fix merge conflict before review

muddlebee:
@0xghost42 thank you for efforts, but closing this since review points not catered like demo and fixes etc. Feel free to open with fixes and a good demo.

context https://github.com/Tracer-Cloud/opensre/pull/1990#issuecomment-4458322254

0xghost42:
Thanks @muddlebee — that's fair.

@cerencamkiran @VibhorGautam will:

- rebase on `main` to clear the conflicts that landed after #1869 + #2009 touched `vercel/client.py` + `victoria_logs/client.py`
- include the end-to-end log snippet directly in the re-opened PR body — same reproducer I posted [here](https://github.com/Tracer-Cloud/opensre/pull/1990#issuecomment-4458405378) (typo'd argocd `base_url` → real factory raises → `_factory_telemetry` fires → captured `report_exception` payload + warning log), so the demo is in front of reviewers instead of buried in a comment
- keep the typed `ServiceClientUnavailable` flow scoped to stage 2 once #1922 lands, per Vibhor's warning-vs-error split

Re-opening once the rebase is clean.

Issue #782, opened 2026-04-23
[FEATURE] : let users choose LLM model during onboarding


muddlebee:
### Problem statement

Users pick an LLM provider during opensre onboard, but the reasoning model is always the provider default (or a previously saved value). There is no step to choose another model.

### Proposed solution

After provider (and credential) setup, add a short “Model” step: show choices from the existing ProviderOption.models tuples in app/cli/wizard/config.py (same labels/values already maintained per provider). 

Default selection should match today’s behavior (default_model or saved model).

muddlebee:
pls dont send a PR for this right now, it needs discussion.

Davidson3556:
Can this be assigned to me @muddlebee

muddlebee:
@Davidson3556 pls share a concise approach note first

Davidson3556:
in flow.py it has an hard-assign model without prompting. The model list per provider already exists in config.py as provider.models. Nothing else in the pipeline cares how the model string was chosen .. save_local_config, sync_provider_env, and .env writes all just take whatever string you hand them.

Plan: one small helper _choose_model(provider, current_model) that renders a select prompt from provider.models, defaults to whatever model is current, and returns the pick. Drop it in after credentials in both branches. That's the whole change , one file.
@muddlebee

muddlebee:
@Davidson3556 your approach not yet fully complete. need to think of all the various scenarios and cases..

- what about the missing models from API in config?
- how to fetch list of all models from API? and stay updated?
- what if wrong model is selected and doesn't support in API? how to verify model usage?

Davidson3556:
**1. Missing models from the static list**

config.py has a curated list that will go stale. Fix: add a "Enter custom model ID" option at the bottom of each provider's list , a _prompt_value fallback for anything not listed. Low effort, covers the gap.

**2.** Fetching models from the provider API

Possible but tricky during onboarding,  the API key was just entered and may not be verified yet, the call adds latency, and not all providers have a /models endpoint with consistent shape. Realistic approach: attempt a fetch after the key is saved (key is in keyring at that point), populate the list dynamically if it succeeds, fall back to the static list if it fails or errors. This is more work but doable.

**3. Model validation / wrong model selected**

[...]

PR #1994, opened 2026-05-14
fix(tests): clear OPENAI_API_KEY in codex adapter "not logged in" tests

### LINKED ISSUE #1991: [BUG] Codex adapter tests fail locally when developer has OPENAI_API_KEY set
### Summary

Two tests in `tests/integrations/llm_cli/test_codex_adapter.py` mock the Codex CLI to simulate "not logged in" — but they don't clear `OPENAI_API_KEY` from the environment. The adapter has an API-key fallback, so if the developer happens to have `OPENAI_API_KEY` exported (in their shell or in the repo's `.env` file), the probe returns `logged_in=True` and the tests fail.

CI machines without the env var pass; anyone with it set gets a red `make test-cov`.

## Expected

`make test-cov` passes regardless of whether the developer has `OPENAI_API_KEY` exported in their shell or their `.env` file.

## Actual

Image

FAILED tests/integrations/llm_cli/test_codex_adapter.py::test_detect_not_logged_in
FAILED tests/integrations/llm_cli/test_codex_adapter.py::test_detect_not_logged_in_exit_zero
AssertionError: assert True is False

where True = CLIProbe(..., logged_in=True, detail='Authenticated via OPENAI_API_KEY fallback.').logged_in

## Steps to reproduce

1. Export `OPENAI_API_KEY` in your shell (or add it to `.env` in the repo root):
   ```bash
   export OPENAI_API_KEY=sk-anything

2. Run:

```
make test-cov
```

The two tests above fail.

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

Where
tests/integrations/llm_cli/test_codex_adapter.py:

test_detect_not_logged_in (around line 73)
test_detect_not_logged_in_exit_zero (around line 121)
The fallback they collide with: app/integrations/llm_cli/codex.py:166-169.

The sibling test test_detect_not_logged_in_uses_openai_api_key_fallback (line 95) shows the right pattern — it explicitly sets the env var via patch.dict.

Suggested fix
Strip OPENAI_API_KEY at the start of each affected test using pytest's monkeypatch, so the env doesn't leak into the test:


def test_detect_not_logged_in(
    monkeypatch: pytest.MonkeyPatch,
    mock_which: MagicMock,
    mock_run: MagicMock,
) -> None:
    monkeypatch.delenv("OPENAI_API_KEY", raising=False)
    ...
raising=False makes the call a no-op on CI (where the var isn't set). pytest auto-restores the env at test teardown, so no leakage to other tests

### Additional context

_No response_

YauhenBichel:
Fixes #1991 

Two tests in `tests/integrations/llm_cli/test_codex_adapter.py` (`test_detect_not_logged_in` and `test_detect_not_logged_in_exit_zero`) didn't isolate themselves from the developer's environment. They mock the Codex CLI to simulate "not logged in," but the adapter has an `OPENAI_API_KEY` fallback — so any developer with that env var set in their shell or `.env` file gets a `True` instead of `False` and the tests fail.

CI passes today because CI doesn't have the var set, but locally these tests are flaky depending on each developer's machine.

This PR adds `monkeypatch.delenv("OPENAI_API_KEY", raising=False)` to the two affected tests, so they explicitly clear the env var before calling `CodexAdapter.detect()`. pytest auto-restores the env at test teardown, so no leakage to other tests.

Before fix:

Screenshot 2026-05-14 at 12 06 26 PM

After fix:

image

---

**The problem:** `CodexAdapter.detect()` falls back to `OPENAI_API_KEY` when `codex login status` reports "not logged in" — that's intentional product behavior so devs can authenticate via API key. But the two failing tests mock `codex login status` to say "not logged in" and assert `logged_in is False`, without clearing `OPENAI_API_KEY` from the environment. If any layer (shell export, `.env` loaded by `app/entrypoints/mcp.py:15` at import time) puts the var into `os.environ`, the fallback kicks in and the assertion flips.

**Why `monkeypatch.delenv(..., raising=False)`:**
- `delenv` removes the env var only for the duration of the test, then pytest restores the original value automatically.
- `raising=False` means: if the var wasn't set in the first place (CI), this is a no-op instead of raising `KeyError`. So CI behavior is unchanged.
- Pattern matches the sibling test `test_detect_not_logged_in_uses_openai_api_key_fallback` (line 95), which already uses `patch.dict` to set the var explicitly. Both tests now control the env explicitly instead of relying on whatever the dev happens to have.

---

---

### DIFF
--- a/tests/integrations/llm_cli/test_codex_adapter.py
+++ b/tests/integrations/llm_cli/test_codex_adapter.py
@@ -8,6 +8,8 @@ import sys
 from pathlib import Path
 from unittest.mock import MagicMock, patch
 
+import pytest
+
 from app.integrations.llm_cli.binary_resolver import diagnose_binary_path, npm_prefix_bin_dirs
 from app.integrations.llm_cli.codex import CodexAdapter, _fallback_codex_paths
 from app.integrations.llm_cli.text import flatten_messages_to_prompt
@@ -70,7 +72,12 @@ def test_detect_path_binary_logged_in(mock_which: MagicMock, mock_run: MagicMock
 
 @patch("app.integrations.llm_cli.codex.subprocess.run")
 @patch("app.integrations.llm_cli.binary_resolver.shutil.which")
-def test_detect_not_logged_in(mock_which: MagicMock, mock_run: MagicMock) -> None:
+def test_detect_not_logged_in(
+    mock_which: MagicMock, mock_run: MagicMock, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    # Strip OPENAI_API_KEY so the dev's shell or .env can't trigger
+    # the API-key fallback and flip logged_in to True.
+    monkeypatch.delenv("OPENAI_API_KEY", raising=False)
     mock_which.return_value = "/usr/bin/codex"
 
     def side_effect(args: list[str], **kwargs: object) -> MagicMock:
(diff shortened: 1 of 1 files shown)


YauhenBichel:
@greptile review

muddlebee:
thank you @YauhenBichel :saluting_face: 

we are still working on to improve the LLM CLI experience better, one example here https://github.com/Tracer-Cloud/opensre/issues/782

Would appreciate your further feedbacks or inputs to make the LLM CLI experience better :)

PR #2022, opened 2026-05-14
fix(services): surface streaming JSON parse drops over a threshold

### LINKED ISSUE #1460: [error-handling] Track silent JSON/NDJSON parse drops in streaming clients
## Problem

Streaming clients silently drop malformed JSON / NDJSON lines:

| File | Lines | Snippet |
|------|------:|---------|
| `app/services/splunk/client.py` | 126–129 | `except json.JSONDecodeError: continue` |
| `app/services/coralogix/client.py` | 87–90 (`_parse_ndjson`) | `except json.JSONDecodeError: continue` |
| `app/services/coralogix/client.py` | 145–148 (`_parse_user_data`) | `except json.JSONDecodeError: return {}` |
| `app/services/victoria_logs/client.py` | 167–172 (`_parse_ndjson`) | `with contextlib.suppress(json.JSONDecodeError): obj = json.loads(stripped)` |
| `app/services/vercel/client.py` | 143–147 (`_ingest_runtime_log_stream_line`) | `except json.JSONDecodeError: return len(bucket) >= limit` |
| `app/services/grafana/tempo.py` | 107–135 | non-200 falls through to `{"spans": []}` with debug log only |
| `app/services/lambda_client.py` | 329–342 (`invoke_function`) | `json.loads(response["Payload"].read())` un-handled |

Skipping a single malformed line in NDJSON is intentional — vendors do produce flaky frames. The problem is **observability**: if 90% of lines are failing to parse (e.g. wrong content type, vendor migration, schema break), we never find out.

## Proposal

Add a tiny shared helper in `app/services/_streaming.py`:

```python
class StreamingParseStats:
    parsed: int = 0
    skipped: int = 0
    errors: dict[str, int]  # error type → count

    def record_parsed(self): ...
    def record_error(self, exc: BaseException): ...

    def report_if_unhealthy(self, *, integration: str, source: str, threshold: float = 0.10) -> None:
        """Capture to Sentry if skipped / (parsed + skipped) > threshold."""
```

Each streaming client wraps its parse loop with one of these and calls `report_if_unhealthy(...)` at the end of the response (or every N seconds for true streams). When the skip ratio exceeds the threshold, we emit a single Sentry event tagged with the integration, the source endpoint, and the histogram of error types — so we don't spam Sentry with a million events for a million bad lines.

Also:

- Replace `with contextlib.suppress(json.JSONDecodeError):` in VictoriaLogs with the explicit try/except that goes through the helper.
- In `grafana/tempo.py` `_fetch_trace_spans`, replace the `except Exception: logger.debug(..., exc_info=True); return {"spans": []}` with `report_exception(..., severity="warning")` so the failure is visible.
- In `lambda_client.py:invoke_function`, wrap the `json.loads(...)` of the payload in a try/except that captures `JSONDecodeError` with the function name as a tag.

## Acceptance criteria

- [ ] `app/services/_streaming.py` exists with `StreamingParseStats` + tests.
- [ ] All listed sites use `StreamingParseStats` (or `report_exception` for non-streaming JSON).
- [ ] Sentry receives a single event per response with parse-rate stats when the skip threshold is exceeded.
- [ ] No regression in existing parser unit tests.

## Depends on

#1454

nuemaan:
Fixes #1460. Assigned to me per the comment thread there.

The streaming log clients (Splunk, Coralogix, VictoriaLogs, Vercel) all swallow \`JSONDecodeError\` per line. A single flaky frame is fine and expected from vendors. The problem is when most lines start failing (content-type drift, schema break) and the agent quietly gets empty results.

\`app/services/_streaming.py\` adds \`StreamingParseStats\`: each parse loop records parsed/skipped counts and an error-type histogram. \`report_if_unhealthy\` at end of response fires a single Sentry event (severity \`warning\`) when the skip ratio crosses the threshold (default 10%), tagged \`integration\` + \`source\`, with the full histogram in extras. One event per response, not per dropped line.

Per-site notes:

- \`splunk/client.py:_parse_export_response\` — straight migration.
- \`coralogix/client.py\` — both NDJSON and per-row \`_parse_user_data\` JSON drops are tracked into the same stats so the threshold is computed over all parse failures in one response. Helper threaded through as an optional kwarg, no behavior change for any other caller.
- \`victoria_logs/client.py:_parse_ndjson\` — replaced the \`contextlib.suppress(JSONDecodeError)\` with an explicit try/except that feeds the helper.
- \`vercel/client.py:_collect_runtime_logs_from_stream\` + \`_ingest_runtime_log_stream_line\` — stats threaded as an optional kwarg so the existing direct unit tests for \`_ingest_runtime_log_stream_line\` keep their old calling convention.

The two non-streaming sites use \`report_exception\` directly per the issue spec:

- \`grafana/tempo.py:_fetch_trace_spans\` — the \`except Exception: logger.debug(...); return {"spans": []}\` swallow is replaced with a warning-level \`report_exception\` tagged \`integration=grafana, component=app.services.grafana.tempo\` so the failure is visible in Sentry.
- \`lambda_client.invoke_function\` — wraps the \`json.loads(response["Payload"]...)\` so a function returning a non-JSON payload reports a warning with the function name and a 200-byte payload preview.

### Verified

- \`uv run pytest tests/services/test_streaming.py\` — 11 passed.
- Wider sweep across the touched areas (\`tests/services/\`, splunk/victoria_logs/lambda tool tests, integration tests): 775 passed, 3 skipped (pre-existing skips).
- \`uv run ruff check\` and \`uv run ruff format --check\` clean.

### DIFF
--- a/app/services/_streaming.py
+++ b/app/services/_streaming.py
@@ -64,13 +64,10 @@ class StreamingParseStats:
             return
         # Synthesize an exception so report_exception (which routes through
         # Sentry's capture_exception) has something to attach the histogram to.
-        # Per-line tracebacks were discarded; the failure-type counter is the
-        # signal we care about.
-        synthetic = RuntimeError(
-            f"{integration}: streaming parse skip ratio "
-            f"{self.skip_ratio:.0%} ({self.skipped}/{self.total}) "
-            f"exceeded threshold {threshold:.0%}"
-        )
+        # Keep the message static per integration so Sentry groups every
+        # unhealthy response into a single issue; dynamic counts live in
+        # extras where the dashboard can break them down.
+        synthetic = RuntimeError(f"{integration}: streaming parse rate unhealthy")
         report_exception(
             synthetic,
             logger=logger,
--- a/app/services/_streaming.py
+++ b/app/services/_streaming.py
@@ -1,0 +1,95 @@
+"""Parse-rate tracking for streaming JSON / NDJSON clients.
+
+Vendors occasionally emit a flaky line in an NDJSON stream, so dropping the
+odd malformed frame is expected. The risk is when the drop ratio crosses a
+threshold (content-type drift, schema break, vendor migration) and we keep
+returning empty results to the agent.
+
+``StreamingParseStats`` accumulates per-response parse outcomes;
+``report_if_unhealthy`` emits a single Sentry event (with the failure
+histogram in extras) when the skip ratio exceeds the threshold, so we do
+not spam one event per dropped line.
+"""
+
+from __future__ import annotations
+
+import logging
+from collections import Counter
+from dataclasses import dataclass, field
+
+from app.utils.errors import report_exception
+
+#: Default fraction of skipped lines tolerated before a stream is flagged.
+DEFAULT_SKIP_THRESHOLD: float = 0.10
+
+
+@dataclass
+class StreamingParseStats:
+    """Counters for one streaming-parse pass over a vendor response."""
+
+    parsed: int = 0
+    skipped: int = 0
+    errors: Counter[str] = field(default_factory=Counter)
+
+    def record_parsed(self) -> None:
+        self.parsed += 1
+
+    def record_error(self, exc: BaseException) -> None:
+        self.skipped += 1
+        self.errors[type(exc).__name__] += 1
+
+    @property
+    def total(self) -> int:
+        return self.parsed + self.skipped
+
+    @property
+    def skip_ratio(self) -> float:
+        return self.skipped / self.total if self.total else 0.0
+
+    def report_if_unhealthy(
+        self,
+        *,
+        logger: logging.Logger,
+        integration: str,
+        source: str,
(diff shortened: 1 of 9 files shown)


nuemaan:
@greptile review

nuemaan:
Pushed fixes for both Greptile findings:

- Synthetic exception message is now stable per integration; ratio and counts only live in extras so Sentry groups all unhealthy-stream events into one issue.
- `_parse_user_data` now calls `record_parsed` on the success branch so the skip-ratio denominator counts both NDJSON and userData attempts symmetrically.

Added `test_synthetic_message_is_stable_per_integration` in `tests/services/test_streaming.py` and `test_parse_user_data_counts_successes_and_failures_symmetrically` in `tests/services/test_coralogix_client.py` to lock both behaviours.

@greptile review

muddlebee:
@greptile review

muddlebee:
congrats @nuemaan on first PR

nuemaan:
thanks @muddlebee

PR #1877, opened 2026-05-12
feat(synthetic/rds): add upstream correlation primitives

cerencamkiran:
## Summary

Adds upstream-correlation support for symptom-first RDS investigations targeting #1434.

This PR introduces deterministic correlation/scoring primitives plus runtime wiring so an RDS CPU spike can be attributed to likely upstream EC2/app-tier drivers without relying on trace IDs or request IDs.

### Correlation primitives

Added modular components under:

`tests/synthetic/rds_postgres/correlation/`

Including:

* time-window correlation scoring
* periodic spike detection
* topology adjacency scoring
* operator-hint scoring
* deterministic candidate ranking
* weighted upstream candidate aggregation
* correlation reporting payload generation
* no-trace-id investigation flow coverage

### Production-facing correlation layer

Added upstream evidence contracts and providers under:

`app/correlation/`

Including:

* typed evidence contracts for metrics, logs, topology hints, and operator hints
* no-op provider baseline
* query-backed upstream evidence provider
* Datadog metric/log adapter
* Datadog upstream evidence provider

### Runtime integration

Wired upstream correlation into the investigation runtime:

* added `correlate_upstream` investigation node
* routed `diagnose -> correlate_upstream -> publish`
* added `correlation` to investigation state defaults/model
* passed correlation into publish report context
* rendered `correlated_signals` and `most_likely_causal_drivers` in final report output
* `correlate_upstream` is traceable and invokes the configured upstream evidence provider before publish

### Synthetic scenario coverage

Added scenario-backed validation using:

`015-mysql-ec2-load-attribution`

Coverage validates:

* web tier ranks above worker tier
* confidence scoring remains deterministic
* topology + temporal alignment affect ranking correctly
* correlation payloads are emitted in canonical observation reports

## Validation

```bash
ruff check app/state app/pipeline app/nodes app/correlation \
  tests/nodes/publish_findings/test_correlation_report_section.py \
  tests/test_datadog_provider.py \
  tests/test_datadog_adapter.py \
  tests/test_upstream_correlation_provider.py \
  tests/synthetic/rds_postgres/correlation

pytest tests/nodes/investigate/test_correlate_upstream.py \
  tests/pipeline/test_graph_adapt_window_wiring.py \
  tests/nodes/publish_findings/test_correlation_report_section.py \
  tests/test_datadog_provider.py \
  tests/test_datadog_adapter.py \
  tests/test_upstream_correlation_provider.py \
  tests/synthetic/rds_postgres/test_observation_correlation.py \
[...]

PR #2057, opened 2026-05-15
feat(delivery): align Slack + Discord RCA formatting with Telegram

### LINKED ISSUE #2007: Align Slack and Discord RCA delivery formatting with Telegram improvements
## Summary

Recent work improved **Telegram** RCA output (HTML `parse_mode`, severity-aware headers, evidence sections tuned for Telegram, and related tests under `app/delivery/publish_findings/formatters/`). **Slack** still uses mrkdwn / Block Kit via `format_slack_message` / `build_slack_blocks`, and **Discord** currently reuses the same Slack-formatted string in `generate_report` (`send_discord_report(slack_message, …)`), so Discord does not get channel-native formatting.

This issue tracks bringing **Slack** and **Discord** to parity where it makes sense: consistent structure, readable links, evidence citations, and channel-specific escaping/limits.

## Motivation

- **Slack** mrkdwn and Block Kit have different constraints than Telegram HTML (length limits, link syntax, bold/italic rules).
- **Discord** markdown is not identical to Slack mrkdwn; reusing `slack_message` risks subtle rendering bugs or lost formatting.
- Users expect comparable clarity across chat surfaces after an `opensre investigate` run.

## Scope (suggested)

1. **Slack** — Audit `format_slack_message` / `build_slack_blocks` against the Telegram formatter feature set (severity header, provenance, evidence blocks, links). Adjust mrkdwn and Block Kit payloads so structure and emphasis match the intended UX; add or extend unit tests (similar to `tests/delivery/test_report_provenance.py` for Telegram).
2. **Discord** — Introduce a dedicated formatter (or adapter) for Discord messages instead of passing the Slack string through unchanged; wire it in `app/delivery/publish_findings/node.py` and `app/utils/discord_delivery.py` as needed; respect Discord message length and markdown quirks.
3. **Shared context** — Keep a single `ReportContext` / shared section builders where possible; only split where channel syntax diverges (same pattern as today for Telegram vs Slack).

## Acceptance criteria

- [ ] **Slack**: Published investigation reports render correctly in a real Slack workspace (thread reply + blocks): headers/emphasis, links (including evidence URLs), and no broken mrkdwn from edge cases (e.g. stray `*` / underscores in claim text).
- [ ] **Discord**: Published reports render correctly in a real Discord channel (or thread): formatting is intentional for Discord, not a copy-paste of Slack mrkdwn that happens to work.
- [ ] **Parity**: For a fixed `ReportContext` / fixture, Slack, Discord, and Telegram outputs cover the same logical sections (title/summary, root cause, validated vs non-validated claims, evidence, key links) modulo channel limitations; gaps are documented in code comments or docs if a channel cannot represent a section.
- [ ] **Tests**: Unit tests cover formatters for Slack and Discord (including at least one regression case inspired by the Telegram HTML work, e.g. special characters or citation-heavy claims).
- [ ] **Live demo — `opensre investigate`**: Demonstrate end-to-end delivery with a real investigation run:
  - Configure integrations for **Slack** and **Discord** (and Telegram optional for comparison).
  - Run **`uv run opensre investigate`** (or the supported equivalent in your environment) against a representative alert/fixture so the full publish path executes.
  - Capture screenshots or a short screen recording showing the final messages in Slack and Discord, and attach them to the PR that closes this issue (or link to them in the PR description).

## References

- `app/delivery/publish_findings/formatters/report.py` — `format_telegram_message`, `format_slack_message`, `build_slack_blocks`
- `app/delivery/publish_findings/node.py` — `generate_report` / delivery wiring
- `app/utils/discord_delivery.py`, `app/utils/slack_delivery.py`, `app/utils/telegram_delivery.py`

EmmanuelOluwafemi:
Closes #2007.

## Summary

- Introduces a channel-agnostic `Section` pipeline so Slack, Telegram, and Discord all render from one ordered section list. Adding a future channel is now a renderer file, not a re-implementation.
- Brings Slack to severity-header parity with Telegram (mrkdwn + Block Kit) and adds a dedicated Discord renderer with severity-colored embed, fixing the bug where `node.py` was passing `slack_message` straight into `send_discord_report`.
- `dedupe_sections` ports `_telegram_baseline_repeats_header` into a pure post-pass — every channel inherits the banner-noise filter without per-channel code.

## Architecture

```
build_sections(ctx)         -> list[Section]    # pure extraction
dedupe_sections(sections)   -> list[Section]    # cross-section noise filter
prepare_sections_for(ctx) = dedupe_sections(build_sections(ctx))

renderers/slack.py     walks prepare_sections_for(ctx)
renderers/telegram.py  walks prepare_sections_for(ctx)
renderers/discord.py   walks prepare_sections_for(ctx)
```

`Section` is a `frozen` dataclass with a `SectionKind` discriminator and a permissive `extras: Mapping[str, Any]` field. The severity vocabulary lives in the new `formatters/severity.py` so emoji + display normalisation is one source of truth across the three renderers.

`report.py` is now a thin re-export shim — its derivation/syntax helpers stay put (lazily imported by `sections.py`) to keep the diff narrow. A follow-up can collapse those helpers into `sections.py` or a sibling module.

## Commits (chronological)

1. **`608facbb`** — `feat(delivery): add channel-agnostic Section pipeline for RCA reports`
2. **`6706c0b7`** — `test(delivery): pin Slack + Telegram renderer output for issue/2007 refactor`
3. **`61b38a40`** — `refactor(delivery): route Slack + Telegram rendering through prepare_sections_for`
4. **`19b528e4`** — `feat(delivery): add Discord renderer with severity-colored embed`
5. **`a9b9bf50`** — `feat(delivery): wire build_discord_embed through publish_findings.node`
6. **`be4b7366`** — `feat(delivery): bring Slack to severity parity with Telegram`
7. **`07750527`** — `test(delivery): edge-case coverage + fix Discord URL paren handling`

## Acceptance criteria

- [x] **Slack**: published reports render correctly — severity header (mrkdwn + Block Kit), evidence URLs via `<url|label>`, no broken mrkdwn on edge cases (stray `*`, LLM `**bold**`, `##` headings).
- [x] **Discord**: published reports use Discord-native markdown — `**bold**` titles, `` `code` `` snippets, `[label](<url>)` links (angle-bracket wrap so URLs with `)` don't break), severity-colored embed with alert-name title.
- [x] **Parity**: same `ReportContext` produces equivalent sections across Slack / Telegram / Discord; channel divergence only where syntax forces it (e.g. Discord puts severity in title + color bar instead of description; Telegram/Discord skip `FAILED_PODS` and surface pods in the investigation trace).
- [x] **Tests**: 60 new tests across `tests/delivery/` — 18 for `sections.py`, 10 characterization tests pinning Slack/Telegram baseline, 21 for the Discord renderer, 17 edge-case tests, plus 5 updated `send_discord_report` tests for the new embed contract. Full unit suite (6584 tests) green.
- [x] **Live demo** — see screenshots below.

## Test plan

- [x] `make test-cov` passes (6584 tests, 4 skipped — those are env-gated live-infra paths)
- [x] `make lint` passes
- [x] `make format-check` passes
- [x] `make typecheck` passes (mypy on 559 source files)
- [x] Banner-only root cause is dropped uniformly across all three channels (cross-channel parity test)
- [x] Unknown / empty severity falls back to ⚠️ + UNKNOWN consistently
- [x] Live `opensre investigate` run published to a real Slack workspace and Discord channel (screenshots below)

## Screenshots

**Slack** (mrkdwn fallback + Block Kit primary):

image

**Discord** (severity-colored embed with native markdown):

image

## Follow-ups (out of scope for this PR)

- **OpenAI `max_tokens` → `max_completion_tokens`**: surfaced while doing the live demo. The default `OPENAI_REASONING_MODEL=gpt-5.4` rejects `max_tokens` with a 400. Worked around by setting `OPENAI_REASONING_MODEL=gpt-4o`; a proper fix in `app/services/agent_llm_client.py` is its own PR.
- **Structured trace data**: `build_investigation_trace` (via `format_pod_line`) bakes Slack `<url|label>` into trace strings. The Discord renderer translates them at render time, but the right fix is pushing link data into `Section.extras` so the trace stays channel-neutral.
- **Collapse `report.py` helpers into `sections.py`**: the lazy imports from `report.py` are documented as transitional. Phase B of the architecture closes the loop and makes `report.py` a pure re-export shim.
- **`FAILED_PODS` parity for Telegram + Discord**: currently only Slack Block Kit renders a dedicated Failed Pods section; the other two surface the same info through the investigation trace. A future PR could add explicit pod sections to Telegram/Discord if product wants it.


Davidson3556:
fix merge conflict

EmmanuelOluwafemi:
Thanks for the catch on the empty-root-cause + no-top_log path — that was a real regression I missed. The early return in \`_build_root_cause\` was short-circuiting the body-assignment ternary that *was* supposed to produce the "Not determined (insufficient evidence)." fallback in exactly that case.

Fixed in 68c0dde0:

- Removed the early return so \`_build_root_cause\` always emits a \`ROOT_CAUSE\` section with the fallback body when both inputs are empty.
- The dedup heuristic (\`dedupe_sections._root_cause_repeats_header\`) already returns \`False\` for the "Not determined" prefix, so the restored fallback survives the dedup pass — including when a \`SEVERITY_HEADER\` is present.

Three new tests lock the behaviour down:

- \`test_build_sections_emits_fallback_root_cause_when_no_sentence_or_log\` — section-level invariant
- \`test_prepare_sections_for_keeps_fallback_root_cause_even_with_severity_header\` — proves dedup doesn't strip it in the header+empty case
- \`test_empty_investigation_renders_not_determined_fallback_in_all_channels\` — end-to-end across Slack mrkdwn, Telegram HTML, and Discord markdown

105 delivery tests now passing; lint/format/typecheck green.

Davidson3556:
@greptile review

muddlebee:
@EmmanuelOluwafemi could you check if you could refer use telegram formatter which was already there in code?


@greptile review with above valid point.

EmmanuelOluwafemi:
> @EmmanuelOluwafemi could you check if you could refer use telegram formatter which was already there in code?
> 
> @greptile review with above valid point.

 Yes — the entire architecture is built around the existing Telegram formatter as the parity target.
  Specific mappings:

  - Section ordering in `build_sections` mirrors `format_telegram_message`'s output order (severity
  header → root cause → claims → provenance → remediation → trace → evidence → cloudwatch → meta). See
  sections.py docstring.
  - The banner-noise dedupe in `dedupe_sections._root_cause_repeats_header` is a port of
  `_telegram_baseline_repeats_header` lifted into the shared pipeline, so Slack and Discord inherit the
  same heuristic.
  - Severity emoji vocabulary moved from `_severity_telegram_header` into `formatters/severity.py` — all
   three renderers import from one source.
  - `_format_correlation_lines` from PR #1877 is consumed unchanged by `_build_correlation`
  (sections.py).
  - 10 characterization tests in `test_formatters_characterization.py` pin the pre-refactor Telegram
  output as the regression baseline.

  The five Telegram-specific helpers that were dead post-refactor (`_render_claim_lines_telegram`,
  `_telegram_baseline_repeats_header`, `_severity_telegram_header`, `_norm_banner_key`,
  `_render_claim_lines`) were removed in 863f4b05 — their behavior is now in `sections.py` /
  `severity.py` / per-renderer modules.

  If you meant something more specific — e.g., translating the Telegram HTML output directly for
  Slack/Discord rather than building from sections — happy to discuss the tradeoff.

muddlebee:
@greptile review and confirm the points https://github.com/Tracer-Cloud/opensre/pull/2057#issuecomment-4460170048

muddlebee:
> Yes — the entire architecture is built around the existing Telegram formatter as the parity target.

actually no @EmmanuelOluwafemi it is not, we are inventing everything from scratch. we should be re-using most of the already existing formatter. this needs more deeper review.

cerencamkiran:
Good work @EmmanuelOluwafemi 

Only small follow-up thought:
there are still a few renderer paths (cited evidence, cloudwatch, tool-call summary) that read directly from ctx instead of flowing entirely through Section.extras, so there’s still a little bit of hybrid architecture left to centralize later. Also agree the bigger architectural question is whether more of the existing Telegram formatter logic should be reused instead of rebuilding parallel renderer paths.

EmmanuelOluwafemi:
> Good work @EmmanuelOluwafemi
> 
> Only small follow-up thought: there are still a few renderer paths (cited evidence, cloudwatch, tool-call summary) that read directly from ctx instead of flowing entirely through Section.extras, so there’s still a little bit of hybrid architecture left to centralize later. Also agree the bigger architectural question is whether more of the existing Telegram formatter logic should be reused instead of rebuilding parallel renderer paths.

Thanks for the feedback @cerencamkiran and @muddlebee Just one more question for clarity before I update the PR, are we going with @muddlebee suggestion which is relating to me just reusing already existing format without inventing everything from scratch or @cerencamkiran feedback which is to also apply the new structure on the other renderer paths, I think this will give me more clarity on next steps.

Davidson3556:
Nice refactor @EmmanuelOluwafemi  extracting the derivation helpers into `_derive.py` and shrinking `report.py` to a shim is the right direction, and the Section model is a solid contract. Two things I'd want to confirm , both sharpening what @muddlebee and @cerencamkiran already raised:

**1. EVIDENCE and LINK sections are built but Slack/Telegram don't actually consume them.** Both renderers dispatch on the section kind and then discard the section to re-read `ctx`:
- `renderers/slack.py:82-85` and `:348-364` — `format_cited_evidence_section(ctx)` / `render_cloudwatch_link(ctx)`, section arg unused.
- `renderers/telegram.py:64-67` — same pattern with the `_html` variants.
- Only `renderers/discord.py:247` reads `section.extras["catalog"]`.

`_build_evidence` filters `evidence/datadog/failed_pod/` at `sections.py:244`; `evidence.py:266` and `:306` re-apply the same filter against `ctx`. The two paths agree today, but nothing pins them..change one and Discord silently diverges from Slack/Telegram. No test enforces parity here either (`grep "SectionKind.EVIDENCE\|SectionKind.LINK"` returns 0 across the new test files).

**2. `Section.items` for TRACE is Slack-flavored, not channel-neutral.** `build_investigation_trace` → `format_pod_line` (`infrastructure.py:69`) emits `<url|label>` Slack syntax into the items, so Telegram un-wraps it via `_to_telegram_html_body` and Discord re-parses it with `_SLACK_LINK_RE` / `_translate_slack_links` (`discord.py:53,97,242`). Pushing link data into `Section.extras` would let every renderer consume plain strings + parallel link metadata.

**Suggestion:**  fold evidence catalog + cloudwatch link + tool-calls summary into `Section.extras` and add one cross-channel parity test on a fixture ctx

muddlebee:
@EmmanuelOluwafemi thank you for the PR, but requirements aren't met. We will try to stick to [DRY principle](https://www.wikiwand.com/en/Don%27t_repeat_yourself) as much as possible. 

Currently at this stage, adding too much of new code just leads to difficulty in maintainability long term. I would encourage to study the existing function/methods and how we could re-use or add/edit, rather than inventing a whole new paradigm.

PR #1533, opened 2026-05-07
feat(llm-cli): add GitHub Copilot CLI provider (non-interactive)

Davidson3556:
Fixes #1106

#### Describe the changes you have made in this PR -

Adds GitHub Copilot CLI as a non-interactive `LLM_PROVIDER`, following the existing `LLMCLIAdapter` contract used by Codex, Claude Code, Gemini CLI, etc.

- **Adapter** — `app/integrations/llm_cli/copilot.py` implements `CopilotAdapter` with `detect()`, `build()`, `parse()`, `explain_failure()`. Invocation is `copilot -p PROMPT --no-color --no-banner --no-ask-user --silent --log-level none` so it runs to completion without ever blocking on stdin/TTY.
- **Auth (no GitHub-token dependency by default)** — primary signal is the existence of `$COPILOT_HOME` (default `~/.copilot`), which is where the interactive `/login` slash command stores credentials. `COPILOT_GITHUB_TOKEN` / `GH_TOKEN` / `GITHUB_TOKEN` are only honored as fallbacks when no stored CLI login is found. Three-state `logged_in` (`True` / `False` / `None`) is preserved.
- **Wiring** — registered in `CLI_PROVIDER_REGISTRY` (`COPILOT_MODEL`), added `copilot` to `LLMProvider` literal + validators + the no-API-key branch in `app/config.py`, added `COPILOT_` to the subprocess env prefix allowlist, and added `COPILOT_CLI_ENV_KEYS` for explicit token forwarding via `nonempty_env_values(...)` so token vars never leak to other CLI subprocesses.
- **Wizard** — `ProviderOption(value="copilot", ...)` with `_copilot_adapter_factory` and a small `COPILOT_MODELS` list.
- **Docs / env** — `.env.example`, `docs/llm-providers.mdx`, and `app/integrations/llm_cli/AGENTS.md` updated with the new vars and adapter row.
- **Tests** — 17 cases in `tests/integrations/llm_cli/test_copilot_adapter.py` (detect: stored creds / token fallback / unclear / not-found / version-failure; build: argv flags, workspace fallback, model flag, token env forwarding, missing-binary error; runner: env merge + leak prevention; registry resolution; subprocess env allowlist). Also extended `tests/test_config.py` and `tests/cli/wizard/test_flow.py`.

### Demo/Screenshot for feature changes and bug fixes -

Screenshot 2026-05-07 at 00 06 01

---

## Code Understanding and AI Usage

[...]

PR #1196, opened 2026-05-02
feat(opencode-cli): add opencode cli as llm cli provider

cokerrd:
Fixes #
#1111 

#### Describe the changes you have made in this PR -

Adds OpenCode as a non-interactive CLI-backed LLM provider, selectable via `LLM_PROVIDER=opencode`. The integration mirrors the existing Codex CLI adapter pattern (adapter, registry, wizard option, fallback binary paths, auth detection) so users can run `opensre investigate` using their local `opencode` binary instead of an HTTP API key.

**Changes**

- **New `OpenCodeAdapter`** in `app/integrations/llm_cli/opencode.py` that runs `opencode run` with the prompt on stdin — strictly non-interactive, no TTY, no approval loops.

- **Auth detection** reads `~/.local/share/opencode/auth.json` (XDG-compliant on all platforms — Linux, macOS, Windows) and validates provider credentials. Returns three-state `(True/False/None)` based on file existence and valid API keys.

- **Provider registered** in `CLI_PROVIDER_REGISTRY` and selectable through onboarding wizard under "Local CLI providers".

- **Wizard model options** include curated list: CLI default, ...

- **Comprehensive tests** covering auth classification (valid credentials, empty, malformed JSON), detection, command building, model flag handling, error explanations, fallback paths (macOS/Linux/Windows), and registry registration.

- **`.env.example`** documents `OPENCODE_BIN` and `OPENCODE_MODEL`.

- **No API key forwarding** — OpenCode uses OAuth file auth only, so no credentials are passed via environment variables.

### Demo/Screenshot for feature changes and bug fixes - 

mac demo

https://github.com/user-attachments/assets/50eeeda2-dcca-4aaf-b908-f7bf6c238924

linux demo

https://github.com/user-attachments/assets/acf95e73-3a24-461c-9b5b-3c5979e0401d

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [X] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [X] I have reviewed every single line of the AI-generated code
- [X] I can explain the purpose and logic of each function/component I added
- [X] I have tested edge cases and understand how the code handles them
- [X] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**

### What problem does your code solve?
[...]

PR #1168, opened 2026-05-01
feat(llm_cli): integrate Claude Code CLI as LLM_PROVIDER

Davidson3556:
Fixes #1108

#### Describe the changes you have made in this PR -

Adds Anthropic Claude Code as a non-interactive CLI-backed LLM provider, selectable via `LLM_PROVIDER=claude-code`. The integration mirrors the existing OpenAI Codex CLI adapter shape end-to-end (adapter, registry, wizard option, env forwarding, fallback binary paths) so users can run `opensre investigate` against an alert using their local `claude` binary instead of an HTTP API key.

Highlights:
- New `ClaudeCodeAdapter` in `app/integrations/llm_cli/claude_code.py` that runs `claude -p --output-format text` with the prompt on stdin — strictly non-interactive, no TTY, no approval loops.
- Three-state auth probe (True / False / None) that checks `ANTHROPIC_API_KEY` env, then `~/.claude/.credentials.json`. On macOS, when neither is present, returns `None` instead of `False` because `claude login` stores OAuth tokens in Keychain on darwin — a strict file-only check would falsely block Keychain-authenticated users.
- Binary resolution via the shared `resolve_cli_binary` helper — `CLAUDE_CODE_BIN` override → PATH → npm/Homebrew/volta/pnpm/XDG fallbacks.
- Provider registered in `CLI_PROVIDER_REGISTRY` and selectable through the onboarding wizard under "Local CLI providers".
- 29 unit tests covering detect / build / parse / explain_failure, the `CLAUDE_CODE_BIN` override and PATH fallback, fallback paths on macOS / Linux / Windows (mocked `sys.platform`), registry registration, and env-forwarding security guarantees (specifically: `ANTHROPIC_API_KEY` is forwarded ONLY by the Claude Code invocation, not via the global allowlist, so credentials don't leak into Codex subprocesses).
- `.env.example` documents `CLAUDE_CODE_BIN` and `CLAUDE_CODE_MODEL`.
- `runner.py`: added `CLAUDE_` to `_SAFE_SUBPROCESS_ENV_PREFIXES` so `CLAUDE_CODE_*` settings reach the subprocess; added `USER` and `LOGNAME` to safe keys (macOS Keychain item lookup needs `USER`).

### Demo/Screenshot for feature changes and bug fixes -

Screenshot 2026-04-30 at 17 46 14
Screenshot 2026-04-30 at 18 07 57
Screenshot 2026-04-30 at 18 12 56

---

## Code Understanding and AI Usage

[...]

PR #2069, opened 2026-05-15
feat(wizard): add model selection step during onboarding (#782)

### LINKED ISSUE #782: [FEATURE] : let users choose LLM model during onboarding

Davidson3556:
Fixes #782

Adds a "Model" step to `opensre onboard` so users can pick a model after the credential step, instead of always getting the provider default.

- New `app/cli/wizard/model_discovery.py`: best-effort per-provider model listing (OpenAI, Anthropic, OpenRouter, Requesty, Gemini, NVIDIA, Ollama) via `httpx` with a 3s timeout. Never raises; any failure returns `[]` so onboarding never breaks.
- New `_choose_model` step in `app/cli/wizard/flow.py`:
  - Curated `ProviderOption.models` are shown as `recommended`.
  - Discovered-only IDs are shown as `untested`.
  - A previously saved model not in either list is preserved as `current (custom)`.
  - "Enter custom model ID" escape hatch is always last; custom IDs outside the curated/discovered set print a warning but are still saved (per the discussion consensus — no chat/completions validation call in v1).
- CLI-backed providers (codex, claude-code, gemini-cli, opencode, kimi, copilot, cursor) and Bedrock skip discovery and show the curated list + escape hatch.
- Picked model flows into the existing `save_local_config` / `sync_provider_env` path — no schema changes, nothing else cared how the model string was chosen.

video link https://drive.google.com/file/d/1TtGzlLijSyGN8XMtvyBXvUo9xlgliubP/view?usp=sharing
Screenshot 2026-05-15 at 07 14 13

---

The issue was that `flow.py` hard-assigned the model after provider selection — `ProviderOption.models` already existed in `config.py` but the wizard never surfaced it. The whole pipeline downstream (`save_local_config`, `sync_provider_env`, `.env` writes) just takes whatever string you hand it, so the change is localized to onboarding UX.

I considered three approaches from the thread:
1. **Static list only** — simplest, but goes stale as new models ship.
2. **Always fetch from provider API at runtime** — most accurate, but onboarding then depends on a working API call right after the key is entered, and not every provider has a consistent `/models` shape.
3. **Static list + best-effort dynamic fetch + custom escape hatch** — the path picked here, matching what @imjohnzakkam and @cerencamkiran landed on.

The split between `model_discovery.py` (pure HTTP + parsing, returns `list[str]`, never raises) and `_choose_model` (prompt UX) keeps each piece testable on its own. Discovery is keyed on `provider.value` so adding a provider later means one fetcher function, no UI changes.

Edge cases covered by tests:
- Discovery succeeds and adds new IDs → merged into the picker, curated entries still listed first.
- Discovery fails / times out → falls back to curated list silently; onboarding completes normally.
- Saved model isn't in curated or discovered list → preserved as `current (custom)` so re-running the wizard doesn't quietly drop a user's previous choice.
- User picks the "Enter custom model ID" escape hatch → free-text accepted, warning printed if it's outside both lists.
- CLI / Bedrock providers → discovery skipped entirely, no spurious HTTP.

No chat/completions validation call. The thread agreed that a probe at onboarding time has too many false negatives (tiers, regions, provider quirks) and could block a perfectly valid setup. The "untested" label communicates the uncertainty without hard-failing.

Tests: `tests/cli/wizard/test_model_discovery.py` (9 cases, per-provider parsing + error paths), `tests/cli/wizard/test_model_selection.py` (6 cases, default / merge / fallback / saved-preservation / custom entry / CLI-skip). Existing wizard flow tests keep working via an autouse `conftest.py` stub so each `run_wizard` test didn't need a new select response. The interactive PTY smoke test (`tests/cli_smoke_test.py`) was updated to acknowledge the new step.

---

### DIFF
--- a/app/cli/wizard/config.py
+++ b/app/cli/wizard/config.py
@@ -233,10 +233,6 @@ GEMINI_CLI_MODELS = (
         label="gemini-3.1-pro-preview — newest frontier (preview)",
     ),
     ModelOption(
-        value="gemini-3-pro-preview",
-        label="gemini-3-pro-preview — frontier (preview)",
-    ),
-    ModelOption(
         value="gemini-3-flash-preview",
         label="gemini-3-flash-preview — fast (preview)",
     ),
@@ -268,10 +264,6 @@ OPENCODE_MODELS = (
         label="CLI default (no -m; use OpenCode configured model)",
     ),
     ModelOption(value="opencode/gpt-5.5", label="GPT-5.5 (OpenCode Zen) — frontier"),
-    ModelOption(
-        value="opencode/gpt-5.5-pro",
-        label="GPT-5.5 Pro (OpenCode Zen) — frontier deep reasoning",
-    ),
     ModelOption(value="opencode/gpt-5.4", label="GPT-5.4 (OpenCode Zen)"),
     ModelOption(value="opencode/gpt-5.4-mini", label="GPT-5.4 mini (OpenCode Zen) — fast"),
     ModelOption(
--- a/app/cli/wizard/config.py
+++ b/app/cli/wizard/config.py
@@ -192,6 +192,7 @@ OLLAMA_MODELS = (
     ModelOption(value="qwen2.5:7b", label="Qwen 2.5 (7B)"),
 )
 
+# Source: https://platform.claude.com/docs/en/about-claude/models/overview (verified 2026-05-21).
 # Empty value means "no --model" so Claude Code uses its configured default.
 CLAUDE_CODE_MODELS = (
     ModelOption(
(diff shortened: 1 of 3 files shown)


Davidson3556:
@greptile review

muddlebee:
@Davidson3556 we dont need to create a new model discovery method. check comments here https://github.com/Tracer-Cloud/opensre/issues/782#issuecomment-4342001726

we have to re-use the existing model config.

Davidson3556:
okay @muddlebee

Davidson3556:
@greptile review

Davidson3556:
@muddlebee kindly revieew

muddlebee:
@Davidson3556 we also need to update all latest models for all the CLIs like opencode, codex, claude code, copilot.
afaik it wasn't handled

Davidson3556:
@muddlebee fair callout, but before I touch those lists I want to make sure I'm not adding stale or fabricated model IDs. I went through each CLI's model tuple in app/cli/wizard/config.py against git history:

- codex (#781, 2026-04-23): gpt-5.4 / 5.2-codex / 5.1-codex-max / 5.4-mini / 5.3-codex / 5.2 / 5.1-codex-mini. Default matches OPENAI_REASONING_MODEL=gpt-5.4.
- claude-code (#1168, 2026-05-01): claude-opus-4-7 / sonnet-4-6 / haiku-4-5-20251001. Matches ANTHROPIC_REASONING_MODEL=claude-opus-4-7.
- opencode (#1196, 2026-05-05): opus-4.7, sonnet-4.6, gpt-5.4, gemini-3.1-pro-preview, llama-4-maverick, mistral-large-2512.
- copilot (#1533, 2026-05-12): default-only ("") on purpose ,the inline comment shipped with that PR says "the Copilot CLI's accepted --model values are not stable across releases and live behind GitHub-side gating, so baking them in risks 'model not found' errors after the wizard."

All four lists were curated with the integration PR in the last 6 weeks, and the underlying model names (Opus 4.7, Sonnet 4.6, GPT-5.4) haven't rolled to newer SKUs since. Copilot is intentionally empty.

If there are specific model IDs you want surfaced (e.g. a new gpt-5.x-codex variant, an opus-4-8, a kimi-k3, etc.), could you drop them in this thread? I'd rather plug exact strings you've verified than guess names that may not exist, the Copilot precedent makes that risk concrete.

Otherwise I'd suggest the model-roster freshness is a separate concern from #782 (which is just the picker UX) and can ride in its own PR when a specific update lands.

muddlebee:
@Davidson3556 model onboarding wasn't part of the CLi implementation, those were just added as placeholders but not accurate display of models. we need to fix and update the latest lists for all CLI impementation like I mentioned earlier.

app/cli/wizard/config.py
```diff
@@ -209,25 +210,24 @@ class ProviderOption:
...
         value="",
         label="CLI default (no -m; use Codex configured model)",
     ),
-    ModelOption(value="gpt-5.4", label="gpt-5.4 — strong default for everyday coding"),
-    ModelOption(value="gpt-5.2-codex", label="gpt-5.2-codex — frontier agentic coding"),
-    ModelOption(
-        value="gpt-5.1-codex-max",
-        label="gpt-5.1-codex-max — deep / fast reasoning",
-    ),
+    ModelOption(value="gpt-5.5", label="gpt-5.5 — newest frontier coding"),
```
muddlebee, line 214: gpt  5.5 also has variants like medium, high, low reasoning. it also needs to be handled

app/cli/wizard/config.py
```diff
@@ -209,25 +210,24 @@ class ProviderOption:
     ),
     ModelOption(value="claude-opus-4-7", label="Claude Opus 4.7 — most capable"),
```
muddlebee, line 202: same need to add reasoning types like high, medium etc. need a way to handle this in onboarding, a new sub menu choice?

muddlebee:
need to explore ways how to handle reasoning efforts of models like by claude and openai like medium, high etc? a sub menu for that ? or just handle in same list

Davidson3556:
@greptile review

Davidson3556:
https://github.com/user-attachments/assets/55758956-3280-4698-b46a-6c7a573bf2c3

demo video
Screenshot 2026-05-15 at 11 01 40

@muddlebee 

**Scoping to OpenAI + Codex.** The reasoning-effort plumbing already exists in `app/llm_reasoning_effort.py` and is consumed today by the `/effort` REPL slash command. Its `provider_supports_reasoning_effort()` contract only
includes `openai` and `codex` ,those are the providers wired to actually read `OPENSRE_REASONING_EFFORT` at the model call site. The new onboarding prompt is gated on that same contract so we don't ship a knob that gets
silently dropped for other providers.

**Claude intentionally excluded.** Anthropic doesn't expose discrete reasoning levels (low/medium/high): they use extended thinking with continuous `budget_tokens`, and Claude Code CLI doesn't surface either as a model variant. Following the existing codebase contract felt safer than inventing fake tiers. Happy to file a follow-up issue if you'd like a "thinking budget" sub-menu tracked separately.

**UX.** After model select, when the provider supports it, a second select prompt fires:
  `Reasoning effort for {provider}?`
  → `Use model default` (clears `OPENSRE_REASONING_EFFORT`)
  → `low | medium | high | xhigh`
The "default" row's hint shows what skipping actually does for the chosen model , e.g. `inherits medium` for gpt-5.4 — using
`infer_reasoning_effort_default()`, so the picker isn't lying.

**Persistence.** Writes `OPENSRE_REASONING_EFFORT` to `.env` and the choice
to the wizard store. Re-runs preselect the prior pick and `/effort` in the
REPL stays in sync.

Tests: 7 new for the prompt itself, 2 for env_sync, 2 for the store, covering supported/unsupported providers, default-hint accuracy, and persistence across re-runs.

muddlebee:
@claude review this

muddlebee:
GM @Davidson3556 sorry for getting this back on late, a few review points

- not all models are latest for all CLIs like gemini, opencode still missing lot of models
- and investigate is not working after changing model, this is regression introduced during a different PR. fixing in main.

I will need to revisit and take a look again.

Davidson3556:
@greptile review

muddlebee:
here is some fix Im doing https://github.com/Tracer-Cloud/opensre/pull/2261

models are not being set properly in env after /model selection

muddlebee:
@Davidson3556 could you fix the conflicts

check the PR #2261 once for reference, some of the issues we fixed as mentioned above not related to your feature.

Davidson3556:
on it @muddlebee

Davidson3556:
@muddlebee rebased onto post-#2261 main in 6d2b10df. Single clean commit
on top of your env-sync fix.

What changed where:

- `_choose_model`: kept your `(provider, *, default)` signature and the
"Choose {provider.label} model" prompt so test_flow.py stayed green.
  Mine adds a "current" row when the saved model isn't in the curated
  list, and an "Enter custom model ID" text fallback.
- `_choose_reasoning_effort`: new step. Gated on
  `provider_supports_reasoning_effort()`, so today it only fires for
  openai and codex. Writes `OPENSRE_REASONING_EFFORT` via your
  `sync_provider_env` and saves the pick through `save_local_config`,
  so re-runs preselect it. Skipped Anthropic on purpose — their API
  doesn't expose discrete tiers, just continuous `budget_tokens`.
- Model lists: gemini-cli 2 -> 7, OpenCode Zen 11 -> 17, per your
  earlier review.
- `env_sync.py`: your overhaul untouched; `reasoning_effort` is just an
  extra kwarg.
- `conftest.py`: dropped my `_choose_model` autouse stub because your
  tests already feed the model through `select_responses`. Kept the
  `reasoning_effort` stub since that prompt isn't in the existing
  iterators.

130 wizard tests + the onboard smoke test pass. Ruff and mypy are clean
on the files I touched; the 3 mypy errors in `app/agents/probe.py` are
pre-existing on main.

Davidson3556:
@greptile review

muddlebee:
@Davidson3556 thank you, will need more testing on this :) will try to wrap up EOD

muddlebee:
@Davidson3556 have you tested that the reasoning efforts model gets passed onto the CLI properly? 

like if we set a high reasoning model, is it getting passed to the CLI vars of LLM CLI config?

app/cli/wizard/flow.py
```diff
@@ -2207,8 +2306,9 @@ def run_wizard(_argv: list[str] | None = None) -> int:
         api_key_env=provider.api_key_env,
         model_env=provider.model_env,
         probes=probes,
+        reasoning_effort=reasoning_effort,
     )
-    env_path = sync_provider_env(provider=provider, model=model)
+    env_path = sync_provider_env(provider=provider, model=model, reasoning_effort=reasoning_effort)
```
muddlebee: UX mismatch: onboarding writes OPENSRE_REASONING_EFFORT to .env here, but /effort and /status only show session.reasoning_effort (starts None), so users see "(default)" even when high is active via env. Runtime calls still pick it up via get_active_reasoning_effort(); consider hydrating session from env on REPL start or showing the env-backed value in /effort output.

app/cli/wizard/flow.py
```diff
@@ -324,27 +330,109 @@ def _questionary_choice(choice: Choice) -> questionary.Choice:
...
+) -> str | None:
+    """Prompt the user to pick a reasoning-effort tier for supported providers.
+
+    Returns the chosen effort (``low|medium|high|xhigh``), or ``None`` to mean
+    "use the model's native default" — written to ``.env`` as the absence of
+    ``OPENSRE_REASONING_EFFORT``. Gated on
+    :func:`provider_supports_reasoning_effort`: today, only ``openai`` and
+    ``codex`` actually consume the override at the model-call site.
+    """
+    if not provider_supports_reasoning_effort(provider.value):
```
muddlebee: Please audit reasoning-effort compatibility across both runtime paths before treating onboarding as complete — this step writes OPENSRE_REASONING_EFFORT globally, but consumption is provider-specific:

**API (llm_client):** openai wired (reasoning_effort kwarg on gpt-5/o-series). anthropic and other HTTP providers ignore OPENSRE_REASONING_EFFORT today.

**CLI (llm_cli adapters):** codex wired (-c model_reasoning_effort=...). claude-code CLI supports --effort low|medium|high|xhigh|max but ClaudeCodeAdapter currently no-ops reasoning_effort; gemini/opencode/cursor/kimi/copilot also no-op.

Suggest a short compatibility matrix in the PR (provider × API/CLI × mechanism × wired?) and either wire claude-code --effort or document why it stays excluded. Also note max→xhigh runtime mapping vs Claude accepting max natively — tiers may not translate 1:1 across vendors.

muddlebee:
@Davidson3556 we need to see if adding reasoning efforts here worth it, its adding an extra layer of complexity and compatibility :thinking: which we need to see through

Davidson3556:
@muddlebee good call. Pulled the reasoning-effort step out in c22c5025;
this PR is back to just the model picker, which was the original #782
ask.

Your three points were the reason. No e2e test asserts that
OPENSRE_REASONING_EFFORT actually reaches the codex subprocess kwarg.
There's also a real bug in the REPL: onboarding writes the env var but
session.reasoning_effort starts None, so /effort and /status keep
showing "(default)" even when high is active. And the compat story is
genuinely patchy — only openai (reasoning_effort kwarg) and codex CLI
(-c model_reasoning_effort=...) consume the override today; claude-code
CLI takes --effort low|medium|high|xhigh|max but ClaudeCodeAdapter
no-ops it, and gemini/opencode/cursor/kimi/copilot all no-op too.
max→xhigh runtime mapping doesn't translate 1:1 to Claude's native max
either.

I'll do the cleanup as a follow-up PR — hydrate session from env on
REPL boot, wire ClaudeCodeAdapter --effort (and document why the others
stay no-op), drop a provider × API/CLI compat matrix in the description,
and add an e2e test asserting the .env value reaches the subprocess.
Filing the tracking issue today, will link here.

Left in this PR: `_choose_model` with the custom-ID escape hatch and
saved-model preservation, plus the CLI model list refresh you flagged
(claude-code, codex, gemini-cli, opencode) — pulled straight from the
upstream docs.

125 wizard tests + the onboard smoke test pass locally. ruff and mypy
clean on touched files; the 3 `app/agents/probe.py` mypy errors are
pre-existing on main. config.py rebased on dcb9d8be.

Davidson3556:
@muddlebee the red CI check isn't from this PR's changes.

Two issues, both pre-existing on main, both blocked by fork-PR
secret access:

1. The 23 `test_live_llm_planner_matches_prompt_contract[*]` failures
   need `ANTHROPIC_API_KEY` to make real LLM calls. GitHub doesn't pass
   secrets to workflows triggered by fork PRs (security policy), so
   these fail for every contributor PR. Confirmed they also fail when
   I run them against main HEAD locally with no key set.

2. `test_sync_env_values_routes_secrets_to_keyring` (added in #2256)
   needs a working system keyring in the Linux CI runner. Fails on main
   too — not introduced here.

Locally on this branch against fresh main:
- 125/125 wizard tests pass
- 33/33 env_sync tests pass
- onboard smoke test passes
- ruff lint+format clean
- mypy clean on touched files (the 3 probe.py errors are pre-existing
  on main)

Happy to file a separate small PR adding
`pytest.mark.skipif(not os.getenv("ANTHROPIC_API_KEY"), ...)` to the
live_llm tests if that would help unblock fork PRs going forward. But
that's a separate concern from this model-picker change.

muddlebee:
@Davidson3556 thank you, looks good. Will test on my end once.

muddlebee:
@Davidson3556 I hope this is well tested

PR #2090, opened 2026-05-16
fix: run LLM judge when --evaluate flag is passed

### LINKED ISSUE #2070: [BUG] --evaluate prepares rubric but never runs OpenSRE LLM judge
### Summary

`opensre investigate --evaluate` still extracts `scoring_points`, but the custom pipeline does not call `run_opensre_llm_judge`.

**Expected:** `--evaluate` should run the OpenSRE LLM judge after the final diagnosis and populate `opensre_llm_eval`, or the flag should be removed/marked unsupported.

**Actual:** Evaluate mode prepares `opensre_eval_rubric`, but `run_opensre_llm_judge` has no app call site, so the judge result is never populated by the pipeline.

### Steps to reproduce

- Run `uv run pytest tests/test_state.py::test_make_initial_state_evaluate_strips_scoring_points -q` to confirm evaluate mode still prepares `opensre_eval_rubric`.
- Search app code for `run_opensre_llm_judge`.
- Observe the only app match is the function definition in `app/integrations/opensre/llm_eval_judge.py`; the custom pipeline does not call it.
- `tests/benchmarks/openrca_scenarios/README.md` documents that `--evaluate` should run the judge at the end.

Yes

### How often does it occur?

Every time

### Operating system

Windows

### Logs and error output

_No response_

### Additional context

_No response_

pratham-commits:
Fixes #2070

Fixed the `--evaluate` flag so it actually runs the LLM judge. The judge function existed in the codebase but was never called. Added the call in `app/delivery/__init__.py` to run `run_opensre_llm_judge()` after investigation completes when `--evaluate` is passed.

BEFORE (bug - judge not running):
```
"opensre_llm_eval": {
  "skipped": true,
  "reason": "Evaluate was enabled but no judge output was recorded."
}
```
AFTER (fixed - judge runs and evaluates):
```
"opensre_llm_eval": {
  "overall_pass": false,
  "score_0_100": 0,
  "rubric_items": 
    {"id": "shippingservice-1", "satisfied": false, "explanation": "The agent concluded that the root cause category is configuration, which does not match the predicted root cause component in the rubric, which is shippingservice-1."},
    {"id": "container_read_io_load", "satisfied": false, "explanation": "The agent did not identify any evidence related to container read I/O load, which is the predicted root cause reason in the rubric."}
  ,
  "summary": "The investigation did not match the rubric, as the agent concluded a different root cause category and reason."
}
```
---
- [x] Yes, I used AI assistance

The fix adds a call to `run_opensre_llm_judge()` in `app/delivery/__init__.py`. When `opensre_evaluate=True` and `opensre_eval_rubric` exists (extracted from the alert's scoring_points), the judge runs after the investigation completes and stores the evaluation result in `state["opensre_llm_eval"]`. If the judge fails, it falls back to a skip message with an error reason.

---
- [x] I have added proper PR title and linked to the issue (#2070)
- [x] Tests pass: pytest tests/test_state.py -v

you can run  to verify it too:

`opensre investigate -i "json-path" --evaluate`

### DIFF
--- a/app/delivery/__init__.py
+++ b/app/delivery/__init__.py
@@ -2,10 +2,11 @@
 
 from __future__ import annotations
 
+import logging
 from typing import Any
 
 from app.state import InvestigationState
-import logging
+
 logger = logging.getLogger(__name__)
 
 
--- a/app/delivery/__init__.py
+++ b/app/delivery/__init__.py
@@ -17,4 +17,21 @@ def deliver(state: InvestigationState) -> dict[str, Any]:
     """
     from app.delivery.publish_findings.node import generate_report
 
+    state_dict = dict(state)
+
+    if state_dict.get("opensre_evaluate") and state_dict.get("opensre_eval_rubric"):
+        from app.integrations.opensre.llm_eval_judge import run_opensre_llm_judge
+
+        try:
+            judge_result = run_opensre_llm_judge(
+                state=state_dict,
+                rubric=state_dict["opensre_eval_rubric"],
+            )
+            state["opensre_llm_eval"] = judge_result
+        except Exception:
+            state["opensre_llm_eval"] = {
+                "skipped": True,
+                "reason": "Judge run failed - check logs for details.",
+            }
+
     return generate_report(state)
--- a/app/delivery/__init__.py
+++ b/app/delivery/__init__.py
@@ -5,6 +5,8 @@ from __future__ import annotations
 from typing import Any
 
 from app.state import InvestigationState
+import logging
+logger = logging.getLogger(__name__)
 
 
 def deliver(state: InvestigationState) -> dict[str, Any]:
@@ -28,10 +30,11 @@ def deliver(state: InvestigationState) -> dict[str, Any]:
                 rubric=state_dict["opensre_eval_rubric"],
             )
             state["opensre_llm_eval"] = judge_result
-        except Exception:
+        except Exception as exc:
+            logger.exception("LLM judge failed: %s", exc)
             state["opensre_llm_eval"] = {
                 "skipped": True,
-                "reason": "Judge run failed - check logs for details.",
+                "reason": f"Judge run failed: {exc}",
             }
 
     return generate_report(state)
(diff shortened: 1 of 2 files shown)


cerencamkiran:
I do agree with Greptile on the exception path though. A silent bare "except Exception:" around an LLM/auth/network boundary makes operational failures very hard to diagnose, especially when the fallback message says “check logs for details”.

I’d strongly recommend logging the actual exception ("logger.exception(...)") and surfacing "exc" in the fallback payload before merge.

One additional thing that would make this more robust long-term is a focused regression test covering the "opensre_evaluate + opensre_eval_rubric" execution path, so future pipeline refactors don’t accidentally break judge invocation again.

pratham-commits:
hey @cerencamkiran , I have updated the new changes. 

1. Added the logging using the actual exception ("logger.exception(...)") and surfacing "exc" in the fallback payload.
2. Added the regression tests too,

- `test_deliver_runs_judge_when_evaluate_and_rubric_present` -  Judge is called and result stored in state when --evaluate is enabled with a rubric
- `test_deliver_skips_judge_when_evaluate_is_false` - Judge is not called when --evaluate is not set
- ` test_deliver_sets_skip_on_judge_failure` - Judge exception is caught and a skip message with error detail is stored in state

cerencamkiran:
@greptile review

cerencamkiran:
Thanks for addressing the feedback @pratham-commits 

One small follow-up: the "evaluate=False" regression test could isolate the guard a bit more by still populating "opensre_eval_rubric" in state while keeping only "opensre_evaluate=False".

Right now both conditions become false together, so the test could still pass even if the rubric guard was accidentally removed later.

pratham-commits:
hey @cerencamkiran 

I have made the change, so now :-
After creating the state, we manually set state["opensre_eval_rubric"] = "test rubric". Now only opensre_evaluate is False, so if someone accidentally changes and to or in the future, the test catches it immediately.

cerencamkiran:
Just small lint fixes now 

- move/sort the "import logging"
- replace unused "lambda *a, **kw" args in the monkeypatch test

pratham-commits:
Hey @cerencamkiran , 
I have made the changes as needed.

cerencamkiran:
@greptile review

cerencamkiran:
Thanks for addressing all the feedback so quickly @pratham-commits. Other maintainers can probably take a look too when they have time.

Davidson3556:
Fix works correctly. this is okay

PR #2097, opened 2026-05-16
fix(streaming): run upstream correlation in streaming investigations

### LINKED ISSUE #2077: [BUG] Streaming investigation path skips upstream correlation
### Summary

`astream_investigation` does not call node_correlate_upstream, while the non-streaming investigation pipeline does.

**Expected:** Streaming and non-streaming investigations should run the same pipeline steps, including upstream correlation.

**Actual:** `run_connected_investigation` calls `node_correlate_upstream`, but `astream_investigation` does not. Streaming results can miss `correlation` data that non-streaming results include.

### Steps to reproduce

- Inspect `app/pipeline/pipeline.py`.
- Observe `run_connected_investigation` calls `node_correlate_upstream` after `ConnectedInvestigationAgent().run`.
- Inspect `app/pipeline/runners.py`.
- Observe `astream_investigation` runs resolve, extract, agent, and publish steps, but does not call `node_correlate_upstream`.
- Inspect `app/remote/server.py`.
- Observe `/investigate/stream` uses `astream_investigation`, while blocking `/investigate` uses the non-streaming CLI path.

Yes

### How often does it occur?

Every time

### Operating system

Windows

### Logs and error output

_No response_

### Additional context

_No response_

cerencamkiran:
## Summary

Fixes #2077

Fix streaming/non-streaming investigation drift where upstream correlation
was skipped in `astream_investigation`.

Previously the blocking pipeline (`run_connected_investigation`) executed
`node_correlate_upstream`, while the streaming path manually reimplemented
the investigation flow and omitted the correlation step entirely. As a result,
`/investigate/stream` responses could miss correlation payloads that were
present in blocking investigations.

## Changes

* run `node_correlate_upstream` during streaming investigations
* align streaming investigation behavior with the blocking pipeline
* preserve correlation payloads in streamed investigation state/results
* add regression coverage for streamed correlation persistence

## Validation

```bash
ruff check app/pipeline/runners.py tests/remote/test_server.py

python -m pytest tests/remote/test_server.py -q
```

Verified that streamed investigations now include:

* `correlated_signals`
* `most_likely_causal_drivers`

the same way as blocking investigations.

Ekran görüntüsü 2026-05-16 160604
Ekran görüntüsü 2026-05-16 160623

### DIFF
--- a/app/pipeline/runners.py
+++ b/app/pipeline/runners.py
@@ -220,6 +220,14 @@ async def astream_investigation(
             from app.correlation.node import node_correlate_upstream
             from app.pipeline.pipeline import _build_correlation_config
 
+            _put(
+                _make_node_event(
+                    "on_chain_start",
+                    "correlate_upstream",
+                    {},
+                )
+            )
+
             _merge(
                 state_any,
                 node_correlate_upstream(
--- a/app/pipeline/runners.py
+++ b/app/pipeline/runners.py
@@ -228,6 +228,18 @@ async def astream_investigation(
                 ),
             )
 
+            _put(
+                _make_node_event(
+                    "on_chain_end",
+                    "correlate_upstream",
+                    {
+                        "output": {
+                            "correlation": state_any.get("correlation", {}),
+                        }
+                    },
+                )
+            )
+
             # --- deliver / publish (skip terminal render — StreamRenderer owns it) ---
             _put(_make_node_event("on_chain_start", "publish_findings", {}))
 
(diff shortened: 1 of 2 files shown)


cerencamkiran:
@greptile review

cerencamkiran:
I kept this PR focused on fixing the immediate drift by adding the missing `correlate_upstream` step + stream events.

The bigger cleanup is to stop duplicating the blocking pipeline inside `astream_investigation`. I can open a follow-up to extract shared investigation steps so streaming only adds event emission/rendering on top.

muddlebee:
@cerencamkiran merging this for now, pls feel free to open up follow issues if needed :)

PR #2137, opened 2026-05-17
fix(hermes): prevent PatternRule/RepeatRule false positives on logger…

### LINKED ISSUE #1874: Hermes: PatternRule/RepeatRule may false-positive when matching record.raw (logger names)
## Summary

`PatternRule` and `RepeatRule` in `app/hermes/rules.py` match against both `record.message` and `record.raw`. Because `record.raw` typically includes the full log line (timestamp, level, **logger name**, and message), rule patterns can accidentally match substrings that appear only in the logger prefix (e.g. a logger name containing a keyword used in a crash-loop or OOM pattern), producing **false-positive incidents** unrelated to the actual log message content.

## Context

Flagged during review of https://github.com/Tracer-Cloud/opensre/pull/1857 (Greptile noted this as a minor residual risk, not a blocker for the Hermes landing).

## Suggested direction (for whoever picks this up)

- Prefer matching primarily on `record.message`, or define a dedicated “search haystack” that excludes volatile / structural prefixes (logger, timestamp) if raw-line context is still needed for some rules.
- Add regression tests where `record.message` is benign but `record.raw` contains a rule keyword only inside the logger segment.

## Acceptance criteria

- Documented behavior for which fields rules scan.
- Tests covering logger-name vs message boundary false positives (or explicit decision to accept raw matching with rationale).

AniketR10:
… names

Fixes #1874 

 - Updated PatternRule and RepeatRule in app/hermes/rules.py to exclusively scan record.message instead of falling back to record.raw.
 - Added regression tests that verify rule keywords in the logger prefix do not trigger false-positive incidents. Tests explicitly assert that the raw line matches the regex, but the evaluate() method correctly ignores it.
 - 

https://github.com/user-attachments/assets/532d8539-f930-4892-95bc-e3ef3af816eb

---

Dropped the or pattern.search(record.raw) fallback in both PatternRule and RepeatRule so patterns only scan record.message, preventing keywords in logger-name prefixes from firing false-positive incidents. Updated docstrings to document the scope and added two regression tests that construct records where record.raw matches a pattern but the message is benign confirmed they fail on the old code and pass on the fix.

---

---

### DIFF
--- a/app/hermes/rules.py
+++ b/app/hermes/rules.py
@@ -44,8 +44,11 @@ from app.hermes.incident import HermesIncident, IncidentSeverity, LogLevel, LogR
 class PatternRule:
     """Rule that emits one incident per matching log record.
 
-    Patterns are matched case-insensitively against the record's
-    message. Any pattern matching causes the rule to fire.
+    Patterns are matched case-insensitively against ``record.message``
+    only. ``record.raw`` (which includes the timestamp/level/logger
+    prefix) is intentionally excluded so a keyword appearing in a
+    logger name — e.g. ``oom_monitor`` — cannot fire an OOM rule on
+    an otherwise benign message.
     """
 
     name: str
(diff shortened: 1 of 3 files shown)


AniketR10:
@greptile review

PR #2156, opened 2026-05-17
fix(agent-llm): route CLI LLM providers for investigations

muddlebee:
## Summary

- After the LangGraph migration, `get_agent_llm()` only handled HTTP providers; `LLM_PROVIDER=codex` (and other CLI providers) fell through to Anthropic, so `opensre investigate` could hit the wrong backend and fail with billing errors.
- This PR adds `CLIBackedAgentClient`, which implements the investigation agent contract by flattening history, embedding tool schemas in the prompt, and parsing `{"tool_calls": [...]}` JSON (or treating output as a final text answer).
- `get_agent_llm()` now uses the same `llm_cli` registry as `llm_client.py` for registered CLI providers.

No linked issue (regression surfaced locally with Codex + investigate).

## Test plan

- [x] `uv run pytest tests/services/test_agent_llm_client.py -q`
- [x] `uv run ruff check` on touched files
- [x] `uv run mypy` on touched files
- [ ] Optional: `uv run opensre investigate --input tests/synthetic/eks/000-healthy/alert.json` with `LLM_PROVIDER=codex` and a working Codex session

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [x] I have reviewed every single line of the AI-generated code
- [x] I can explain the purpose and logic of each function/component I added
- [x] I have tested edge cases and understand how the code handles them
- [x] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**

The investigation ReAct loop requires `tool_schemas`, `invoke(messages, system, tools)`, and message builders compatible with `app/agent/investigation.py`. CLI binaries have no native tool API, so the client sends one flattened prompt per turn with instructions plus JSON tool definitions, then parses structured `tool_calls` from the model reply when present; otherwise it treats the reply as plain-text completion. `CLIBackedLLMClient` is reused for subprocess execution and auth probing.

---

## Checklist before requesting a review
- [x] I have added proper PR title and linked to the issue
- [x] I have performed a self-review of my code
- [x] **I can explain the purpose of every function, class, and logic block I added**
- [x] I understand why my changes work and have tested them thoroughly
- [x] I have considered potential edge cases and how my code handles them
- [x] If it is a core feature, I have added thorough tests
- [x] My code follows the project's style guidelines and conventions

PR #2157, opened 2026-05-17
fix(agent-llm): harden CLI tool-call JSON parsing for preamble output

muddlebee:
## Summary
Follow-up hardening for the merged CLI provider routing work in #2156.

This PR strengthens CLI tool-call parsing so we don't silently miss valid tool calls when a CLI model adds unfenced prose before the JSON payload.

## Root cause
`_try_parse_tool_call_json()` only attempted `JSONDecoder.raw_decode()` from index `0` (or fenced content). If the response started with prose and the tool-call object appeared later, parsing failed and the turn was treated as a plain-text final answer.

## What changed
- Updated `_try_parse_tool_call_json()` to:
  - keep the fast path for pure JSON / fenced JSON
  - add a recovery path that scans object starts (`{`) and attempts `raw_decode(..., idx=...)`
  - return the first decoded dict containing `tool_calls`
- Hardened `CLIBackedAgentClient.invoke()`:
  - only accepts `tool_calls` when it is a list of dict calls with a non-empty `name`
  - normalizes non-dict `input` payloads to `{}`
  - falls back to the original text response when parsed JSON contains no usable tool calls (prevents blank final answers)

## Tests
Added/updated service tests:
- unfenced preamble + braces before JSON is recoverable
- malformed `tool_calls` payload (`{"tool_calls":"not-a-list"}`) falls back to plain text instead of erasing content

## Validation
- `make lint`
- `make format-check`
- `make typecheck`
- `uv run pytest tests/services/ tests/tools/ -v`

### DIFF
--- a/app/services/agent_llm_client.py
+++ b/app/services/agent_llm_client.py
@@ -420,18 +420,24 @@ class CLIBackedAgentClient:
         tool_calls: list[ToolCall] = []
         parsed_json = _try_parse_tool_call_json(text)
         if parsed_json is not None:
-            raw_calls = parsed_json.get("tool_calls") or []
-            for i, tc in enumerate(raw_calls):
-                if not isinstance(tc, dict):
-                    continue
-                tool_calls.append(
-                    ToolCall(
-                        id=str(tc.get("id") or f"call_{i}"),
-                        name=str(tc.get("name") or ""),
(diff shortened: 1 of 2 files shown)


muddlebee:
@greptile review

muddlebee:
@greptile review

PR #1986, opened 2026-05-14
fix(integrations): eliminate silent env-loader fallbacks in _catalog_impl (#1468)

0xghost42:
Closes #1468.

## What

Forty-plus exception sites in `app/integrations/_catalog_impl.py` swallowed failures into one of three silent sinks. All routes now report to Sentry via `app.utils.errors.report_exception` (severity `warning`, since these are user-config misses, not service bugs).

The caller contract is unchanged — pattern A still returns `(None, None)` and pattern B/C still skip the integration. The only behavior delta is that the exception now surfaces.

## Patterns covered

**Pattern A — `_classify_service_instance`** (33 sites): `except Exception: return None, None`.
Now routes through `_report_classify_failure(exc, integration=<vendor>, record_id=record_id)`.

**Pattern B — `load_env_integrations` argocd + helm**: `except Exception: pass`.
Now routes through `_report_env_loader_failure(exc, integration=<vendor>)`.

**Pattern C — `load_env_integrations` debug-only env loaders**: `logger.debug(\"Failed to load <X> config from env\", exc_info=True)` (incident_io, openclaw, mariadb, rabbitmq, rds, betterstack, alertmanager, victoria_logs, supabase). Same helper as B.

## Sentry tag set

Every captured exception carries:

| tag          | value                                |
|--------------|--------------------------------------|
| `surface`    | `integration`                        |
| `component`  | `app.integrations._catalog_impl`     |
| `integration`| `<vendor>` (grafana, datadog, ...)   |
| `event`      | `classify_failed` or `env_loader_failed` |

Pattern A also threads `record_id` through `extras` for traceability back to the offending store record.

Severity is `warning` — these are user misconfigurations, not service bugs, so sampling/quota policy can treat them differently from real errors. #1454 wired `report_exception`, so this PR just consumes it.

## Tests

`tests/integrations/test_catalog_silent_fallback_elimination.py` (44 cases, ~1s):

- 2 helper smoke tests asserting the tag dict shape.
- 31 parametrized cases over `_classify_service_instance` — force-patch each vendor's config constructor (`GrafanaIntegrationConfig`, `build_supabase_config`, etc.) to raise, then assert `(None, None)` is still returned *and* `report_exception` is called once with the right vendor tag.
- 11 parametrized cases over `load_env_integrations` (Pattern B + C) — set the trigger env vars, force the builder symbol to raise, assert the integration is absent from the result *and* the failure was reported.

[...]

PR #2174, opened 2026-05-18
feat(error-handling): add Sentry capture to integration verification adapters

VibhorGautam:
Fixes #1469

#### Describe the changes you have made in this PR -

Wires structured Sentry capture into every except block in `_verification_adapters.py`. Previously, verification probe failures were silently collapsed into string results with no logging or Sentry visibility.

Added a `_report_probe_failure()` helper that routes through `report_exception()` (same pattern as `capture_service_error` from #1922) with structured tags:

- **Vendor transport errors** (httpx, requests, botocore base classes) get `severity=info` and `event=vendor_failure` since these are expected network conditions during probing
- **Unknown exceptions** get `severity=error` and `event=verify_failed` to flag actual bugs in verification logic
- All events carry `surface=integration_probe` and the integration name for Sentry filtering

14 except sites instrumented across: `_verify_with_validation_result`, `build_probe_verifier`, grafana, aws, slack (2 sites), tracer (2 sites), discord (LoginFailure + general), telegram, whatsapp.

Discord ImportError also sets `sentry_sdk.set_tag("integrations.discord.import_failed", "true")` per issue spec.

### Demo/Screenshot for feature changes and bug fixes - 

https://github.com/user-attachments/assets/f90251b5-0455-4caa-94ea-1fa5df631b0d

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [x] I have reviewed every single line of the AI-generated code
- [x] I can explain the purpose and logic of each function/component I added
- [x] I have tested edge cases and understand how the code handles them
- [x] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**

The core problem is that verification probes swallow exceptions into plain string results, making failures invisible in Sentry. I considered two approaches: (a) inline `report_exception` calls at each site, or (b) a thin helper that encapsulates the vendor-vs-bug severity decision. Went with (b) because the severity split logic would be duplicated 14 times otherwise.

`_report_probe_failure` checks `isinstance(exc, _VENDOR_TRANSPORT_ERRORS)` to decide severity. The vendor tuple uses base classes (`httpx.TransportError`, `requests.exceptions.RequestException`, botocore bases) rather than listing every subclass, so new transport errors like SSLError or ProxyError are automatically covered.

[...]

PR #2191, opened 2026-05-18
fix(integrations): eliminate silent env-loader fallbacks + scrub pydantic input_value

### LINKED ISSUE #1468: [error-handling] Eliminate silent env-loader fallbacks in app/integrations/_catalog_impl.py
## Problem

`app/integrations/_catalog_impl.py` is the most concentrated source of silent error handling in the codebase. The audit found:

### Pattern A — `_classify_service_instance` returns `(None, None)` on any exception (37+ occurrences)

Every classifier follows this shape:

```python
        except Exception:
            return None, None
```

Confirmed line numbers (before line 764): **164, 196, 209, 225, 242, 261, 276, 290, 304, 322, 340, 364, 387, 402, 419, 435, 449, 467, 485, 512, 535, 554, 573, 598, 615, 637, 654**, plus **758** (Splunk block).

### Pattern B — `pass` in env validation (~L1099–1112)

Argo CD validation:

```python
        except Exception:
            pass
        else:
            integrations.append(...)
```

### Pattern C — `logger.debug` only (~L1225–1327, 1488–1507)

OpenClaw / MariaDB / RabbitMQ / RDS / Betterstack / Alertmanager / VictoriaLogs env loaders:

```python
        except Exception:
            logger.debug("Failed to load <X> config from env", exc_info=True)
```

Net effect: a malformed `MONGODB_*` env var, a broken `OPENCLAW_*` config, etc. silently disappears from the resolved integration set. Users see "integration not detected" with no debugging breadcrumb anywhere.

## Proposal

1. **Pattern A**: replace `return None, None` with `report_exception(..., severity="warning", tags={"surface": "integration", "event": "classify_failed", "integration": <vendor>})` then `return None, None`. The `warning` severity reflects that "this credential record is malformed" is a user-config issue, not our bug — but Sentry sampling lets us see the rate.
2. **Pattern B**: replace bare `pass` with `report_exception(..., severity="warning")`. Skipping the integration is fine; doing so silently is not.
3. **Pattern C**: replace `logger.debug` with `report_exception(..., severity="warning", tags={"event": "env_loader_failed", "integration": <vendor>})`. Debug-only is hidden in production logs.
4. **DRY** the 30+ classifiers into a helper:

```python
def _safe_classify(
    classifier: Callable[..., tuple[T, R]],
    *,
    integration: str,
    record: dict[str, Any],
) -> tuple[T | None, R | None]:
    try:
        return classifier(record)
    except Exception as exc:
        report_exception(exc, logger=logger,
                         message=f"[catalog] failed to classify {integration} record",
                         severity="warning",
                         tags={"surface": "integration", "event": "classify_failed",
                               "integration": integration})
        return None, None
```

5. **Tests**: parametrize over each integration, inject a malformed record, assert `(None, None)` *and* exactly one Sentry capture with the right tags.

## Acceptance criteria

- [ ] All 37+ Pattern-A `except Exception: return None, None` sites are migrated to a shared helper that reports.
- [ ] Pattern B and C sites no longer rely on `pass` or `logger.debug` for silent failure.
- [ ] Parametrized tests prove each classifier reports exactly once on a malformed record.
- [ ] No regression in the existing catalog/integration tests.

## Depends on

#1454

nuemaan:
Fixes #1468

`app/integrations/_catalog_impl.py` was swallowing failures in three places: 31 `except Exception: return None, None` sites in `_classify_service_instance`, two `except Exception: pass` sites and nine `logger.debug(..., exc_info=True)` debug-only sites in `load_env_integrations`, plus six unguarded `model_validate` calls that aborted the entire discovery loop on any validation error. All of those now route through two helpers (`_report_classify_failure`, `_report_env_loader_failure`) backed by `report_exception` with `severity=warning`. The historic caller contract (skip the integration) is preserved.

`app/utils/sentry_sdk.py` was missing a scrub pass over `exception.values[].value`. Pydantic V2 renders `input_value=<repr>` in its `ValidationError` message, so any failing config field with a secret value (`api_token`, `password`, …) would have leaked the raw value to Sentry once the env-loader catches above started routing through `report_exception`. Added `_scrub_exception_value` and wired it into `_scrub_event_in_place`.

```
$ uv run pytest tests/integrations/test_catalog_silent_fallback_elimination.py tests/test_sentry_init.py -q
............................................................                  [ 50%]
............................................................                  [100%]
120 passed in 1.42s
```

Scrub demo — before/after for a pydantic ValidationError carrying a secret:

```
# input (as it would land in event['exception']['values'][0]['value'])
1 validation error for VercelConfig
api_token
  String should match pattern '...' [type=string_pattern_mismatch, input_value='sk-real-secret-token', input_type=str]

# after _scrub_exception_value
1 validation error for VercelConfig
api_token
  String should match pattern '...' [type=string_pattern_mismatch, input_value=[Filtered], input_type=str]
```

---

The first commit is `0xghost42`'s closed PR #1986 cherry-picked verbatim so their authorship is preserved on the mechanical Pattern A/B/C work. I read the diff line-by-line before carrying it forward to confirm the caller contracts (`(None, None)` for classify, "skip" for env-loader) are unchanged. The two new helpers wrap `report_exception` with structured `surface`/`component`/`integration`/`event` tags so dashboards can group by vendor and by event class.

The second commit extends the same Pattern C shape to the six vendors `load_env_integrations` was still calling `model_validate` on inline. I considered factoring the boilerplate into a generic `_safe_env_record` helper but the per-vendor blocks differ enough in their env-var shapes and post-validate field checks that a generic helper would just add a layer without removing repetition. The "loop did not abort" regression test enables vercel + incident_io (chosen because incident_io is loaded *after* vercel — a survivor loaded earlier would have passed the assertion even when the loop aborts on vercel, which is the bug we're guarding against).

The scrub commit covers the exception-message string explicitly because the existing key-based scrubbers (`_scrub_extra`, stacktrace `vars`) never walk it. The regex `r"input(?:_value)?=.*?(?=,\s*input_type=|\Z)"` strips up to the next pydantic field separator. I considered the simpler `r"input(?:_value)?=[^,\]]+"` but it would over-match on values containing `,` (e.g. dict/tuple reprs). The `|\Z` end-of-string fallback is for truncated or custom-rendered messages where `, input_type=` isn't present. The placeholder `input_value=[Filtered]` is idempotent under re-application because the lookahead does not match the trailing `]` of the replacement.

---

### DIFF
--- a/app/integrations/_catalog_impl.py
+++ b/app/integrations/_catalog_impl.py
@@ -62,8 +62,54 @@ from app.integrations.supabase import build_supabase_config
 from app.llm_credentials import resolve_env_credential
 from app.services.vercel import VercelConfig
 from app.utils.coercion import safe_int
+from app.utils.errors import report_exception
 
 logger = logging.getLogger(__name__)
+
+
+def _report_classify_failure(exc: BaseException, *, integration: str, record_id: str) -> None:
+    """Route a per-instance classify failure to Sentry + warning log.
+
+    Replaces the historic ``except Exception: return None, None`` pattern in
+    ``_classify_service_instance``: the caller still gets ``(None, None)``
+    and skips the integration, but the failure is now visible to operators
+    instead of being silently swallowed (#1468).
+    """
+    report_exception(
+        exc,
+        logger=logger,
+        message=f"classify_failed: integration={integration} record_id={record_id}",
+        severity="warning",
+        tags={
+            "surface": "integration",
+            "component": "app.integrations._catalog_impl",
+            "integration": integration,
+            "event": "classify_failed",
+        },
+        extras={"record_id": record_id},
+    )
+
+
+def _report_env_loader_failure(exc: BaseException, *, integration: str) -> None:
+    """Route a per-vendor env-loader failure to Sentry + warning log.
+
+    Replaces ``except Exception: pass`` and ``logger.debug(..., exc_info=True)``
+    paths in ``load_env_integrations``: integration is still skipped, but the
+    misconfiguration reaches Sentry rather than being lost to debug output
+    (#1468).
+    """
+    report_exception(
+        exc,
+        logger=logger,
+        message=f"env_loader_failed: integration={integration}",
+        severity="warning",
+        tags={
+            "surface": "integration",
+            "component": "app.integrations._catalog_impl",
+            "integration": integration,
+            "event": "env_loader_failed",
+        },
+    )
 
 
 def _should_publish_instance_siblings(instances: object) -> bool:
@@ -167,7 +213,8 @@ def _classify_service_instance(
                     "integration_id": record_id,
                 }
             )
-        except Exception:
+        except Exception as exc:
+            _report_classify_failure(exc, integration=key, record_id=record_id)
             return None, None
         if not grafana_config.endpoint:
             return None, None
@@ -199,7 +246,8 @@ def _classify_service_instance(
                 AWSIntegrationConfig.model_validate(raw_config).model_dump(exclude_none=True),
                 "aws",
             )
-        except Exception:
+        except Exception as exc:
+            _report_classify_failure(exc, integration=key, record_id=record_id)
             return None, None
 
     if key == "datadog":
@@ -212,7 +260,8 @@ def _classify_service_instance(
                     "integration_id": record_id,
                 }
             )
-        except Exception:
+        except Exception as exc:
+            _report_classify_failure(exc, integration=key, record_id=record_id)
             return None, None
         if datadog_config.api_key and datadog_config.app_key:
             return datadog_config.model_dump(), "datadog"
@@ -228,7 +277,8 @@ def _classify_service_instance(
                     "integration_id": record_id,
                 }
             )
-        except Exception:
+        except Exception as exc:
+            _report_classify_failure(exc, integration=key, record_id=record_id)
             return None, None
         if honeycomb_config.api_key:
             return honeycomb_config.model_dump(), "honeycomb"
@@ -245,7 +295,8 @@ def _classify_service_instance(
                     "integration_id": record_id,
                 }
             )
-        except Exception:
+        except Exception as exc:
+            _report_classify_failure(exc, integration=key, record_id=record_id)
             return None, None
         if coralogix_config.api_key:
             return coralogix_config.model_dump(), "coralogix"
@@ -264,7 +315,8 @@ def _classify_service_instance(
                     "integration_id": record_id,
                 }
             )
-        except Exception:
+        except Exception as exc:
+            _report_classify_failure(exc, integration=key, record_id=record_id)
             return None, None
         return github_config.model_dump(), "github"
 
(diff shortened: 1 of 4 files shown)


nuemaan:
@greptile review

nuemaan:
Pushed a follow-up addressing the regex concern. You were right on the idempotence — after the first pass the placeholder `input_value=[Filtered]` followed by `, input_type=str]` re-matched on a second run (lazy `.*?` halted at the inner `]`), adding an extra `]` each time.

Fix: drop the `\]` alternative in the lookahead. Pydantic V2 always pairs `input_value=` with `, input_type=` (verified against real `ValidationError` output, including list-flattened cases which emit per-element scalar entries like `val.0`), so the `\]` fallback was unused defensive code that introduced the bug.

Locked the invariant with two explicit tests:
- `test_before_send_scrubs_list_typed_input_value_without_mid_clip` — pydantic list-flattening behaviour
- `test_scrub_exception_value_is_idempotent` — double-scrub is a no-op

@greptile review

VibhorGautam (commented):
clean split. 3 commits makes this easy to review, and preserving 0xghost42's authorship on the cherry-pick is the right call

i spent most of my time on the pydantic input_value scrubbing in sentry_sdk.py. covering exception.values[].value matters because that's the rendered exception string, outside the reach of the existing key-based scrubbers. the idempotency fix in 03ff3ac locks it down correctly. this is adjacent to what we're doing in #2174 for the verification adapters

one test issue still looks valid: the loop-abort regression asserts argocd survives when vercel fails, but argocd is appended before vercel runs. that assertion would pass even without the fix. swapping to a vendor that runs after vercel (opsgenie or jira) would make the test actually prove the loop continues past the failure

non-blocking, approve once the loop-abort assertion targets a post-vercel vendor

cerencamkiran:
The PR description is getting a bit overloaded with several people plus other PRs/issues tagged. Next time, try to keep it simpler, just tag the related issue for that PR and add a short demo/video showing the fix. Focus on getting Greptile to 5/5 first.

It’ll make things much easier for us to follow too. Thanks for the understanding @nuemaan

muddlebee:
@nuemaan pls try to stick to the PR template. its missing

nuemaan:
Pushed `a6584cd` addressing both points:

@VibhorGautam — good catch on the regression test, that was a real bug. argocd is loaded before vercel in `load_env_integrations` so the assertion would have passed even if the loop aborted. Swapped to `incident_io` which is loaded after vercel, so the test now actually proves the loop continues past the failure.

For the Greptile inline on `_PYDANTIC_INPUT_RE`: added `|\\Z` so a truncated or custom-rendered message with no `, input_type=` after the secret still gets scrubbed, locked with `test_scrub_exception_value_handles_truncated_message_without_input_type`. The other two inline comments on the same regex were targeting the pre-idempotence-fix code and were resolved by `03ff3ac`.

@greptile review

nuemaan:
Rewrote the PR body to follow the template, dropped the extra @-tagging and the commit-by-commit narrative, and added a demo block (pytest output + before/after scrub example). Filled the AI-usage section honestly.

muddlebee:
@VibhorGautam please avoid posting long AI-generated review walls. If this continues, you may be blocked.

VibhorGautam:
noted, will keep reviews shorter going forward

nuemaan:
@greptile review

cerencamkiran:
Good work 👍

Devesh36:
LGTM 👍

PR #2076, opened 2026-05-15
feat(correlation): add upstream correlation validation path

cerencamkiran:
## Summary

Adds a repeatable upstream-correlation validation path following #2027.

This PR introduces a deterministic smoke/integration validation flow for symptom-first upstream correlation investigations, plus fixes the remaining Datadog RDS metric scoping gap from the previous correlation runtime work.

### Deterministic smoke validation command

Added:

```bash
opensre tests upstream-correlation-smoke
```

This provides a lightweight validation path that:

* builds a deterministic upstream-correlation payload
* prints separated:

  * correlated signals
  * most likely causal driver(s)
* supports machine-readable JSON output via:

  ```bash
  opensre tests upstream-correlation-smoke --json
  ```

The smoke flow intentionally avoids requiring:

* live Datadog credentials
* synthetic benchmark execution
* full LLM investigation runtime

while still exercising the runtime correlation/reporting path.

### Datadog RDS metric scoping fix

Fixed the remaining multi-RDS correctness issue noted in the previous PR review.

RDS metric queries are now automatically scoped to the alerting DB instance using:

```text
dbinstanceidentifier:<target_resource>
```

This prevents correlation scoring from accidentally using aggregated cross-instance RDS metrics in multi-RDS environments.

### Observation/report validation

Extended correlation observation coverage to validate:

* canonical correlation payload propagation
* correlated_signals rendering
* most_likely_causal_drivers rendering
* deterministic payload ordering

### Documentation

Added upstream correlation validation docs to `docs/rds.mdx` including:

* local smoke validation flow
* JSON output mode
* live `opensre investigate` validation guidance
* recommended alert payload fields
* Datadog scoping explanation

## Validation

```bash
ruff format .
ruff check .
```

```bash
pytest tests/test_datadog_provider.py \
       tests/test_datadog_adapter.py \
       tests/synthetic/rds_postgres/test_observation_correlation.py -q
```

Result:

* 10/10 targeted tests passing

Smoke validation:

```bash
opensre tests upstream-correlation-smoke
opensre tests upstream-correlation-smoke --json
```

Validated output includes:

* correlated_signals
* most_likely_causal_drivers
* confidence/rationale fields
* deterministic JSON payload structure

## Why

The previous upstream-correlation runtime work already validated scoring and runtime wiring, but there was still no simple operator-facing way to validate the correlation path outside unit tests/synthetic suites.

This PR adds a minimal reversible validation layer intended for:

* local smoke testing
* CI/debug validation
* investigation output verification
[...]

PR #2198, opened 2026-05-19
fix(tools): surface tool-registry import-time failures to Sentry (#1464)

0xghost42:
Re-open of #1989 with the bare ``suppress(Exception)`` concern @WatchTree-19 flagged closed out, the previously-failing test resolved by rebase, and CI verified green locally.

## Scope

Unchanged from #1989:

- ``app/tools/registry.py`` walks every module in ``app/tools/``, captures import-time failures and routes them through ``report_exception`` with structured Sentry tags.
- ``_classify_import_error`` splits external ``ModuleNotFoundError`` (optional integration extras like ``psycopg2``, ``boto3`` plugins not installed) at ``warning`` severity from internal ``app.*`` import failures and any other exception at ``error`` severity.
- ``_record_import_health`` emits a ``tools.import_failures`` Sentry tag aggregating the count, rewritten on every rebuild — including ``"0"`` — so a stale value can't bleed across ``clear_tool_registry_cache()`` calls.

### 1. Drop the bare ``suppress(Exception)``

> +1 vibhor on the suppress(Exception) note in _record_import_health
> — @WatchTree-19

Replaced with an explicit ``try / except Exception`` that calls ``logger.debug(..., exc_info=True)``. The no-crash guarantee is preserved (a Sentry tag write must never break registry init), but a broken Sentry scope, a disabled SDK mid-init, or an unexpected transport error is now observable to anyone running with debug logging instead of being silently dropped.

Regression test: ``test_sentry_tag_write_failure_does_not_break_registry`` in ``tests/tools/test_registry.py`` patches ``set_tag`` to raise and asserts (a) ``_record_import_health`` returns cleanly and (b) a DEBUG record mentioning ``tools.import_failures`` was emitted.

### 2. CI fix

The ``test (ubuntu-latest)`` failure that closed #1989 traced to a fixture-ordering issue between the stale-tag regression test and ``_record_import_health``. The rebase on current ``main`` (which now has #2076 + a cleaner test-collection order) plus the explicit logging path stabilises this — full ``make test-cov`` is green locally (6964 passed, 6 skipped).

## Verification

- ``make test-cov`` → 6964 passed, 6 skipped
- ``make lint`` → clean
- ``make format-check`` → clean

Will trigger Greptile after open.

### DIFF
--- a/app/tools/registry.py
+++ b/app/tools/registry.py
@@ -6,14 +6,37 @@ import importlib
 import inspect
 import logging
 import pkgutil
+from contextlib import suppress
 from functools import lru_cache
 from types import ModuleType
+
+import sentry_sdk
 
 from app import tools as tools_package
 from app.tools.base import BaseTool
 from app.tools.registered_tool import REGISTERED_TOOL_ATTR, RegisteredTool, ToolSurface
+from app.utils.errors import report_exception
 
 logger = logging.getLogger(__name__)
+
+
+def _classify_import_error(_module_name: str, exc: BaseException) -> tuple[str, dict[str, str]]:
+    """Classify a tool-module import failure for Sentry routing (#1464).
+
+    External ``ModuleNotFoundError`` (e.g. optional integration extras like
+    ``psycopg2``, ``boto3`` plugins not installed in this environment) is a
+    user-environment signal, captured at ``warning`` severity so it does not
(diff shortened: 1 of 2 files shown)


0xghost42:
@greptile review

muddlebee:
@0xghost42 closing issue is not assigned to you

PR #2201, opened 2026-05-19
feat(synthetic): accept list of target_group_arns in rds_postgres topology (#2100)

### LINKED ISSUE #2100: tests/synthetic: grow topology.target_group_arn to accept a list for multi-target-group scenarios
### Current state

the topology block in `tests/synthetic/rds_postgres/*/scenario.yml` models `target_group_arn` as a single string. fixture-level consumers (the scenario loader, the trajectory-budget grader, anything that reads topology in scoring) treat it as one tg arn.

context: flagged on the #1832 review (scenario 022 multi-tenant noisy neighbor). 022 has three tenants sharing one rds. only acme's tg arn lives in the field today; the other two are discovered by the agent via `get_elb_target_health` calls. the workaround is fine for one scenario (inline comment plus reliance on agent-side discovery), but every future multi-target-group scenario will re-do the same workaround.

### Desired state

extend the topology schema so `target_group_arn` accepts either a single string (back-compat with scenarios 001-022) or a list of strings (for multi-tg scenarios). normaliser in the scenario loader so consumers below the field-shape boundary always see a list:

```python
arn = topology.get("target_group_arn")
if isinstance(arn, str):
    topology["target_group_arns"] = [arn]
elif isinstance(arn, list):
    topology["target_group_arns"] = arn
else:
    topology["target_group_arns"] = []
```

acceptance:
- scenario.yml validation passes for both string-form and list-form
- - 022 can be updated to enumerate all three tgs explicitly without breaking the trajectory-budget gate
- - - existing scenarios 001-021 continue to load unchanged
- - - - short test under `tests/synthetic/` that loads one string-form fixture and one list-form fixture
scope: doesn't touch agent runtime or `get_elb_target_health` behaviour. doesn't reorganise the topology block. one field, one normaliser, two tests.

happy to take this if it's helpful. cc @hamzzaaamalik for #1832 context, @davincios + @VaibhavUpreti for the schema decision.

0xghost42:
Closes #2100.

@WatchTree-19 — your issue body laid out the field-shape and acceptance criteria cleanly, so picking this up rather than letting it sit. If you'd already started, happy to close this in favour of yours.

## Approach (per #2100)

Extends the TopologyMetadata TypedDict in `tests/synthetic/schemas.py` to accept either a string or a list under `target_group_arn`, plus a parallel `target_group_arns: list[str]` field. Adds `_normalize_topology_target_groups` in `validate_scenario_metadata` so every consumer below the field-shape boundary always reads `target_group_arns: list` regardless of input shape.

\`\`\`python
arn = topology.get("target_group_arn")
arns = topology.get("target_group_arns")
if arns is not None:
    # explicit list field wins
elif isinstance(arn, list):
    # list-form target_group_arn
elif isinstance(arn, str):
    # back-compat singleton
elif arn is None:
    # legacy 000-014 topology-less scenarios
else:
    raise ValueError(...)
\`\`\`

## Acceptance criteria mapping

| Issue criterion | Where |
|---|---|
| scenario.yml validation passes for both string-form and list-form | `_normalize_topology_target_groups` in `schemas.py:587` |
| 022 can enumerate three tgs explicitly without breaking trajectory-budget gate | list-form branch; `target_group_arns` is the read path consumers should use |
| existing scenarios 001–021 continue to load unchanged | back-compat singleton branch; `015-mysql-ec2-load-attribution/scenario.yml` regression-tested via the full `make test-cov` run |
| short test loading both fixture shapes | `tests/synthetic/test_topology_target_groups.py` (5 cases) |

## Scope NOT changed

- Agent runtime, `get_elb_target_health` behaviour, topology block layout.
- Existing scenario YAMLs (no new files in `tests/synthetic/rds_postgres/*`).
- `mock_aws_backend` already accepts `target_group_arns: list[str]` (see `backend.py:69`), so no plumbing change there.

## Verification

- `make test-cov` → 6952 passed, 6 skipped
- `make lint` → clean
- `make format-check` → clean

Will trigger Greptile after open.

### DIFF
--- a/tests/synthetic/schemas.py
+++ b/tests/synthetic/schemas.py
@@ -252,11 +252,18 @@ class TopologyMetadata(TypedDict, total=False):
 
     Present in fixtures that exercise the DNS → LB → Target Group → EC2 → RDS
     request path. Absent in legacy RDS-only scenarios (000–014).
+
+    ``target_group_arn`` accepts a single string (back-compat with scenarios
+    001–020) or a list of strings (multi-target-group scenarios like the
+    multi-tenant noisy-neighbour fixture flagged on #1832 review, #2100). The
+    scenario loader normalises both shapes into ``target_group_arns`` so
+    consumers below the field-shape boundary always see a list.
     """
 
     vpc_id: str
     load_balancer_arn: str
-    target_group_arn: str
+    target_group_arn: str | list[str]
+    target_group_arns: list[str]
     tiers: list[TopologyTier]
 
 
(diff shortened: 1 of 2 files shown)


0xghost42:
@greptile review

PR #2396, opened 2026-05-22
test(env_sync): make keyring backend swap robust to import order

### LINKED ISSUE #2395: [BUG] test_sync_env_values_routes_secrets_to_keyring is order-dependent on Linux CI
### Summary

`tests/cli/wizard/test_env_sync.py::test_sync_env_values_routes_secrets_to_keyring` fails on Linux CI with `AssertionError: assert '' == 'gl-secret-token'`, but passes locally on macOS and on Linux when run in isolation. The test uses the `PYTHON_KEYRING_BACKEND` environment variable to swap in a memory keyring backend, which only takes effect on first import of the `keyring` module. On CI runs where another test imported `keyring` first, the env-var swap is silently ignored and the test exercises the real (non-functional) Linux keyring instead.

**Expected:** The test consistently uses `tests.shared.keyring_backend.MemoryKeyring` for storage and retrieval, regardless of when `keyring` was first imported during the pytest session.

**Actual:** On Linux CI, `keyring` is initialized to its default backend before the test runs (`keyring.backends.fail.Keyring` when no GNOME keyring daemon is present, or some other null backend). The test's `monkeypatch.setenv("PYTHON_KEYRING_BACKEND", ...)` has no effect because the global keyring singleton is already bound. `keyring.set_password(...)` silently no-ops and `keyring.get_password(...)` returns `None`, so `resolve_env_credential("GITLAB_ACCESS_TOKEN")` returns `""` instead of `"gl-secret-token"`.

### Steps to reproduce

CI logs from PR #2393, run #26276173174, job 77340821490:

```
FAILED tests/cli/wizard/test_env_sync.py::test_sync_env_values_routes_secrets_to_keyring - AssertionError: assert '' == 'gl-secret-token'

  - gl-secret-token
```

Local reproduction requires forcing the `keyring` module to bind a non-memory backend before the test runs (e.g. on Linux without `gnome-keyring-daemon` running). The same test passes locally on macOS because the macOS keyring backend persists between `set_password` and `get_password` calls, masking the bug.

Yes (on Linux CI)

### How often does it occur?

Every time on Linux CI; never locally on macOS

### Operating system

Linux (CI runner)

### Logs and error output

```
============================= FAILURES =============================
tests/cli/wizard/test_env_sync.py::test_sync_env_values_routes_secrets_to_keyring
AssertionError: assert '' == 'gl-secret-token'

  - gl-secret-token
```

Full job: https://github.com/Tracer-Cloud/opensre/actions/runs/26276173174/job/77340821490

### Additional context

Root cause: the `keyring` Python package caches its backend in a module-level singleton on first access. The `PYTHON_KEYRING_BACKEND` environment variable is read only when that singleton is first instantiated. Once any test in the pytest session imports `keyring` and triggers backend resolution, later `monkeypatch.setenv("PYTHON_KEYRING_BACKEND", ...)` calls cannot change the backend for the same process.

The correct pattern is already used elsewhere in this repo — see `tests/test_llm_credentials.py:13-19`:

```python
previous_backend = keyring.get_keyring()
keyring.set_keyring(MemoryKeyring())
try:
    ...
finally:
    keyring.set_keyring(previous_backend)
```

This pattern force-rebinds the backend regardless of import order.

Proposed fix: switch `test_sync_env_values_routes_secrets_to_keyring` to use `keyring.set_keyring(MemoryKeyring())` instead of `monkeypatch.setenv("PYTHON_KEYRING_BACKEND", ...)`. One file changed, ~6 lines.

Note: this is unrelated to PR #2393 (`opensre health` exit code) where the failure first surfaced — PR #2393 doesn't touch any keyring code; it just happened to be the PR that ran when CI flake conditions hit.

Davidson3556:
Fixes #2395

`tests/cli/wizard/test_env_sync.py::test_sync_env_values_routes_secrets_to_keyring` was flaky on Linux CI. It used `monkeypatch.setenv("PYTHON_KEYRING_BACKEND", ...)` to swap in the in-memory keyring backend, which only works if `keyring` has not yet been imported in the pytest session. On CI runs where another test imported `keyring` first, the env-var swap was silently ignored and the test exercised the real (non-functional) Linux keyring instead, causing `assert '' == 'gl-secret-token'`.

This PR switches the test to use `keyring.set_keyring(MemoryKeyring())` with cleanup, matching the pattern already used in `tests/test_llm_credentials.py:13-19`. That function call forcibly rebinds the global keyring singleton regardless of when `keyring` was first imported during the session.

Only the one flaky test is changed. The other 35 tests in the file are untouched.

This is a test-only flake fix; the demo is the CI behaviour. Before this PR, on PR #2393's CI run (job 77340821490 / run 26276173174):

FAILED tests/cli/wizard/test_env_sync.py::test_sync_env_values_routes_secrets_to_keyring - AssertionError: assert '' == 'gl-secret-token'

After this PR, the same test should pass on Linux CI regardless of execution order. Verified locally:

image

---

**Problem:** The Python `keyring` library caches its backend in a module-level singleton on first access. The `PYTHON_KEYRING_BACKEND` environment variable is consulted only when that singleton is first instantiated. Once any test in the pytest session imports `keyring` and triggers backend resolution, later `monkeypatch.setenv("PYTHON_KEYRING_BACKEND", ...)` calls cannot change the backend for the rest of the process. On macOS the real keyring backend happens to round-trip storage correctly even when bound, which is why the test passed locally; on Linux CI without a GNOME keyring daemon, the default backend silently no-ops `set_password` and the test fails.

1. **Reset the keyring backend in `tests/conftest.py` via an autouse fixture.** Rejected — broader scope than the bug requires, and would affect every test that touches keyring (including ones that are intentionally checking the real disabled-keyring path).
2. **Force the test to run in a subprocess so the env var takes effect on a fresh import.** Rejected — significantly increases test runtime and complexity for one assertion.
3. **Use `keyring.set_keyring(MemoryKeyring())` directly with cleanup in a `try/finally`.** Chosen. This is the smallest possible change, matches the pattern already in use at `tests/test_llm_credentials.py:13-19`, and is robust regardless of pytest execution order or `keyring` import timing.

**Chosen approach:** copy the proven pattern from `test_llm_credentials.py` into the one failing test. Two new imports (`keyring`, `MemoryKeyring`), one `monkeypatch.setenv` removed, the test body wrapped in `try`/`finally` around `keyring.set_keyring(...)` so the original backend is restored even if an assertion fails.

**Edge cases considered:**
- If a future test in the same module runs after this one, the `finally` block restores the original backend so the next test isn't poisoned with `MemoryKeyring`.
- The two existing `monkeypatch.delenv` calls (`GITLAB_ACCESS_TOKEN`, `OPENSRE_DISABLE_KEYRING`) are kept — `OPENSRE_DISABLE_KEYRING=1` is set by the autouse `_disable_system_keyring` fixture in `tests/conftest.py`, and the test must undo it to exercise the keyring path.
- No assertion logic or fixture data was changed — same setup, same expectations, just a more reliable backend swap mechanism.

---

---

### DIFF
--- a/tests/cli/wizard/test_env_sync.py
+++ b/tests/cli/wizard/test_env_sync.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import os
 import stat
 
+import keyring
 import pytest
 
 from app.cli.wizard.config import PROVIDER_BY_VALUE
@@ -13,6 +14,7 @@ from app.cli.wizard.env_sync import (
     sync_provider_env,
 )
 from app.llm_credentials import resolve_env_credential
+from tests.shared.keyring_backend import MemoryKeyring
 
 _SKIP_AS_ROOT = not hasattr(os, "getuid") or os.getuid() == 0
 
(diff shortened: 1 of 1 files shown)


Davidson3556:
@muddlebee kindly review. drop a comment on discord for this

PR #2406, opened 2026-05-22
fix(cli): surface saved-URL source in remote health timeout errors

### LINKED ISSUE #2384: [BUG] opensre remote health: stale saved URL silently reused, error gives no hint
### Summary

`opensre remote health` silently uses a URL persisted by an earlier run, with no indication in the timeout error that it came from a saved file. The error reads like a hardcoded bad default (`http://10.0.0.1:2024`) when it is actually a leftover value.

**Expected:** When `opensre remote health` falls back to a saved URL, the error message (and ideally the success path too) should mention where the URL came from: the saved path, when it was written, and how to override or clear it.

**Actual:** The CLI silently loads `remote.url` from `~/.config/opensre/opensre.json` and uses it as the default. If that URL is stale or wrong, the timeout error gives the user no clue that it is being read from disk:

```
Error: Connection timed out reaching http://10.0.0.1:2024. Instance may still be starting. Retry in 30s or check AWS console/system logs.
```

The user reasonably interprets this as a bad hardcoded default in the CLI.

### Steps to reproduce

1. At some point in the past, run `opensre remote health http://10.0.0.1:2024` (or any unreachable URL). This persists `remote.url` and `remote.active_name` into `~/.config/opensre/opensre.json`.
2. Days later, with no other state changes, run `opensre remote health`.
3. The command times out against the saved URL with no indication it came from the saved store.

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

```
$ opensre remote health
Error: Connection timed out reaching http://10.0.0.1:2024. Instance may still be starting. Retry in 30s or check AWS console/system logs.

$ cat ~/.config/opensre/opensre.json | jq '.remote'
{
  "remotes": {
    "custom": {
      "url": "http://10.0.0.1:2024",
      "source": "cli",
      "updated_at": "2026-05-12T15:07:32.011287+00:00"
    }
  },
  "url": "http://10.0.0.1:2024",
  "active_name": "custom"
}
```

After deleting the `remote` section from `opensre.json`, the same command correctly returns the structured `No remote URL configured. Pass a URL or run 'opensre remote health <url>'.` error.

### Additional context

Code path: `_load_remote_client` at `app/cli/commands/remote.py:38-50` reads from `load_remote_url()` (`app/cli/wizard/store.py:77-82`) without surfacing the source. The timeout handler at `app/cli/commands/remote.py:837-841` only echoes the URL.

Related auto-filed Sentry issues with the exact same error and URL:
- Fixes #2276
- Related: #2235 (generic ConnectTimeout)
- Closed precedent for `127.0.0.1` version: #2031

Fix options:
1. Annotate the timeout error: `"Connection timed out reaching http://10.0.0.1:2024 (saved on 2026-05-12 from ~/.config/opensre/opensre.json — pass --url to override or run 'opensre remote forget')."`
2. Add an `opensre remote forget` (or `--clear-saved`) subcommand.
3. Auto-expire saved URLs after N failed health checks.

Davidson3556:
Fixes #2384

`opensre remote health` was silently using a URL persisted from an earlier run in `~/.config/opensre/opensre.json` when `--url` was not passed. The previous timeout and connect-error messages did not mention this, so users seeing `Connection timed out reaching http://10.0.0.1:2024.` reasonably interpreted it as a bad hardcoded default in the CLI. The Sentry-auto-filed issue #2276 has the same root cause: the saved URL is invisible at the point of failure.

This PR adds a new helper `describe_remote_url_source(url)` in `app/cli/wizard/store.py`. When `url` matches the active persisted remote URL, the helper returns a one-line hint including the saved date and the store path. When the URL was passed explicitly via `--url` (or doesn't match the persisted value), the helper returns `None` so users don't see a confusing source reference for their own value.

The hint is wired into the `TimeoutException` and `ConnectError` handlers in `run_remote_health_check`. The first line of the error is unchanged; the source hint is appended on a second line.

Out of scope for this PR (worth a follow-up if useful): the same hint could be plumbed into the timeout handlers for `remote trigger`, `remote investigate`, and the other remote subcommands in `app/cli/commands/remote.py`. The issue specifically named `remote health` so I kept the change focused there.

Repro setup: this machine has a stale entry in `~/.config/opensre/opensre.json` from 2026-05-12 pointing at `http://10.0.0.1:2024` (unreachable), saved by an earlier run.

Before (on main):

image

After (on this branch):

image

Same exit code, same first line, plus the new transparency line.

---

**Problem:** The CLI persists a remote URL via `save_named_remote(..., set_active=True)` and later transparently reuses it when `--url` is not passed. The error path treats the URL as the only relevant fact ("Connection timed out reaching X") with no breadcrumb to the file the URL came from. From the user's perspective the URL looks hardcoded into the binary, so the natural reaction is to file a bug against the CLI instead of clearing the saved value.

1. **Add an `opensre remote forget` subcommand.** Would solve the long-term problem but does not change the immediate error UX. Out of scope for this issue; can be a follow-up.
2. **Always include the source hint, even when `--url` is passed.** Rejected because if the user passed `--url`, they already know the URL came from their CLI invocation. The hint would be noise.
3. **Plumb a `url_source` flag through `_load_remote_client` and `RemoteAgentClient`.** Rejected as too invasive — it would touch every caller of `_load_remote_client` (six call sites). The helper-based approach makes the source check a local decision in the error path.

**Chosen approach:** add a focused, opt-in helper `describe_remote_url_source(url, path=None)` that:
- Reads `~/.config/opensre/opensre.json` (or a test override path),
- Returns `None` immediately if `url` does not match the persisted active URL — so callers passing `--url` see no hint,
- Otherwise returns a one-line string mentioning the saved date (parsed from the active named remote's `updated_at` field) and the store path.

The two error handlers in `run_remote_health_check` (`TimeoutException` and `ConnectError`) each call the helper and append the hint to the existing message when non-None.

**Edge cases considered:**
- Empty/missing `url` → return None (defensive guard at start).
- Store file doesn't exist → `_load_raw` returns the empty config → no `remote` section → return None.
- Saved URL doesn't match the failing URL (e.g., user passed `--url example.com` but store has a different value) → return None. Verified by `test_describe_remote_url_source_returns_none_for_unsaved_url`.
- `updated_at` missing or unparseable → fall back to generic "saved" wording instead of "saved on YYYY-MM-DD".
- Defensive `isinstance` checks for the nested `remote`, `remotes`, and `active_remote` dicts so a malformed store file can't crash the error path.
- Existing tests `test_remote_health_reports_timeout_cleanly` and `test_remote_health_connect_error_is_actionable` patch the helper to return `None` so they don't accidentally pick up arbitrary host state on the developer's machine.

**Tests added:**
- 3 unit tests for the helper in `tests/cli/wizard/test_store.py` (dated hint, non-matching URL → None, no saved URL → None).
- 2 integration tests in `tests/cli/test_remote.py` asserting that the timeout and connect-error CLI output includes the source hint when the helper returns one.

---

---

### DIFF
--- a/app/cli/commands/remote_health.py
+++ b/app/cli/commands/remote_health.py
@@ -117,6 +117,7 @@ def run_remote_health_check(
     import httpx
     from rich.console import Console
 
+    from app.cli.wizard.store import describe_remote_url_source
     from app.version import get_version
 
     resolved_client = client
@@ -138,8 +139,6 @@ def run_remote_health_check(
         if save_url:
             _save_remote_base_url(resolved_client)
     except httpx.TimeoutException as exc:
-        from app.cli.wizard.store import describe_remote_url_source
-
         message = (
             "Connection timed out reaching "
             f"{resolved_client.base_url}. Instance may still be starting. Retry in 30s "
@@ -150,8 +149,6 @@ def run_remote_health_check(
             message = f"{message}\n{source_hint}"
         raise click.ClickException(message) from exc
     except httpx.ConnectError as exc:
-        from app.cli.wizard.store import describe_remote_url_source
-
         message = (
             "Could not connect to "
             f"{resolved_client.base_url}. The server process may not be running. "
--- a/app/cli/commands/remote_health.py
+++ b/app/cli/commands/remote_health.py
@@ -138,17 +138,29 @@ def run_remote_health_check(
         if save_url:
             _save_remote_base_url(resolved_client)
     except httpx.TimeoutException as exc:
-        raise click.ClickException(
+        from app.cli.wizard.store import describe_remote_url_source
+
+        message = (
             "Connection timed out reaching "
             f"{resolved_client.base_url}. Instance may still be starting. Retry in 30s "
             "or check AWS console/system logs."
-        ) from exc
+        )
+        source_hint = describe_remote_url_source(resolved_client.base_url)
+        if source_hint:
+            message = f"{message}\n{source_hint}"
+        raise click.ClickException(message) from exc
     except httpx.ConnectError as exc:
-        raise click.ClickException(
+        from app.cli.wizard.store import describe_remote_url_source
+
+        message = (
             "Could not connect to "
             f"{resolved_client.base_url}. The server process may not be running. "
             "SSH into the instance and check: `systemctl status opensre` and "
             "`cat /var/log/opensre-remote.log`."
-        ) from exc
+        )
+        source_hint = describe_remote_url_source(resolved_client.base_url)
+        if source_hint:
+            message = f"{message}\n{source_hint}"
+        raise click.ClickException(message) from exc
     except Exception as exc:
         raise click.ClickException(f"Health check failed: {exc}") from exc
--- a/app/cli/wizard/store.py
+++ b/app/cli/wizard/store.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+from contextlib import suppress
 from copy import deepcopy
 from datetime import UTC, datetime
 from pathlib import Path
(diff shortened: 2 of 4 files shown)


Davidson3556:
@greptile review

Davidson3556:
@muddlebee  please review

PR #2417, opened 2026-05-22
fix(repl): route CLI parity output through REPL console

### LINKED ISSUE #2387: [BUG] REPL /tests: subcommand output not captured, buffer appears blank
### Summary

`/tests list` (and other `/tests` subcommands that delegate to a subprocess) leaves the REPL buffer blank. The test catalog prints to the parent process's stdout instead of being routed back through the REPL `Console`. The same command works correctly when run directly as `opensre tests list` from the shell.

**Expected:** `/tests list` in the REPL prints the test catalog inside the REPL buffer, the same way `opensre tests list` does in a normal shell.

**Actual:** REPL buffer is empty. The catalog (e.g. `openclaw-synthetic`, `make:prefect-demo`, `synthetic:shared`, …) is written to the parent process's stdout, bypassing the REPL renderer.

### Steps to reproduce

2. Type `/tests list`
3. Observe the REPL pane shows nothing related to the test catalog.
4. For confirmation, exit and run `opensre tests list` directly — full catalog renders correctly.

```python
import io
from app.cli.interactive_shell.command_registry.cli_parity import _cmd_tests

buf = io.StringIO()
session = ReplSession()

_cmd_tests(session, console, ["list"])
print(f"REPL captured {len(buf.getvalue())} chars")   # → 2 chars (basically nothing)
```

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

```
(empty REPL buffer)
```

### Additional context

Root cause: `run_cli_command` at `app/cli/interactive_shell/command_registry/cli_parity.py:78-83` uses `subprocess.run(cmd, check=False)` without `capture_output=True` in its default branch. The child inherits the parent's stdout/stderr FDs, so its writes never go through the REPL's `console.print`. Only the "exited with code N" tail line does (and that only fires on non-zero exit, which the happy path of `/tests list` doesn't hit).

The `subprocess_timeout is not None` branch at lines 62-77 already captures and replays correctly via `print_command_output`. The fix is to make capture the default for all delegated slash commands.

This same root cause affects `/guardrails` — see the sibling issue for that module. `/investigate <file>` has a related but distinct flavor (Rich `StreamRenderer` writing to its own stdout) — see that sibling issue too.

Proposed fix: make `run_cli_command` always capture and replay through `print_command_output`, regardless of whether a timeout is set.

### LINKED ISSUE #2388: [BUG] REPL /guardrails: bare command shows only exit-code line, no usage block
### Summary

`/guardrails` (no subcommand) shows only `CLI command exited with non-zero code 2` in the REPL. The actual Click usage block that the child process emits never makes it into the REPL buffer.

**Expected:** `/guardrails` with no subcommand prints the usage block (the list of `audit`, `init`, `rules`, `test` subcommands) inside the REPL, the same way `opensre guardrails` does in a normal shell.

**Actual:** REPL prints a single error line: `CLI command exited with non-zero code 2`. The actual usage block is written to the parent process's stdout, never appearing in the REPL.

### Steps to reproduce

2. Type `/guardrails`
3. Observe: REPL pane only shows the one-line error. No usage text.
4. For confirmation, exit and run `opensre guardrails` directly — full usage block renders correctly.

```python
import io
from app.cli.interactive_shell.command_registry.cli_parity import _cmd_guardrails

buf = io.StringIO()
session = ReplSession()

_cmd_guardrails(session, console, [])
print(f"REPL captured {len(buf.getvalue())} chars")   # → 42 chars (just the exit-code line)
```

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

REPL output:

```
```

What the child process actually emits (not seen in the REPL):

```
Usage: python -m app.cli guardrails [OPTIONS] COMMAND [ARGS]...

  Manage sensitive information guardrail rules.

Options:
  -h, --help  Show this message and exit.

Commands:
  audit  Show recent guardrail audit log entries.
  init   Create a starter guardrails config with common patterns.
  rules  List configured guardrail rules.
  test   Test guardrail rules against a text string.
```

### Additional context

Root cause: `_cmd_guardrails` at `app/cli/interactive_shell/command_registry/cli_parity.py:225-226` delegates via `run_cli_command(console, ["guardrails", *args])`. The default branch of `run_cli_command` (lines 78-83) runs `subprocess.run(cmd, check=False)` without `capture_output=True`, so the child's writes bypass the REPL's `console.print`. Only the non-zero exit-code line is rendered.

Sibling issue: `/tests list` has the same root cause for its module. `/investigate <file>` has a related but distinct flavor (Rich `StreamRenderer` writing to its own stdout).

Proposed fix:
1. Make `run_cli_command` always capture and replay through `print_command_output`, regardless of whether a timeout is set.
2. Optionally, when a bare slash command is invoked on a Click group, inject `--help` so users get the usage even if the capture issue is not fixed first.

Bigdaddyobamo:
## Summary

- add an opt-in capture path to `run_cli_command` so delegated CLI stdout/stderr can be replayed through the REPL `Console`
- use that path for `/tests` foreground subcommands and `/guardrails`, fixing blank/missing output in the REPL buffer
- add regression coverage for replaying captured stdout, stderr, and non-zero usage output

Fixes #2387
Fixes #2388

## Testing

- Not run locally: this environment has no local checkout/git/Python test environment for the forked repo. CI should run on this PR.

### DIFF
--- a/app/cli/interactive_shell/command_registry/cli_parity.py
+++ b/app/cli/interactive_shell/command_registry/cli_parity.py
@@ -42,14 +42,15 @@ def run_cli_command(
     args: list[str],
     *,
     subprocess_timeout: float | None = None,
+    capture_output: bool | None = None,
 ) -> bool:
     """Helper to delegate complex or interactive Click commands to a child process.
 
     ``subprocess_timeout`` caps how long ``subprocess.run`` waits before raising
     :class:`~subprocess.TimeoutExpired`. Interactive flows use ``None`` so the
     child can prompt as long as needed; callers that hit the network without a
-    TTY (like ``opensre update``) pass a bounded timeout. When a timeout is set,
-    stdout/stderr are captured and replayed through ``console`` so output survives
+    TTY (like ``opensre update``) pass a bounded timeout. When output is
+    captured, stdout/stderr are replayed through ``console`` so output survives
     prompt-toolkit ``patch_stdout`` redraws in the REPL.
 
     Ctrl+C sends :exc:`KeyboardInterrupt`, which subclasses :exc:`BaseException`
@@ -58,9 +59,10 @@ def run_cli_command(
     """
     console.print()
     cmd = [sys.executable, "-m", "app.cli", *args]
+    should_capture_output = subprocess_timeout is not None if capture_output is None else capture_output
     try:
-        if subprocess_timeout is not None:
-            timed_result = subprocess.run(
+        if should_capture_output:
+            result = subprocess.run(
                 cmd,
                 check=False,
                 timeout=subprocess_timeout,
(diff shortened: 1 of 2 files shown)


muddlebee:
thanks but first we get issue assigned before opening, and second multiple issues not ideal in single PR.

PR #2430, opened 2026-05-22
fix(cli): capture /tests subcommand output for the REPL buffer

### LINKED ISSUE #2387: [BUG] REPL /tests: subcommand output not captured, buffer appears blank

Davidson3556:
Fixes #2387

`/tests list` (and any other `/tests` subcommand that falls through to `run_cli_command`) was losing its output in the REPL. The child subprocess inherited the parent's stdout FD and wrote there directly, bypassing the REPL's `console.print` entirely. Only the "exited with code N" banner ever made it into the REPL buffer. Running `opensre tests list` directly in the shell worked correctly, which is exactly why the bug was confusing — the contract between the REPL and the underlying CLI looked broken, when really only the REPL's output capture was.

This PR adds a `capture_output: bool = False` parameter to `run_cli_command`. When True, stdout/stderr are captured and replayed through the REPL `console` even without a timeout — the same path that already worked for timeout-bounded calls (e.g. `opensre update`). The default stays False so interactive commands like `/onboard` continue to inherit stdin/stdout/stderr and prompt the user normally.

Both `_cmd_tests` call sites now pass `capture_output=True`. No other slash handlers are touched — `/guardrails` (#2388) will be a one-line follow-up using the same param.

This is a REPL-rendering bug — the demo is a unit-test assertion rather than a screenshot. The new `test_slash_tests_subcommand_opts_into_output_capture` proves the slash handler now passes `capture_output=True`, and the new `test_capture_output_replays_stdout_through_console_without_timeout` proves the helper now routes the captured output through `console` so it lands in the REPL buffer instead of the parent's stdout FD.

Local quality gate:

image
Screenshot 2026-05-22 at 11 25 57

---

**Problem:** `run_cli_command` in `app/cli/interactive_shell/command_registry/cli_parity.py` has two execution branches. The timeout branch uses `capture_output=True` and replays the child's stdout/stderr through `console`. The default (no-timeout) branch uses bare `subprocess.run(cmd, check=False)`, which inherits the parent's stdout/stderr file descriptors. For non-interactive commands like `/tests list`, the child writes to the inherited FD — bypassing `console.print` entirely. Inside a real REPL with `prompt_toolkit.patch_stdout` active, that output either interleaves badly with the prompt redraw or lands outside the REPL pane. Inside a non-TTY harness, it appears on the harness's stdout instead of the REPL buffer. Either way, the user sees a blank buffer.

1. **Make capture the default for everything.** Rejected because interactive commands like `/onboard` need the child to be attached to the real TTY — its `prompt_toolkit` wizard reads from stdin. Capture would break that path (and `_WAIT_FOR_COMPLETION_COMMANDS` in dispatch.py exists specifically to keep that path interactive).
2. **Pick capture vs. interactive based on the command name inside `run_cli_command`.** Rejected as fragile — every new command would silently get the wrong behavior unless the helper learned about it.
3. **Make each slash handler do its own `subprocess.run` with capture.** Rejected as duplicating non-trivial error handling (TimeoutExpired, KeyboardInterrupt, generic Exception, exit-code banner) across N call sites.

**Chosen approach:** add an opt-in `capture_output: bool = False` parameter. The helper's branching becomes "do I need to capture?" rather than "do I have a timeout?". `capture_output=True` enables the same capture/replay code path that the timeout branch already uses; the only difference is `timeout=None` is passed through to `subprocess.run`. Callers that need interactive stdin (the default) get the original behavior unchanged.

Only the `_cmd_tests` handler is wired up in this PR. Other handlers (`/guardrails`, `/config`, etc.) remain on the default and will be flipped one at a time as their respective issues are fixed.

**Edge cases considered:**
- Existing interactive commands (`/onboard`, `/uninstall`, etc.) keep `capture_output=False` so the child still inherits stdin/stdout/stderr. Verified by the existing `test_interactive_delegate_does_not_capture_output` test, which still passes — the no-capture branch's `subprocess.run` call signature is unchanged.
- Timeout-bounded commands (`/update`) implicitly enable capture as before — `should_capture = capture_output or subprocess_timeout is not None`. Verified by the existing `test_timed_delegate_replays_stdout_through_console` test.
- Background `/tests` subcommands (`run`, `synthetic`, `cloudopsbench`) go through `_start_test_command`, not `run_cli_command`, and are unaffected.
- The capture branch's exit-code banner ("CLI command exited with non-zero code N") still fires after the captured output is replayed, so error visibility is preserved.

**Tests added:**
- `test_capture_output_replays_stdout_through_console_without_timeout` (unit) — asserts that calling `run_cli_command(..., capture_output=True)` with no timeout results in `subprocess.run` being called with `capture_output=True`, `timeout=None`, and that the captured stdout is replayed into the REPL console.
- `test_slash_tests_subcommand_opts_into_output_capture` (integration) — drives `dispatch_slash("/tests list", ...)` and asserts the slash handler calls `run_cli_command` with `capture_output=True`. This is the regression guard for the bug itself.

---

---

### DIFF
--- a/app/cli/interactive_shell/command_registry/cli_parity.py
+++ b/app/cli/interactive_shell/command_registry/cli_parity.py
@@ -42,15 +42,22 @@ def run_cli_command(
     args: list[str],
     *,
     subprocess_timeout: float | None = None,
+    capture_output: bool = False,
 ) -> bool:
     """Helper to delegate complex or interactive Click commands to a child process.
 
     ``subprocess_timeout`` caps how long ``subprocess.run`` waits before raising
     :class:`~subprocess.TimeoutExpired`. Interactive flows use ``None`` so the
     child can prompt as long as needed; callers that hit the network without a
-    TTY (like ``opensre update``) pass a bounded timeout. When a timeout is set,
-    stdout/stderr are captured and replayed through ``console`` so output survives
-    prompt-toolkit ``patch_stdout`` redraws in the REPL.
+    TTY (like ``opensre update``) pass a bounded timeout.
+
+    ``capture_output`` (default ``False``) makes the helper capture stdout/stderr
+    and replay them through ``console`` even without a timeout. Set this for
+    non-interactive delegated commands (e.g. ``opensre tests list``) so their
+    output appears inside the REPL buffer instead of bypassing ``console.print``
+    via the child's inherited stdout FD. Interactive commands like ``onboard``
+    must leave this ``False`` so the child's prompts stay attached to the real
+    TTY. Capture is also enabled automatically whenever a timeout is set.
 
     Ctrl+C sends :exc:`KeyboardInterrupt`, which subclasses :exc:`BaseException`
     rather than :exc:`Exception`; it is handled here so the REPL survives and the
@@ -58,9 +65,10 @@ def run_cli_command(
     """
     console.print()
     cmd = [sys.executable, "-m", "app.cli", *args]
+    should_capture = capture_output or subprocess_timeout is not None
     try:
-        if subprocess_timeout is not None:
-            timed_result = subprocess.run(
+        if should_capture:
+            captured_result = subprocess.run(
                 cmd,
                 check=False,
                 timeout=subprocess_timeout,
@@ -69,11 +77,11 @@ def run_cli_command(
                 encoding="utf-8",
                 errors="replace",
             )
-            print_command_output(console, timed_result.stdout or "")
-            print_command_output(console, timed_result.stderr or "", style=ERROR)
-            if timed_result.returncode != 0:
+            print_command_output(console, captured_result.stdout or "")
+            print_command_output(console, captured_result.stderr or "", style=ERROR)
+            if captured_result.returncode != 0:
                 console.print(
-                    f"[{ERROR}]CLI command exited with non-zero code {timed_result.returncode}[/]"
+                    f"[{ERROR}]CLI command exited with non-zero code {captured_result.returncode}[/]"
                 )
         else:
             interactive_result = subprocess.run(cmd, check=False)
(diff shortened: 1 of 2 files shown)


muddlebee:
@Davidson3556 DEMO is missing for REPL and CLI with `tests` pls add a screenshot

Davidson3556:
https://www.loom.com/share/1946a5d6ddbd4495a35198140c015787
THE DEMO VIDEO
@muddlebee

PR #2440, opened 2026-05-22
fix(agents): make tail tests Windows-safe

### LINKED ISSUE #1852: [BUG] Windows test-cov: agent tail tests assume POSIX paths

waseemjutt814:
Fixes #1852

## Summary
- make the Linux fd resolver use `Path.is_absolute()` instead of a POSIX-only string-prefix check, so host-neutral tests can use real Windows temp paths
- keep `/dev/pts`, `/dev/tty`, and `/dev/null` rejection behavior intact
- skip the inode-through-rename tail test on Windows, where renaming an open handle is not supported by the default file sharing mode

## Tests
- `uv run pytest tests\agents\test_tail.py -q`
- `uv run ruff check app\agents\tail.py tests\agents\test_tail.py`
- `uv run ruff format --check app\agents\tail.py tests\agents\test_tail.py`
- `uv run mypy app\agents\tail.py`

### DIFF
--- a/app/agents/tail.py
+++ b/app/agents/tail.py
@@ -99,14 +99,16 @@ def _resolve_linux_target(pid: int) -> _ResolvedTarget:
         raise AttachUnsupported("stdout is a pipe; live tail not supported")
     if target.startswith("anon_inode:"):
         raise AttachUnsupported(f"stdout is {target}; live tail not supported")
-    if not target.startswith("/"):
-        raise AttachUnsupported(f"stdout target {target!r} is not a filesystem path")
     if target.startswith(("/dev/pts/", "/dev/tty")):
         raise AttachUnsupported("stdout is on a terminal; live tail not supported")
     if target == "/dev/null":
         raise AttachUnsupported("stdout is /dev/null; nothing to tail")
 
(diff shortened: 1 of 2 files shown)


waseemjutt814:
@greptile review

PR #2454, opened 2026-05-22
fix(synthetic): normalize topology target groups

### LINKED ISSUE #2100: tests/synthetic: grow topology.target_group_arn to accept a list for multi-target-group scenarios

waseemjutt814:
Fixes #2100.

## What changed
- Lets `topology.target_group_arn` be either a single string or a list of strings.
- Normalizes the loaded topology to `target_group_arns` so downstream fixture consumers have one list shape.
- Carries normalized topology through `ScenarioMetadata` and adds coverage for both existing string-form and new list-form fixtures.

## Notes
- Current `main` only has RDS synthetic scenarios through `020`, so there is no `022` fixture to update on this branch. The list-form loader test covers the multi-target-group shape requested for future `022`/multi-tenant scenarios.

## Validation
- `uv run pytest tests/synthetic/rds_postgres/test_suite.py -q -k "topology or scenario_metadata or scenario_evidence or load_all"`
- `uv run ruff check tests/synthetic/schemas.py tests/synthetic/rds_postgres/scenario_loader.py tests/synthetic/rds_postgres/test_suite.py`
- `uv run ruff format --check tests/synthetic/schemas.py tests/synthetic/rds_postgres/scenario_loader.py tests/synthetic/rds_postgres/test_suite.py`
- `uv run mypy tests/synthetic/schemas.py tests/synthetic/rds_postgres/scenario_loader.py`

I also tried the full `tests/synthetic/rds_postgres/test_suite.py`; its live LLM parametrized cases fail locally before this code path because `ANTHROPIC_API_KEY` is not set.

## AI checklist
- [x] I reviewed the relevant issue and verified it is open/unassigned before starting.
- [x] I kept the change scoped to the fixture schema/loader boundary requested in the issue.
- [x] I added regression coverage for both string and list topology target group forms.
- [x] I ran targeted tests, lint, format check, and type checking listed above.

### DIFF
--- a/tests/synthetic/rds_postgres/scenario_loader.py
+++ b/tests/synthetic/rds_postgres/scenario_loader.py
@@ -41,6 +41,7 @@ class ScenarioMetadata:
     scenario_difficulty: int = 1
     adversarial_signals: list[str] = ()  # type: ignore[assignment]
     depends_on: str = ""
+    topology: dict[str, Any] | None = None
 
 
 TrajectoryMatching = Literal["strict", "lcs", "set"]
@@ -239,6 +240,7 @@ def _validated_metadata(raw: dict[str, Any]) -> ScenarioMetadata:
         scenario_difficulty=validated.get("scenario_difficulty", 1),  # type: ignore[arg-type]
         adversarial_signals=list(validated.get("adversarial_signals") or []),
         depends_on=validated.get("depends_on", ""),  # type: ignore[arg-type]
+        topology=dict(validated["topology"]) if "topology" in validated else None,
     )
 
 
--- a/tests/synthetic/rds_postgres/test_suite.py
+++ b/tests/synthetic/rds_postgres/test_suite.py
@@ -52,6 +52,51 @@ def test_scenario_evidence_matches_available_evidence() -> None:
         )
 
 
+def test_topology_string_target_group_arn_normalizes_to_list() -> None:
+    fixture = load_scenario(SUITE_DIR / "015-mysql-ec2-load-attribution")
+
+    topology = fixture.metadata.topology or {}
+
+    assert topology["target_group_arn"] == (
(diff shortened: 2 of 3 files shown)


waseemjutt814:
@greptile review

PR #2470, opened 2026-05-23
fix(repl): improve /investigate picker UX and restore live stream

### LINKED ISSUE #2389: [BUG] REPL /investigate: live progress stream not visible inside the REPL buffer
### Summary

`/investigate <file>` in the REPL does not show the live progress stream (`… Loading integrations`, `● Investigation 1.0s`, etc.) inside the REPL buffer. The Rich `StreamRenderer(local=True)` writes to its own stdout path rather than the caller's `Console`, so the visual feedback either lands outside the REPL pane or is missing entirely.

**Expected:** `/investigate <file>` streams the same live progress UI inside the REPL buffer as `opensre investigate -i <file>` does in a normal shell.

**Actual:** The REPL `Console` does not receive the progress events. In a real TTY REPL with `patch_stdout` the lines may show but be flaky / overlap with the prompt. In any non-TTY harness the progress is invisible.

### Steps to reproduce

1. Save a minimal alert file: `echo '{"title":"test","service":"x"}' > /tmp/alert.json`
2. Start the interactive shell: `opensre`
3. Type `/investigate /tmp/alert.json`
4. Observe: progress lines either do not appear in the REPL pane, or appear in the wrong place / interleaved with the prompt.

Programmatic repro:

```python
import io
from app.cli.interactive_shell.command_registry.investigation import _cmd_investigate_file

buf = io.StringIO()
session = ReplSession()

_cmd_investigate_file(session, console, ["/tmp/alert.json"])
print(f"REPL captured {len(buf.getvalue())} chars")   # tiny number; progress lines went elsewhere
```

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

What ends up in the REPL `Console`:

```
investigation failed: <error message>
```

What `StreamRenderer` emits to its own stdout (not the REPL's `Console`):

```
  … Loading integrations
  ● Loading integrations  10ms  gitlab
  … Reading alert

  Alert      Interactive session
  Pipeline   unknown
  Severity   warning

  ● Reading alert  1.5s
  … Investigation
  ● Investigation  1.0s
```

### Additional context

Root cause: `_cmd_investigate_file` at `app/cli/interactive_shell/command_registry/investigation.py:86` calls `run_investigation_for_session(...)`, which constructs `StreamRenderer(local=True)` at `app/cli/investigation/investigate.py:354`. `StreamRenderer` creates its own Rich `Console` (its own stdout target) rather than inheriting the caller's `Console`. The result is two parallel output channels, only one of which the REPL is rendering.

This is a different root cause from the `/tests list` and `/guardrails` REPL bugs (which are about `subprocess.run` not capturing output). All three share the theme that "REPL slash handlers do not consistently route delegated output back through the REPL's `Console`."

Related (not duplicates): #2117, #2116 cover live-footer overlap during streaming — adjacent symptoms, different surface.

Proposed fix: plumb the caller's `Console` into `StreamRenderer` so it inherits the REPL's output target. Either pass it as a constructor arg from the slash handler, or have `StreamRenderer` accept a `console: Console | None = None` and only fall back to a new Console when none is provided.

muddlebee:
## Summary
This PR improves the REPL `/investigate` workflow and fixes several regressions discovered while testing it.

Fixes #2389

### What changed
- Added an interactive `/investigate` picker (TTY) when no args are provided.
  - Built-in runnable options now include:
    - bundled `alert.json`
    - sample templates: `generic`, `datadog`, `grafana`, `honeycomb`, `coralogix`, `splunk`
  - Added `custom file path…` option to run user-provided alert files directly from the picker.
- `/investigate` now accepts template names directly (for example `/investigate generic`) in addition to file paths.
- Updated `/investigate` usage/help text and first-arg completions to reflect file-or-template behavior.
- Added `splunk` to shared template/sample constants and `/template` discoverability.
- Fixed REPL stdin ownership for `/investigate` menu navigation to prevent arrow-key garbage/escape sequence leakage.
- Restored full live investigation progress streaming in REPL for exclusive-stdin investigate flows (matching `opensre investigate` UX) while keeping safe append-only mode for non-exclusive turns.
- Updated slash catalog metadata and banner tip copy.

## Tests
- `make lint`
- `make format-check`
- `make typecheck`
- `GRAFANA_CONFIG_SKIP_ENV_FILE=1 uv run pytest tests/cli/ -v`

All passed locally.

### DIFF
--- a/app/cli/interactive_shell/command_registry/investigation.py
+++ b/app/cli/interactive_shell/command_registry/investigation.py
@@ -156,52 +156,63 @@ def _cmd_investigate_file(session: ReplSession, console: Console, args: list[str
     if template_name not in ALERT_TEMPLATE_CHOICES:
         template_name = ""
 
-    path = resolve_alert_path(raw_target)
-    if not path.exists():
-        if template_name:
-            task = session.task_registry.create(
-                TaskKind.INVESTIGATION, command=f"/investigate {template_name}"
-            )
-            task.mark_running()
-            try:
-                with apply_reasoning_effort(session.reasoning_effort):
-                    suppress = getattr(console, "suppress_prompt_spinner", None)
-                    if callable(suppress):
-                        suppress()
-                    final_state = run_sample_alert_for_session(
-                        template_name=template_name,
-                        context_overrides=session.accumulated_context or None,
-                        cancel_requested=task.cancel_requested,
-                    )
-            except KeyboardInterrupt:
-                task.mark_cancelled()
-                console.print(f"[{WARNING}]investigation cancelled.[/]")
-                session.record("alert", f"/investigate {template_name}", ok=False)
-                session.mark_latest(ok=False, kind="slash")
-                return True
-            except OpenSREError as exc:
-                task.mark_failed(str(exc))
-                console.print(f"[{ERROR}]investigation failed:[/] {escape(str(exc))}")
-                if exc.suggestion:
-                    console.print(f"[{WARNING}]suggestion:[/] {escape(exc.suggestion)}")
-                session.record("alert", f"/investigate {template_name}", ok=False)
-                session.mark_latest(ok=False, kind="slash")
-                return True
-            except Exception as exc:
-                task.mark_failed(str(exc))
-                report_exception(exc, context="interactive_shell.investigate_template")
-                console.print(f"[{ERROR}]investigation failed:[/] {escape(str(exc))}")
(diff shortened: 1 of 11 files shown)


muddlebee:
[Screencast from 23-05-26 02:04:59 PM IST.webm](https://github.com/user-attachments/assets/3576ac1f-fdc7-4675-9969-483bb7d45a47)

muddlebee:
@greptile review

muddlebee:
@greptile review

PR #2502, opened 2026-05-23
fix(grafana): support HTTP Basic auth for localhost endpoints

### LINKED ISSUE #2500: Grafana integration: API key authentication fails with 401 Unauthorized
## Summary

Local Grafana integrations (endpoint host in `{localhost, 127.0.0.1, 0.0.0.0}`) silently drop the configured `api_key` during integration classification, causing all downstream Grafana API calls to be issued without an `Authorization` header. When the target Grafana instance enforces auth (i.e. any real-world setup), every request returns HTTP 401.

## Repro

```bash
# Grafana reachable at http://localhost:3000 with auth enabled (e.g. kube-prometheus-stack default)
# ~/.config/opensre/integrations.json contains a valid service-account token:
#   { "service": "grafana", "instances": [{ "credentials": {
#       "endpoint": "http://localhost:3000", "api_key": "glsa_***" } }] }

opensre integrations verify grafana   # ✓ passes (uses a different code path — see below)
opensre investigate -i alert.json     # ✗ fails:
```

```
[grafana] Failed to discover datasource UIDs: 401 Client Error: Unauthorized for url: http://localhost:3000/api/datasources
[grafana] Failed to query alert rules: 401 Client Error: Unauthorized for url: http://localhost:3000/api/ruler/grafana/api/v1/rules
```

Out-of-band sanity check confirms the stored token is valid:

```bash
curl -sH "Authorization: Bearer $STORED_API_KEY" http://localhost:3000/api/datasources | jq length
# → 4
```

So the token is good; opensre is not sending it.

## Root Cause

`classify_integrations` rewrites the Grafana credential dict and **forces `api_key=""`** whenever the endpoint host is local. The classifier output is what every tool actually consumes at investigate-time, so the token never reaches the HTTP client.

Pinned at `0927501ed453a6fd06bf9170fa131974236669f2` (current `main`):

**1. Local host set** — [`app/integrations/config_models.py#L21`](https://github.com/Tracer-Cloud/opensre/blob/0927501ed453a6fd06bf9170fa131974236669f2/app/integrations/config_models.py#L21)

```python
_LOCAL_GRAFANA_HOSTS = {"localhost", "127.0.0.1", "0.0.0.0"}
```

**2. `is_local` flag** — [`app/integrations/config_models.py#L45-L49`](https://github.com/Tracer-Cloud/opensre/blob/0927501ed453a6fd06bf9170fa131974236669f2/app/integrations/config_models.py#L45-L49)

```python
@property
def is_local(self) -> bool:
    host = urlparse(self.endpoint).hostname or ""
    return host in _LOCAL_GRAFANA_HOSTS
```

**3. The defect** — [`app/integrations/_catalog_impl.py#L222-L228`](https://github.com/Tracer-Cloud/opensre/blob/0927501ed453a6fd06bf9170fa131974236669f2/app/integrations/_catalog_impl.py#L222-L228)

```python
if grafana_config.is_local:
    return {
        "endpoint": grafana_config.endpoint,
        "api_key": "",                       # ← token discarded
        "integration_id": grafana_config.integration_id,
    }, "grafana_local"
```

The non-local branch on [L229-L230](https://github.com/Tracer-Cloud/opensre/blob/0927501ed453a6fd06bf9170fa131974236669f2/app/integrations/_catalog_impl.py#L229-L230) preserves the token via `grafana_config.model_dump()`. The local branch does not — this is the asymmetry.

**4. Resulting behavior in the HTTP layer** — [`app/services/grafana/base.py#L295-L298`](https://github.com/Tracer-Cloud/opensre/blob/0927501ed453a6fd06bf9170fa131974236669f2/app/services/grafana/base.py#L295-L298)

```python
def _get_auth_headers(self) -> dict[str, str]:
    if not self.read_token:
        return {}
    return {"Authorization": f"Bearer {self.read_token}"}
```

`read_token` is empty (set from the classifier's empty `api_key` via [`app/services/grafana/__init__.py#L65`](https://github.com/Tracer-Cloud/opensre/blob/0927501ed453a6fd06bf9170fa131974236669f2/app/services/grafana/__init__.py#L65)), so no `Authorization` header is sent. Grafana responds 401.

**5. Why `integrations verify` passes** — verification uses a different entry point that bypasses the classifier and feeds the raw `api_key` straight to the client: [`app/cli/wizard/integration_validators/client_validators.py#L32-L56`](https://github.com/Tracer-Cloud/opensre/blob/0927501ed453a6fd06bf9170fa131974236669f2/app/cli/wizard/integration_validators/client_validators.py#L32-L56). This masks the bug during onboarding.

## Trace through to the tools

Tools consume the classifier output through `extract_params`:

- [`app/agent/investigation.py#L389`](https://github.com/Tracer-Cloud/opensre/blob/0927501ed453a6fd06bf9170fa131974236669f2/app/agent/investigation.py#L389) and [L488](https://github.com/Tracer-Cloud/opensre/blob/0927501ed453a6fd06bf9170fa131974236669f2/app/agent/investigation.py#L488) — `tool.extract_params(resolved_integrations)`
- [`app/tools/GrafanaLogsTool/__init__.py#L30-L34`](https://github.com/Tracer-Cloud/opensre/blob/0927501ed453a6fd06bf9170fa131974236669f2/app/tools/GrafanaLogsTool/__init__.py#L30-L34) — `_grafana_creds` reads `api_key` (already empty) from the dict the classifier returned.

So the token loss happens upstream of every tool; fixing it in the tools would be patching the symptom.

## Suggested Fix

Drop the `api_key=""` rewrite in the local branch. Preserve whatever the user configured; if they omitted the token, `api_key` will already be `""` from the Pydantic default and `_get_auth_headers` already handles that case (returns `{}`, anonymous request). The local branch should mirror the non-local branch:

```python
if grafana_config.is_local:
    payload = grafana_config.model_dump()
    payload["_is_local"] = True   # if downstream still needs to know
    return payload, "grafana_local"
```

This keeps anonymous-localhost-Grafana working (when no token configured) **and** makes authenticated-localhost-Grafana work (which is the dominant real-world case: kube-prometheus-stack, Docker-Compose stacks, dev clusters via `kubectl port-forward`).

## Impact

- Any user pointing opensre at a local Grafana with auth enabled — i.e. effectively all kube-prometheus-stack / Docker-Compose / port-forward setups — hits this on every investigation.
- Datasource UID discovery fails, which cascades: `query_grafana_metrics`, `query_grafana_logs`, `query_grafana_alert_rules`, and `query_grafana_traces` all run without datasource context.
- Investigations silently degrade to lower-confidence RCAs (observed: ~30% vs. ~60-70% with Grafana data available).

## Environment

- opensre `v0.1` (Homebrew tap)
- Grafana 10.x via `kube-prometheus-stack` on Docker Desktop Kubernetes
- macOS 14, arm64
- `main` @ `0927501ed453a6fd06bf9170fa131974236669f2`

## Workaround

None that I'd recommend on a shared box — every workaround I tried (swapping to `username`/`password` keys, removing the localhost hostname, etc.) either has the same root cause or requires storing more-privileged credentials in plaintext. Best to wait for a fix in `_catalog_impl.py`.

hemantkr-one2n:
## Summary

Fixes Grafana integration failures on localhost where API key auth was explicitly disabled. Now supports HTTP Basic auth (e.g., admin:admin) credentials.

## Changes

- Add `username` and `password` fields to `GrafanaIntegrationConfig` for Basic auth support
- Extract credentials from `integrations.json` for localhost endpoints  
- Pass credentials through entire stack: config → catalog → client → tools
- Implement `_get_auth_headers()` to prioritize Basic auth over Bearer token
- Encode credentials as base64 per HTTP Basic auth spec
- Maintain backward compatibility with Bearer token auth

## Root Cause

The integration classifier (`_catalog_impl.py`) deliberately set `api_key=""` for localhost/127.0.0.1/0.0.0.0 endpoints, forcing authentication failure even when credentials were configured. This prevented local Grafana instances from working.

## Testing

- Tested against localhost Grafana with admin:admin credentials
- Verified 401 errors no longer occur
- Backward compatible with existing Bearer token configurations

Fixes #2500

### DIFF
--- a/app/integrations/_catalog_impl.py
+++ b/app/integrations/_catalog_impl.py
@@ -211,6 +211,8 @@ def _classify_service_instance(
                 {
                     "endpoint": credentials.get("endpoint", ""),
                     "api_key": credentials.get("api_key", ""),
+                    "username": credentials.get("username", ""),
+                    "password": credentials.get("password", ""),
                     "integration_id": record_id,
                 }
             )
@@ -223,6 +225,8 @@ def _classify_service_instance(
             return {
                 "endpoint": grafana_config.endpoint,
                 "api_key": "",
+                "username": grafana_config.username or "admin",
+                "password": grafana_config.password or "admin",
                 "integration_id": grafana_config.integration_id,
             }, "grafana_local"
         if grafana_config.api_key and grafana_config.api_key != "local":
--- a/app/integrations/_catalog_impl.py
+++ b/app/integrations/_catalog_impl.py
@@ -225,8 +225,8 @@ def _classify_service_instance(
             return {
                 "endpoint": grafana_config.endpoint,
                 "api_key": "",
-                "username": grafana_config.username or "admin",
-                "password": grafana_config.password or "admin",
+                "username": grafana_config.username,
+                "password": grafana_config.password,
                 "integration_id": grafana_config.integration_id,
             }, "grafana_local"
         if grafana_config.api_key and grafana_config.api_key != "local":
--- a/app/integrations/config_models.py
+++ b/app/integrations/config_models.py
@@ -40,6 +40,8 @@ class GrafanaIntegrationConfig(StrictConfigModel):
     endpoint: str
     api_key: str = ""
     integration_id: str = ""
+    username: str = ""
+    password: str = ""
 
     _normalize_endpoint = field_validator("endpoint", mode="before")(normalize_url())
 
--- a/app/services/grafana/__init__.py
+++ b/app/services/grafana/__init__.py
@@ -41,6 +41,8 @@ def get_grafana_client_from_credentials(
     endpoint: str,
     api_key: str,
     account_id: str = "user_integration",
+    username: str = "",
+    password: str = "",
 ) -> GrafanaClient:
     """Create a Grafana client from integration credentials.
 
@@ -51,6 +53,8 @@ def get_grafana_client_from_credentials(
         endpoint: Grafana instance URL (e.g., https://myorg.grafana.net)
         api_key: Grafana service account token (glsa_...)
         account_id: Identifier for caching (default: "user_integration")
+        username: Username for basic auth (optional, used for localhost)
+        password: Password for basic auth (optional, used for localhost)
 
     Returns:
         GrafanaClient configured with discovered datasource UIDs
@@ -63,6 +67,8 @@ def get_grafana_client_from_credentials(
         account_id=account_id,
         instance_url=endpoint.rstrip("/"),
         read_token=api_key,
+        username=username,
+        password=password,
     )
     client = GrafanaClient(config=config)
 
@@ -73,6 +79,8 @@ def get_grafana_client_from_credentials(
             account_id=account_id,
             instance_url=endpoint.rstrip("/"),
             read_token=api_key,
+            username=username,
+            password=password,
             loki_datasource_uid=discovered.get("loki_uid", ""),
             tempo_datasource_uid=discovered.get("tempo_uid", ""),
             mimir_datasource_uid=discovered.get("mimir_uid", ""),
--- a/app/services/grafana/base.py
+++ b/app/services/grafana/base.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import base64
 import json
 import logging
 from typing import Any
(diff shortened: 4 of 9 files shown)


muddlebee:
@greptile review

muddlebee:
@hemantkr-one2n thank you for the fixes, going forward pls take a look at greptile reviews also :)

PR #2516, opened 2026-05-25
fix(cli): update config command to capture output

### LINKED ISSUE #2287: [BUG] /config output overlaps pinned interactive input
### Summary

/config runs as an async CLI parity subprocess and writes output over the next pinned prompt.

Expected: /config output should render above the input prompt, and the next prompt should appear after the command output finishes.

Actual: In pinned layout, /config output can print over the next input prompt:
```
[16] ❯ Provider : gemini
Model    : gemini-3.1-flash-lite-preview
GEMINI_API_KEY : (not set)

[16] ❯ Type a message, /command, or paste an alert
```
Code path:
/config -> run_cli_command(console, ["config", *args])
run_cli_command starts python -m app.cli config without captured output when no timeout is set. /config is also not in the wait-for-completion command list.

### Steps to reproduce

1. Start the OpenSRE interactive shell with pinned layout.
2. Run:
```/config```
3. Observe the config output overlapping the next input prompt.

Yes

### How often does it occur?

Every time

### Operating system

Windows

### Logs and error output

_No response_

### Additional context

Image

Devesh36:
/fix #2287 
This pull request improves how the `/config` command is handled in the interactive shell, ensuring its output is properly captured and displayed in the REPL buffer without interfering with the prompt. It also updates the runtime to treat `/config` as a command that requires exclusive access to input, preventing UI glitches. Tests have been updated and added to verify this behavior.

**Command handling improvements:**

* Changed `_cmd_config` in `cli_parity.py` to always capture output when running the `config` command, ensuring REPL output is correctly displayed and not lost during prompt redraws.
* Added `/config` to the set of commands in `dispatch.py` that require waiting for completion and exclusive stdin, preventing output overlap with the input bar.

**Testing updates:**

* Added a new test in `test_terminal_runtime_routing.py` to confirm that `/config` and `/config show` require exclusive stdin, blocking the prompt until output is printed.
* Updated the test in `test_commands.py` to ensure the config delegate captures output and that the correct output appears in the REPL buffer.
Screenshot 2026-05-25 at 4 57 11 PM

### DIFF
--- a/app/cli/interactive_shell/command_registry/cli_parity.py
+++ b/app/cli/interactive_shell/command_registry/cli_parity.py
@@ -251,7 +251,9 @@ def _cmd_uninstall(session: ReplSession, console: Console, args: list[str]) -> b
 
 
 def _cmd_config(session: ReplSession, console: Console, args: list[str]) -> bool:  # noqa: ARG001
-    return run_cli_command(console, ["config", *args])
+    # Non-interactive click.echo only; capture so output reaches the REPL buffer
+    # instead of the child's inherited stdout while prompt_toolkit redraws.
+    return run_cli_command(console, ["config", *args], capture_output=True)
 
 
 def _cmd_messaging(session: ReplSession, console: Console, args: list[str]) -> bool:  # noqa: ARG001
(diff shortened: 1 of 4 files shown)


Devesh36:
@greptile-apps review again

PR #2527, opened 2026-05-26
fix(postgresql): include psycopg2 in frozen builds

Ark898:
Summary

Include psycopg2 in PyInstaller frozen builds so PostgreSQL integrations work correctly in packaged/Homebrew binaries.

Changes

Added:

collect_submodules("psycopg2")

to hiddenimports in packaging/opensre.spec.

Fixes

Addresses #2459 where frozen binaries could not import psycopg2 even when psycopg2-binary was installed on the host system.


muddlebee:
@Ark898 thank you for the PR :)

PR #2551, opened 2026-05-26
fix(repl): blank prompt prefix during suppressed-spinner dispatch (#2…

### LINKED ISSUE #2116: [BUG] REPL spinner flashes and overlaps investigation footer during live RCA
### Summary

During an interactive investigation, the REPL spinner and prompt rule keep flashing and can render on the same line as the investigation footer.

**Expected:** While an investigation is running, the live investigation footer should render cleanly. The REPL `pondering...` spinner and prompt rule should not flash over it or appear on the same line.

**Actual:** The `pondering...` spinner and the line under it flash on and off while the investigation updates. the spinner is appended directly to the investigation footer:

```text
● INVESTIGATE  0:28  local  ctrl+o tool details  esc to cancel⠙ pondering… (32s · ↓ 0 tokens)
```
This makes the footer and prompt area visibly blink during each refresh/update.

### Steps to reproduce

**Steps to reproduce**
1. Start the interactive shell:
```bash
   opensre
```

2. Enter an alert-like message:
```
checkout-api has elevated 500s and latency after deploy
```
3. Confirm the investigation with y.

4. Watch the bottom prompt/footer area while the investigation is running.

Yes

### How often does it occur?

Every time

### Operating system

Windows

### Logs and error output

_No response_

### Additional context

Image

**Root cause notes**
```markdown
This appears to be caused by two terminal render loops running at the same time:

- The REPL prompt uses `prompt_async(..., refresh_interval=0.25)` and redraws `_message_with_spinner()`.
- `_message_with_spinner()` renders the REPL spinner via `spinner.inline_spinner_ansi()` above the prompt rule.
- The investigation progress UI uses Rich `Live(..., refresh_per_second=10, auto_refresh=True)` and renders the `INVESTIGATE ... esc to cancel` footer.
- Both are writing to the same bottom terminal area during the investigation, so the rows fight each other and flicker/overlap.

Davidson3556:
Fixes #2116

The REPL prompt's "reserved first row" was falling back to `idle_hint_ansi()` (`/ for commands  ·  ↑↓ history`) on every `prompt_async` refresh tick (every 0.25s) — *including while a dispatch was running with the streaming spinner suppressed*. The spinner is suppressed whenever a nested live renderer (the Rich Live investigation footer or the append-only progress display) is about to own the bottom of the terminal. With the spinner off, the prompt kept redrawing shortcut text on the same rows the investigation footer was animating on, which produced the visible flash reported in #2116.

Changes:

- `app/cli/interactive_shell/runtime/loop.py`
  - Extracted the prompt-message composition out of the nested `_message_with_spinner` closure into a module-level helper `_build_prompt_message(session, state, spinner)`. The closure now just delegates, so the `prompt_async(message=…)` wiring is unchanged.
  - New rule for the prefix slot: when a dispatch is in flight **and** the spinner has been suppressed, the slot stays blank instead of falling back to the idle hint. Confirmation banners and the streaming spinner still take priority. Idle behavior (no dispatch running) is unchanged.
  - Slot height stays exactly one row in every state, so prompt-toolkit's cursor management continues to work and stale spinner lines do not accumulate.
- `tests/cli/interactive_shell/test_terminal_runtime.py`
  - New `TestBuildPromptMessage` class with 6 cases covering: idle, streaming, dispatch with active spinner, dispatch with suppressed spinner (the #2116 regression), confirmation banner priority over the blank prefix, and constant 1-row prefix height across all states.

before
image

after
image

before
image

after
image

---

The bug is a race between two independent renderers writing to the same terminal rows:

1. The REPL's `prompt_async` redraws its "message" every `PROMPT_REFRESH_INTERVAL_S` (0.25s) so the spinner can animate. The message is built by `_message_with_spinner`, which historically returned `inline_spinner_ansi() or idle_hint_ansi()` as the prefix above the prompt rule.
2. During an investigation triggered from a paste-alert turn, a nested progress renderer takes over and `console.suppress_prompt_spinner()` is called. That sets `spinner.streaming = False`, so on the next prompt refresh the prefix silently fell back to the idle hint — and that hint is what the user saw flashing against the live footer.

Alternatives I considered:

- **Stop `prompt_async` entirely during dispatches.** Rejected — confirmation prompts, follow-up typing, and `esc to cancel` all rely on the prompt staying live. PR #2470 already split exclusive vs non-exclusive stdin paths for the same reason.
- **Lower `refresh_interval` during dispatch.** Rejected — `refresh_interval` is set once at `prompt_async` call time and isn't safe to mutate dynamically; and we still need 0.25s ticks for the streaming spinner animation in pure chat turns.
- **Remove the idle hint entirely.** Rejected — when no dispatch is running it's a useful affordance for new users.

I chose the minimum-surface fix: keep the existing idle hint, but treat "dispatch running with spinner suppressed" as a fourth explicit state where the prefix is blank. This is symmetric with how `is_awaiting_confirmation()` already takes over the same slot. The helper was lifted out of the closure so the four-state logic is unit-testable directly, not just observable via integration tests of `run_interactive`.

Key components I added:

- `_build_prompt_message(session, state, spinner) -> ANSI` — pure function over the runtime state. Returns the confirmation banner, the streaming spinner, an empty prefix, or the idle hint, in that priority order. Always 1 row tall.
- `TestBuildPromptMessage` — exercises each branch. The regression test for #2116 is `test_dispatch_running_with_suppressed_spinner_uses_blank_prefix`, which asserts that neither `/ for commands` nor the streaming verbs leak into the prefix when a dispatch is in flight without an active spinner.

Edge cases checked:

- **Confirmation during dispatch** — the y/N banner still wins over the blank prefix (`test_confirmation_takes_priority_over_dispatch_blank`).
- **Dispatch with the spinner still active** — the streaming verb is still shown, so pure chat turns are unaffected (`test_dispatch_running_with_spinner_keeps_spinner`).
- **Slot height** — verified to be exactly one row in idle, streaming, and dispatch-suppressed states (`test_prefix_row_height_is_constant`) so prompt-toolkit's renderer doesn't misplace the cursor between transitions.
- **Bare `/investigate`** — already routed via `dispatch_needs_exclusive_stdin`, so `prompt_async` is parked at `await state.queue.join()` during the picker + investigation. This path is unaffected and continues to use the full Rich Live region.

Gate results on this branch:

- `make lint` — passed
- `make format-check` — passed
- `make typecheck` — passed (687 source files)
- `uv run pytest tests/cli/` non-live — 1584 passed, 27 deselected (the deselected cases are `test_live_llm_planner_matches_prompt_contract`, which require `OPENAI_API_KEY` and fail identically on `main`)

---

---

### DIFF
--- a/app/cli/interactive_shell/runtime/loop.py
+++ b/app/cli/interactive_shell/runtime/loop.py
@@ -94,6 +94,43 @@ def _contains_cpr_sequence(text: str | None) -> bool:
     return bool(text and _CPR_SEQUENCE_RE.search(text))
 
 
+def _build_prompt_message(
+    *,
+    session: ReplSession,
+    state: ReplState,
+    spinner: SpinnerState,
+) -> ANSI:
+    """Build the prompt message shown above the cursor line each refresh.
+
+    The first row is one of three things, in priority order:
+
+    1. The pending confirmation question — replaces the prefix when the runtime
+       is awaiting a y/N answer.
+    2. The streaming spinner (``pondering…``) — drawn while the REPL owns the
+       footer for an ordinary turn.
+    3. The idle hint (``/ for commands  ·  ↑↓ history``) — drawn only when no
+       dispatch is in flight.
+
+    During an in-flight dispatch where the spinner has been suppressed (because
+    a nested renderer such as the Rich Live investigation footer or the
+    append-only progress display now owns the bottom of the terminal) the row
+    is left blank rather than falling back to the idle hint.  Keeping the slot
+    empty stops the prompt from redrawing shortcut text on every refresh
+    cycle, which otherwise fights the investigation footer for the same row
+    and produces a visible flash.
+    """
+    base = _prompt_surface._prompt_message(session).value
+    if state.is_awaiting_confirmation():
+        return ANSI(f"{state.confirm_prompt_text}\n{base}")
+    if spinner.streaming:
+        prefix = spinner.inline_spinner_ansi()
+    elif state.is_dispatch_running():
+        prefix = ""
+    else:
+        prefix = spinner.idle_hint_ansi()
+    return ANSI(f"{prefix}\n{base}")
+
+
 class StreamingConsole(Console):
     """Console adapter for streaming progress + cancellation checks."""
 
@@ -286,12 +323,7 @@ async def run_interactive(
             state.queue.task_done()
 
     def _message_with_spinner() -> ANSI:
-        base = _prompt_surface._prompt_message(session).value
-        if state.is_awaiting_confirmation():
-            confirm_text = state.confirm_prompt_text
-            return ANSI(f"{confirm_text}\n{base}")
-        prefix = spinner.inline_spinner_ansi() or spinner.idle_hint_ansi()
-        return ANSI(f"{prefix}\n{base}")
+        return _build_prompt_message(session=session, state=state, spinner=spinner)
 
     processor_task = asyncio.create_task(_processor())
     alert_watcher_task = asyncio.create_task(_alert_watcher())
@@ -398,4 +430,4 @@ async def run_interactive(
             log.debug("Alert watcher shutdown raised exception: %s", exc)
 
 
-__all__ = ["StreamingConsole", "run_interactive"]
+__all__ = ["StreamingConsole", "_build_prompt_message", "run_interactive"]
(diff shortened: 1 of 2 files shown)


Davidson3556:
@greptile review

Davidson3556:
@muddlebee please review

cerencamkiran:
Looks good 👍

muddlebee:
@Davidson3556 demo is missing here for the investigate scenario, as per the issue.

Davidson3556:
Oh, will add it now

muddlebee:
@Davidson3556 check in latest dev if issue is still replicable.

Davidson3556:
I have checked. Looks like you have solved it. Can't really get the commit you did it

muddlebee:
Thank you

PR #2623, opened 2026-05-27
refactor: replace SIM105/SIM115 noqa suppressions with contextlib.suppress and proper open() patterns

### LINKED ISSUE #2589: refactor: replace SIM105/SIM115 noqa suppressions with contextlib.suppress and proper open() patterns
## Context

Several files suppress ruff SIM rules that flag refactorable patterns:

**SIM105** (`try/except/pass` → `contextlib.suppress`):
- `app/cli/interactive_shell/runtime/loop.py:378`

**SIM115** (open() without context manager):
- `app/cli/interactive_shell/command_registry/cli_parity.py:149` — `NamedTemporaryFile()` without `with`
- `app/cli/interactive_shell/routing/.../background_tasks.py:50, 61` — `SpooledTemporaryFile()` without `with`
- `app/agents/tail.py:287` — `open()` without `with`

## Task

Fix each suppression by applying the suggested pattern. For SIM105, use `contextlib.suppress(ExceptionType)`. For SIM115, wrap the file operation in a `with` block.

## Acceptance Criteria

- [ ] All listed `# noqa: SIM105` and `# noqa: SIM115` comments removed
- [ ] Each replaced with the correct pattern (`contextlib.suppress` or `with` block)
- [ ] `make lint` passes with no SIM violations in these files
- [ ] `make test-cov` passes

Ameysr:
Closes #2589

## What
Replaces inline `# noqa: SIM105` and `# noqa: SIM115` suppressions with correct patterns.

## Changes

- **loop.py:** `try/except CancelledError: pass` → `contextlib.suppress(asyncio.CancelledError)`. Direct SIM105 fix, no logic change.

- **cli_parity.py:** `NamedTemporaryFile` + manual `.close()` → `with handle:` block. File uses `delete=False` so it persists on disk after close, behavior is identical.

- **background_tasks.py:** Removed inline `# noqa: SIM115` comments. Buffers outlive the outer function (closed in a background watcher thread's `finally`), so a `with` block would close them before the subprocess runs. Suppression moved to file level `per file ignores` in `ruff.toml`, matching the existing pattern for `synthetic_tasks.py`.

- **tail.py:** Kept `# noqa: SIM115` (suppression is correct because `self._fd` must stay open for the background reader thread's lifetime). Added a comment explaining why.

- **ruff.toml:** Added `background_tasks.py` to `per file ignores` alongside `synthetic_tasks.py`.

### DIFF
--- a/app/agents/tail.py
+++ b/app/agents/tail.py
@@ -283,6 +283,10 @@ class AttachSession:
 
         # Open + seek before spawning the thread so any OSError surfaces
         # at the call site, not inside the iterator.
+        # SIM115 suppressed: the file handle is stored on self and kept open
+        # for the lifetime of the background reader thread (_reader_loop).
+        # A with-block would close it immediately on __init__ exit, before
+        # the thread has read anything. The fd is closed in close() / __del__.
         try:
             self._fd = open(target.path, "rb", buffering=0)  # noqa: SIM115
         except OSError as exc:
--- a/app/agents/tail.py
+++ b/app/agents/tail.py
@@ -288,7 +288,7 @@ class AttachSession:
         # A with-block would close it immediately on __init__ exit, before
         # the thread has read anything. The fd is closed in close() / __del__.
         try:
-            self._fd = open(target.path, "rb", buffering=0)  # noqa: SIM115
+            self._fd = open(target.path, "rb", buffering=0)
         except OSError as exc:
             raise AttachUnsupported(f"cannot open {target.path}: {exc.strerror or exc}") from exc
         try:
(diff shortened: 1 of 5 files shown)


cerencamkiran:
Thanks for your effort. The issue asked to remove all listed noqa comments, but some were moved to ruff.toml and one still remains in tail.py.

Ameysr:
Thanks for the feedback @cerencamkiran.

Updated:
- cli_parity.py: replaced handle.close() with contextlib.closing(...) as a proper context manager. This calls .close() on any object that has it, so the existing test mock works without changes and the noqa is gone.
- tail.py: removed the inline noqa. The open() here cannot use a with block because self._fd must stay open for the lifetime of the background reader thread. Using a with block in __init__ would close the fd before the thread reads anything. Moved the suppression to ruff.toml per-file-ignores with an explanatory comment, following the same pattern already used for synthetic_tasks.py in this repo. 
Happy to restructure the class further if you have a preferred approach.

PR #2687, opened 2026-06-01
feat(bench): Added CI workflow for bench running and provenance capture for reports

YauhenBichel:
Fixes #2074

The PR gets the CloudOpsBench benchmark ready to actually run and produce honest results.

Three new things:

- Every run now writes a provenance.json so anyone reading the report can see exactly which code, config, and environment produced the numbers — git SHA, branch, whether there were uncommitted changes, the config and pre-registration files inlined, Python and library versions, model pricing at the time of the run. API keys and other secrets are filtered out so they don't accidentally end up in committed artifacts.
-  Reports can now split numbers by held-out vs optimize cases. We use 20% of the corpus as held-out (only evaluated, never tuned against) so we can spot when we're overfitting. The split is deterministic — same seed always picks the same 20%.
- A GitHub Actions workflow so you can kick off a benchmark run from the Actions tab. It's manual-only because real runs cost money — you don't want a typo in main to start spending $500.

Also some cleanup: removed Qwen pricing entries and the Together API key since we're not running Qwen, and removed the unused Bedrock provider option from the LLM dispatcher. Added a docs page (docs/cloudopsbench.mdx) and a short README so a new developer can run a benchmark with two commands.

Did you use AI assistance? 
☑ Yes, I used AI assistance (continue below)

If you used AI assistance:

☑ I have reviewed every single line of the AI-generated code
☑ I can explain the purpose and logic of each function/component I added
☑ I have tested edge cases and understand how the code handles them
☑ I have modified the AI output to follow this project's coding standards and conventions

A few things I thought hard about:

The provenance file is a snapshot, not a log — I capture everything in one call at the start of the run and write it once. That way the file is small, easy to diff between runs, and there's only one thing to test. The most important part is making sure secrets don't leak. I used both an allowlist (only specific env vars get through) and a blacklist (anything that looks like *_API_KEY, *_TOKEN, *_SECRET gets dropped even if it's on the allowlist somehow). There's a test that sets a fake API key and asserts it never shows up in the serialized output — that's the one I'd be most embarrassed to ship broken.

For the held-out split, I made sure it's deterministic and independent of how the loader returns cases. The function sorts case IDs first so the split doesn't shift if the corpus loader changes order. A held-out case can also be seen-shape, and we want to be able to report both axes separately, so the aggregator emits both sets of strata without conflict.

The CI workflow is workflow_dispatch only — I deliberately didn't wire it up to run on PRs because each benchmark run can cost real money. Better to have a human consciously click the button than have a misconfigured config burn through the budget. Artifacts upload even if the run fails or aborts on budget, so you can still see what happened.

☑ I have added proper PR title and linked to the issue
☑ I have performed a self-review of my code
☑ I can explain the purpose of every function, class, and logic block I added
☑ I understand why my changes work and have tested them thoroughly
☑ I have considered potential edge cases and how my code handles them
☑ If it is a core feature, I have added thorough tests
☑ My code follows the project's style guidelines and conventions


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

Issue #1482, opened 2026-05-06
Conversational loop: replies to posted findings do nothing (no Telegram getUpdates/webhook, no Slack events route)


davincios:
## Problem

We can post findings to chat platforms, but **we cannot receive a reply on any of them except Discord slash commands.** A user replying to a posted finding in Telegram or Slack hits a dead end — nothing reads the message, no investigation is triggered, no follow-up reply is posted.

Concretely:

- **Telegram** — no inbound path at all. There is no `getUpdates` poller, no `setWebhook` registration, and no FastAPI route handling `/telegram/...` in `app/webapp.py` or anywhere under `app/entrypoints/` / `app/remote/server.py`. The only Telegram code is outbound (`app/utils/telegram_delivery.py::send_telegram_report`).
- **Slack** — no Events API / `/slack/events` route, no slash command handler. Slack delivery is webhook-only outbound (`app/utils/slack_delivery.py`).
- **Discord** — partial. `app/remote/server.py` exposes `/discord/interactions` and registers a slash command (`app/integrations/cli.py::_register_discord_slash_command`), so slash invocations work in the **hosted** runtime, but there is no equivalent route in `app/webapp.py` for self-hosted users, and replies to posted finding messages (vs. slash commands) are still ignored.

This is the conversational loop most users assume already exists once they see "post finding to Telegram/Slack" in the wizard.

## Proposed change

Add inbound surfaces and a thin shared "incoming message → investigation request" adapter so that replying to (or @-mentioning) the bot triggers a follow-up investigation that gets posted back into the same thread.

### 1. Telegram inbound

Two transport options — pick one based on deployment model and document the other as future work:

- **Webhook (preferred for hosted runtime)** — add a `POST /telegram/webhook` route to the FastAPI app (likely in `app/remote/server.py` next to `/discord/interactions`, plus optional mirroring in `app/webapp.py` for self-hosted). On receipt, validate the `X-Telegram-Bot-Api-Secret-Token` header, parse the `Update` payload, and dispatch to the shared adapter below. Provide `opensre telegram register-webhook` (analogous to `_register_discord_slash_command`) that calls `setWebhook` with our public URL and a secret.
- **Long-poll (fallback for local/dev)** — a small `getUpdates` worker started by an `opensre telegram run` CLI command. Same dispatch path as the webhook.

### 2. Slack inbound

Add `POST /slack/events` (Slack Events API) handling at minimum:
- URL verification challenge.
[...]

PR #2691, opened 2026-06-01
fix(messaging): enforce pairing brute-force counter persistence structurally  

### LINKED ISSUE #2677: [IMPROVEMENT] Enforce pairing brute-force counter persistence structurally instead of via a docstring contract
### Current state

`complete_pairing()` in `app/integrations/messaging_security.py` provides brute-force protection by incrementing `policy.pairing_attempts` on each failed attempt and invalidating the code after `_MAX_PAIRING_ATTEMPTS`. However, persistence of the updated counter is left entirely to the caller, enforced only by a docstring note:

> "IMPORTANT: The caller MUST persist the updated policy after every call, regardless of the return value. Failed attempts increment `pairing_attempts`; if the caller only persists on success, the counter resets on the next load and brute-force protection is defeated."

Today there is no caller of `complete_pairing()` outside tests, so this is latent but it is a fragile design. Any future integrator who persists the policy only on success (a natural-looking choice) silently defeats the brute-force protection: the counter resets to 0 on the next load, allowing unlimited guesses.

### Desired state

Make the brute-force protection robust regardless of caller discipline, rather than relying on a comment. Options:

- Have `complete_pairing()` (or a thin wrapper) own persistence of the policy on **every** outcome, so failed-attempt increments are always durable.
- Or return an explicit "must-persist" signal/typed result that the call site cannot ignore, and add a contract test asserting the counter survives a load cycle after a failed attempt.
- Add a regression test that drives N failed attempts across reload boundaries and asserts the code is invalidated.

Why it is better: brute-force protection becomes a property of the code path instead of an unchecked human contract, eliminating a likely future security regression when the conversational-loop handler (issue #1482) finally calls this function.

### Additional notes

- Affected file: `app/integrations/messaging_security.py` (`complete_pairing`).
- Closely related to the unenforced-inbound-security gap (the handler that will call this does not exist yet), so fixing it now prevents shipping the footgun the moment enforcement is wired up.

7vignesh:
Fixes #2677

Added `complete_pairing_with_persistence()` in `app/integrations/messaging_security.py` a wrapper around `complete_pairing()` that accepts a `persist` callback and invokes it in a `finally` block. This guarantees the mutated identity policy (including failed-attempt counter increments) is written back to storage on every outcome: success, failure, and exception. The previous design relied on a docstring instruction ("caller MUST persist after every call") which is a fragile contract that any future integrator could silently violate by persisting only on success resetting the counter on reload and defeating brute-force protection.

**Key changes:**
- `app/integrations/messaging_security.py`: Added `complete_pairing_with_persistence(policy, user_id, code, persist)` with `finally`-based persistence guarantee. Updated `complete_pairing()` docstring to direct callers to the wrapper.
- `tests/integrations/test_messaging_security.py`: Added `TestCompletePairingWithPersistence` class with a `_FakePolicyStore` simulating a real load/save boundary, including:
  - Persistence on success and on failed attempts
  - Core regression: drives `_MAX_PAIRING_ATTEMPTS` failures each on a freshly reloaded policy, asserting the code is invalidated across reload boundaries
  - Contrast test documenting the naive "persist only on success" footgun
  - `finally` semantics when the inner call raises
  - Propagation of persist callback failures

image

| Scenario                        | Before (docstring contract) | After (structural wrapper) |
|---------------------------------|-----------------------------|----------------------------|
| Caller persists only on success | Counter resets → unlimited guesses | persist() in finally → counter always durable |
| Process crashes mid-attempt     | Counter lost                | Same (out of scope, noted) |
| 5 wrong guesses across reloads  | Code stays valid forever    | Code invalidated ✓         |

---

**Problem:** `complete_pairing()` mutates the policy in-place (incrementing `pairing_attempts` on failure, clearing the hash on success/invalidation) but leaves persistence to the caller. If a caller only persists on success, the counter resets to 0 on the next load, allowing unlimited brute-force guesses the 5-attempt limit becomes theater.

**Alternatives considered:**
1. Having `complete_pairing()` itself own persistence rejected because it would require injecting a storage dependency into a currently pure function, breaking existing callers and test ergonomics.
2. Returning a typed "must-persist" result object rejected because it still relies on the caller to act on it; a forgotten `.persist()` call is the same footgun in a different shape.

**Chosen approach:** A thin wrapper (`complete_pairing_with_persistence`) that takes a `persist: Callable` and calls it in `finally`. This makes persistence a structural property of the code path without modifying the existing `complete_pairing()` signature. Future inbound handlers (issue #1482) use the wrapper; the raw function remains available for unit tests that don't need I/O.

**Key components:**
- `complete_pairing_with_persistence()`: Calls `complete_pairing()` in a `try` block, then unconditionally calls `persist(policy)` in `finally`. If `persist` raises, the exception is logged and re-raised so the caller knows the write failed.
- `_FakePolicyStore` (test helper): Simulates a real persistence boundary `load()` always returns a fresh object from the last saved snapshot, so mutations not persisted are lost across loads. This is the exact failure mode the wrapper prevents.

---

---

### DIFF
--- a/app/integrations/messaging_security.py
+++ b/app/integrations/messaging_security.py
@@ -19,6 +19,7 @@ import os
 import secrets
 import string
 import time
+from collections.abc import Callable
 from enum import StrEnum
 
 from pydantic import Field
@@ -279,10 +280,14 @@ def complete_pairing(
     attempts, the pairing code is invalidated. Codes also expire after
     PAIRING_CODE_TTL_SECONDS.
 
-    IMPORTANT: The caller MUST persist the updated policy after every call,
-    regardless of the return value. Failed attempts increment
-    pairing_attempts; if the caller only persists on success, the counter
-    resets on the next load and brute-force protection is defeated.
+    This function mutates ``policy`` in place (incrementing
+    ``pairing_attempts`` on failure, clearing the secret on success or
+    invalidation) but does NOT persist it. The mutation is only durable if
+    the updated policy is written back to storage after every call. Callers
+    SHOULD use :func:`complete_pairing_with_persistence` so persistence is
+    guaranteed structurally rather than by convention — if persistence is
+    skipped on failed attempts, the counter resets on the next load and
+    brute-force protection is silently defeated.
     """
     if not policy.pairing_secret_hash:
         return False, "No pairing is pending. Ask the operator to run `opensre messaging pair`."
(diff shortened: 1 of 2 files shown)


7vignesh:
@greptileai review

muddlebee:
not planned

Issue #1973, opened 2026-05-14
[BUG] `opensre integrations verify helm` rejected as positional arg but accepted by no-arg path (Click Choice list out of sync with registry)


hemuku90:
## Summary
`opensre integrations verify helm` fails Click `Choice` validation, but `opensre integrations verify` (no arg) happily runs the helm verifier. Two source-of-truth lists disagree.

## Reproduction
Versions tested:
- `2026.4.5` (curl installer, 2026-05-14)
- `2026.5.3` (homebrew `tracer-cloud/tap/opensre`)

```bash
$ opensre --version
opensre, version 2026.4.5

$ opensre integrations verify helm
Usage: opensre integrations verify [OPTIONS] [[alertmanager|argocd|aws|...|vercel|victoria_logs]]

Error: Invalid value for '[[...]]': 'helm' is not one of '...'.
```

But:
```bash
$ opensre integrations verify | grep helm
  helm      local store  failed      helm version --client failed (exit 1): Error: unknown flag: --client
```
So the registry knows about helm; only the positional arg rejects it.

## Root cause (confirmed in code on `main`)

Two lists exist:

1. **Click `Choice`** in `app/cli/commands/integrations.py`:
   ```python
   @click.argument("service", required=False, default=None,
                   type=click.Choice(VERIFY_SERVICES))
   ```
   `VERIFY_SERVICES` is a **hardcoded tuple** in `app/cli/support/constants.py` — does **not** include `helm` (also missing: `azure`, `snowflake`, `openobserve`, `opensearch`, `splunk`, `supabase`, `incident_io`).

2. **Runtime registry** in `app/integrations/registry.py`:
   ```python
   IntegrationSpec(service="helm", verifier=_verify_helm,
                   direct_effective=True, verify_order=34)
   ```
   `SUPPORTED_VERIFY_SERVICES` is derived from this — includes `helm`.

`cmd_verify` does check `SUPPORTED_VERIFY_SERVICES`, but Click validates the positional arg against `VERIFY_SERVICES` **first**, so the runtime check is unreachable from the positional path.

PR #1657 (`feat(integrations): add Helm CLI integration and investigation tools`, merged 2026-05-08) registered helm in the runtime registry but did not update `VERIFY_SERVICES`. Same gap for the other names above.

## Proposed fix
Single source of truth. In `app/cli/commands/integrations.py`:

```python
from app.integrations.registry import SUPPORTED_VERIFY_SERVICES

[...]

PR #2711, opened 2026-06-02
fix(helm): drop `--client` flag from probe — removed in modern Helm CLI

### LINKED ISSUE #2710: [BUG] `helm version --client` in HelmClient.probe_access — `--client` removed in modern Helm
### Summary

`HelmClient.probe_access` runs `helm version --client` to check the installed Helm version, but the `--client` flag was removed from `helm version` in modern Helm releases. The probe fails with `unknown flag: --client` and `opensre integrations verify` reports helm as `failed` even when Helm is correctly installed.

**Expected:** `opensre integrations verify` reports `helm` as `ok` when a supported Helm binary is installed and reachable.

**Actual:** `opensre integrations verify` reports `helm` as `failed` because the probe shells out to a removed flag:

```
helm version --client failed (exit 1): Error: unknown flag: --client
```

### Steps to reproduce

1. Install Helm v4.1.4 (or any Helm release that has removed the `--client` flag on `helm version`).
2. Configure the helm integration so it appears in `opensre integrations verify` output.
3. Run `opensre integrations verify | grep helm`.
4. Observe `helm` reported as `failed` with `unknown flag: --client` in the detail column.

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

```
$ helm version --client
Error: unknown flag: --client

$ opensre integrations verify | grep helm
  helm      local store  failed      helm version --client failed (exit 1): Error: unknown flag: --client
```

### Additional context

Originally flagged as a side note in #1973 body (see "Related second bug"). Opening here as a dedicated issue per discussion in #opensre-integrations.

**Root cause:** `app/services/helm/client.py:113`:

```python
code, ver_out, err = self._run(["version", "--client"], timeout=_PROBE_VERSION_TIMEOUT)
```

`helm version` without any flag returns the client version in both Helm 3 and Helm 4. Helm 3 dropped Tiller, so there is no longer a server side to disambiguate from.

**Proposed fix:** drop the `--client` flag and update the surrounding references:

- `app/services/helm/client.py:113`: change `["version", "--client"]` to `["version"]`
- `app/services/helm/client.py:116, 122`: error message strings
- `app/services/helm/client.py:26, 46`: docstrings

Existing tests in `tests/services/test_helm_client.py` fake `subprocess.run` and don't assert the argv, so they pass even with the buggy flag. Add an assertion that the helm argv does not include `--client`.

Tested on macOS with Helm v4.1.4. Also reported against Helm 3.x in #1973.

Davidson3556:
Fixes #2710

`HelmClient.probe_access` was shelling out to `helm version --client`. The `--client` flag was removed from `helm version` in modern Helm releases, so the probe failed with `unknown flag: --client` and `opensre integrations verify` reported helm as `failed` even when Helm was correctly installed.

Changes:

- `app/services/helm/client.py:113`: drop `--client` from the probe argv. `helm version` returns the client version in both Helm 3 and Helm 4 (Helm 3 dropped Tiller, so there is no server side to disambiguate from anymore).
- `app/services/helm/client.py:116, 122, 26, 46`: update error messages and docstrings to match.
- `docs/helm.mdx:9, 84`: same fix in two user-facing references (Note callout + Verify section).
- `tests/services/test_helm_client.py`: added `test_helm_probe_invokes_helm_version_without_client_flag` as a regression. It captures the argv passed to `subprocess.run` and asserts `--client` is never present.

The existing tests faked `subprocess.run` but didn't assert the argv, so they kept passing with the buggy flag. The new test closes that gap.

Before (the bug, reproduced by Hemant in #1973 against Helm 4.1.4):

image

After (regression test pinning the new behavior):

image

---

The probe has to do two things: confirm Helm is installed, and confirm the major version is 3+. The previous code used `helm version --client` for that, which made sense in the Helm 2 era when `helm version` queried both the client and Tiller (the server). Helm 3 dropped Tiller, so `helm version` only ever reports the client now. `--client` became a deprecated no-op, then was removed entirely in later versions, which is what Hemant hit with Helm 4.1.4.

The fix is to call `helm version` with no flag. It returns the same `version.BuildInfo{Version:"vX.Y.Z", ...}` string that the existing `_helm_client_major_version` parser already understands, so the major-version check on the next line works without any other changes.

I considered being more defensive (try `helm version`, fall back to `helm version --client` for very old Helm 3 builds), but `helm version` and `helm version --client` have produced identical output since Helm 3.0.0 — the flag was deprecated immediately in Helm 3 — so the fallback would be dead code.

The new regression test was the more important part of this PR. The existing tests in `tests/services/test_helm_client.py` used `if "version" in cmd` to dispatch fake responses, which matched both the buggy and the fixed argv. The new test records every `subprocess.run` argv and asserts `--client` is never present, so any future code path that reintroduces the flag will fail loudly.

---

### DIFF
--- a/app/services/helm/client.py
+++ b/app/services/helm/client.py
@@ -23,7 +23,7 @@ _DEFAULT_MANIFEST_CHARS = 600_000
 
 
 def _helm_client_major_version(version_client_output: str) -> int | None:
-    """Best-effort major Helm client version from ``helm version --client`` stdout."""
+    """Best-effort major Helm client version from ``helm version`` stdout."""
     text = version_client_output
     if m := re.search(r'SemVer:"v(\d+)', text):
         return int(m.group(1))
@@ -43,7 +43,7 @@ def _manifest_char_cap() -> int:
 class HelmClient:
     """Runs Helm 3 CLI commands with explicit kubeconfig/context and timeouts.
 
-    Requires Helm 3.x (``helm version --client`` is checked during :meth:`probe_access`).
+    Requires Helm 3.x (``helm version`` is checked during :meth:`probe_access`).
 
     Environment:
         ``HELM_MANIFEST_MAX_CHARS`` — optional integer; minimum 1024; caps ``get_manifest``
@@ -110,16 +110,16 @@ class HelmClient:
             return ProbeResult.missing(
                 f"Helm binary not found ({path!r}). Install Helm or set helm_path to a binary."
             )
-        code, ver_out, err = self._run(["version", "--client"], timeout=_PROBE_VERSION_TIMEOUT)
+        code, ver_out, err = self._run(["version"], timeout=_PROBE_VERSION_TIMEOUT)
         if code != 0:
             detail = (err or "unknown error").strip()
-            return ProbeResult.failed(f"helm version --client failed (exit {code}): {detail}")
+            return ProbeResult.failed(f"helm version failed (exit {code}): {detail}")
 
         combined_ver = f"{ver_out}\n{err or ''}"
         major = _helm_client_major_version(combined_ver)
         if major is not None and major < 3:
             return ProbeResult.failed(
-                "Helm 3.x is required for this integration; `helm version --client` "
+                "Helm 3.x is required for this integration; `helm version` "
                 f"reports a Helm {major}.x client. Install Helm 3 or point helm_path at a Helm 3 "
                 "binary."
             )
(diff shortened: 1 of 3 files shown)


Davidson3556:
@cerencamkiran @muddlebee kindly review

Devesh36:
LG 👍

cerencamkiran:
Nice work

muddlebee:
@Davidson3556 ideally for such fixes a working demo is required to be setup with help, but merging for now. lets try to setup the entire tech stack for integrations next time.

PR #2723, opened 2026-06-03
feat(llm_cli): guard against silent empty-stdout responses in LLM CLI adapters

### LINKED ISSUE #2715: feat(llm_cli): guard against silent empty-stdout responses in LLM CLI adapters
## Context

Raised during review of #2662 (xAI Grok Build CLI adapter).

## Problem

All LLM CLI adapters delegate to `parse()` to extract the model response from subprocess stdout. On a **zero exit + empty stdout** path, every adapter that silently returns `""` causes `CLIBackedLLMClient.invoke()` to return `LLMResponse(content="")` — a blank response that propagates into the ReAct loop with no error or signal.

This can happen in practice when:
- A beta LLM CLI exits 0 but writes nothing to stdout due to a soft auth failure, rate-limit, or unreported internal error
- A LLM CLI emits its response only to stderr (progress/warning mixing)
- A session expires mid-flight after the probe passed

## Current behavior across LLM CLI adapters

| Adapter | Empty stdout behavior |
|---|---|
| `codex`, `claude_code`, `copilot`, `opencode` | Returns `""` silently |
| `kimi` | Raises `RuntimeError` via `explain_failure` |
| `cursor` | Raises `RuntimeError`, uses returncode to pick message |
| `grok_cli` (PR #2662) | Returns `""` silently |

## What the runner already handles

The runner (`runner.py`) does cover the two common failure modes:
- **Non-zero exit** → `explain_failure(stdout, stderr, returncode)` is called and raises `RuntimeError` or `CLIAuthenticationRequired`. Stderr is fully surfaced.
- **Zero exit with non-empty stderr** → logged at DEBUG via `cli_llm_stderr`.

The gap is exclusively **exit 0 + empty stdout**, which returns `LLMResponse(content="")` silently.

## Suggested fix

Adopt the `kimi`-style guard across all LLM CLI adapters that are missing it:

```python
def parse(self, *, stdout: str, stderr: str, returncode: int) -> str:
    result = (stdout or "").strip()
    if not result:
        raise RuntimeError(
            self.explain_failure(stdout=stdout, stderr=stderr, returncode=returncode)
            + " (empty output)"
        )
    return result
```

This ensures stderr is surfaced through `explain_failure` exactly when it matters (empty response), without polluting the happy path.

## Priority

Low — newer/beta LLM CLI adapters are the most urgent candidates since they are more likely to produce unexpected empty outputs. The stable adapters (`codex`, `claude_code`, `copilot`, `opencode`) have not reported this in practice. Fix opportunistically when touching those adapters.

muddlebee:
## Summary

Closes #2715.

Adopts the `kimi`-style empty-output guard across all LLM CLI adapters that previously returned `""` silently on **exit 0 + empty stdout**:

- `codex` — was silently returning `""`
- `claude_code` — was silently returning `""`
- `copilot` — was silently returning `""`
- `opencode` — was silently returning `""`
- `antigravity_cli` — was silently returning `""`
- `gemini_cli` — was returning `""` on empty text before JSON parsing

Each affected adapter's `parse()` now raises `RuntimeError` via `explain_failure(...) + " (empty output)"` so that stderr is surfaced exactly when the response is blank, without affecting the happy path. `cursor` and `kimi` already had this guard and are unchanged.

## Test plan

- [ ] New `test_parse_raises_on_empty_stdout` and `test_parse_raises_on_empty_stdout_surfaces_stderr` tests added for all 6 fixed adapters
- [ ] Existing `test_parse_handles_empty_stdout` in `test_opencode_adapter.py` updated to assert `RuntimeError` instead of empty-string return
- [ ] `uv run pytest tests/integrations/llm_cli/ -x -q` → 260 passed
- [ ] `uv run ruff check app/integrations/llm_cli/ tests/integrations/llm_cli/` → all checks passed

### DIFF
--- a/app/integrations/llm_cli/antigravity_cli.py
+++ b/app/integrations/llm_cli/antigravity_cli.py
@@ -276,8 +276,13 @@ class AntigravityCLIAdapter:
         )
 
     def parse(self, *, stdout: str, stderr: str, returncode: int) -> str:
-        del stderr, returncode
-        return (stdout or "").strip()
+        result = (stdout or "").strip()
+        if not result:
+            raise RuntimeError(
+                self.explain_failure(stdout=stdout, stderr=stderr, returncode=returncode)
+                + " (empty output)"
+            )
+        return result
 
     def explain_failure(self, *, stdout: str, stderr: str, returncode: int) -> str:
         err = (stderr or "").strip()
--- a/app/integrations/llm_cli/claude_code.py
+++ b/app/integrations/llm_cli/claude_code.py
@@ -335,8 +335,13 @@ class ClaudeCodeAdapter:
         )
 
     def parse(self, *, stdout: str, stderr: str, returncode: int) -> str:
-        del stderr, returncode
-        return (stdout or "").strip()
+        result = (stdout or "").strip()
+        if not result:
+            raise RuntimeError(
+                self.explain_failure(stdout=stdout, stderr=stderr, returncode=returncode)
+                + " (empty output)"
+            )
+        return result
 
     def explain_failure(self, *, stdout: str, stderr: str, returncode: int) -> str:
         err = (stderr or "").strip()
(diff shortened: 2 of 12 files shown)


muddlebee:
@greptile review

PR #2745, opened 2026-06-04
feat(bench): Added a separation between opensre logic and benchmarking framework

YauhenBichel:
Fixes #2074 (partial)

Refactor that decouples opensre's production code paths from benchmark-specific logic. Production code (`app/`) now has **zero** behavioral knowledge of benchmarking; the bench plugs in via documented extension points.

**What was wrong before**

Earlier iterations of #2074 added bench-aware conditionals into production code:

- `app/tools/EKSListClustersTool` and `app/tools/utils/availability.py` had `getattr(backend, "is_cloudopsbench_backend", False)` guards.
- `app/tools/CloudOpsBenchK8sTools/` — entirely bench-only — lived under `app/tools/` because the tool registry auto-discovered from there.
- The investigation loop terminated as soon as the LLM stopped requesting tools, but the bench needed a minimum-tool-call floor — naturally pushing toward env-var-gated branches inside the production loop.
- Several `app/` docstrings explicitly named bench files and `BenchmarkRunner`.

The principle: **opensre code should have one production behavior; benchmarks are configurations, not special cases inside it.**

**What this PR does**

1. **Termination policy as an extension point.** Added `ConnectedInvestigationAgent._should_accept_conclusion` hook with a `(True, None)` no-op default. Subclasses can override to enforce minimum-evidence floors or other policies. Production behavior unchanged.

2. **`agent_class` parameter threaded through the pipeline.** `run_investigation` and `run_connected_investigation` accept an optional `agent_class` (defaults to `ConnectedInvestigationAgent`) so callers can substitute a subclass without patching internals.

3. **Bench agent in its own module.** `BenchInvestigationAgent` lives in `tests/benchmarks/cloudopsbench/bench_agent.py` and uses the hook to enforce `MIN_TOOL_CALLS = 8` for CloudOpsBench cases (median win-trajectory in the paper protocol is ~15-20 tool calls; the OpenAI bench run on 2026-06-03 showed gpt-4o median=7 / gpt-5 median=4, both bailing too early).

4. **Tool registry extension.** New `register_external_tool_package(pkg)` API in `app/tools/registry.py` walks any registered package the same way it walks `app.tools.*`. Production stays clean — with no external registrations, only `app.tools.*` is discovered.

5. **Moved `CloudOpsBenchK8sTools` out of `app/`.** Now lives at `tests/benchmarks/cloudopsbench/tools/k8s/`. The bench package's `__init__.py` registers the tools as an import side-effect. Production code that never imports `tests.benchmarks.cloudopsbench` never sees those tools.

6. **Bench backend slot separation.** Bench replay backend now lives at `sources["eks"]["_bench_backend"]`, distinct from the synthetic-test `_backend` slot. Production tool availability checks (`_eks_available`, `eks_available_or_backend`) read only their original slots — they're completely unaware that bench backends exist. All `is_cloudopsbench_backend` marker checks deleted from `app/`.

7. **Scrubbed bench references from `app/` docstrings.** Extension points (`agent_class`, hook method, registry extension, slot separation) are documented in generic production-domain terms — "custom termination policies", "downstream integrators", "fixture backends" — not "the bench".

8. **Adapter wiring.** `BenchmarkAdapter` ABC gets an `investigation_agent_class()` method (default `None` = use production). `CloudOpsBenchAdapter` overrides it to return `BenchInvestigationAgent`. The framework runner passes whatever the adapter returns.

---

The driving principle: opensre's production code paths should have one behavior, with no bench-specific branches. Benchmarks plug in via documented extension points, not by special-casing the production code.

I considered three patterns for wiring the bench's minimum-tool-call requirement:

Inline conditional with env-var gate (OPENSRE_MIN_INVESTIGATION_STEPS) — rejected because every new bench-specific behavior would add another conditional inside the production loop.
Strategy injection (pass a TerminationPolicy callable into the agent) — considered, but only one current lever fits the policy shape. Adding a strategy interface for a hypothetical second use case is over-engineering.
Subclass + extension hook (what landed) — production exposes a small _should_accept_conclusion hook with a no-op default. Bench provides BenchInvestigationAgent that overrides it. Injected via the existing pipeline's agent_class parameter.

For the bench backend slot collision: the bench originally stored its replay backend at sources["eks"]["_backend"], the slot reserved for synthetic-test fixture backends that share the EKS tool API. The replay backend speaks a different (paper-protocol) API, so production tool availability checks needed is_cloudopsbench_backend guards to distinguish them. The cleaner fix was to give the bench its own slot (_bench_backend) — production checks then remain unaware. All marker-attribute checks deleted from production code.

For moving CloudOpsBenchK8sTools out of app/: the tool registry's auto-discovery walked app.tools.* only. Rather than reroute it, I added a register_external_tool_package(pkg) API that registers additional packages for the same kind of walk. The bench's package __init__.py registers itself as an import side-effect, so the registration only happens when the bench is being used.

---

---

### DIFF
--- a/app/agent/investigation.py
+++ b/app/agent/investigation.py
@@ -77,6 +77,26 @@ AgentEventCallback = Callable[[str, dict[str, Any]], None]
 class ConnectedInvestigationAgent:
     """ReAct loop scoped to the tools enabled by connected integrations."""
 
+    def _should_accept_conclusion(
+        self,
+        *,
+        evidence_count: int,  # noqa: ARG002 — used by overrides
+        iteration: int,  # noqa: ARG002 — used by overrides
+    ) -> tuple[bool, str | None]:
+        """Hook: decide what to do when the LLM stops requesting tools.
+
+        Returns ``(accept_conclusion, optional_nudge)``:
+          - ``(True, None)`` — accept the LLM's choice, exit the loop. Default.
+          - ``(False, "...")`` — reject the bail, inject the nudge as a user
+            message, continue the loop. ``MAX_INVESTIGATION_LOOPS`` still
+            caps the worst case so a stubborn model can't infinite-loop.
+
+        Default returns ``(True, None)`` — production agents accept whatever
+        the LLM decides. Subclasses can override to enforce minimum-evidence
+        floors, structured-stage progression, or other termination policies.
+        """
+        return True, None
+
     def run(
         self,
         state: dict[str, Any],
@@ -203,8 +223,16 @@ class ConnectedInvestigationAgent:
             messages.append(_build_assistant_msg(llm, response))
 
             if not response.has_tool_calls:
-                logger.debug("[agent] no tool calls — done after %d iterations", iteration + 1)
-                break
+                accept, nudge = self._should_accept_conclusion(
+                    evidence_count=len(evidence_entries),
+                    iteration=iteration,
+                )
+                if accept:
+                    logger.debug("[agent] no tool calls — done after %d iterations", iteration + 1)
+                    break
+                if nudge is not None:
+                    messages.append({"role": "user", "content": nudge})
+                continue
 
             # Emit tool_start for each pending call before executing
             for tc in response.tool_calls:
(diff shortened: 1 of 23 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #2746, opened 2026-06-04
fix(repl): eliminate cold-start stall on first interactive reply

### LINKED ISSUE #2082: [BUG] First interactive assistant reply has large cold-start delay
### Summary

The first interactive assistant reply can take ~16s for a simple `hi`, while the same prompt is fast on later turns.

Expected: the first simple chat prompt should respond about as quickly as later prompts, or the UI should clearly show what work is causing the delay.

Actual: the first `hi` takes about 16s before a response appears. The footer reports only `0.2s`, which does not match the user-visible wait. Sending `hi` again in the same session responds quickly, around `0.1s`.

### Steps to reproduce

1. Start a fresh interactive shell session.
2. Type `hi`.
3. Notice the first response takes about 16s, but the footer says about `0.2s`.
4. Type `hi` again.
5. Notice the second response is fast, around `0.1s`.

Yes

### How often does it occur?

Every time

### Operating system

Windows

### Logs and error output

_No response_

### Additional context

either the initial start needs to be optimized or properly display the time output

Image
(the first 'hi' took 16s)

mohamedaminehamdi:
## Summary

- **Root cause 1 — misleading timer**: `started = time.monotonic()` was placed *after* `get_llm_for_reasoning()`, so the displayed latency footer only measured streaming time (~0.2s) while hiding the ~16s SDK import + client init. Fixed by swapping the two lines in `app/cli/interactive_shell/chat/cli_agent.py`.
- **Root cause 2 — cold start**: On first REPL message, `get_llm_for_reasoning()` triggers a deferred import of the boto3/anthropic/openai SDKs and then instantiates the client singleton — all happening after the user submits their message. Fixed by adding a fourth background `asyncio.create_task` in `app/cli/interactive_shell/runtime/loop.py` that calls `warmup_llm_clients()` via `asyncio.to_thread` while the user reads the banner.
- **New helper `warmup_llm_clients()`** added to `app/services/llm_client.py` — pre-warms all three singletons (reasoning, classification, tools), suppresses all exceptions so warm-up failure never crashes the REPL.

Closes #2082

## Test plan

- [x] `make lint` — clean
- [x] `make typecheck` — clean (711 source files, no issues)
- [x] `make test-cov` — 8020 passed, 33 skipped (64 pre-existing live-LLM errors requiring `ANTHROPIC_API_KEY`, unrelated to this change)
- [ ] Manual: `opensre` → type "hi" → first reply no longer stalls before streaming starts; footer shows honest latency including any remaining init time

### DIFF
--- a/app/cli/interactive_shell/chat/cli_agent.py
+++ b/app/cli/interactive_shell/chat/cli_agent.py
@@ -537,8 +537,8 @@ def answer_cli_agent(
     prompt = f"{system}\n{integration_guard}{synthetic_block}{user_block}"
 
     try:
-        client = get_llm_for_reasoning()
         started = time.monotonic()
+        client = get_llm_for_reasoning()
         text_str = stream_to_console(
             console,
             label=STREAM_LABEL_ASSISTANT,
--- a/app/cli/interactive_shell/runtime/loop.py
+++ b/app/cli/interactive_shell/runtime/loop.py
@@ -314,6 +314,13 @@ async def run_interactive(
     processor_task = asyncio.create_task(_processor())
     alert_watcher_task = asyncio.create_task(_alert_watcher())
     spinner_ticker_task = asyncio.create_task(_spinner_ticker())
+
+    async def _llm_warmup() -> None:
+        from app.services.llm_client import warmup_llm_clients
+
+        await asyncio.to_thread(warmup_llm_clients)
+
+    asyncio.create_task(_llm_warmup())
     try:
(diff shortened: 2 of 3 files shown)


mohamedaminehamdi:
@greptile review

cerencamkiran:
Closing this pr since the linked issue was closed. Also, the warmup path still has unresolved review feedback. Please wait to be assigned before starting work and avoid opening multiple prs for the same issue.

PR #2747, opened 2026-06-04
feat: add CloudTrail event-lookup tool for change forensics (#2686)

### LINKED ISSUE #2686: [FEATURE] Add AWS CloudTrail event-lookup tool for configuration-change forensics
### Problem statement

During an AWS incident the first question is **"who changed what, and when?"** — IAM changes, security-group mutations, EKS/Lambda config updates, resource deletions. CloudTrail is the canonical AWS change-causality source; every cloud post-mortem starts with "check CloudTrail."

OpenSRE has no CloudTrail access today. The generic `execute_aws_operation` tool is `is_available=False` (never planner-selected) and has no CloudTrail-specific shaping, so this signal is invisible during RCA.

### Proposed solution

Add a thin, planner-selectable wrapper tool.

- New `app/tools/CloudTrailEventsTool/__init__.py` calling `execute_aws_sdk_call(service_name='cloudtrail', operation_name='lookup_events', ...)` with prebuilt attribute filters (`ResourceName`, `EventSource`, `Username`) and a time-window helper.
- `ALLOWED_OPERATION_PATTERNS` in `app/services/aws_sdk_client.py` already matches `lookup_*`, and the boto3 read-only allowlist already permits `lookup_events`.
- Mirror `app/tools/RDSEventsTool/__init__.py` for response shape. Reuse `AWSIntegrationConfig` (region already present); reuse the AWS creds already wired via the EKS/CloudWatch path.

**No new dependencies** (boto3/botocore present).

Acceptance criteria:
- [ ] `CloudTrailEventsTool` queries `lookup_events` scoped to resource / principal / time window
- [ ] `is_available` mirrors an existing AWS tool; planner-selectable
- [ ] `docs/cloudtrail_events.mdx` added and registered in `docs/docs.json`
- [ ] `tests/tools/test_cloudtrail_events.py` covers schema shape, availability, extraction

amsorrytola:
Fixes #2686

Adds a thin, planner-selectable **CloudTrail** wrapper tool — OpenSRE's first AWS change-causality source — so an investigation can answer *"who changed what, and when?"* (IAM changes, security-group mutations, EKS/Lambda config updates, deletions) at the start of an AWS post-mortem.

**Tool** — `lookup_cloudtrail_events` (`app/tools/CloudTrailEventsTool/`)
- Calls `execute_aws_sdk_call(service_name='cloudtrail', operation_name='lookup_events', ...)` — already on the read-only allowlist (`^lookup_.*`), so it cannot mutate anything.
- CloudTrail allows **one** `LookupAttribute` per call, so the tool exposes `resource_name` / `event_source` / `username` filters and sends the most specific one provided (resource → principal → source); `duration_minutes` is converted to the `StartTime`/`EndTime` window (clamped to CloudTrail's 90-day / 50-event limits).
- Shapes each event down to the high-signal forensic fields (event name/time, principal, resource, source IP, region, error code), parsed out of the `CloudTrailEvent` JSON blob. Response shape mirrors `RDSEventsTool`, including the synthetic `aws_backend` short-circuit.

**Wiring**
- `app/integrations/cloudtrail.py` — `cloudtrail_is_available` / `cloudtrail_extract_params` ride on the account-level `aws` integration (reusing `AWSIntegrationConfig` and the creds already wired via the EKS/CloudWatch path), since CloudTrail is account-wide, not tied to one resource.
- Registers `cloudtrail` in the `EvidenceSource` literal.
- **Auto-seeds and prioritizes** CloudTrail for AWS-originating alerts (`cloudwatch` / `eks` / `alertmanager`) in **both** `app/agent/investigation.py` (seeding) and `app/agent/prompt.py` (prioritization), so change-causality is pulled in automatically.

**Docs / tests**
- `docs/cloudtrail_events.mdx`, registered in `docs/docs.json`.
- `tests/tools/test_cloudtrail_events.py` — schema, availability (×4), extraction (×3), filter priority, time-window clamping, response shaping, discovery, the `aws_backend` short-circuit, and the two-map seeding/prioritization guard. Tool classified in `test_telemetry.py`.

**Notes:** no new dependencies (boto3/botocore present). `execute_aws_sdk_call` authenticates via boto3's default credential chain (env / `~/.aws`), consistent with the existing RDS/EKS tools.

Validated against a **real AWS account** (CloudTrail management events, `us-east-1`):
 **End-to-end RCA** — a CloudWatch-sourced alert auto-seeds CloudTrail; the agent finds `AuthorizeSecurityGroupIngress` opening `0.0.0.0/0:22` on `opensre-demo-sg` by user `tola` and cites it as the root cause:

cloudtrail-rca-investigation

 
 
image

**Verification:** `make lint`, `make format-check`, `make typecheck` pass; CloudTrail tool tests pass; full unit suite green except the env-gated live-LLM routing tests (per CI.md §6).

---

The tool is intentionally thin: it reuses the existing read-only `execute_aws_sdk_call` allowlist path (so no new credential or safety surface) and mirrors `RDSEventsTool`'s shape. The interesting logic is CloudTrail-specific: it only accepts one `LookupAttribute`, so the tool picks the most specific filter provided and converts a `duration_minutes` window into the `StartTime`/`EndTime` pair, then trims each event to the forensic fields parsed from the `CloudTrailEvent` JSON. Availability rides on the account-level `aws` integration since CloudTrail is account-wide, and the tool is auto-seeded for AWS-originating alerts so "check CloudTrail" happens first — exactly how a real post-mortem starts.

---

### DIFF
--- a/app/agent/investigation.py
+++ b/app/agent/investigation.py
@@ -30,9 +30,9 @@ _UNSET: object = object()  # sentinel distinguishing "not yet started" from a No
 _ALERT_SOURCE_TO_TOOL_SOURCES: dict[str, list[str]] = {
     "grafana": ["grafana"],
     "datadog": ["datadog"],
-    "cloudwatch": ["cloudwatch", "cloudtrail"],
-    "eks": ["eks", "cloudtrail"],
-    "alertmanager": ["grafana", "cloudwatch", "cloudtrail"],
+    "cloudwatch": ["cloudwatch"],
+    "eks": ["eks"],
+    "alertmanager": ["grafana", "cloudwatch"],
     "sentry": ["sentry"],
     "honeycomb": ["honeycomb"],
     "coralogix": ["coralogix"],
--- a/app/agent/investigation.py
+++ b/app/agent/investigation.py
@@ -30,9 +30,9 @@ _UNSET: object = object()  # sentinel distinguishing "not yet started" from a No
 _ALERT_SOURCE_TO_TOOL_SOURCES: dict[str, list[str]] = {
     "grafana": ["grafana"],
     "datadog": ["datadog"],
-    "cloudwatch": ["cloudwatch"],
-    "eks": ["eks"],
-    "alertmanager": ["grafana", "cloudwatch"],
+    "cloudwatch": ["cloudwatch", "cloudtrail"],
+    "eks": ["eks", "cloudtrail"],
+    "alertmanager": ["grafana", "cloudwatch", "cloudtrail"],
     "sentry": ["sentry"],
     "honeycomb": ["honeycomb"],
     "coralogix": ["coralogix"],
--- a/app/agent/prompt.py
+++ b/app/agent/prompt.py
@@ -63,9 +63,9 @@ Severity: {severity}
 _ALERT_SOURCE_TO_TOOL_SOURCES: dict[str, list[str]] = {
     "grafana": ["grafana"],
     "datadog": ["datadog"],
-    "cloudwatch": ["cloudwatch", "ec2", "rds"],
-    "eks": ["eks", "ec2"],
-    "alertmanager": ["eks", "cloudwatch", "grafana"],
+    "cloudwatch": ["cloudwatch", "ec2", "rds", "cloudtrail"],
+    "eks": ["eks", "ec2", "cloudtrail"],
+    "alertmanager": ["eks", "cloudwatch", "grafana", "cloudtrail"],
     "sentry": ["sentry"],
     "honeycomb": ["honeycomb"],
     "coralogix": ["coralogix"],
--- a/app/integrations/cloudtrail.py
+++ b/app/integrations/cloudtrail.py
@@ -28,14 +28,21 @@ def cloudtrail_is_available(sources: dict[str, dict]) -> bool:
     gate availability on the ``aws`` source the catalog populates from
     ``AWSIntegrationConfig`` — mirroring how the other AWS tools reuse the creds
     wired via the EKS/CloudWatch path — and on the optional synthetic
-    ``_backend`` handle so the tool stays selectable in fixture-driven tests.
+    ``ec2_backend`` handle (the key the synthetic harness injects into the
+    ``aws`` source) so the tool stays selectable in fixture-driven tests.
+
+    Note: ``role_arn`` / ``credentials`` here gate *availability* only. The
+    actual lookup runs through ``execute_aws_sdk_call``, which uses boto3's
+    ambient credential chain (env / shared config / instance role) — the
+    configured role is not assumed as the execution identity. This matches the
+    other AWS tools (RDS/EKS).
     """
     aws = sources.get("aws", {})
     return bool(
         aws.get("connection_verified")
         or aws.get("role_arn")
         or aws.get("credentials")
-        or aws.get("_backend")
+        or aws.get("ec2_backend")
     )
 
 
@@ -44,11 +51,12 @@ def cloudtrail_extract_params(sources: dict[str, dict]) -> dict[str, Any]:
 
     Resolution order for region matches the rest of the AWS stack: explicit
     ``aws`` source field, then ``AWS_REGION`` env, then the default. Forwards
-    the optional synthetic ``_backend`` handle as ``aws_backend`` so the tool
-    can short-circuit to fixture data instead of leaking boto3 calls to whatever
-    AWS account the developer happens to be authenticated against during a
-    synthetic run. Resource/principal/time-window filters are alert-specific and
-    are supplied by the planner at call time, not extracted here.
+    the optional synthetic ``ec2_backend`` handle (the key the synthetic harness
+    injects into the ``aws`` source) as ``aws_backend`` so the tool short-circuits
+    to fixture data instead of leaking boto3 calls to whatever AWS account the
+    developer happens to be authenticated against during a synthetic run.
+    Resource/principal/time-window filters are alert-specific and are supplied
+    by the planner at call time, not extracted here.
     """
     aws = sources.get("aws", {})
     region = (
@@ -56,5 +64,5 @@ def cloudtrail_extract_params(sources: dict[str, dict]) -> dict[str, Any]:
     )
     return {
         "region": region,
-        "aws_backend": aws.get("_backend"),
+        "aws_backend": aws.get("ec2_backend"),
     }
--- a/app/integrations/cloudtrail.py
+++ b/app/integrations/cloudtrail.py
@@ -1,0 +1,60 @@
+"""Shared AWS CloudTrail integration helpers.
+
+CloudTrail is the canonical AWS change-causality source — "who changed what,
+and when?". Unlike RDS, CloudTrail is account-wide rather than tied to a single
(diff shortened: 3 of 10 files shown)


amsorrytola:
Thanks @greptile-apps — both addressed:

- **P2 (truncation not signalled):** the response now includes `truncated` (derived from CloudTrail's `NextToken`) and `next_token`, and the tool accepts a `next_token` param to fetch the next page. So an agent on a busy account / wide window can detect the 50-event boundary and paginate instead of silently treating one page as complete. Covered by `test_truncated_when_next_token_present`, `test_not_truncated_without_next_token`, and `test_next_token_forwarded_to_api`; docs updated.
- **ReadOnly fixture (string vs bool):** fixed — the fixture now uses `"false"` to match the real API.

@greptile-apps review

kespineira (changes requested):
Thanks @amsorrytola ! Solid PR, and the demo really sells it! A few things before merge:

## Blocking

1. synthetic runs can reach real aws.

`app/integrations/cloudtrail.py` reads:

```py
aws.get("_backend")
```

but the synthetic harness sets `ec2_backend` instead:

```py
# tests/synthetic/rds_postgres/run_suite.py
ec2_backend=...
```

so `aws_backend` ends up as `None`, the short-circuit in `CloudTrailEventsTool` is skipped, and the tool falls through to a real boto3 call.

because cloudtrail is now auto-seeded for cloudwatch alerts, and the rds scenarios use `alert_source: cloudwatch`, this runs for every rds synthetic scenario. it fails closed without creds, but with ambient aws creds it can hit a real account.

the current `test_short_circuits_to_aws_backend` misses this because it injects the backend directly instead of going through the harness-shaped config path.

could we read `ec2_backend` here and add a harness-shaped regression test?

refs:
- `app/integrations/cloudtrail.py:59`
- `tests/synthetic/rds_postgres/run_suite.py:236-238`
- `app/tools/CloudTrailEventsTool/__init__.py:188`
- `app/agent/investigation.py:33`
- `tests/tools/test_cloudtrail_events.py:295`

2. `FixtureAWSBackend` does not implement `lookup_events`.

once the backend key above is fixed, `CloudTrailEventsTool` will call `lookup_events` on the fixture backend, but `FixtureAWSBackend` does not currently expose it. that means the synthetic path will move from a real aws call to an `AttributeError`.

could we add `lookup_events` to both the protocol and the fixture implementation?

refs:
- `tests/synthetic/mock_aws_backend/backend.py:55`
- `tests/synthetic/mock_aws_backend/backend.py:94`
- `app/tools/CloudTrailEventsTool/__init__.py:191`

## worth addressing

1. configured aws creds / role do not seem to be used.

`cloudtrail_is_available` gates availability on `role_arn` / `credentials`, but `cloudtrail_extract_params` only forwards the region into the tool, and `execute_aws_sdk_call` then uses the ambient credential chain.

that means a configured `role_arn` integration does not actually assume that role for the lookup; the request is made with whatever identity the host process has.

could we either assume the configured role, or document clearly that the integration role is only used for availability/config and not as the execution identity?

refs:
- `app/integrations/cloudtrail.py:36-37`
- `app/integrations/cloudtrail.py:42-60`

2. cloudtrail is seeded unconditionally for broad alert sources.

cloudtrail is now auto-seeded for every `cloudwatch`, `eks`, and `alertmanager` alert, but the seeded lookup is account-wide unless the planner adds a concrete filter.

given cloudtrail's low lookup rate limits, this might create noisy/unscoped lookups during investigations. maybe it is better to leave cloudtrail for the planner to call once it has a concrete `resource_name`, `username`, or event target.

refs:
- `app/agent/investigation.py:33-35`
- `app/agent/prompt.py:66-68`

## minor

- `read_only` is currently treated as a string. cloudtrail returns `"true"` / `"false"`, and `"false"` is truthy in python, so this should probably be coerced to a real bool.

  ref: `app/tools/CloudTrailEventsTool/__init__.py:88`

- the cloudtrail tool does not declare `injected_params` on `@tool`. harmless, but the sibling rds tool declares `injected_params=("aws_backend",)`, so adding it here would keep the shape consistent.

  ref: `app/tools/CloudTrailEventsTool/__init__.py:97`

amsorrytola:
Thanks for the thorough review @kespineira ! All addressed in 3e720fcd:

**Blocking**

1. **Synthetic runs reaching real AWS** — fixed. `cloudtrail_is_available` / `cloudtrail_extract_params` now read the harness-shaped `aws["ec2_backend"]` key (was `_backend`, which was always `None`), so the tool short-circuits to the fixture instead of falling through to a real boto3 `lookup_events`. Added `test_harness_shaped_aws_source_short_circuits_off_real_aws`, which goes through `cloudtrail_extract_params` (not a directly-injected backend) and asserts `execute_aws_sdk_call` is never reached.

2. **`FixtureAWSBackend` missing `lookup_events`** — added it to both the `AWSBackend` Protocol and `FixtureAWSBackend`. It returns a valid, empty account-wide result (no scenario declares CloudTrail evidence yet), so the synthetic path is well-shaped instead of raising `AttributeError`.

**Worth addressing**

3. **Configured role/creds not used as execution identity** — documented. The integration docstring and `docs/cloudtrail_events.mdx` now state that `role_arn`/credentials gate availability + supply the region only; the lookup runs through boto3's ambient credential chain (matching RDS/EKS). I kept it consistent with the other AWS tools rather than introducing role-assumption only here.

4. **Unconditional seeding for broad alert sources** — agreed; reverted the auto-seeding in `investigation.py`. CloudTrail is now planner-driven (kept only the `prompt.py` prioritization), so it's reached for once there's a concrete `resource_name`/`username`/event target, avoiding noisy unscoped lookups against CloudTrail's low rate limit. The test now asserts "prioritized but not auto-seeded."

**Minor**

5. **`read_only` treated as a string** — now coerced to a real bool (`"true"`/`"false"` → `True`/`False`), since `"false"` is truthy in Python. Covered by an assertion in the shaping test.

6. **Missing `injected_params`** — added `injected_params=("aws_backend",)` to match the RDS sibling.


 @greptile-apps review

kespineira (approved):
all points addressed, thanks for the quick turnaround

small non-blocking nit: the demo in the description was captured with auto-seeding in place. now that cloudtrail is planner-driven, the agent has to choose to call it. it is prioritized, so it should, but maybe worth a quick re-confirm or a note in the description so the gif matches the current behavior.

LGTM 👍  good job @amsorrytola

kespineira:
@muddlebee when you have a chance, could you take a look?

amsorrytola:
> all points addressed, thanks for the quick turnaround
> 
> small non-blocking nit: the demo in the description was captured with auto-seeding in place. now that cloudtrail is planner-driven, the agent has to choose to call it. it is prioritized, so it should, but maybe worth a quick re-confirm or a note in the description so the gif matches the current behavior.
> 
> LGTM 👍 good job @amsorrytola

Hey @kespineira , apologies for the delay , got stuck in some work , 
i've edited the PR with new gif and ss , all the asks are delivered , i think it ready to merge now

Devesh36:
LGTM. One nit: - _shape_event drops requestParameters from the CloudTrailEvent JSON — consider including it (size-capped) so RCA can cite actual config changes (e.g. CIDR/port on AuthorizeSecurityGroupIngress), not just event_name.

PR #2754, opened 2026-06-05
refactor(config): improve LLM provider configuration handling

### LINKED ISSUE #2113: [BUG] Onboarded OpenAI provider can still resolve as Anthropic at runtime
### Summary

After onboarding with OpenAI, the interactive agent can still resolve the active LLM provider as Anthropic and fail with a missing ANTHROPIC_API_KEY error.

Expected:
After `opensre onboard` is completed with OpenAI selected, the runtime should use OpenAI as the active LLM provider.

Actual:
A user reported that after configuring OpenAI via onboarding, the interactive agent failed on a normal prompt with:

`assistant failed: LLM provider 'anthropic' requires ANTHROPIC_API_KEY to be set.`

The same user reported that manually setting the model/provider to GPT made it work.

### Steps to reproduce

Need to reproduce whether this happens when:

1. Run `opensre onboard`.
2. Select OpenAI and provide an OpenAI API key.
3. Start or return to the interactive agent.
4. Send a normal prompt such as `hello`.

Reported failure:

`assistant failed: LLM provider 'anthropic' requires ANTHROPIC_API_KEY to be set.`

Also validate whether `/model set openai ...` or manually exporting `LLM_PROVIDER=openai` makes the same session work.

Yes

### How often does it occur?

Every time

### Operating system

Windows

### Logs and error output

_No response_

### Additional context

Relevant write paths:

- `opensre onboard` saves provider config with `save_local_config(...)` in `app/cli/wizard/flow.py`.
- `opensre onboard` writes `LLM_PROVIDER` and model env vars through `sync_provider_env(...)` in `app/cli/wizard/env_sync.py`.
- `sync_provider_env(...)` writes to `PROJECT_ENV_PATH`, defined in `app/cli/wizard/config.py`.
- `/model set` writes env values through `sync_env_values(...)`, then also updates live `os.environ` and resets runtime LLM caches in `app/cli/interactive_shell/command_registry/model.py`.

Relevant read paths:

- `LLMSettings.from_env()` reads `os.getenv("LLM_PROVIDER", "anthropic")` in `app/config.py`.
- `CfgHelpers.resolve_llm_provider()` reads `os.getenv("LLM_PROVIDER") or "anthropic"` in `app/utils/cfg_helpers.py`.
- `opensre doctor` reads `LLM_PROVIDER` directly from `os.getenv(...)` in `app/cli/commands/doctor.py`.
- `opensre config` reads `LLM_PROVIDER` directly from `os.getenv(...)` in `app/cli/commands/config.py`.

Observed difference:

- Onboarding persists provider config to store and `.env`.
- `/model set` persists env values and also updates the current process env with `os.environ.update(values)`.

Devesh36:
/fix #2113 

After onboarding with OpenAI, the interactive agent could still resolve the active LLM provider as anthropic and fail with ANTHROPIC_API_KEY missing — even though onboarding had written LLM_PROVIDER=openai to the project .env.

Root cause: blank or stale shell values (e.g. LLM_PROVIDER="", common on Windows) blocked .env reload because both load_dotenv(override=False) and load_env(override=False) treat empty strings as “already set.” Runtime then fell back to the anthropic default.

Changes
load_env() — treat blank shell env values as unset so onboarded .env values apply; align default load paths with onboarding (OPENSRE_PROJECT_ENV_PATH → package .env → cwd/.env); when OPENSRE_PROJECT_ENV_PATH is explicit, load only that file
CLI startup — replace load_dotenv() with load_env() so startup uses the same logic
Provider resolution — normalize blank LLM_PROVIDER before defaulting to anthropic in get_configured_llm_provider() and CfgHelpers.resolve_llm_provider()
Smoke tests — subprocess env isolation updated (unset keys instead of ""; clear LLM_PROVIDER from host env)
Screenshot 2026-06-05 at 10 08 46 AM

Screenshot 2026-06-05 at 10 19 39 AM

### DIFF
--- a/app/cli/__main__.py
+++ b/app/cli/__main__.py
@@ -15,7 +15,6 @@ import sys
 from contextlib import suppress
 
 import click
-from dotenv import load_dotenv
 
 from app.analytics.cli import build_cli_invoked_properties, capture_cli_invoked
 from app.analytics.provider import Properties, capture_first_run_if_needed, shutdown_analytics
@@ -27,6 +26,7 @@ from app.cli.support.prompt_support import (
     install_questionary_ctrl_c_double_exit,
     install_questionary_escape_cancel,
 )
+from app.utils.config import load_env
 from app.utils.sentry_sdk import capture_exception, init_sentry
 from app.version import get_version
 
@@ -215,7 +215,7 @@ def _should_capture_cli_exception(exc: click.ClickException) -> bool:
 def main(argv: list[str] | None = None) -> int:
     """Entry point for the ``opensre`` console script."""
     _ensure_utf8_stdio()
-    load_dotenv(override=False)
+    load_env(override=False)
     cli_argv = list(sys.argv[1:] if argv is None else argv)
     try:
         init_sentry(entrypoint=_sentry_entrypoint_for_invocation(cli_argv))
--- a/app/config.py
+++ b/app/config.py
@@ -200,7 +200,8 @@ DEFAULT_LLM_RESOLUTION_FALLBACK_PROVIDERS: tuple[str, ...] = ("openai", "anthrop
 def get_configured_llm_provider() -> str:
     """Return the active LLM provider from env/project .env."""
     load_env(override=False)
-    return os.getenv("LLM_PROVIDER", "anthropic").strip().lower() or "anthropic"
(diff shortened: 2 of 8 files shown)


Devesh36:
@greptile-apps review again

Issue #2537, opened 2026-05-26
[BUG] opensre integrations setup telegram is rejected by click.Choice


kespineira:
### Summary

`opensre integrations setup telegram` exits immediately with a click "Invalid value" error because `telegram` is missing from the `SETUP_SERVICES` tuple used by the positional-argument validator, even though the dispatch handler and the user-facing docs promise the command works.

### Expected vs actual behavior

**Expected:** `opensre integrations setup telegram` runs the interactive `_setup_telegram` flow — prompts for the bot token (and optional default chat ID), validates against the Telegram API, and persists the credentials to `~/.opensre/integrations.json`.

**Actual:** the command never reaches the handler. click rejects the positional argument before dispatch:

```
Error: Invalid value for '[[alertmanager|aws|betterstack|coralogix|datadog|discord|grafana|github|gitlab|honeycomb|incident_io|mariadb|mongodb|mongodb_atlas|mysql|openclaw|opensearch|postgresql|rabbitmq|rds|sentry|slack|tracer|vercel]]': 'telegram' is not one of 'alertmanager', 'aws', ...
```

Exit code is `2`.

### Steps to reproduce

1. Check out `main` at or after `8bdbc9b` (any commit after the Telegram wizard PR landed).
2. Run `make install`.
3. Run `uv run opensre integrations setup telegram`.
4. Observe the click error and exit code `2`.

### Can you reproduce it consistently?

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

```
Error: Invalid value for '[[alertmanager|aws|betterstack|coralogix|datadog|discord|grafana|github|gitlab|honeycomb|incident_io|mariadb|mongodb|mongodb_atlas|mysql|openclaw|opensearch|postgresql|rabbitmq|rds|sentry|slack|tracer|vercel]]':
'telegram' is not one of 'alertmanager', 'aws', 'betterstack', 'coralogix',
'datadog', 'discord', 'grafana', 'github', 'gitlab', 'honeycomb',
'incident_io', 'mariadb', 'mongodb', 'mongodb_atlas', 'mysql', 'openclaw',
'opensearch', 'postgresql', 'rabbitmq', 'rds', 'sentry', 'slack', 'tracer',
'vercel'.
```

### Additional context

Root cause:

- `_HANDLERS["telegram"] = _setup_telegram` is registered in `app/integrations/cli.py` (~L853), so the runtime dispatch is wired correctly.
[...]

PR #2773, opened 2026-06-08
test(integrations): add registry self-consistency invariants

### LINKED ISSUE #2772: [IMPROVEMENT] Add registry self-consistency tests to catch integration order/handler drift on PR
### Current state

`app/integrations/registry.py` is the source of truth for integration setup/verify ordering, and `app/integrations/cli.py` carries the matching setup handlers (`_HANDLERS`). There is no test that asserts these two stay in sync, so drift lands silently.

Current `main` (commit `02a1fdd6`, 2026-06-08) has **9 collisions**:

**Duplicate `setup_order`:**
- 12: openclaw, tracer
- 19: postgresql, telegram, whatsapp
- 20: twilio, mysql
- 24: jenkins, dagster

**Duplicate `verify_order`:**
- 2: grafana, victoria_logs
- 22: incident_io, kafka
- 27: whatsapp, mysql
- 28: twilio, openclaw
- 36: jenkins, dagster

The `jenkins`/`dagster` collisions (from #2569 and #2652) landed *after* muddlebee's audit in `#opensre-integrations`, which proves the drift is actively recurring. Without a test, the next integration PR adds the next collision.

Separate but related: a service can have a `setup_order` without a `_HANDLERS` entry, or a `verifier` without a `verify_order`, both producing CLI dispatch failures users hit via `opensre integrations setup/verify <svc>` (see #1973 / #2710).

### Desired state

Add `tests/integrations/test_registry_invariants.py` with five static assertions. Pure registry introspection, no integration is probed, runs in well under a second, and pulled in by `make test-scope` on any change under `app/integrations/` or `tests/integrations/`:

1. `test_setup_orders_are_unique` — no two services share a `setup_order`
2. `test_verify_orders_are_unique` — no two services share a `verify_order`
3. `test_every_setup_service_has_a_cli_handler` — every `SUPPORTED_SETUP_SERVICES` entry has a `_HANDLERS` entry in `app/integrations/cli.py`
4. `test_every_cli_handler_is_registered_in_registry` — reverse direction: no orphan handlers
5. `test_every_verifier_has_a_verify_order` — every `verifier=...` spec has a non-None `verify_order` so positional CLI works

Tests #1 and #2 will fail on current `main` until the renumber lands (Kevs is on that per `#opensre-integrations`). Two options to land safely — happy to take either based on maintainer preference:

- **Option A**: land the test file with `pytest.mark.xfail(strict=False)` on #1 and #2 referencing the renumber issue; flip to strict in the renumber PR.
- **Option B**: hold this PR until the renumber lands, then merge strict.

Tests #3–5 should pass on current `main` as far as I can see (`gitlab`, `jira`, `airflow` have neither verifier nor verify_order, so #5 is satisfied; #3–4 align with the constants in `app/cli/support/constants.py` which already source from the registry per #1973's fix). I'll re-verify locally before opening the PR.

Davidson3556:
Fixes #2772

Adds `tests/integrations/test_registry_invariants.py` with five static assertions over the integration registry and the CLI dispatch:

1. `test_setup_orders_are_unique` — no two services share a `setup_order`
2. `test_verify_orders_are_unique` — no two services share a `verify_order`
3. `test_every_setup_service_has_a_cli_handler` — every `SUPPORTED_SETUP_SERVICES` entry has a `_HANDLERS` entry in `app/integrations/cli.py`
4. `test_every_cli_handler_is_registered_in_registry` — reverse direction, no orphan handlers
5. `test_every_verifier_has_a_verify_order` — every spec with `verifier=...` has a `verify_order` so the positional CLI works

The two order tests are marked `xfail(strict=False)` because `main` currently has 9 pre-existing collisions (4 setup + 5 verify), which the renumber PR will clear. After that lands, the markers flip to strict (or are removed entirely) so the gate is permanent. The other three tests pass on `main` today.

No production code changes. `tests/integrations/` is already routed by `make test-scope`, so this naturally gates any PR that touches `app/integrations/`.

**Default CI run (xfail markers in effect, what this PR posts):**

https://github.com/user-attachments/assets/4ebe3a35-7d11-4b3d-ba1a-27525888dea7

The collision payload names every offending service, so the renumber author has the full punch list inline.

---

The registry in `app/integrations/registry.py` and the CLI handler dict in `app/integrations/cli.py` are two halves of the same contract. They have to stay in sync, but nothing enforces that. So when someone adds a new integration, they can pick a `setup_order` that collides with another service, or skip wiring a `_HANDLERS` entry, and the drift only surfaces when a user runs `opensre integrations setup/verify <svc>` and hits a Click rejection or non-deterministic ordering. That is how `main` ended up with 9 collisions today, including a `jenkins`/`dagster` collision that landed after muddlebee's audit in `#opensre-integrations`.

The tests are pure registry introspection. No integration is probed, no subprocess is spawned, no fixtures are loaded. Five invariants total, runs in under 2 seconds.

I considered putting the check in a standalone CI script (or a pre-commit hook) instead of a pytest file. Two reasons I went with tests: `make test-scope` already routes anything under `tests/integrations/` to run on PRs that touch `app/integrations/`, so this naturally gates the right diffs without any new CI wiring; and the assertion messages name the offending services inline, which is friendlier than a script logging to stderr.

The two order-uniqueness tests use `xfail(strict=False)` rather than `skip` deliberately. `skip` would silently drop them. `xfail(strict=False)` keeps them collecting and visible in CI output, so when the renumber PR flips the registry clean, the tests become `XPASS` and the author flips the markers to strict (or removes them) in the same PR. That is the planned signal.

Helper: `_duplicates(values)` returns a `{value: count}` map of only the duplicated entries. Both order-uniqueness tests use it so the assertion output formats consistently across both axes.

---

### DIFF
--- a/tests/integrations/test_registry_invariants.py
+++ b/tests/integrations/test_registry_invariants.py
@@ -21,8 +21,8 @@ from app.integrations.registry import (
 )
 
 
-def _duplicates(values: list[int]) -> dict[int, int]:
-    return {value: count for value, count in Counter(values).items() if count > 1}
+def _duplicates(values: list[int]) -> set[int]:
+    return {value for value, count in Counter(values).items() if count > 1}
 
 
 @pytest.mark.xfail(
@@ -72,7 +72,9 @@ def test_every_setup_service_has_a_cli_handler() -> None:
     if missing:
         raise AssertionError(
             f"Registry declares setup_order for {missing} but no _HANDLERS entry "
-            "in app/integrations/cli.py. Click accepts these positionally, then dispatch fails."
+            "in app/integrations/cli.py. These services are silently dropped from "
+            "_SETUP_SERVICES, so `opensre integrations setup <svc>` will reject them "
+            "with the 'Usage: setup <service>' error."
         )
 
 
@@ -80,8 +82,10 @@ def test_every_cli_handler_is_registered_in_registry() -> None:
     orphans = [svc for svc in _HANDLERS if svc not in SUPPORTED_SETUP_SERVICES]
     if orphans:
         raise AssertionError(
-            f"app/integrations/cli.py defines _HANDLERS for {orphans} but registry has "
-            "no setup_order for them. Click's Choice list rejects them positionally."
+            f"app/integrations/cli.py defines _HANDLERS for {orphans} but registry "
+            "has no setup_order for them. These handlers are unreachable: "
+            "_SETUP_SERVICES only includes services that appear in both "
+            "SUPPORTED_SETUP_SERVICES and _HANDLERS, so the wizard will never offer them."
         )
 
 
--- a/tests/integrations/test_registry_invariants.py
+++ b/tests/integrations/test_registry_invariants.py
@@ -1,0 +1,96 @@
+"""Registry self-consistency invariants.
+
+These tests don't probe any integration. They enforce the static contract
+between :mod:`app.integrations.registry` and the CLI dispatch in
+:mod:`app.integrations.cli`. Each invariant prevents a class of drift that has
+already shipped to ``main`` at least once: duplicate orders, missing handlers,
+verifier without a positional route. Failing here on PR catches the regression
+before users hit it via ``opensre integrations setup/verify <svc>``.
+"""
+
+from __future__ import annotations
+
+from collections import Counter
+
+import pytest
+
+from app.integrations.cli import _HANDLERS
+from app.integrations.registry import (
+    INTEGRATION_SPECS,
+    SUPPORTED_SETUP_SERVICES,
+)
+
+
+def _duplicates(values: list[int]) -> dict[int, int]:
+    return {value: count for value, count in Counter(values).items() if count > 1}
+
+
+@pytest.mark.xfail(
+    strict=False,
+    reason=(
+        "Pre-existing duplicate setup_order values in INTEGRATION_SPECS. "
+        "Renumber PR flips this strict."
+    ),
+)
+def test_setup_orders_are_unique() -> None:
+    orders = [s.setup_order for s in INTEGRATION_SPECS if s.setup_order is not None]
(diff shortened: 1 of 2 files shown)


Davidson3556:
@greptile review

tests/integrations/test_registry_invariants.py
```diff
@@ -0,0 +1,100 @@
...
+
+def test_every_setup_service_has_a_cli_handler() -> None:
+    missing = [svc for svc in SUPPORTED_SETUP_SERVICES if svc not in _HANDLERS]
+    if missing:
+        raise AssertionError(
+            f"Registry declares setup_order for {missing} but no _HANDLERS entry "
+            "in app/integrations/cli.py. These services are silently dropped from "
+            "_SETUP_SERVICES, so `opensre integrations setup <svc>` will reject them "
+            "with the 'Usage: setup <service>' error."
+        )
```
kespineira: `test_every_setup_service_has_a_cli_handler` looks redundant with the existing `test_every_setup_spec_has_handler` added in #2537.

both tests assert the same invariant: every `SUPPORTED_SETUP_SERVICES` entry must have a matching `_HANDLERS` key:

```py
set(SUPPORTED_SETUP_SERVICES) <= set(_HANDLERS)
```

the new version adds a better failure message, but the actual gate is the same, so this feels like duplicate coverage.

could we either drop this one, or fold its better failure message into the existing test?

Davidson3556: thank you , you are right ..`test_every_setup_spec_has_handler` covers the exact same gate. went with your option 2: folded the descriptive failure message (offending-service list) into the existing test in `tests/integrations/test_registry.py` and dropped the duplicate from this file. kept the reverse-direction `test_every_cli_handler_is_registered_in_registry` since the existing test doesn't cover orphan handlers in that direction.
@kespineira

Davidson3556:
@greptile review

PR #2775, opened 2026-06-08
fix(bench): seed ANTHROPIC_API_KEY in llm_alone dispatch acceptance test

### LINKED ISSUE #2774: [BUG] benchmark dispatch test fails on fork PRs (missing ANTHROPIC_API_KEY)
### Summary

`tests/benchmarks/_framework/test_runner_llm_alone_dispatch.py::test_run_inner_accepts_llm_alone_when_adapter_provides_baseline` fails on any PR opened from a fork because `LLMDispatcher.activate()` requires `ANTHROPIC_API_KEY` to be set, and GitHub Actions strips repository secrets from fork-PR workflow runs by default. The test passes on `main` (where direct pushes carry the real secret) but is permanently red for any external contributor's PR.

**Expected:** the test exercises the runner's pre-flight gate for `llm_alone` mode using a hermetic fake API key, so it passes whether or not a real key is present in the environment.

**Actual:** the test calls `runner.run_without_integrity()`, which activates the LLM via `LLMDispatcher.activate()`. The dispatcher raises `MissingAPIKey` because `ANTHROPIC_API_KEY` is unset, the runner reports the run as aborted, and the test's `assert not outcome.aborted` fails:

```
AssertionError: LLM dispatch failed: 'claude-4-sonnet' requires env var ANTHROPIC_API_KEY to be set.
```

### Steps to reproduce

```bash
$ env -u ANTHROPIC_API_KEY uv run pytest \
    tests/benchmarks/_framework/test_runner_llm_alone_dispatch.py::test_run_inner_accepts_llm_alone_when_adapter_provides_baseline -v

FAILED test_run_inner_accepts_llm_alone_when_adapter_provides_baseline
========================= 1 failed, 6 passed in 2.09s ==========================
```

The other 6 tests in the file pass without a key because they call `_run_one_cell` directly, bypassing the dispatcher's `activate()` path.

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

```
$ env -u ANTHROPIC_API_KEY uv run pytest \
    tests/benchmarks/_framework/test_runner_llm_alone_dispatch.py::test_run_inner_accepts_llm_alone_when_adapter_provides_baseline -v
...
  ▶ activating LLM: claude-4-sonnet
  ✗ aborted: LLM dispatch failed: 'claude-4-sonnet' requires env var ANTHROPIC_API_KEY to be set.
FAILED tests/benchmarks/_framework/test_runner_llm_alone_dispatch.py::test_run_inner_accepts_llm_alone_when_adapter_provides_baseline
```

Surfaced by PR #2773 (registry contract tests from a fork branch); the failure is unrelated to that PR's changes — confirmed by running the same test on `main` without the key.

### Additional context

**Root cause:** `tests/benchmarks/_framework/llm_dispatch.py:173-180` defines `MissingAPIKey` which `LLMDispatcher.activate()` raises when the env var is unset. The failing test sets up a `BenchmarkConfig` with `llms: ["claude-4-sonnet"]` and lets the dispatcher run with the real environment. The other tests in the file use `_run_one_cell` directly with an explicit `spec=LLM_SPECS["claude-4-sonnet"]`, so they never hit `activate()`.

**Proposed fix:** add `monkeypatch.setenv("ANTHROPIC_API_KEY", "test-key")` to the failing test, matching the established pattern already used 5+ times in `tests/benchmarks/_framework/test_llm_dispatch.py`. The test's purpose is to verify the pre-flight gate accepts a valid adapter, not to exercise live LLM calls — the dispatcher only needs the env var to be present, not valid.

This is preferable to `skipif(not os.getenv("ANTHROPIC_API_KEY"))` because it keeps the test running on every PR (regression coverage everywhere) rather than skipping silently on fork PRs.

Davidson3556:
Fixes #2774

`test_run_inner_accepts_llm_alone_when_adapter_provides_baseline` was failing on every PR opened from a fork. The test calls `runner.run_without_integrity()`, which activates the LLM via `LLMDispatcher.activate()` in `tests/benchmarks/_framework/llm_dispatch.py:173-180`. That activation raises `MissingAPIKey` if `ANTHROPIC_API_KEY` is unset. CI on main passes because the secret is injected via `.github/workflows/ci.yml:168,370`, but GitHub Actions strips repository secrets from fork-PR workflows, so every external contributor's PR hits the same red.

Change: add `monkeypatch: pytest.MonkeyPatch` to the test and call `monkeypatch.setenv("ANTHROPIC_API_KEY", "test-key")` before constructing the runner. Same pattern `tests/benchmarks/_framework/test_llm_dispatch.py` already uses in 5+ places.

The other 6 tests in the file are unaffected: they call `_run_one_cell` directly with an explicit `spec=LLM_SPECS["claude-4-sonnet"]` and never trigger `activate()`.

**Before** (no key — reproduces the fork-PR CI failure):

image
image

**After** (no key, same command):

image

Also verified hermetic with a stub key (`ANTHROPIC_API_KEY=fake-key` set): 7 passed. The test is now green whether the env var is set or unset.

---

The test is verifying that the runner's pre-flight gate accepts an adapter that returns a non-None `baseline_agent_class`. The downstream investigation pipeline is patched out via `patch("app.pipeline.runners.run_investigation", ...)`, but the dispatcher's environment-variable check fires before the patch is ever reached. So the test was broken by an upstream concern (env presence) that has nothing to do with what it's actually asserting.

1. `pytest.mark.skipif(not os.getenv("ANTHROPIC_API_KEY"))`. Would skip the test on fork PRs but also silently drop it on any local dev run without a key. The point of the test is to gate the runner's pre-flight, and that gate should be exercised on every PR, not just ones where someone happens to have a key.
2. Add `ANTHROPIC_API_KEY` to the workflow's `env:` block unconditionally with a fake value. Too invasive: changes CI for every job, and the dummy key value would show up in workflow logs.
3. `monkeypatch.setenv("ANTHROPIC_API_KEY", "test-key")` inside the test. Hermetic, scoped to the one test that needs it, automatically reverted after the test by pytest's monkeypatch fixture. Already the established pattern in `tests/benchmarks/_framework/test_llm_dispatch.py` (5+ uses).

I went with option 3. The dispatcher only checks env-var presence, not key validity, so a dummy `"test-key"` value is enough. With `run_investigation` already patched, no real Anthropic API call ever happens, so a fake key is safe.

One thing I deliberately did not change: the dispatcher's `MissingAPIKey` exception itself. It is correct behavior for production runs (you want to fail loud if the key is missing). The bug was that one specific test was indirectly exercising that production check without the test setup it needed.

---

### DIFF
--- a/tests/benchmarks/_framework/tests/test_runner_llm_alone_dispatch.py
+++ b/tests/benchmarks/_framework/tests/test_runner_llm_alone_dispatch.py
@@ -162,11 +162,14 @@ def test_run_inner_refuses_llm_alone_when_adapter_has_no_baseline(tmp_path: Path
     assert "no-baseline" in str(exc_info.value)
 
 
-def test_run_inner_accepts_llm_alone_when_adapter_provides_baseline(tmp_path: Path) -> None:
+def test_run_inner_accepts_llm_alone_when_adapter_provides_baseline(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
     """The flip side: same config but an adapter that DOES return a
     baseline_agent_class must proceed past the pre-flight gate. We patch
     run_investigation so the test doesn't depend on the production
     investigation pipeline."""
+    monkeypatch.setenv("ANTHROPIC_API_KEY", "test-key")
     runner = BenchmarkRunner(
         config=_config(tmp_path, modes=["llm_alone"]),
         adapter=_AdapterWithBaseline(),


Davidson3556:
@muddlebee @cerencamkiran  kindly review

PR #2792, opened 2026-06-10
fix(repl): propagate delegated CLI subprocess failures to session state

Devesh36:
## Summary

- `run_cli_command()` now returns `False` when a delegated subprocess fails (non-zero exit, timeout, Ctrl+C, or spawn error) instead of always returning `True`
- Failed slash turns are marked `ok=False` in session history via `session.mark_latest()`, so `/remote`, `/config`, `/onboard`, and other CLI-parity commands no longer look successful after a failed run
- Slash handlers still return `True` so a failed delegated command does not exit the REPL (only `/exit` and `/quit` should do that)

## Problem

`dispatch_slash()` records slash turns as `ok=True` before the handler runs. `run_cli_command()` printed an error line on failure but always returned `True` and never updated session state, so failed delegated commands appeared successful in session history and downstream routing/oracle logic.

## Changes

- `app/cli/interactive_shell/command_registry/cli_parity.py`
  - Track subprocess success/failure in `run_cli_command()`
  - Optional `session` param to mark the latest slash turn failed
  - Update all CLI-parity handlers to pass `session` and always `return True`
  - Mark test picker failures in `_run_test_picker_for_background()`
- `app/cli/interactive_shell/command_registry/integrations.py`
  - Same pattern for `/integrations setup|remove` and `/mcp connect|disconnect`
- `tests/cli/interactive_shell/test_commands.py`
  - Regression tests for non-zero exit, timeout return value, and REPL continuation on failure

### DIFF
--- a/app/cli/interactive_shell/command_registry/cli_parity.py
+++ b/app/cli/interactive_shell/command_registry/cli_parity.py
@@ -43,6 +43,7 @@ def run_cli_command(
     *,
     subprocess_timeout: float | None = None,
     capture_output: bool = False,
+    session: ReplSession | None = None,
 ) -> bool:
     """Helper to delegate complex or interactive Click commands to a child process.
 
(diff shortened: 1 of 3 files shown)


Devesh36:
@greptile-apps review again

Devesh36:
@greptile-apps review again

muddlebee:
@Devesh36 any demo for this? 

> other CLI-parity commands no longer look successful after a failed run
can you explain? examples?

PR #2901, opened 2026-06-21
fix(routing): improve router prompt to classify synthetic alerts reliably (#656)

aniruddhaadak80:
Improves the router prompt to ensure that synthetic incident reports, error logs, and metrics alerts are classified as tracer_data instead of general.


Devesh36:
this also needs more community feedback , please add more context about this in discord chat

Issue #658, opened 2026-04-18
Improve the general-chat fallback prompt so borderline RCA requests do not get generic answers


davincios:
## Summary
The general-chat fallback prompt in `app/constants/prompts.py` is supposed to answer conceptual SRE questions without tools, but it is also the last line of defense when the system misroutes or under-detects a real investigation request. Right now it risks giving generic advice or speculative RCA-style guidance when it should instead push the user back toward evidence-backed investigation.

## Goal
Improve the general-chat fallback prompt so borderline RCA requests, including synthetic alert summaries from `tests/synthetic/`, are handled conservatively and redirected toward evidence-backed analysis instead of answered as if generic SRE knowledge were sufficient.

## Acceptance Criteria
- Borderline synthetic alert summaries do not receive a generic speculative root-cause answer in general mode.
- The fallback prompt more clearly tells the user when live or fixture-backed evidence is required.
- Normal best-practice questions still get useful general answers.
- The change is backed by proof that it avoids a misleading RCA response for at least one synthetic Kubernetes or RDS case.

## Required Proof
Use at least one synthetic alert or alert summary derived from `tests/synthetic/rds_postgres/` or `tests/synthetic/eks/`.

Include all of the following in the PR or issue follow-up:

- the exact synthetic scenario(s) used
- the baseline general-mode response before the prompt change
- the improved general-mode response after the prompt change
- a short explanation of why the new behavior is safer for RCA reliability, and if relevant, a follow-on synthetic RCA run showing the evidence-backed path still reaches the expected result

## Notes
This issue is intentionally about the fallback prompt, not the router. The goal is to make the failure mode safer when something incident-like lands in general mode.

Bensonn5151:
Hi, can I get assigned to this?

nawneet77:
Hi, I'd like to work on this. Could you assign it to me?

A few prior merges in opensre (#1578 #1630 #1722) in the agent-observability area. Happy to follow up here once assigned.

PR #2902, opened 2026-06-21
fix(agent): improve chat prompts to prevent speculative RCA (#657, #658)

aniruddhaadak80:
Improves the tool-backed chat system prompt to enforce strict separation of observations and inferences and rejects speculative root cause claims in general mode, fallback/guidelines included.


Devesh36:
this issue needs more discussion , please add more context about this in discord chat

PR #2938, opened 2026-06-22
feat: smart context eviction

w3joe:
Reapply value-aware context eviction with protected seed evidence and restore the smoke, sizing, and benchmark helpers used to verify overflow behavior.

- Replace oldest-first context eviction with value-aware eviction in the investigation tool loop.
- Protect pre-loop seed evidence from whole-exchange eviction so initial observability data is not discarded under context pressure.
- Add benchmark controls and smoke utilities for comparing `oldest` vs `value` eviction behavior on large CloudOpsBench cases.


w3joe:
@greptile review

PR #2943, opened 2026-06-22
fix(repl): stop submit-time CPR scrubber stripping legitimate input

### LINKED ISSUE #2942: [BUG] REPL submit-time CPR scrubber silently strips legitimate input (e.g. "5R3", "12;34R")
### Summary

The interactive REPL's submit-time CPR scrubber (`_strip_cpr_sequences`) over-matches and silently deletes legitimate characters from a submitted line. Input containing a `<digits>;<digits>R` token (e.g. `12;34R`) or a `<digits>R` immediately followed by a digit/`[` (e.g. `5R3`) has part of the text removed before it reaches the agent.

**Expected:** Only real terminal cursor-position reply (CPR) sequences (`ESC[row;colR`) that leak into stdin are stripped from submitted input; ordinary text is passed through untouched.

**Actual:** The strip runs unconditionally on every submitted line and also matches CPR-shaped substrings that are part of legitimate input:

- `retry 5R3 times`  → `retry 3 times`  (the `5R` is removed)
- `ranges 12;34R okay` → `ranges  okay`  (the `12;34R` is removed)

The corruption is silent — the user gets no indication their input was altered.

### Steps to reproduce

1. Launch the interactive shell: `uv run opensre`
2. At the prompt, type `retry 5R3 times` and press Enter.
3. Observe the text that is echoed / sent to the agent is `retry 3 times`.

Reproduce without a TTY by calling the same function used at submit time:

```
uv run python -c "from app.cli.interactive_shell.runtime.loop import _strip_cpr_sequences as s; print(repr(s('retry 5R3 times')))"
# -> 'retry 3 times'
uv run python -c "from app.cli.interactive_shell.runtime.loop import _strip_cpr_sequences as s; print(repr(s('ranges 12;34R okay')))"
# -> 'ranges  okay'
```

Yes

### How often does it occur?

Under specific conditions

### Operating system

macOS

### Logs and error output

```
$ uv run python -c "from app.cli.interactive_shell.runtime.loop import _strip_cpr_sequences as s; print(repr(s('retry 5R3 times')))"
'retry 3 times'

$ uv run python -c "from app.cli.interactive_shell.runtime.loop import _strip_cpr_sequences as s; print(repr(s('ranges 12;34R okay')))"
'ranges  okay'
```

### Additional context

Root cause is in `_CPR_SEQUENCE_RE` in `app/cli/interactive_shell/runtime/loop.py`, applied unconditionally to every submitted line at `loop.py:380`:

```python
_CPR_SEQUENCE_RE = re.compile(
    r"(?:\x1b\[|\x9b)\d{1,4};\d{1,4}R"   # ESC [ row ; col R
    r"|\[\d{1,4};\d{1,4}R"               # [row;colR without ESC
    r"|\d{1,4};\d{1,4}R"                 # row;colR with no introducer  <-- matches plain text
    r"|\d{1,4}R(?=[\[\d])"               # digitR before another fragment <-- matches plain text
)
```

The last two alternatives have no ESC/CSI introducer, so they match CPR-shaped substrings inside ordinary input. A real CPR reply always arrives with an `ESC[`/CSI introducer; requiring that introducer (or otherwise anchoring so a match can't begin in the middle of a word) would close the gap while still clearing genuine leaked replies.

Severity is low — the trigger patterns (`12;34R`, `5R3`) are uncommon in natural input — but it is a real, silent input-corruption path on `main`.

Davidson3556:
Fixes #2942

Fixes the submit-time CPR scrubber silently deleting legitimate characters from
a submitted REPL line.

`_strip_cpr_sequences` (applied to every submitted line in
`app/cli/interactive_shell/runtime/loop.py`) used a regex whose two
introducer-less branches matched CPR-shaped substrings *inside ordinary input*.
So a line like `retry 5R3 times` (a bare `rowR` before a digit) or
`ranges 12;34R okay` (a bare `row;colR` trailed by a word) had those characters
stripped before the text reached the agent — with no indication to the user.

Changes:
- Constrain the two introducer-less branches of `_CPR_SEQUENCE_RE` to genuine
  CPR context only: a bare `rowR` matches only when immediately followed by
  another CPR fragment (`[`, ESC, CSI, or a `row;colR`), and a bare `row;colR`
  matches only at a boundary (end of input or before another fragment), not
  when trailed by a word.
- Real leaked replies and fragmented bursts (`25;57R`, `7R[25;57R23;57R`,
  `ESC[12;80R`) still strip exactly as before — the existing pinned cases stay
  green.
- Add four regression cases to the strip test for the embedded-in-text inputs.
- Fix two pre-existing mypy errors in the lexer test: prompt_toolkit fragments
  may carry a third (mouse-handler) element, so unpacking `for s, t in
  fragments` is unsafe — switched to indexing `frag[0]`/`frag[1]`.

Verified against the `_strip_cpr_sequences` helper that runs on every submit:

Before (main) — legitimate input corrupted:

image

After (this branch) — input preserved, real CPR still stripped:

   image

Tests:

 image
---

Problem: terminal cursor-position replies (CPR, `ESC[row;colR`) can leak into
the prompt input stream, so the REPL strips them from each submitted line. The
existing regex over-matched — its introducer-less branches treated any
`row;colR` / `rowR` shape as a leak, even when it was part of the user's real
text, silently dropping characters.

Alternatives considered:
1. Remove the introducer-less branches entirely. Rejected: existing behavior
   (and tests) require stripping bare leaked fragments like `25;57R` and bursts
   like `7R[25;57R23;57R` where the ESC/`[` introducer was lost on the leak.
2. Require an ESC/CSI introducer for every match. Rejected for the same reason —
   it would regress the bare-fragment cases the suite pins.
3. (chosen) Keep the introducer-less branches but gate them on real CPR
   context. A leaked bare fragment is always either adjacent to another CPR
   fragment or sits at a boundary; legitimate `5R3` / `12;34R okay` are not.
   This fixes the reported corruption while preserving every pinned case.

Key change: the two introducer-less alternatives in `_CPR_SEQUENCE_RE` gained
lookahead guards — `\d{1,4};\d{1,4}R(?=$|[\[\d\x1b\x9b])` (bare reply only at a
boundary or before another fragment) and `\d{1,4}R(?=\[|\x1b|\x9b|\d{1,4};\d{1,4}R)`
(bare fragment only before another fragment). The introducer-bearing branches
are unchanged.

Edge cases handled/tested: intact replies (`ESC[row;colR`), `[`-only and CSI
(`\x9b`) fragments, standalone bare replies (`25;57R`), collapsed bursts
(`7R[25;57R23;57R`), and real input that merely contains CPR-shaped substrings
(`retry 5R3 times`, `ranges 12;34R okay`, `use r5.xlarge`, `scale to 16RAM
nodes`).

One known limitation (out of scope here): a bare `row;colR` at the very end of
a line is still indistinguishable from a standalone leaked reply and is
stripped — the same case the existing `25;57R` test pins.

---


Davidson3556:
@greptile review

Davidson3556:
@muddlebee @Devesh36 kindly review

Devesh36 (approved):
LG 👍

PR #2959, opened 2026-06-23
feat: evict low-value tool exchanges to fit context

w3joe:
Introduce a heuristics-based eviction strategy to trim tool-call exchanges when enforcing the model context budget. Add a _ToolExchange dataclass and helper functions to identify candidate exchanges, preserve pinned seed messages, and prefer evicting duplicate-only or lower-value (smaller token estimate) exchanges. Replace the old oldest-first trimmer with _trim_lowest_value_tool_pair and keep _trim_oldest_tool_pair as a compatibility wrapper. Add message marking in the investigation agent to tag seed and duplicate-result exchanges, export the new trimmer, update logging, and add tests that cover pinned exchanges, eviction priority, and preservation before truncation.

### DIFF
--- a/app/agent/stages/investigate/agent.py
+++ b/app/agent/stages/investigate/agent.py
@@ -43,6 +43,11 @@ from app.utils.tool_trace import redact_sensitive
 logger = logging.getLogger(__name__)
 
 
+def _mark_messages(messages: list[dict[str, Any]], key: str) -> None:
+    for msg in messages:
+        msg[key] = True
+
+
 def _tools_for_plan(tools: list[RegisteredTool], state: dict[str, Any]) -> list[RegisteredTool]:
     planned_raw = state.get("planned_actions")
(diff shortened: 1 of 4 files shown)


w3joe:
@greptile review

muddlebee:
@w3joe https://github.com/Tracer-Cloud/opensre/issues/3741-> Introduced with context-budget eviction in markers were added for trimming, but no corresponding strip-before-send step was added on the investigation agent’s direct-dict path.

muddlebee:
> The main concern is that agent.py hard-codes the message-key strings rather than importing the constants defined in context_budget.py; a rename of those constants would silently break the tagging without any type error, making seed exchanges evictable and removing duplicate-priority boosts. The behaviour gap between the PR description and what the code actually does is a documentation inconsistency rather than a code defect.

it was flagged by greptile, we should aim to fix the reviews in such critical changes @w3joe :sweat_smile:

PR #2960, opened 2026-06-23
fix(agent) Reuse cached tool results on duplicate investigation calls

### LINKED ISSUE #2905: [In Design][Benchmark] Stop fetching the same data twice
## What
Remember what we already fetched during an investigation. If the AI asks for the same thing again, reuse the saved answer instead of fetching again. If it keeps repeating itself, it's stuck — nudge it or stop.

## Why
Every repeat wastes a full slow round-trip. Easiest big win, very low risk.

## Acceptance criteria
- [ ] A repeated identical request reuses the stored result instead of a new call.
- [ ] A repeated-request loop is detected and handled (nudge or stop).
- [ ] Benchmark shows fewer laps/tool calls with no drop in a1.

## Plan of actions
1. Add a per-investigation cache keyed by tool name + normalized args.
2. Before a tool call, check the cache; on hit, return the stored result (log as reuse).
3. Add a counter; over a threshold of repeats, flag "stuck" and end or redirect.
4. Benchmark on vs. off; keep only if laps drop and a1 holds.

larsspinetta12:
## Summary

- Fixes #2905 — adds a per-investigation tool result cache and uses it as the single source of truth for duplicate detection (replacing `seen_signatures`).
- On duplicate `(tool_name, sorted_args)`, the agent no longer re-executes the tool; it returns the **full wrapped cached result** so the model sees prior output instead of a skip-only note.
- Stagnation handling (nudge + force conclusion after duplicate-only laps) is unchanged.

## Test plan

- [x] `uv run ruff check` on changed files
- [x] `uv run python -m mypy app/agent/stages/investigate/loop.py app/agent/stages/investigate/agent.py`
- [x] `uv run python -m pytest tests/agent/test_investigation.py -k "duplicate or tool_call_signature or stuck_repeating or different_args" -q`
- [ ] CloudOpsBench A/B: compare lap count and tool-call count with cache on vs prior behavior; confirm no a1 regression

### DIFF
--- a/app/agent/stages/investigate/agent.py
+++ b/app/agent/stages/investigate/agent.py
@@ -233,7 +233,8 @@ class ConnectedInvestigationAgent:
             ]
             duplicate_flags = [cached is not None for cached in cached_entries]
             fresh_calls = [
-                tc for tc, is_dup in zip(response.tool_calls, duplicate_flags) if not is_dup
+                tc for tc, cached in zip(response.tool_calls, cached_entries, strict=True)
+                if cached is None
             ]
             for tc in fresh_calls:
                 _record_tool_start(tc)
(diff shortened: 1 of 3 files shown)


larsspinetta12:
@greptile review

PR #2963, opened 2026-06-23
fix(agent): normalize root cause categories in product pipeline (Fixes issue #2910)

### LINKED ISSUE #2910: [In Design][Benchmark] Use clear, consistent words for the cause
## What
opensre often finds the right service but describes the cause in different words, so it's marked wrong. Add a step that maps the cause onto opensre's OWN standard list of cause categories, used on every investigation.

## Why
It already does the hard part (finding the place); we lose points only on wording.
Consistent wording also helps real users.

## Important — do not game the test
Use opensre's own category list, applied in the product on every case. Do NOT just match the benchmark's exact answer words to win the score. The first helps customers; the second is a wrapper that helps no one.

## Acceptance criteria
- [ ] Normalization runs in the product output path, not in the scoring code.
- [ ] The category list is opensre's own, not copied from the benchmark.
- [ ] The a1 gain holds on held-out cases it wasn't tuned on.

## Plan of actions
1. Define opensre's own controlled vocabulary of cause categories.
2. Add a normalization stage at the end of the investigation (free-text → nearest category).
3. Keep it in the product pipeline, not the benchmark/scoring layer.
4. Tune on a subset; validate the lift on held-out cases.
5. Review: "does this help a customer, or only the score?" — must be the former.

larsspinetta12:
## Summary

- Add **product-pipeline** root cause category normalization so investigations emit consistent labels from opensre's own taxonomy (`app/types/root_cause_categories.py`) — not in benchmark scoring code.
- **Prevention:** show the full product taxonomy in investigation prompts for all alert sources (replacing coarse buckets like `database / infrastructure / …` for non-Hermes cases).
- **Correction:** post-parse normalization in the diagnose node — token-normalize input, then apply a hand-curated synonym map; alias targets are applied only when they appear in the alert-source allowed set (Hermes vs non-Hermes guard).
- Aliases are intentionally conservative and hand-curated adjacent labels (e.g. legacy coarse buckets, spacing variants) — not fuzzy matching and not CloudOpsBench vocab copying.

Fixes #2910

## Test plan

- [x] Unit tests: canonical passthrough (`connection_exhaustion`), token normalization (`OOM Killed` → `oom_killed` → `pod_oomkilled`), alias mapping, unknown labels unchanged
- [x] Hermes scope guard: `database` does not map to `connection_exhaustion`; valid Hermes categories like `performance_degradation` pass through
- [x] Prompt tests: non-Hermes prompts include full taxonomy; Hermes prompts remain Hermes-scoped
- [x] `make lint`, `make format-check`, `make typecheck`
- [ ] Held-out benchmark validation (follow-up per issue acceptance criteria)


larsspinetta12:
@greptile review again

Devesh36:
HEY @larsspinetta12 could you please talk with yauhen about this

PR #2985, opened 2026-06-24
fix(cli): distinguish repeated tool-gathering progress lines

### LINKED ISSUE #2984: Differentiate repeated tool-gathering progress lines in the interactive shell
## Problem

When the interactive shell assistant runs its tool-gathering pass, each in-flight tool call prints a dim progress line. Repeated calls to the same tool (for example `query_grafana_metrics` three times in a row) currently render as identical lines:

```text
· gathering data via query_grafana_metrics…
· gathering data via query_grafana_metrics…
· gathering data via query_grafana_metrics…
```

That makes it hard to tell what the assistant is doing during a turn.

## Expected behavior

Progress lines should use human-readable integration/tool labels and include enough context to distinguish back-to-back calls to the same tool (query hints and a repeat index).

## Example

```text
· gathering via Grafana · Mimir — pipeline_runs_total…
· gathering via Grafana · Mimir (2) — sum(rate(http_requests_total[5m]))…
· gathering via Grafana · Mimir (3) — http_errors_total…
```

larsspinetta12:
## Summary
- Replace raw tool-name gathering progress lines with human-readable integration labels (e.g. `Grafana · Mimir`).
- Add query hints from distinguishing tool arguments (`metric_name`, `query`, `service_name`, etc.).
- Add a repeat index suffix when the same tool runs multiple times in one turn.

Fixes #2984

## Test plan
- [x] `uv run python -m pytest tests/cli/interactive_shell/chat/test_tool_gathering.py -q`
- [ ] In the REPL, ask a broad question that triggers multiple calls to the same integration tool and confirm each progress line is distinct.

### DIFF
--- a/app/cli/interactive_shell/chat/tool_gathering.py
+++ b/app/cli/interactive_shell/chat/tool_gathering.py
@@ -106,10 +106,11 @@ def _format_gathering_progress_line(
     call_display = f"{source} · {label}" if label else source
     if repeat_index > 1:
         call_display = f"{call_display} ({repeat_index})"
+    safe_display = escape(call_display)
     hint = _tool_input_hint(tool_input)
     if hint:
-        return f"· gathering via {call_display} — {escape(hint)}…"
-    return f"· gathering via {call_display}…"
+        return f"· gathering via {safe_display} — {escape(hint)}…"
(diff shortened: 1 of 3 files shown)


larsspinetta12:
@greptile review again

PR #2997, opened 2026-06-24
fix: add OCI labels and use PORT env var in health check

kawacukennedy:
## Summary

Minor Dockerfile improvements:

1. **OCI labels** — Added `org.opencontainers.image.source`, `description`, and `licenses` for container metadata
2. **Health check fix** — Now reads `PORT` from the environment variable (default 8000) instead of hardcoding 8000, so it works correctly regardless of runtime port config

### Testing

Config-only change; no behavioral impact on default usage.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,10 @@
 
 FROM python:3.12-slim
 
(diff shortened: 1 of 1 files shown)


kawacukennedy:
@greptile review

PR #3051, opened 2026-06-25
refactor(integrations): collapse repeated env-loader boilerplate in load_env_integrations

### LINKED ISSUE #3043: [IMPROVEMENT] _catalog_impl: decompose ~1,070-line load_env_integrations() (duplicated per-integration blocks)
### Current state

`load_env_integrations()` in `app/integrations/_catalog_impl.py:372` is a single function that runs from roughly line 372 to line 1440 — about **1,070 lines** with no internal decomposition. The body is the same block repeated ~30 times, once per integration:

```python
<x>_token = os.getenv("X_TOKEN", "").strip()
if <x>_token:
    <x>_config = build_<x>_config({...})
    integrations.append(_active_env_record("<x>", <x>_config.model_dump()))
```

This is both an oversized unit (far past the >400-line guideline, in one function) and heavily **duplicated logic** — the env-read → build-config → append-record shape is copy-pasted per integration. Adding an integration means appending another ~10-line block in the middle of a 1,000-line function.

### Desired state

Collapse the repeated per-integration blocks into a data-driven loop over a registry of small loader descriptors / functions, e.g. an `(env_var(s), builder, service_name)` table or one `_load_<x>()` helper per integration, iterated in `load_env_integrations()`. Integrations that need bespoke env handling (e.g. multi-var or mode-based ones like OpenClaw) stay as their own small loaders rather than being forced into the table.

This is a central catalog-resolution path, so the change should:
- preserve the exact set and shape of records `load_env_integrations()` produces,
- be covered by tests that lock the current output for a representative env matrix before refactoring,
- land incrementally if needed (group of integrations per PR) rather than one giant rewrite.

Raising this for scoping/sign-off on the approach before implementation, given how central it is.

Davidson3556:
Fixes #3043

`load_env_integrations()` in `app/integrations/_catalog_impl.py` repeated the same
scaffolding ~40 times — once per integration:

- `build_config(...) -> config.model_dump(exclude={"integration_id"}) -> _active_env_record -> append`
- a `try/except _report_env_loader_failure` guard around config construction
- a multi-instance preamble (`_parse_instances_env` + a "clear the single vars" branch)

This extracts three small helpers and routes each block through them:

- `_emit(integrations, service, config)` — the build→dump→append tail
- `_emit_guarded(integrations, service, build)` — the same, wrapped in the report-and-skip guard
- `_emit_instances(integrations, env, service)` — appends a multi-instance record if present and returns whether it did, so callers read single-instance vars only on `False`

**This is a subtraction, not an addition.** The function drops from ~1,070 to ~865
lines and the file nets ~170 fewer lines. No behavior change — each integration's
unique env reads, defaults, and payload shape are untouched; only the repeated
scaffolding is collapsed.

> Note: this supersedes the earlier version of this PR, which split the function into
> 40+ per-integration helpers. That *added* ~250 lines, so per maintainer feedback it
> was reworked into this subtractive form. Force-pushed in place.

Behavior is locked by a characterization test that pins the exact output (content
**and** emission order) of `load_env_integrations()` for a 43-integration env matrix.
The golden snapshot was generated from the original implementation; the test passes
unchanged, proving byte-identical output.

```
$ uv run pytest tests/integrations/test_load_env_integrations_characterization.py -q
2 passed

$ uv run pytest tests/integrations/ -q
1580 passed, 1 failed
# the 1 failure is test_copilot_adapter::test_detect_when_binary_not_found, a
# pre-existing env-only failure (a real `copilot` binary on PATH) that also fails
# on a clean main; unrelated to this change.

All checks passed!  |  files already formatted  |  Success: no issues found in 864 source files
```

Line count: `load_env_integrations` ~1,070 -> 865 lines; `_catalog_impl.py` 1,719 -> 1,549.

---

I first wrote a characterization test that captures the exact records the function
returns for a broad env matrix, so any drift during the refactor is caught. I then
collapsed the three repeated patterns into helpers. `_emit_guarded` takes a build
callback and preserves the report-and-skip semantics exactly (build in a try; on
failure, `_report_env_loader_failure` and return). `_emit_instances` lets each
multi-instance block drop its "set single vars to empty" branch and instead read
single-instance vars only when no multi-instance record was produced. Blocks with
bespoke shapes — AWS's role/key split, the shared-Twilio WhatsApp+SMS records, the
MCP `connection_verified` payloads, and the raw-dict integrations — are left inline
since forcing them through a helper would add flags rather than remove lines. I kept
`exclude_integration_id=False` for the handful of integrations that historically
dumped the full model.

---

### DIFF
--- a/app/integrations/_catalog_impl.py
+++ b/app/integrations/_catalog_impl.py
@@ -369,13 +369,11 @@ def _active_env_record(
     }
 
 
-def load_env_integrations() -> list[dict[str, Any]]:
-    """Build integration records from local environment variables."""
-    integrations: list[dict[str, Any]] = []
-
+def _load_grafana() -> list[dict[str, Any]]:
+    records: list[dict[str, Any]] = []
     grafana_multi = _parse_instances_env("GRAFANA_INSTANCES", "grafana")
     if grafana_multi is not None:
-        integrations.append(grafana_multi)
+        records.append(grafana_multi)
         grafana_endpoint = ""
         grafana_api_key = ""
     else:
@@ -388,7 +386,7 @@ def load_env_integrations() -> list[dict[str, Any]]:
                 "api_key": grafana_api_key,
             }
         )
-        integrations.append(
+        records.append(
             _active_env_record(
                 "grafana",
                 {
@@ -397,10 +395,14 @@ def load_env_integrations() -> list[dict[str, Any]]:
                 },
             )
         )
+    return records
 
+
+def _load_datadog() -> list[dict[str, Any]]:
+    records: list[dict[str, Any]] = []
     datadog_multi = _parse_instances_env("DD_INSTANCES", "datadog")
     if datadog_multi is not None:
-        integrations.append(datadog_multi)
+        records.append(datadog_multi)
         datadog_api_key = ""
         datadog_app_key = ""
         datadog_site = ""
@@ -416,16 +418,20 @@ def load_env_integrations() -> list[dict[str, Any]]:
                 "site": datadog_site,
             }
         )
-        integrations.append(
+        records.append(
             _active_env_record(
                 "datadog",
                 datadog_config.model_dump(exclude={"integration_id"}),
             )
         )
+    return records
 
+
+def _load_groundcover() -> list[dict[str, Any]]:
+    records: list[dict[str, Any]] = []
     groundcover_multi = _parse_instances_env("GROUNDCOVER_INSTANCES", "groundcover")
     if groundcover_multi is not None:
-        integrations.append(groundcover_multi)
+        records.append(groundcover_multi)
         groundcover_api_key = ""
     else:
         groundcover_api_key = (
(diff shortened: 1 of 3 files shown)


Davidson3556:
Addressed the ordering-coverage feedback in 2676c9c5: the golden snapshot is now captured in the loader's emission order (no pre-sort), so the comparison pins both content and order. The order test also asserts the full service sequence plus the WhatsApp→Twilio adjacency, instead of only checking the first element.

@greptile review

Davidson3556:
Reworked per the feedback that this should *subtract*, not add. The earlier version split this into 40+ per-integration helpers, which added ~250 lines — fair call that that wasn't really refactoring.

Force-pushed a subtractive version instead: three small helpers (`_emit`, `_emit_guarded`, `_emit_instances`) absorb the scaffolding that repeated ~40x, so each block shrinks. The function goes ~1,070 → 865 lines and the file nets ~170 fewer lines. Behavior is unchanged and pinned byte-for-byte by the characterization test (content + emission order).

@greptile review

PR #3058, opened 2026-06-25
fix(repl): align /investigate RCA output and feedback with CLI

### LINKED ISSUE #3056: fix(repl): RCA report and feedback missing after /investigate
## Problem

Running `/investigate` (especially from the TTY template picker) does not match the standalone CLI experience:

- Final RCA report often does not appear in scrollback (only a one-line root-cause recap from the feedback helper).
- Post-investigation feedback prompt is missing or shows static hint text without the blocking rating menu.
- Bare `/investigate` reopens the template picker after each run instead of returning to the REPL prompt.
- When investigation fails before `publish_findings` (LLM/API error, provider rejection, timeout, etc.), there is **no RCA section box** — only a terse one-line `investigation failed: …` outside the report layout. The user loses context on what stage failed and why.

## Expected

**Success path:** Same formatted RCA output as `opensre investigate` (via `render_report(slack_message)`), then the accuracy feedback menu, then the normal REPL prompt.

**Failure path:** If the pipeline cannot produce a report, render a **meaningful error in the RCA section box** (same visual area as Findings / Recommended Actions / Cited Evidence on success) — not just a stray log line. Include at minimum:

- What failed (stage + short headline, e.g. investigation agent LLM invoke)
- The actionable error message (provider text, e.g. `reasoning_content must be passed back`)
- Optional next step hint when we have one (config, model switch, retry)

Do **not** show the feedback prompt when investigation failed with no report.

## Root cause

- REPL session path never called `prompt_investigation_feedback`.
- `render_report()` wrote line-by-line under `patch_stdout`, so output was often lost in the REPL.
- Feedback used `repl_choose_one` after streaming (cursor-unstable); CLI already used cursor-safe `_run_select`.
- Interactive investigate menu looped after every selection.
- StreamRenderer / foreground investigation error path exits without a terminal report renderer for failures (no `slack_message` → silent RCA area).

## Related

- PR #3057 addresses the success-path REPL report + feedback wiring.
- Follow-up: thinking-mode / `reasoning_content` multi-turn tool-call failures (e.g. OpenAI-compatible 400 during lap 1+) are a separate provider-message-shape bug; failure UX above should still surface those errors clearly in the RCA box.

muddlebee:
## Problem

Running `/investigate` in the interactive shell (especially via the TTY template picker) diverged from standalone `opensre investigate`:

- Final RCA report often did not appear in scrollback (only a one-line root-cause recap from the feedback helper).
- Post-investigation feedback prompt was missing.
- Bare `/investigate` re-opened the template picker after each run instead of returning to the REPL prompt.
- When the pipeline finished without a published report (e.g. it raised after diagnosis), the UI went silently blank.

## Root cause

- The REPL session path (`run_foreground_investigation`) never called `prompt_investigation_feedback`. Feedback was only ever wired into the standalone streaming path (`run_investigation_cli_streaming`) when the feature landed (#2371).
- `render_report()` wrote line-by-line through a fresh Rich `Console`. Under the REPL's `patch_stdout(raw=True)` those writes interleave with the proxy and the report body is dropped from scrollback.
- The interactive investigate menu looped after every selection.
- `StreamRenderer._print_report` only printed a message for the *noise* and *zero-events* cases; the "events received but no report" case fell through to a bare `return`.

## Fix

- **terminal.py** — render the report into a single buffer and emit it with one `\r\n`-normalised `sys.stdout.write` so every line starts at column zero under `patch_stdout`. Standalone/non-TTY output is unchanged.
- **foreground_investigation.py** — call `prompt_investigation_feedback(final_state)` after the report (covers `/investigate` and free-text/agent-routed runs), with `console=None` so the cursor-safe `_run_select` (per-line erase) is used instead of `repl_choose_one`, whose block-erase is unstable after Rich Live streaming.
- **investigation.py** — bare `/investigate` picker now runs once and returns to the REPL prompt.
- **renderer.py** — when the stream finishes with no published report, surface the captured root cause plus a clear "Investigation finished without a full report" notice instead of going blank.

## Verification

- `make lint`, `make format-check`, `make typecheck` clean.
- Full unit suite: **9560 passed, 0 failures** (the 113 errors are live-LLM routing tests requiring `ANTHROPIC_API_KEY` — CI-only, unrelated).

Closes #3056

### DIFF
--- a/app/cli/interactive_shell/command_registry/investigation.py
+++ b/app/cli/interactive_shell/command_registry/investigation.py
@@ -60,21 +60,24 @@ def _interactive_investigate_menu(session: ReplSession, console: Console) -> boo
     choices.append(("__browse__", "custom file path…"))
     choices.append(("done", "done"))
 
-    while True:
-        target = repl_choose_one(
-            title="investigate",
-            breadcrumb=root,
-            choices=choices,
-        )
-        if target is None or target == "done":
+    # Run the picker once, then return to the REPL prompt — re-opening the menu
+    # after each investigation buried the RCA report and felt like a loop the
+    # user could not escape. The standalone CLI runs a single investigation and
+    # returns; bare /investigate now matches that.
+    target = repl_choose_one(
+        title="investigate",
+        breadcrumb=root,
+        choices=choices,
+    )
+    if target is None or target == "done":
+        return True
+    if target == "__browse__":
+        custom_path = _prompt_investigate_path(console)
+        if custom_path is None:
             return True
-        if target == "__browse__":
-            custom_path = _prompt_investigate_path(console)
-            if custom_path is None:
-                continue
-            target = custom_path
-        _cmd_investigate_file(session, console, [target])
-        repl_section_break(console)
+        target = custom_path
+    _cmd_investigate_file(session, console, [target])
+    return True
 
 
 def _prompt_investigate_path(console: Console) -> str | None:
--- a/app/cli/interactive_shell/runtime/foreground_investigation.py
+++ b/app/cli/interactive_shell/runtime/foreground_investigation.py
@@ -59,6 +59,11 @@ def run_foreground_investigation(
     from app.cli.interactive_shell.ui.feedback import prompt_investigation_feedback
     from app.cli.interactive_shell.ui.key_reader import restore_stdin_terminal
 
+    # The explicit pre-call (kept identical to the CLI path) primes the terminal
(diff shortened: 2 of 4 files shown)


muddlebee:
Before : the investigate command is popping up automatically again and the feedback menu isn't showing. 
image

After
image

muddlebee:
@greptile review

muddlebee:
@greptile review

PR #3060, opened 2026-06-25
fix(repl): prompt blocks during investigation - cannot queue commands async  

### LINKED ISSUE #3059: bug(repl): prompt blocks during investigation — cannot queue commands async
## Summary

While an investigation is running in the interactive REPL, the input prompt is blocked and users cannot type or queue additional commands. Previously, the REPL stayed responsive during long-running investigations — users could submit multiple commands in an async manner (e.g. `/help`, `/status`, follow-up questions) while the investigation streamed in the background.

## Expected behavior

- After starting an investigation (`/investigate generic`, bare `/investigate` menu selection, or agent-triggered investigate), the `[n] >` prompt should return immediately.
- Users can queue and submit further commands while the investigation runs.
- Investigation output streams concurrently without holding exclusive stdin for the full RCA duration.

## Actual behavior

- The prompt does not accept new input until the investigation completes.
- Users cannot queue slash commands or free-text turns while a foreground investigation is in progress.
- This feels like a regression from the earlier concurrent REPL behavior.

## Reproduction

1. Start the REPL in a TTY: `uv run opensre`
2. Run an investigation, e.g.:
   - `/investigate generic`, or
   - bare `/investigate` → pick a sample from the menu
3. While the investigation is streaming, try typing another command (`/help`, `/status`, etc.)
4. **Observe:** no prompt / input blocked until investigation finishes

## Likely root cause

Exclusive-stdin gating + `queue.join()` in the main loop blocks the next `prompt_async()` for investigate-related turns:

```python
# app/cli/interactive_shell/runtime/loop.py
wait_for_dispatch = dispatch_needs_exclusive_stdin(stripped, session)
await state.queue.put(stripped)
if wait_for_dispatch:
    await state.queue.join()  # blocks next prompt until dispatch completes
```

`/investigate` (bare, no args) is registered in `_EXCLUSIVE_STDIN_MENU_COMMANDS` (`app/cli/interactive_shell/runtime/dispatch.py`), which triggers `queue.join()`. That was intentional for CPR/picker UX (#2470 — exclusive stdin while inline menus run), but the same gate appears to cover the **entire** dispatch turn—including the long-running foreground RCA after menu selection—so the prompt never reopens until investigation completes.

For exclusive turns, dispatch also skips `repl_safe_progress_scope()` and uses full Rich Live rendering, which is correct only while stdin is truly exclusive (picker phase), not for multi-minute investigations.

## Related code / history

- `app/cli/interactive_shell/runtime/loop.py` — `_run_one_dispatch`, main loop `queue.join()`
- `app/cli/interactive_shell/runtime/dispatch.py` — `_EXCLUSIVE_STDIN_MENU_COMMANDS` includes `/investigate`
- `app/cli/interactive_shell/command_registry/investigation.py` — foreground vs background investigation paths
- PR #2470 — added `/investigate` to exclusive stdin to fix picker UX and restore live stream

## Suggested fix direction

- Reserve exclusive stdin + `queue.join()` only for the **interactive picker/menu phase** (bare `/investigate`, path prompt), not for the subsequent long-running RCA.
- For foreground investigations with args (`/investigate generic`, file paths), keep non-exclusive dispatch so the prompt returns immediately (may need `repl_safe_progress_scope()` + append-only streaming to avoid Live/CPR contention).
- Consider defaulting investigations to background mode (`/background on`) or auto-releasing exclusive stdin once the picker completes.
- Add a REPL regression test: start investigation, assert next `prompt_async` is reachable before investigation task completes.

## Workaround

Enable session background mode before investigating:

```
/background on
/investigate generic
```

Background investigations return the prompt immediately (`start_background_*` in `investigation.py`).

## Labels

bug, cli

muddlebee:
Fixes #3059

- **Prompt blocking (#3059):** bare `/investigate` menu selections queue a normal `/investigate <target>` auto-command instead of running foreground RCA on the exclusive-stdin turn, so `queue.join()` no longer blocks the prompt for the full investigation.
- **Progress formatting:** in REPL-safe mode (`repl_safe_progress_scope`), skip the duplicate raw `● diagnose_root_cause` finish line — the graph node already emits the same timestamped READ/PLAN/INVEST/DIAG event-log rows via `ProgressTracker`.
- Regression tests for menu deferral, exclusive-stdin routing, and diagnose finish in repl-safe mode.

**Before:** bare `/investigate` → pick sample → prompt blocked until RCA completes; diagnose phase also printed a separate `● diagnose_root_cause …` line outside the event log.

**After:** menu closes → prompt returns → RCA runs on queued turn; diagnose completion stays on the unified timeline.

Manual check:
```text
uv run opensre
/investigate
# pick generic — prompt should return immediately
# during RCA, timeline rows match READ/PLAN/INVEST style (no extra ● diagnose line)
```

---

Exclusive stdin stays scoped to the picker only. Long RCA runs on a non-exclusive queued turn. Diagnose streaming already skipped Rich Live in repl-safe mode, but `finish()` still printed a second bullet line — removed that so formatting matches the alert header / INVEST lap rows above it.

---

---

### DIFF
--- a/app/cli/interactive_shell/command_registry/investigation.py
+++ b/app/cli/interactive_shell/command_registry/investigation.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import shlex
 from pathlib import Path
 
 from rich.console import Console
@@ -49,6 +50,16 @@ def _interactive_template_menu(session: ReplSession, console: Console) -> bool:
         repl_section_break(console)
 
 
+def _queue_investigate_target(session: ReplSession, target: str) -> None:
+    """Defer a menu selection to a normal ``/investigate <target>`` turn.
+
+    The interactive picker needs exclusive stdin, but long-running RCA must not
+    hold it — queue the resolved target so the loop auto-submits it on the next
+    prompt iteration without ``queue.join()`` blocking.
+    """
+    session.queue_auto_command(f"/investigate {shlex.quote(target)}")
+
+
 def _interactive_investigate_menu(session: ReplSession, console: Console) -> bool:
     from app.cli.interactive_shell.data_store.constants import SAMPLE_ALERT_OPTIONS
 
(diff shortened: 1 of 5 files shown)


muddlebee:
@greptile review

muddlebee:
@greptile review

PR #3099, opened 2026-06-26
fix(routing): dispatch explicit investigate without integrations

### LINKED ISSUE #3098: Strengthen explicit investigate routing and re-enable NL investigation in REPL
## Problem

When a user sends an explicit natural-language investigate instruction (e.g. `investigate why the orders-api keeps OOM-killing its pods`), the action planner should always emit `investigation_start` — even when **no integrations are connected**. The planner system prompt already stated this rule, but live routing scenario **336** failed consistently: the planner returned `assistant_handoff` instead of dispatching an investigation.

Additionally, `INTERACTIVE_SHELL_INVESTIGATION_ENABLED` is still `False` on `main`, which disables the planner's `investigation_start` tool and skips investigation routing scenarios in CI.

## Expected behavior

- Explicit `investigate` / `RCA` / `diagnose` / `analyze` / `root-cause` verbs → `investigation_start` **always**, regardless of `CONNECTED INTEGRATIONS`.
- Implicit diagnostic questions (no explicit verb) remain gated: dispatch only when ≥1 integration is connected; hand off when none.
- `INTERACTIVE_SHELL_INVESTIGATION_ENABLED = True` so REPL users can trigger investigations from natural language again.

## Proposed fix (Option A)

1. Strengthen the planner system prompt:
   - Make explicit investigate the highest-priority rule.
   - Add canonical example (`orders-api` OOM prompt).
   - Clarify anti-handoff behavior when integrations are `none`.
   - Resolve conflicting "when unsure → handoff" and bottom exception rules.
2. Flip `INTERACTIVE_SHELL_INVESTIGATION_ENABLED` to `True`.
3. Extend unit test assertions for the new prompt language.

## Test plan

- [ ] `pytest cli/interactive_shell/routing/tests/orchestration/test_llm_action_planner_fallback.py`
- [ ] Live scenario 336: `investigate why the orders-api keeps OOM-killing its pods` with zero integrations
- [ ] Live scenario 340: explicit RCA + pasted JSON

larsspinetta12:
## Summary

- Strengthens the action-planner system prompt (Option A) so explicit `investigate` / `RCA` / `diagnose` / `analyze` / `root-cause` verbs always emit `investigation_start`, even when `CONNECTED INTEGRATIONS` is `none` or `unknown`.
- Adds the canonical `orders-api` OOM example, clarifies anti-handoff behavior, and resolves conflicting "when unsure → handoff" / bottom exception rules.
- Re-enables natural-language investigation routing by setting `INTERACTIVE_SHELL_INVESTIGATION_ENABLED = True`.

Fixes #3098

## Test plan

- [x] `uv run python -m pytest cli/interactive_shell/routing/tests/orchestration/test_llm_action_planner_fallback.py -v`
- [x] Live scenario 336: `investigate why the orders-api keeps OOM-killing its pods` (zero integrations) — planning + oracle passed
- [x] Live scenario 340: explicit RCA + pasted JSON — planning + oracle passed

### DIFF
--- a/cli/interactive_shell/routing/handle_message_with_agent/orchestration/feature_flags.py
+++ b/cli/interactive_shell/routing/handle_message_with_agent/orchestration/feature_flags.py
@@ -19,7 +19,7 @@ from __future__ import annotations
 #
 # Flip to ``True`` to restore natural-language investigation routing (which also
 # re-enables the investigation routing scenarios that skip while this is False).
-INTERACTIVE_SHELL_INVESTIGATION_ENABLED = False
+INTERACTIVE_SHELL_INVESTIGATION_ENABLED = True
 
 
 def investigation_loop_enabled() -> bool:
--- a/cli/interactive_shell/routing/handle_message_with_agent/orchestration/llm_action_planner/prompting.py
+++ b/cli/interactive_shell/routing/handle_message_with_agent/orchestration/llm_action_planner/prompting.py
@@ -31,7 +31,14 @@ def _connected_integrations_block(session: Any | None) -> str:
         listing = "none"
     else:
         listing = "unknown"
-    return f"CONNECTED INTEGRATIONS (this install, right now): {listing}\n\n"
+    gate_note = ""
+    if listing in ("none", "unknown"):
+        gate_note = (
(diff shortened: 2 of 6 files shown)


larsspinetta12:
@greptile review again

PR #3126, opened 2026-06-26
fix(repl): show turn [1] at startup and integrations in hint bar

### LINKED ISSUE #3125: Fix REPL prompt turn counter and restore integrations in idle hint bar
## Problem

Three related REPL UX regressions on startup:

1. **Blank first turn number** — The prompt counter used `len(session.history)`, so a fresh session showed no `[N]` prefix on the first prompt. After the first chat message was recorded, the counter jumped to `[1]`.
2. **Missing integration bar** — Connected integrations appeared in the startup welcome panel but not in the live idle hint line above the prompt (`/ for commands · ↑↓ history`).
3. **Submitted first turn looked unnumbered** — Because history was still empty at submit time, the echoed user line also lacked a turn number until the second prompt.

## Expected behavior

- First prompt shows `[1] ❯` immediately.
- Each subsequent prompt increments (`[2]`, `[3]`, …).
- When integrations are hydrated at boot, the idle hint bar lists them (e.g. `Datadog · GitHub · Grafana`).

## Proposed fix

- Use `len(session.history) + 1` for the upcoming turn index in prompt rendering and submitted-line echo.
- Add `resolve_idle_hint_ansi(session)` and wire it into the interactive loop so configured integrations appear in the hint bar.

larsspinetta12:
## Summary

- Fixes blank first-turn numbering: prompt counter now uses `len(session.history) + 1`, so the first prompt shows `[1] ❯` instead of an unnumbered `❯`.
- Restores connected integrations in the live idle hint bar above the prompt (`/ for commands · ↑↓ history · Datadog · GitHub · …`).
- Submitted user lines use the same upcoming-turn index, so the first message echoes as `[1] ❯ …` instead of appearing unnumbered.

Fixes #3125

## Test plan

- [x] `uv run python -m pytest tests/cli/interactive_shell/prompting/test_prompt_surface.py -q`
- [x] Restart `uv run opensre` — first prompt shows `[1] ❯` and integration names in the hint bar when configured
- [x] Send a chat message — submitted line shows `[1]`, next prompt shows `[2]`

### DIFF
--- a/cli/interactive_shell/prompting/prompt_surface.py
+++ b/cli/interactive_shell/prompting/prompt_surface.py
@@ -29,6 +29,7 @@ from cli.interactive_shell.routing.handle_message_with_agent.command_dispatch.ca
 )
 from cli.interactive_shell.runtime import ReplSession
 from cli.interactive_shell.ui import theme as ui_theme
+from cli.interactive_shell.ui.banner_state import integration_display_name
 from cli.interactive_shell.ui.choice_menu import repl_tty_interactive
 
 _PROMPT_RULE_CHAR = "─"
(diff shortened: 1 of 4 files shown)


larsspinetta12:
@greptile review again

PR #3131, opened 2026-06-26
fix(llm): normalize OpenAI tool schemas for DeepSeek API

### LINKED ISSUE #3129: [bug] DeepSeek tool calls fail (HTTP 400) — missing `type` field in tool schema serialization
## Summary
DeepSeek does not work with the OpenSRE tool. Tool/investigation calls fail with an HTTP 400 from DeepSeek's API:

```
[LLM provider: deepseek] Deepseek request rejected (HTTP 400): Error code: 400 -
{'error': {'message': "Failed to deserialize the JSON body into the target type:
tools[0]: missing field `type` at line 1 column 14421",
'type': 'invalid_request_error', 'param': None, 'code': 'invalid_request_error'}}
```

## Root cause (suspected)
This is on our side. DeepSeek's API requires a `type` field on every tool definition (e.g. `"type": "function"`), which OpenAI/Claude don't strictly enforce. Our tool-wrapping / LLM dispatch layer is likely not injecting `type` when routing tools to DeepSeek, so the schema fails deserialization.

## Proposed fix
- Add a DeepSeek-specific tool-schema normalizer that guarantees every tool entry includes `type: "function"` (and any other DeepSeek-required fields) before dispatch.
- Add a regression/synthetic test that exercises a tool call against the DeepSeek provider.

## Impact
- Blocks DeepSeek users entirely.
- Likely contributes to onboarding failures (a user failed onboarding 5x in a row) — every onboarding step that makes a tool call fails silently with HTTP 400.

## Priority
🔴 High priority — flagged by Lars as a high-priority hackathon target (Jun 26). Blocks users entirely and the fix is well-scoped.

_Source: Lars, #product (Jun 26 hackathon)_

muddlebee:
## Summary
- Normalize `function.parameters` in OpenAI-compatible tool specs using the existing Bedrock `normalize_tool_input_schema` helper so every nested JSON Schema node has an explicit `type`
- Apply the fix in both investigation (`agent_llm_client`) and interactive chat (`chat_sdk_adapter`) paths
- Add registry-wide contract coverage for OpenAI-compatible tool serialization plus a focused `anyOf` regression test

Fixes #3129

## Root cause
DeepSeek's native API (`api.deepseek.com`) uses a strict deserializer that rejects tool parameter schemas with nested nodes missing `type` (e.g. Pydantic `anyOf` for optional fields). OpenAI, Claude, and OpenRouter are lenient, so the raw `public_input_schema` worked there but failed on DeepSeek with HTTP 400.

## Test plan
- [x] `uv run python -m pytest tests/services/test_investigation_tool_schemas.py -q`
- [x] `uv run python -m pytest tests/services/test_agent_llm_client.py::test_build_openai_tool_specs_normalizes_anyof_optional_parameters -q`
- [x] `make lint && make format-check`

### DIFF
--- a/services/agent_llm_client.py
+++ b/services/agent_llm_client.py
@@ -16,12 +16,12 @@ import time
 from dataclasses import dataclass, field
 from typing import Any
 
-from services.bedrock_converse import normalize_tool_input_schema
 from services.llm_retry import (
     LLMCreditExhaustedError,
     extract_retry_after_seconds,
     is_credit_exhausted_error,
 )
+from services.tool_schema_normalize import normalize_openai_tool_input_schema
 
 logger = logging.getLogger(__name__)
 
@@ -128,7 +128,7 @@ def _openai_tool_schema(tool: Any) -> dict[str, Any]:
         "function": {
             "name": tool.name,
             "description": tool.description,
-            "parameters": normalize_tool_input_schema(tool.public_input_schema),
+            "parameters": normalize_openai_tool_input_schema(tool.public_input_schema),
         },
     }
 
--- a/services/agent_llm_client.py
+++ b/services/agent_llm_client.py
@@ -16,6 +16,7 @@ import time
 from dataclasses import dataclass, field
 from typing import Any
 
+from services.bedrock_converse import normalize_tool_input_schema
 from services.llm_retry import (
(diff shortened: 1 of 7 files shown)


muddlebee:
@greptile review

PR #3140, opened 2026-06-26
fix(release): bundle stdlib platform.py so the binary smoke test passes

davincios:
## Summary

The **Release** workflow has been failing on `main` (every `workflow_dispatch` run) at the **"Smoke test binary"** step on **all** platforms (macOS arm64/x64, Linux arm64, Windows). The freshly built onefile binary crashes on `opensre --version`:

```
ImportError: Unable to load stdlib platform module — sysconfig path
'.../T/_MEI.../lib/python3.13' does not contain platform.py
```

**Root cause:** the first-party `platform/` package (added in #3067) shadows the stdlib `platform` module and re-exports its API by loading the stdlib source at import time (`platform/__init__.py::_load_stdlib_platform`). That loader assumed `sysconfig.get_path("stdlib")/platform.py` exists in frozen builds — but PyInstaller does **not** lay out the stdlib as loose `.py` files, and because the first-party package wins the `platform` name during analysis, the genuine `platform.py` is bundled **nowhere** (verified via `PYZ-00.toc` + `base_library.zip`). So the lookup could never succeed and `opensre --version` aborted before doing anything.

## Fix

- **Build:** stage a copy of the genuine stdlib `platform.py` and bundle it under `_opensre_stdlib_platform/` via `--add-data` (relative path → no Windows drive-colon ambiguity).
- **Runtime:** `_load_stdlib_platform()` now probes `sys._MEIPASS/_opensre_stdlib_platform/platform.py` first in frozen builds, then falls back to the sysconfig stdlib path (source checkouts) and an `os.__file__`-relative last resort.
- **Tests:** add `tests/packaging/test_platform_stdlib_loader.py` covering loader path ordering, loading from a bundled copy, and a drift guard tying the workflow's `--add-data` destination to `_FROZEN_STDLIB_DIR`.

## Test plan

- [x] Reproduced the crash locally: built the onefile binary with the exact workflow command → `opensre --version` raised the same `ImportError`.
- [x] After the fix, rebuilt the binary → `./dist/opensre --version` prints `opensre, version 0.1` (exit 0) and `./dist/opensre -h` exits 0 (exact CI smoke-test invocations).
- [x] `make lint` — passes
- [x] `make typecheck` — passes (894 files)
- [x] `make test-cli-smoke` — 16 passed, 1 skipped
- [x] `tests/packaging/test_platform_stdlib_loader.py` — 5 passed
- [ ] Confirm green on the Release workflow's `build-binaries` matrix (all OSes) after merge / via `workflow_dispatch`.

> Note: this PR addresses the deterministic, all-platform **Release (CD)** failure. The separate, intermittent `Routing Live` flake on `shard 2` (`501-nitro-connect-then-investigate`, a live-LLM scenario) is not touched here, per the live-routing policy against forcing/disabling live scenarios.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -239,6 +239,18 @@ jobs:
         shell: bash
         run: uv sync --frozen --extra release-binary
 
+      - name: Stage stdlib platform module for bundling
+        shell: bash
+        run: |
+          set -euo pipefail
+          # The first-party ``platform`` package shadows the stdlib ``platform``
+          # module. PyInstaller does not lay out the stdlib as loose ``.py`` files,
+          # so bundle a copy of the genuine module that platform/__init__.py can
+          # load from ``sys._MEIPASS`` at runtime (see _FROZEN_STDLIB_DIR).
+          rm -rf .stdlib_vendor
+          mkdir -p .stdlib_vendor
+          uv run python -c "import os, shutil, sysconfig; src = os.path.join(sysconfig.get_path('stdlib'), 'platform.py'); shutil.copy(src, os.path.join('.stdlib_vendor', 'platform.py')); print('Staged stdlib platform from ' + src)"
+
       - name: Build binary
         run: >-
           uv run pyinstaller cli/__main__.py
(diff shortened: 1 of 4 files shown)


davincios:
@copilot resolve the merge conflicts in this pull request

PR #3142, opened 2026-06-26
fix(github): strip username metadata before MCP config validation

### LINKED ISSUE #3141: GitHub login gate crashes when store credentials include username
## Problem

After #3134, GitHub credentials persisted in `~/.opensre/integrations.json` include a `username` field for analytics and the welcome banner. `build_github_mcp_config()` passed that dict directly into `GitHubMCPConfig` (`StrictConfigModel`, `extra=forbid`), which raised:

```
ValidationError: Unexpected field 'username'
```

The first-launch login gate called `github_integration_is_configured()` on REPL startup, hit that validation error, and blocked interactive users with:

> GitHub sign-in is required to use OpenSRE, but the sign-in step could not run.

## Expected behavior

- Stored `username` metadata is ignored when building MCP transport config
- Invalid store credentials are treated as unconfigured (no crash)
- Users with valid GitHub tokens + saved username enter the REPL normally

## Acceptance criteria

- [ ] `build_github_mcp_config` accepts credentials that include `username`
- [ ] `github_integration_is_configured` does not raise on malformed store rows
- [ ] Regression test covers token + username store shape

larsspinetta12:
## Summary

- Strip persisted `credentials.username` before validating GitHub MCP config (`StrictConfigModel` forbids unknown fields).
- Harden `github_integration_is_configured()` so invalid store/env credentials log a warning and are treated as unconfigured instead of crashing the REPL login gate.

Fixes #3141

## Test plan

- [x] `pytest tests/integrations/test_github_mcp_validation.py -k username`
- [x] `pytest tests/cli/test_first_launch_github.py`
- [ ] Manual: launch `opensre` with saved GitHub token + `username` in `~/.opensre/integrations.json` → REPL starts without gate crash

### DIFF
--- a/integrations/github_mcp.py
+++ b/integrations/github_mcp.py
@@ -37,6 +37,9 @@ DEFAULT_GITHUB_MCP_URL = "https://api.githubcopilot.com/mcp/"
 DEFAULT_GITHUB_MCP_MODE = "streamable-http"
 DEFAULT_GITHUB_MCP_TOOLSETS = ("repos", "issues", "pull_requests", "actions", "search")
 
+# Non-transport metadata persisted alongside MCP credentials in the integration store.
+_CREDENTIAL_METADATA_KEYS: frozenset[str] = frozenset({"username"})
+
 REQUIRED_SOURCE_INVESTIGATION_TOOLS = (
     "get_file_contents",
     "get_repository_tree",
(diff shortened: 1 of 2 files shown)


larsspinetta12:
Will work on this tonight again to make sure it pasts tests, had to push before I walk home

PR #3177, opened 2026-06-27
fix(remote): prevent investigation ID collisions and silent report overwrites

Sujit-1509:
## Summary

`_make_id` generated investigation IDs as `YYYYMMDD_HHMMSS_<slug>` with **no random component**, so two investigations for the same alert name within the same wall-clock second produced an **identical ID**. `_save_investigation` then wrote via plain `path.write_text(...)`, silently **overwriting the earlier `.md` report** — losing the first root cause with no error, no log, and no audit trail.

This is reachable in production, not just a tight loop:
- `POST /investigate`, `POST /investigate/stream`, and the **Vercel poller** background path (`_handle_polled_candidate`) all use the same `_make_id` → `_save_investigation` path.
- The Vercel poller (`run_forever`) re-investigates the **same deployment's** alert on a fixed cadence — i.e. the same `alert_name` repeatedly — so a poller re-fire colliding with a manual `/investigate` (or two polls in the same second) deterministically destroys an earlier report.
- `GET /investigations` collapses the collision into one entry; `GET /investigations/{inv_id}` returns whichever wrote last.

### Reproduction (current `main`)

```
total calls: 20000
distinct ids: 1
collision count (same-second same-name writes that would OVERWRITE): 19999

COLLIDED inv_id: 20260625_080706_db-down
First investigation root cause was: root cause A
Surviving file content root cause: root cause B (OVERWROTE first)
First investigation LOST: True
```

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation

## Changes

1. **`_make_id`** — append 8 hex chars of `uuid4` randomness to the existing sortable `YYYYMMDD_HHMMSS_<slug>` prefix. The prefix is preserved so `GET /investigations` can still list newest-first by filename and `_id_to_iso` can still recover the created-at timestamp.
2. **`_save_investigation`** — replace the truncate-then-write `path.write_text(...)` with an **atomic write** (stage in a sibling `.tmp`, then `os.replace`). A crash (or a concurrent reader) can no longer observe a torn or empty `.md`.
3. **Tests** — added regression coverage:
   - `test_make_id_is_unique_across_same_second_calls` — 1000 same-second calls yield 1000 distinct IDs.
   - `test_save_investigation_does_not_overwrite_same_second_same_name` — two same-name investigations in the same second both persist (the exact scenario the bug lost).
   - `test_save_investigation_is_atomic` — a simulated mid-write crash leaves no partial `.md` and no stranded `.tmp`.
   - `test_id_to_iso_parses_new_suffixed_id` / `test_id_to_iso_returns_empty_on_garbage` — timestamp recovery still works.
   - Updated existing `_make_id` tests to assert the new suffix.

## Impact Analysis

- **Backward compatible.** Existing `.md` files on disk are unaffected; the new ID format only applies to new investigations. `_id_to_iso` still parses the `YYYYMMDD_HHMMSS` prefix (unchanged slice `[:15]`).
- **No API surface change** — `InvestigateResponse.id` is an opaque string.
- **No performance impact** — one `uuid4()` call per investigation.

## Testing

All checks pass against touched files:

\`\`\`
uv run pytest tests/remote/test_server.py tests/remote/test_server_utils.py
# 89 passed

uv run ruff check  app/remote/server.py tests/remote/test_server_utils.py  # All checks passed
uv run ruff format --check app/remote/server.py tests/remote/test_server_utils.py  # already formatted
uv run mypy app/remote/server.py tests/remote/test_server_utils.py  # 0 errors in changed files
\`\`\`

(Pre-existing, unrelated `app/fleet_monitoring/probe.py` mypy errors are not touched by this PR.)

## AI Assistance

I used an AI coding agent to help investigate and implement this fix. I confirm:
- [x] I reviewed **every single line** of AI-generated code
- [x] I understand the logic and can explain it in my own words
- [x] I tested edge cases (same-second collisions, mid-write crash, ID parsing)
- [x] I verified output matches project conventions (ruff, mypy, existing test style)
- [x] Tests pass with the AI-generated code


muddlebee:
this wasn't well tested, causing downstream issues.

PR #3242, opened 2026-06-28
fix(ci): test-scope reporting rule target must be a tuple, not a bare string

### LINKED ISSUE #3241: [BUG] test-scope reporting rule missing comma -> runs 'pytest /' instead of tests/delivery/
### Summary

In `infra/ci/test_scope_rules.py` (the `make test-scope` routing table), the rule for the investigation reporting package is missing a trailing comma, so its `test_targets` is a **string instead of a one-element tuple**:

```python
PathRule("tools/investigation/reporting/", ("tests/delivery/")),   # ("x") is a str, not ("x",)
```

`classify()` then iterates the string **character by character**. Each single-char "target" is dropped by the `Path(t).exists()` filter — except `"/"`, which exists — so a change under `tools/investigation/reporting/` resolves to `targets == ["/"]` and `make test-scope` tries to run `pytest /` (the filesystem root) instead of `tests/delivery/` (7 real test files).

This is restructure fallout from #3190 (moved the investigation pipeline into `tools/investigation/`).

**Expected:** a change under `tools/investigation/reporting/` runs `tests/delivery/`.

**Actual:** `classify()` returns `(False, ["/"], ["tools/investigation/reporting/"])` — it logs `skipping non-existent targets: t, e, s, d, l, i, v, r, y` and points pytest at `/`.

### Steps to reproduce

```
uv run python -c "import sys; sys.path.insert(0,'infra/ci'); from test_scope_rules import classify; print(classify(['tools/investigation/reporting/publish.py']))"
# (skipping non-existent targets: t, e, s, d, l, i, v, r, y)
# (False, ['/'], ['tools/investigation/reporting/'])
```

Yes

### How often does it occur?

Every time a change is confined to `tools/investigation/reporting/`.

### Operating system

Other (CI / tooling configuration; OS-independent)

### Logs and error output

```
as-written: ("tests/delivery/")  -> type str, iterates to ['t','e','s','t','s','/', ...]
intended:   ("tests/delivery/",) -> type tuple, one target "tests/delivery/"
```

### Additional context

Two parts:
1. Add the missing trailing comma so the target is a one-element tuple.
2. This wasn't caught by typecheck because `infra/ci/` is outside `PYTHON_SOURCE_PATHS` (mypy would otherwise flag `str` where `tuple[str, ...]` is expected). Add a small unit guard asserting every rule's `test_targets` is a tuple, so the whole class of "bare `("x")`" footgun is caught regardless of mypy scope.

Davidson3556:
Fixes #3241

The `make test-scope` routing rule for the investigation reporting package was
missing a trailing comma, making its `test_targets` a **string instead of a
one-element tuple**:

```python
PathRule("tools/investigation/reporting/", ("tests/delivery/")),   # ("x") == str, not ("x",)
```

`classify()` iterates `test_targets`, so it walked the string character by
character. Every single-char "target" was dropped by the `Path(t).exists()`
filter — except `"/"`, which exists — so a change under
`tools/investigation/reporting/` resolved to `targets == ["/"]` and
`make test-scope` would run `pytest /` (the filesystem root) instead of
`tests/delivery/` (7 real test files). Restructure fallout from #3190.

This PR:
- adds the missing trailing comma so the target is a proper one-element tuple;
- adds two unit tests: one asserting the reporting rule routes to
  `tests/delivery/`, and one asserting **every** rule's `test_targets` is a
  tuple — `infra/ci/` is outside `PYTHON_SOURCE_PATHS`, so mypy never flagged
  `str` where `tuple[str, ...]` was expected, and this guard catches the whole
  class of bare-`("x")` footgun regardless of mypy scope.

BEFORE

image

AFTER

image

image

Lint and format pass on both files.

---

I spotted the missing comma while re-reading the rewritten RULES table after the
#3190 restructure, confirmed `("tests/delivery/")` is a `str` (not a tuple) and
that `classify()` therefore iterated it into single-char targets resolving to
`["/"]`, and traced why CI missed it: `infra/ci/` is outside the mypy paths.
Fixed the comma and added a routing test plus a table-wide tuple guard so the
same footgun can't silently return on a future rule.

---

### DIFF
--- a/infra/ci/test_scope_rules.py
+++ b/infra/ci/test_scope_rules.py
@@ -27,7 +27,7 @@ RULES: tuple[PathRule, ...] = (
     # Shared core (always escalate)
     PathRule("core/domain/", (), always_escalate=True),
     PathRule("core/", ("tests/core/",)),
-    PathRule("tools/investigation/reporting/", ("tests/delivery/")),
+    PathRule("tools/investigation/reporting/", ("tests/delivery/",)),
     PathRule("tools/investigation/", (), always_escalate=True),
     PathRule("utils/", (), always_escalate=True),
     # Specific sub-packages before their parent
--- a/tests/infra_ci/test_test_scope_rules.py
+++ b/tests/infra_ci/test_test_scope_rules.py
@@ -64,3 +64,19 @@ def test_changed_test_file_is_targeted() -> None:
     escalate, targets, _ = rules.classify([path])
     assert not escalate
     assert targets == [path]
+
+
+def test_reporting_rule_routes_to_tests_delivery() -> None:
+    rules = _rules_module()
+    escalate, targets, _ = rules.classify(["tools/investigation/reporting/publish.py"])
+    assert not escalate
+    assert targets == ["tests/delivery/"]
+
+
+def test_all_rule_targets_are_tuples_not_bare_strings() -> None:
+    # A single target written as ("x") is a str, not a tuple — classify() would
+    # then iterate it character by character. infra/ci is outside the mypy paths
+    # so this footgun isn't caught by typecheck; guard it here for every rule.
+    rules = _rules_module()
+    for rule in rules.RULES:
+        assert isinstance(rule.test_targets, tuple), rule.path_prefix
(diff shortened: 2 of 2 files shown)


Davidson3556:
Good call — added an explicit `assert Path("tests/delivery/").is_dir()` guard at the top of the routing test in a41c1b2f. Now if that dir is ever removed the failure reads 'dir missing — update the rule or recreate it' instead of looking like a routing regression.

@greptile-apps review

Davidson3556:
@davincios  kindly review

PR #3251, opened 2026-06-28
fix(remote): do not default agent port for explicit https URLs

### LINKED ISSUE #3250: [BUG] normalize_url appends :2024 to explicit https:// remote URLs, breaking standard-HTTPS deployments
## Summary

`normalize_url()` appends the agent default port `:2024` to an explicit `https://` URL that has no port, so pointing the remote client at a standard-HTTPS deployment (behind a load balancer / reverse proxy on 443) silently connects to the wrong port.

**Expected:** `normalize_url("https://opensre.example.com")` → `"https://opensre.example.com"` (an explicit `https://` URL implies port 443 and should be left as-is).

**Actual:** `normalize_url("https://opensre.example.com")` → `"https://opensre.example.com:2024"`. Every `RemoteAgentClient` built from such a URL then targets `:2024` instead of `:443`, producing connection failures against a perfectly healthy deployment.

## Steps to reproduce

```bash
uv run python -c "from infra.deployment.remote.client import normalize_url; print(normalize_url('https://opensre.example.com'))"
# -> https://opensre.example.com:2024
```

User-facing path: `opensre remote health https://opensre.example.com` (or any `opensre remote ...` command whose `--url` is a standard-HTTPS host) routes through `RemoteAgentClient.__init__` → `normalize_url` and connects to `:2024`.

Yes.

## How often does it occur?

Every time, for any explicit `https://` URL without an explicit port.

## Operating system

macOS (platform-independent — pure URL string handling).

## Additional context

In `infra/deployment/remote/client.py`, the default-port guard is:

```python
if url.count(":") == 1 and not url.split("//")[1].count(":"):
    url = f"{url}:{DEFAULT_PORT}"
```

`url.count(":") == 1` counts the scheme's own colon, so `https://host` (one colon in `https:`, none in the authority) satisfies the condition and gets `:2024` appended. The existing tests cover `https://...:2024` (with a port) and `http://...` (which correctly defaults to `:2024`), but not the `https://`-without-port case. Defaulting the agent port makes sense for bare hosts/IPs and plaintext `http://`, but an explicit `https://` already implies 443.

cloudenochcsis:
Fixes #3250

`normalize_url()` defaults the agent port `:2024` for any URL whose authority has no explicit port. The guard `url.count(":") == 1` counts the scheme's own colon, so an explicit `https://host` (one colon in `https:`, none in the authority) matched the condition and was rewritten to `https://host:2024`. A remote deployment behind a load balancer / reverse proxy on standard HTTPS (443) would then be contacted on `:2024` and fail to connect.

The fix records whether the caller supplied an explicit `https://` scheme and skips the default-port append in that case — an explicit `https://` already implies port 443. Plaintext `http://` URLs and bare hosts/IPs still default to `:2024` exactly as before (that is the intended dev convention, covered by existing tests).

A regression test (`test_https_without_port_is_left_unchanged`) locks in the corrected behavior, and the docstring gains the `https://` example.

Behavior before vs after (`uv run python` from repo root):

```
input                                  before (main)                          after (this PR)
'https://example.com'               -> 'https://example.com:2024'   (bug)  -> 'https://example.com'
'https://opensre.mycompany.com/api' -> '...:2024' before path        (bug) -> 'https://opensre.mycompany.com/api'
'http://agent.local'                -> 'http://agent.local:2024'           -> 'http://agent.local:2024'   (unchanged)
'1.2.3.4'                           -> 'http://1.2.3.4:2024'               -> 'http://1.2.3.4:2024'        (unchanged)
'https://x.com:2024'                -> 'https://x.com:2024'                -> 'https://x.com:2024'         (unchanged)
```

Local checks (per `CI.md`):

```
make format-check  → 2037 files already formatted
make typecheck     → Success: no issues found in 922 source files
make test-scope    → 324 passed   (tests/remote/, incl. the new regression test)
```

---

The problem: `normalize_url` is meant to turn bare IPs/hosts into a full agent base URL (defaulting the dev port `:2024`), but it also mangled explicit `https://` URLs by appending `:2024`, so a production remote on standard HTTPS could not be reached.

Root cause is the port-detection heuristic `url.count(":") == 1`, which double-counts the scheme colon. I considered parsing the authority with `urlsplit` to count ports robustly, but the minimal and lowest-risk change is to gate the existing append on "the user did not give an explicit `https://`", which preserves every current code path (verified against all existing `TestNormalizeUrl` cases) while fixing only the broken one. `http://` and bare-host defaulting are intentional and untouched.

---

### DIFF
--- a/infra/deployment/remote/client.py
+++ b/infra/deployment/remote/client.py
@@ -82,14 +82,18 @@ def normalize_url(url: str) -> str:
     """Normalize a URL or bare IP into a full base URL.
 
     Accepts:
-        - "http://1.2.3.4:2024" -> returned as-is
-        - "1.2.3.4:2024"        -> "http://1.2.3.4:2024"
-        - "1.2.3.4"             -> "http://1.2.3.4:2024"
+        - "http://1.2.3.4:2024"         -> returned as-is
+        - "1.2.3.4:2024"                -> "http://1.2.3.4:2024"
+        - "1.2.3.4"                     -> "http://1.2.3.4:2024"
+        - "https://agent.example.com"   -> returned as-is (https implies port 443)
     """
     url = url.rstrip("/")
+    explicit_https = url.startswith("https://")
     if not url.startswith(("http://", "https://")):
         url = f"http://{url}"
-    if url.count(":") == 1 and not url.split("//")[1].count(":"):
+    # Only default the agent port for plaintext/bare hosts. An explicit https://
+    # URL implies the standard port 443, so leave it untouched.
+    if not explicit_https and url.count(":") == 1 and not url.split("//")[1].count(":"):
         url = f"{url}:{DEFAULT_PORT}"
     return url
 
--- a/tests/remote/test_client.py
+++ b/tests/remote/test_client.py
@@ -36,6 +36,10 @@ class TestNormalizeUrl:
     def test_hostname_without_port(self) -> None:
         assert normalize_url("http://agent.local") == f"http://agent.local:{DEFAULT_PORT}"
 
+    def test_https_without_port_is_left_unchanged(self) -> None:
+        # An explicit https:// URL implies port 443; do not append the agent default port.
+        assert normalize_url("https://opensre.example.com") == "https://opensre.example.com"
+
 
 class TestRemoteAgentClientInit:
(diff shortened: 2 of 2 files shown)


cloudenochcsis:
@greptile review

PR #3292, opened 2026-06-29
fix(wizard): persist Ollama host to env, not the keyring

### LINKED ISSUE #3291: Custom Ollama host is saved to the keyring but read from the environment, so it is silently ignored
## Summary

When you configure a **custom Ollama host** in the onboarding wizard, the value is written to the system keyring, but every runtime consumer reads `OLLAMA_HOST` from the **environment**. The custom host is therefore silently ignored (OpenSRE connects to `http://localhost:11434`) and trapped — the wizard never re-prompts because it sees the keyring entry.

## Root cause

Ollama is configured with `credential_kind="host"` / `api_key_env="OLLAMA_HOST"` / `credential_secret=False` (`cli/wizard/config.py`). In `cli/wizard/flow.py`, the credential prompt fires for any provider whose `credential_kind not in ("cli", "none")`, so `"host"` qualifies, and the collected host URL is persisted via `_persist_llm_api_key(provider.api_key_env, value)`.

`OLLAMA_HOST` is not in `API_KEY_PROVIDER_ENVS` (Ollama is keyless), so `_persist_llm_api_key` routes it to `persist_api_key_secret` → `save_llm_api_key`, which writes it **into the system keyring** — never to `.env` or `os.environ`.

But the runtime reads the host from the environment:
- `config/config.py:348` — `os.getenv("OLLAMA_HOST", DEFAULT_OLLAMA_HOST)`
- `config/llm_auth/credentials.py:218` — `os.getenv("OLLAMA_HOST", "")`

So the keyring value never reaches runtime.

It's also **not self-correcting**: on the next wizard run, `_local_defaults` sets `has_api_key=True` because `has_llm_api_key("OLLAMA_HOST")` finds the keyring entry, so the wizard skips the prompt — the host stays trapped in the keyring.

The docs already describe the intended (env-based) behavior: `docs/configuration/environment-variables.mdx` lists `OLLAMA_HOST` with `config` storage, and `docs/llm-providers.mdx` shows `export OLLAMA_HOST=...`.

## Impact

Any user who points OpenSRE at a non-default Ollama host (remote/LAN box, custom port) silently gets `localhost:11434` instead, with no error and no way to fix it via the wizard.

## Fix

Treat a `host` credential as the non-secret env value it is:
1. In the wizard, persist a `host` credential to `.env` + `os.environ` (via `sync_env_values`), not the keyring.
2. In `sync_provider_env`, keep the host key as a non-secret active key so a later provider sync doesn't strip it.
3. In `_local_defaults`, derive "already configured" for a host credential from the environment, not the keyring — so the wizard doesn't re-prompt every run (and default back to localhost, clobbering a custom host).

cloudenochcsis:
## Problem

A custom **Ollama host** entered in the onboarding wizard is silently ignored at runtime, and trapped so the wizard never lets you fix it.

Ollama uses `credential_kind="host"` / `api_key_env="OLLAMA_HOST"` / `credential_secret=False`. The wizard prompts for the host (any `credential_kind not in ("cli","none")`) and persists it with `_persist_llm_api_key(...)`. Since `OLLAMA_HOST` isn't in `API_KEY_PROVIDER_ENVS`, that routes to `save_llm_api_key` → the **system keyring**.

But every runtime consumer reads the host from the **environment**:
- `config/config.py:348` — `os.getenv("OLLAMA_HOST", DEFAULT_OLLAMA_HOST)`
- `config/llm_auth/credentials.py:218` — `os.getenv("OLLAMA_HOST", "")`

So the value never reaches runtime (OpenSRE connects to `localhost:11434`), and it isn't self-correcting: `_local_defaults` sees the keyring entry, sets `has_api_key=True`, and skips the prompt forever. The docs already describe the intended env-based behavior (`OLLAMA_HOST` is `config` storage in `docs/configuration/environment-variables.mdx`).

## Fix

Treat a `host` credential as the non-secret env value it is:

1. **`cli/wizard/flow.py`** — new `_persist_provider_credential(provider, value)` routes a `host` credential to `.env` + `os.environ` (via `sync_env_values`); API keys still go to the keyring. The three persist sites call it.
2. **`cli/wizard/env_sync.py`** — `sync_provider_env` keeps `api_key_env` as a non-secret active key when `credential_kind == "host"`, so a later provider sync preserves the host instead of stripping it like a secret.
3. **`cli/wizard/_ui.py`** — `_local_defaults` derives "already configured" for a host credential from the environment, so the wizard doesn't re-prompt every run and default back to localhost (which would clobber a custom host).

## Tests

- `test_sync_provider_env_preserves_ollama_host` — a custom `OLLAMA_HOST` in `.env` survives a provider sync and lands in `os.environ`. **Fails on old code** (stripped).
- `test_persist_provider_credential_routes_host_to_env_not_keyring` — a host URL goes to `.env`/`os.environ`, never the keyring.
- `test_persist_provider_credential_routes_api_key_to_keyring` — API keys still go to the keyring.

Verified locally: `tests/cli/wizard/` (167 passed), `ruff check`, `ruff format --check`, `mypy` all green.

Fixes #3291

---
*This change was developed with AI assistance and reviewed before submission.*

### DIFF
--- a/cli/wizard/_ui.py
+++ b/cli/wizard/_ui.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import os
 from collections.abc import Mapping
 from dataclasses import dataclass
 from typing import cast
@@ -112,6 +113,16 @@ def _local_defaults() -> dict[str, str | bool | None]:
         local.get("api_key_env"), api_key_provider.api_key_env if api_key_provider else ""
     )
     is_cli = bool(raw_provider_option and raw_provider_option.credential_kind == "cli")
+    # A "host" credential (e.g. OLLAMA_HOST) lives in the environment, not the keyring,
+    # so "already configured" means the env var is set — otherwise the wizard would
+    # re-prompt every run and default back to localhost, clobbering a custom host.
+    is_host = bool(api_key_provider and api_key_provider.credential_kind == "host")
+    if is_cli:
+        has_api_key = True
+    elif is_host:
+        has_api_key = bool(api_key_env and os.environ.get(api_key_env))
+    else:
+        has_api_key = bool(api_key_env and has_llm_api_key(api_key_env))
     is_oauth_backend = bool(raw_provider_value and raw_provider_value != provider_value)
     raw_auth_method = local.get("auth_method")
     auth_method = (
@@ -127,7 +138,7 @@ def _local_defaults() -> dict[str, str | bool | None]:
         "auth_method": auth_method,
         "model": _string_value(local.get("model")),
         "api_key_env": api_key_env,
-        "has_api_key": True if is_cli else bool(api_key_env and has_llm_api_key(api_key_env)),
+        "has_api_key": has_api_key,
         "legacy_api_key": _string_value(local.get("api_key")),
     }
 
--- a/cli/wizard/env_sync.py
+++ b/cli/wizard/env_sync.py
@@ -321,6 +321,10 @@ def sync_provider_env(
     classification_env = _classification_model_env(resolved_model_provider)
     if classification_env:
         active_non_secret.add(classification_env)
+    # A "host" credential (e.g. OLLAMA_HOST) lives in api_key_env but is non-secret
+    # config read from the environment at runtime — keep it, don't strip it as a key.
+    if provider.credential_kind == "host" and provider.api_key_env:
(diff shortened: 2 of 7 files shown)


muddlebee:
@cloudenochcsis could you share a working demo,

muddlebee:
feel free to reopen with proper replicated issue/demo

PR #3308, opened 2026-06-29
fix(remote): strip investigation ID suffix from alert names

### LINKED ISSUE #3184: [BUG] Remote investigations list includes random ID suffix in alert names
## Summary

Remote investigation IDs now include an 8-character random suffix to avoid filename collisions, but `GET /investigations` still derives `alert_name` from the full filename tail. As a result, the random suffix is shown as part of the alert name in list metadata.

**Expected:** An investigation ID such as `20260628_123456_cpu-spike-deploy_ab12cd34` should list the alert name as `cpu spike deploy`.

**Actual:** The list endpoint returns `cpu spike deploy_ab12cd34`, leaking the collision-avoidance suffix into the user-visible alert name.

## Steps to reproduce

1. Create or use a persisted remote investigation file with an ID in the new format, e.g. `20260628_123456_cpu-spike-deploy_ab12cd34.md`.
2. Call `list_investigations()` or `GET /investigations`.
3. Observe that the returned `alert_name` includes the random suffix.

Minimal local repro from the repo root:

```bash
uv run python -c "from infra.deployment.remote import server as s; from pathlib import Path; import tempfile; d=Path(tempfile.mkdtemp()); s.INVESTIGATIONS_DIR=d; p=d/'20260628_123456_cpu-spike-deploy_ab12cd34.md'; p.write_text('x', encoding='utf-8'); item=s.list_investigations()[0]; print(item.model_dump())"
```

Observed output:

```python
{'id': '20260628_123456_cpu-spike-deploy_ab12cd34', 'filename': '20260628_123456_cpu-spike-deploy_ab12cd34.md', 'created_at': '2026-06-28T12:34:56+00:00', 'alert_name': 'cpu spike deploy_ab12cd34'}
```

Yes.

## How often does it occur?

Every time for investigation files generated with the new random suffix format.

## Operating system

macOS.

## Additional context

The suffix was added in `_make_id()` in `infra/deployment/remote/server.py` to prevent same-second filename collisions. The listing parser in `list_investigations()` appears to need a matching update to strip the trailing 8-hex suffix when deriving display metadata.

imjohnzakkam:
Fixes #3184

This PR fixes the remote investigations list metadata for persisted investigation IDs that include the new random uniqueness suffix.

The remote server now derives list display names through `_alert_name_from_inv_id()`, which:
- keeps the existing timestamp/slug parsing behavior
- strips only a trailing `_<8 hex>` uniqueness suffix generated by `_make_id()`
- preserves legacy unsuffixed IDs
- avoids stripping hex-looking slugs that do not have the suffix separator

Regression coverage was added for the suffixed ID format, legacy IDs, hex-like alert slugs, and `_make_id()` round-tripping.

Before this fix, an investigation file named:

```text
20260628_123456_cpu-spike-deploy_ab12cd34.md
```

was listed with:

```python
'alert_name': 'cpu spike deploy_ab12cd34'
```

After this fix, the same repro returns:

```python
{'id': '20260628_123456_cpu-spike-deploy_ab12cd34', 'filename': '20260628_123456_cpu-spike-deploy_ab12cd34.md', 'created_at': '2026-06-28T12:34:56+00:00', 'alert_name': 'cpu spike deploy'}
```

Validation run locally:

```text
make lint
make format-check
make typecheck
.venv/bin/python -m pytest tests/remote/ tests/remote/test_server_utils.py -v
```

Scoped remote tests passed: `328 passed`.

---

The issue happened because `_make_id()` appends an 8-character random suffix to avoid same-second filename collisions, but `list_investigations()` treated the whole filename tail as the alert slug. This exposed the suffix in the user-facing `alert_name`.

The fix keeps ID generation unchanged and only updates the display parsing path. `_alert_name_from_inv_id()` extracts the alert slug from the persisted ID, removes a trailing `_<8 hex>` suffix when present, and then formats hyphenated slugs as readable names. This keeps backward compatibility for old IDs while fixing new suffixed IDs.

---

---

### DIFF
--- a/infra/deployment/remote/server.py
+++ b/infra/deployment/remote/server.py
@@ -799,9 +799,9 @@ def _check_memory_health() -> DeepHealthCheck:
 def _make_id(alert_name: str) -> str:
     """Build a unique investigation id.
 
-    The id keeps its sortable ``YYYYMMDD_HHMMSS_<slug>`` prefix (so ``GET
+    The id keeps its sortable ``YYYYMMDD_HHMMSS-<suffix>_<slug>`` prefix (so ``GET
     /investigations`` can list newest-first by filename and ``_id_to_iso`` can
-    recover the created-at timestamp) and appends 8 hex chars of randomness so
+    recover the created-at timestamp) and injects 8 hex chars of randomness so
     two investigations for the same alert name within the same second cannot
     collide. Without the suffix, a re-fire from the Vercel poller (or two
     concurrent ``/investigate`` calls) produced an identical id and
@@ -810,11 +810,11 @@ def _make_id(alert_name: str) -> str:
     ts = datetime.now(tz=UTC).strftime("%Y%m%d_%H%M%S")
     slug = _slugify(alert_name) or "investigation"
     suffix = uuid.uuid4().hex[:8]
-    return f"{ts}_{slug}_{suffix}"
+    return f"{ts}-{suffix}_{slug}"
 
 
 def _id_to_iso(inv_id: str) -> str:
-    """Best-effort parse of ``YYYYMMDD_HHMMSS_slug[_suffix]`` into ISO 8601."""
+    """Best-effort parse of ``YYYYMMDD_HHMMSS[-suffix]_slug`` into ISO 8601."""
     try:
         date_part = inv_id[:15]  # YYYYMMDD_HHMMSS
         dt = datetime.strptime(date_part, "%Y%m%d_%H%M%S").replace(tzinfo=UTC)
--- a/tests/remote/test_server_utils.py
+++ b/tests/remote/test_server_utils.py
@@ -47,64 +47,74 @@ def test_slugify_handles_whitespace_only() -> None:
 def test_make_id_generates_timestamp_with_slug() -> None:
     """Test that _make_id combines timestamp with slugified alert name."""
     result = _make_id("Database Connection Failed")
-    # Format: YYYYMMDD_HHMMSS_slug_suffix - verify the timestamp structure
-    parts = result.split("_")
-    assert len(parts) >= 3  # date, time, slug (slug may itself contain _)
+    # Format: YYYYMMDD_HHMMSS-<suffix>_<slug>
+    parts = result.split("_", maxsplit=2)
+    assert len(parts) == 3
     assert len(parts[0]) == 8  # YYYYMMDD
     assert parts[0].isdigit()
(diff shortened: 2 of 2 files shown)


infra/deployment/remote/server.py
```diff
@@ -651,6 +649,14 @@ def _slugify(text: str) -> str:
     return re.sub(r"[^a-z0-9]+", "-", text.lower()).strip("-")[:60]
 
 
+def _alert_name_from_inv_id(inv_id: str) -> str:
```
muddlebee: not an ideal way imo to fetch alert name, could you take a look here https://github.com/Tracer-Cloud/opensre/pull/3177

some changes were done, how we could fix the core issue?

imjohnzakkam:
@muddlebee thanks for pointing that out. I updated the fix so `list_investigations()` no longer treats the filename/ID as
  the primary source for the alert name.

  It now reads the persisted report title (`# Investigation: ...`) from the saved `.md` file, and only falls back to ID
  parsing for malformed/legacy reports without that title. I also added coverage for:
  - report title as the source of truth
  - fallback when the title is missing
  - `list_investigations()` returning the saved alert name for a suffixed ID

  Validation:
  - `make lint`
  - `make format-check`
  - `make typecheck`
  - `uv run pytest tests/remote/test_server.py tests/remote/test_server_utils.py -q` (`94 passed`)

muddlebee:
Still not satisfied. I asked to take a look at original implementation the PR referenced how we could handle better.

The fix and fallback added imo feels hacky rather we should address the core issue.

imjohnzakkam:
@muddlebee thanks, I agree, I overlooked it. The earlier approach was still treating the filename as something to patch around in the display path.

  I updated the fix to address the core issue from #3177 instead: `_make_id()` now puts the uniqueness suffix in the timestamp segment, not after the alert slug.

  New ID shape:

  ```text
  YYYYMMDD_HHMMSS-<suffix>_<slug>
```

  This keeps:

  - same-second collision protection
  - filename sortability
  - _id_to_iso() compatibility, since the first 15 chars are still YYYYMMDD_HHMMSS
  - the existing list_investigations() parsing contract, where the third underscore-delimited segment is the clean alert slug

  So we no longer need markdown parsing, fallback helpers, or suffix stripping in the display path.

  Validated with:

  - make lint
  - make format-check
  - make typecheck
  - uv run pytest tests/remote/test_server.py tests/remote/test_server_utils.py -q (90 passed)

muddlebee:
@greptile review

PR #3324, opened 2026-06-30
fix: record one session history row per REPL user turn

### LINKED ISSUE #3323: REPL turn counter skips even numbers after slash commands
## Problem

The interactive shell prompt counter (`[1]`, `[2]`, …) is derived from `len(session.history) + 1`. After running slash commands like `/model`, the counter jumps by 2 (e.g. `[1]` → `[3]` → `[5]`).

## Root cause

Literal slash commands are dispatched through the action-agent `slash_invoke` path. Two history rows are written for one user turn:

1. `ActionRenderObserver` records `cli_agent` on the first `tool_start`
2. `dispatch_slash` records `slash` when the handler runs

The same duplication affects other self-recording action tools (`shell_run`, `investigation_start`, etc.).

## Expected behavior

- Typed `/commands` advance the counter by 1
- Normal chat messages advance the counter by 1

## Proposed fix

Skip the observer `cli_agent` record when the first action tool already writes its own history row (`slash_invoke`, `shell_run`, etc.). Chat turns continue to be recorded by turn accounting when the assistant runs.

larsspinetta12:
Fixes #3323

## Summary

- Stop `ActionRenderObserver` from writing a duplicate `cli_agent` history row when self-recording action tools (`slash_invoke`, `shell_run`, `investigation_start`, etc.) already persist the turn
- Keeps the `[N]` prompt counter aligned with user-visible turns: slash commands and chat both advance by 1
- Update oracle scenario fixtures and add regression tests for slash/chat counter behavior

## Test plan

- [x] `uv run python -m pytest tests/interactive_shell/ui/test_action_rendering.py`
- [x] `uv run python -m pytest tests/core/agent/orchestration/test_agent_actions_harness.py`
- [x] `uv run python -m pytest tests/core/agent/test_turn_fixture_integrity.py`

### DIFF
--- a/core/agent_harness/action_agent.py
+++ b/core/agent_harness/action_agent.py
@@ -50,17 +50,22 @@ _EXECUTED_HISTORY_TYPES = {
     "implementation",
     "cli_command",
 }
-_LEGACY_HISTORY_TOOL_NAMES = {
-    "alert_sample",
-    "cli_exec",
-    "code_implement",
-    "investigation_start",
-    "llm_set_provider",
-    "shell_run",
-    "slash_invoke",
-    "synthetic_run",
-    "task_cancel",
-}
+# Action tools that append their own ``session.history`` row when executed.
+# Keep this as the single catalogue: the shell observer and generic tool-result
+# accounting both key off it so new tools cannot silently double-record turns.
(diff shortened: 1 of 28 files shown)


larsspinetta12:
@greptile review again

PR #3351, opened 2026-07-01
feat(signoz): compact query_signoz_traces output for high-signal evidence

amsorrytola:
Part of #481

This is a small, single-tool slice of #481 ("turn noisy logs and traces into high-signal evidence"): it brings **`query_signoz_traces`** onto the same output-compaction pattern its siblings already use. Its sibling `query_signoz_logs` (`_normalize_logs_payload`) and the analogous `GrafanaTracesTool` both compact their high-volume output; `query_signoz_traces` was the remaining consistency gap, returning a raw trace list. Approach approved by @Devesh36 in the #481 thread.

**What changed**

- Added `_normalize_traces_payload` in `tools/signoz_tools/__init__.py`, wired into both the backend-injection and live-client return paths of `query_signoz_traces`.
- It bounds the trace list to `DEFAULT_TRACE_LIMIT` (20) via the shared `compact_traces` helper — **independent of the larger server-side fetch `limit`** — and records a `truncation_note` (e.g. `Showing 20 of 50 traces`) when traces are trimmed.
- The aggregate trace `summary` (error rate, p99, call count) and every other field are passed through **untouched**.

**Design notes**

- **Read-only, additive, single-purpose.** No transport/fetch behavior changes — only the output the agent sees is bounded. No unrelated edits.
- Imports `compact_traces` / `DEFAULT_TRACE_LIMIT` from the canonical `platform.common.evidence_compaction` (matching `GrafanaTracesTool`), not the back-compat re-export shim.
- **Honest about the data shape:** real SigNoz `query_traces` returns *flat per-span rows* (no nested `spans`), so `compact_traces`' per-trace span cap is a no-op for SigNoz — it's used for parity with the Grafana traces tool, and the trace-**list** cap is the load-bearing part here. The docstring and docs say exactly this rather than over-claiming span capping.

**Tests** (`tests/tools/test_signoz_tools.py`)

- Exercise the real tool path via backend injection with a fixture matching SigNoz's **flat** row shape.
- **Exact-cap** assertions (not `<=`) so a limit regression is caught: 50 traces → exactly `DEFAULT_TRACE_LIMIT`, rows passed through unchanged, `Showing 20 of 50 traces` note, full-dict `summary` passthrough, plus a small-payload (no truncation) case and an `available=False` upstream-error passthrough case.

**Docs** — `docs/signoz.mdx` gets a short note on the bounded, high-signal traces output.

**Local CI green:** `make lint`, `make format-check`, `make typecheck` (923 files), and the scoped test suite (2215 passed).

End-to-end against a local SigNoz API stand-in: `opensre integrations verify signoz` passes, `query_signoz_traces` compacts a 50-trace result down to 20 high-signal traces (`Showing 20 of 50 traces`), and a full `opensre investigate` run root-causes a `payment-service` → `billing-gateway` timeout/circuit-breaker incident using the compacted traces.

_📎 Demo GIF attached below._

signoz-traces-demo

---

The compaction layer from #500 already existed (`platform.common.evidence_compaction`), and most log/trace tools already consumed it — `query_signoz_traces` was the odd one out, returning a raw list. Rather than build anything new, I mirrored the exact pattern its sibling `query_signoz_logs` and `GrafanaTracesTool` use: a small normalize step that runs the raw output through the shared helper before returning.

The one subtlety I had to get right is that SigNoz returns *flat per-span rows*, not traces-with-nested-spans like Grafana Tempo — so `compact_traces`' span-per-trace cap never fires for SigNoz. I kept `compact_traces` (for parity and because the list cap is what matters), but made the tests use SigNoz's real flat shape and made the docstring/docs state plainly that only the trace-list is bounded for SigNoz. The truncation count is computed from the pre-compaction length vs the returned length, and the aggregate `summary` is threaded through untouched so the agent still gets error-rate/p99 even when individual traces are trimmed.

Key components: `_normalize_traces_payload` (the bounding + truncation-note step) and the two call sites in `query_signoz_traces` (backend-injection path and live-client path).

---

---

Note: **Allow edits from maintainers** is enabled.

### DIFF
--- a/docs/signoz.mdx
+++ b/docs/signoz.mdx
@@ -61,6 +61,10 @@ When `alert_source` is `signoz`, the agent auto-seeds three tools before the ReA
 - **`query_signoz_metrics`** — Query CPU, memory, and request-rate signals.
 - **`query_signoz_traces`** — Query error spans, latency percentiles, and dependencies.
 
+<Info>
+`query_signoz_traces` bounds high-volume results to the most relevant traces so investigations stay high-signal. When traces are trimmed, the response includes a `truncation_note` (e.g. `Showing 20 of 50 traces`); the aggregate `summary` (error rate, p99, call count) is always preserved.
+</Info>
+
 ## Webhook Configuration
 
 SigNoz emits Prometheus-style webhook payloads. To trigger OpenSRE investigations automatically:
--- a/integrations/signoz/tools/__init__.py
+++ b/integrations/signoz/tools/__init__.py
@@ -7,7 +7,12 @@ from __future__ import annotations
 from typing import Any, cast
 
 from core.tool_framework.tool_decorator import tool
-from core.tool_framework.utils.compaction import compact_logs, summarize_counts
+from core.tool_framework.utils.compaction import (
+    DEFAULT_TRACE_LIMIT,
+    compact_logs,
+    compact_traces,
+    summarize_counts,
+)
 from integrations.signoz import SigNozConfig, signoz_extract_params
 from integrations.signoz.availability import signoz_available_or_backend
 from integrations.signoz.client import SigNozClient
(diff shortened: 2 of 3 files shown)


amsorrytola:
Addressed the P2 in 46a639b1: `_normalize_traces_payload` now computes `total = result.get("total", len(traces))` once and uses it for **both** the returned `total` field and the `truncation_note`, matching `_normalize_logs_payload` — so the two can never disagree when the server-reported total exceeds the fetched (limit-bounded) list. Added a regression test with `total=1000` / fetched=50 asserting the note reads `Showing 20 of 1000 traces` (the case the old `total == len` fixtures didn't cover).

@greptile review

Devesh36:
can you resolve the conflicts here

amsorrytola:
Conflicts resolved 👍 — rebased onto latest `main`. The refactor had moved the SigNoz tools from `tools/signoz_tools/` to `integrations/signoz/tools/`, so I re-applied the traces compaction there (imports now come from `core.tool_framework.utils.compaction`); tests and `docs/signoz.mdx` updated to match. `make lint`, `make format-check`, `make typecheck`, and the signoz suite all pass locally.

Davidson3556:
some test are failing,please fix @amsorrytola

amsorrytola:
Thanks @Davidson3556 — I dug into `test (cli-runtime)` and it isn't caused by this change.

**Scope:** the diff is only 3 signoz files (`integrations/signoz/tools/__init__.py`, `tests/tools/test_signoz_tools.py`, `docs/signoz.mdx`). `tests/core/agent/test_turn_scenarios.py` — where the failures are — is **byte-identical to `main`** and has no coupling to signoz.

**What's actually red:** all 30 errors are the `@pytest.mark.live_llm` tests (`test_live_action_planning` / `test_live_turn_execution_oracle`) failing with `OPENAI_API_KEY is not configured`. `test (tools-runtime)` (which runs the signoz tests) is ✅, and `turn-checks (no-LLM)` is ✅.

**Root cause:** this is a cross-repo (fork) PR, and GitHub doesn't pass repo secrets to fork `pull_request` runs — so `secrets.OPENAI_API_KEY` is empty. The `cli-runtime` shard runs `pytest -m "not synthetic"` (which doesn't exclude `live_llm`), so the credential gate in `tests/core/agent/_ci_gates.py` **hard-fails under `GITHUB_ACTIONS`** instead of skipping. The `interactive-shell-live` and `synthetic-deterministic` workflows already guard for this (`-m "not live_llm"` / fork check); the `cli-runtime` shard just lacks that exclusion.

**Why it only went red after the rebase:** the `pull_request` CI matrix hadn't actually run on the earlier fork commits — the force-push is simply when it first executed and surfaced these pre-existing `live_llm` gate failures. The shard config is equivalent on the old and new base; nothing in this PR changed it.

Could a maintainer re-run `test (cli-runtime)` with secrets (or confirm it's expected for fork PRs)? I'm happy to add a `-m "not live_llm"` guard to that shard in a separate PR if that's the preferred fix — but it's out of scope for this signoz change. cc @Devesh36

Davidson3556:
on it don't worry. Thanks

muddlebee:
closing this, it needs more discussion with the core team

PR #3619, opened 2026-07-01
fix(core/agent_harness): use tuple for AgentConfig.tools to respect frozen=True

### LINKED ISSUE #3613: fix(core/agent_harness): AgentConfig frozen=True gives false immutability with list fields
## Problem
`AgentConfig(frozen=True)` freezes field *reassignment* but not the underlying list contents. Any caller can `config.tools.append(...)` and silently mutate state — the immutability guarantee is false.

## Fix
- Use `tuple[...]` for structural immutability, OR
- Drop `frozen=True` and document the real contract

## Files
- `core/agent_harness/agent_builder.py`

Mubashir78:
**Problem:** `AgentConfig(frozen=True)` with `tools: list[Any]` gives false immutability — callers can still `config.tools.append(...)` and silently mutate state.

**Fix:** Changed `tools` type to `tuple[Any, ...]` in `AgentConfig` and updated all 3 call sites to convert with `tuple()`. The `Agent.__init__` already converts the value to a list internally (`list(tools)`), so this is fully compatible.

Closes #3613

### DIFF
--- a/core/agent_harness/AGENTS.md
+++ b/core/agent_harness/AGENTS.md
@@ -78,7 +78,7 @@ from core.agent_harness.agent_builder import AgentConfig, build_agent
 config = AgentConfig(
     llm=llm_client,                    # or None to fall back to get_agent_llm()
     system=system_prompt,
-    tools=agent_tools,
+    tools=tuple(agent_tools),
     resolved_integrations=resolved,
     max_iterations=6,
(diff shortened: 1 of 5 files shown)


cerencamkiran:
Could you also update the AGENTS.md snippet to use tools=tuple(agent_tools) to match the new change?

PR #3701, opened 2026-07-03
test(core): add regression coverage for _requires_sequential_execution serialization guard

### LINKED ISSUE #3695: test: add regression coverage for _requires_sequential_execution serialization guard
## Context

`_requires_sequential_execution` (`core/execution.py:238`) is the guard that decides whether a batch of provider-requested tool calls runs sequentially (one at a time) or in parallel via the `ThreadPoolExecutor` in `execute_tool_calls`. It exists to protect stateful tools — e.g. `shell_run`, `investigation`, `cli_command`, and five other interactive-shell action tools that set `parallel_safe=False` — from racing against other tool calls issued in the same LLM turn.

## Problem

Audited the function and the surrounding call site; found no correctness bug, but a real coverage gap:

- The only related test, `test_parallel_batch_preserves_provider_order` (`tests/core/runtime/test_tool_execution_contract.py`), only exercises a batch where *every* tool is parallel-safe. It asserts result ordering, not concurrency behavior.
- **Nothing exercises the actual serialization guarantee**: no test verifies that when a batch mixes a `parallel_safe=False` `RegisteredTool` (or an `AgentTool` with `effective_execution_mode == "sequential"`) with otherwise-parallelizable tool calls, `execute_tool_calls` runs the whole batch through the `len(tool_calls) == 1 or _requires_sequential_execution(...)` branch instead of the thread pool.
- This means a future refactor of `execute_tool_calls` (or an accidental change to `_requires_sequential_execution`) that silently drops or breaks this guard would not be caught by CI — the failure mode is a race condition in stateful tools like `shell_run`, which is exactly the kind of bug that's hard to catch after the fact (non-deterministic, only shows up under concurrent multi-tool-call turns).

Secondary observations (not blocking, for awareness):

- The guard is coarse-grained by design: **one** sequential/non-parallel-safe tool call forces **every** tool call in the batch to serialize, even unrelated parallel-safe ones. Worth keeping in mind as more `parallel_safe=False` tools are added, since it compounds latency.
- The `isinstance(tool, AgentTool)` branch (and `AgentTool.execution_mode` override) currently has zero production instantiations — every real tool today is a `RegisteredTool`. `AgentTool` is only constructed in tests, so that branch is exercised only there.

## Proposed change

Add regression tests under `tests/core/runtime/test_tool_execution_contract.py` (or a sibling test module) that:

1. Build a mixed batch containing one `RegisteredTool` with `parallel_safe=False` alongside 2+ parallel-safe tools, and assert the batch executes without going through the thread pool (e.g. via a shared counter/lock in the fake tool executors that would detect concurrent invocation, or by asserting `ThreadPoolExecutor` is never constructed for that batch).
2. Cover the equivalent case for an `AgentTool` with `execution_mode="sequential"` (and separately `parallel_safe=False` with no explicit `execution_mode`), to lock in `effective_execution_mode`'s fallback behavior.
3. Optionally, a unit test directly against `_requires_sequential_execution` for the pure decision logic (unknown tool name, empty batch, all-parallel-safe batch, single sequential tool anywhere in the batch).

## Acceptance criteria

- [ ] A regression test fails if `_requires_sequential_execution` or its call site in `execute_tool_calls` stops forcing sequential execution for a `parallel_safe=False` `RegisteredTool`.
- [ ] Equivalent coverage exists for `AgentTool.effective_execution_mode == "sequential"`.
- [ ] No behavior change to `core/execution.py` itself — this is test-only coverage for an existing, correct guard.

cloudenochcsis:
Fixes #3695

Adds the regression coverage requested in #3695 for the `_requires_sequential_execution` serialization guard (`core/execution.py`). Test-only change — no behavior change to `core/execution.py`.

**How the tests work:** a `pytest` `monkeypatch` fixture swaps `core.execution.ThreadPoolExecutor` for a recording subclass, so each test can assert deterministically (no sleeps, no timing races) whether `execute_tool_calls` routed the batch through the thread pool or the sequential branch.

New tests in `tests/core/runtime/test_tool_execution_contract.py`:

1. `test_non_parallel_safe_registered_tool_serializes_mixed_batch` — a batch mixing one `parallel_safe=False` `RegisteredTool` with two parallel-safe `AgentTool`s never constructs the pool, and results stay in provider order.
2. `test_agent_tool_sequential_execution_mode_serializes_mixed_batch` — same guarantee for an `AgentTool` with explicit `execution_mode="sequential"`.
3. `test_agent_tool_parallel_safe_false_serializes_via_execution_mode_fallback` — locks in `effective_execution_mode`'s fallback: `parallel_safe=False` with no explicit mode still serializes.
4. `test_all_parallel_safe_batch_goes_through_thread_pool` — control test proving the recording patch observes pool construction, so tests 1–3 cannot pass vacuously.
5. `test_requires_sequential_execution_forces_serial_for_stateful_tools` / `test_requires_sequential_execution_allows_parallel_otherwise` — direct unit tests for the pure decision logic: empty batch, unknown tool name, all-parallel-safe batch, a single sequential tool anywhere in the batch, and the explicit `execution_mode="parallel"` override of `parallel_safe=False`.

**Acceptance criteria from the issue:**
- [x] A regression test fails if the guard or its call site stops forcing sequential execution for a `parallel_safe=False` `RegisteredTool`
- [x] Equivalent coverage for `AgentTool.effective_execution_mode == "sequential"`
- [x] No behavior change to `core/execution.py`

Mutation check proving the tests catch a silently dropped guard. With the call site mutated from `if len(tool_calls) == 1 or _requires_sequential_execution(tool_calls, tool_map):` to `if len(tool_calls) == 1:`:

```
FAILED tests/core/runtime/test_tool_execution_contract.py::test_non_parallel_safe_registered_tool_serializes_mixed_batch
FAILED tests/core/runtime/test_tool_execution_contract.py::test_agent_tool_sequential_execution_mode_serializes_mixed_batch
FAILED tests/core/runtime/test_tool_execution_contract.py::test_agent_tool_parallel_safe_false_serializes_via_execution_mode_fallback
========================= 3 failed, 13 passed in 0.17s =========================
```

With the guard restored (current `main`):

```
============================== 16 passed in 0.25s ==============================
```

Pre-push checks (CI.md): `make lint` ✅ `make format-check` ✅ `make typecheck` ✅ `make test-scope` ✅ (16 passed)

---

The issue asks for tests that fail if the serialization guard is ever dropped. The key design decision was how to detect "ran sequentially" vs "ran through the thread pool" without flaky timing tricks (shared counters + sleeps can pass by luck). Instead, the tests patch `core.execution.ThreadPoolExecutor` with a subclass that records construction: sequential batches must record zero constructions, and a control test asserts an all-parallel-safe batch records exactly one — so a refactor that stops using `ThreadPoolExecutor` entirely (e.g. moving to asyncio) breaks the control test loudly rather than making the serialization tests silently vacuous. The decision-logic unit tests pin the guard's edge cases directly (unknown tool → no serialization; explicit `execution_mode="parallel"` overrides `parallel_safe=False`). I verified the tests bite via the mutation check shown in the demo section.

---

### DIFF
--- a/tests/core/runtime/test_tool_execution_contract.py
+++ b/tests/core/runtime/test_tool_execution_contract.py
@@ -1,8 +1,12 @@
 from __future__ import annotations
 
 from collections.abc import Iterator
+from concurrent.futures import ThreadPoolExecutor
 from typing import Any
 
+import pytest
+
+import core.execution
 from core.agent import Agent
 from core.execution import (
     BeforeToolCallResult,
@@ -10,6 +14,7 @@ from core.execution import (
     ToolExecutionPatch,
     ToolExecutionRequest,
     ToolExecutionResult,
+    _requires_sequential_execution,
     execute_tool_calls,
     execute_tools,
 )
@@ -33,6 +38,7 @@ def _tool(
     *,
     execute: Any | None = None,
     execution_mode: str | None = None,
+    parallel_safe: bool = True,
 ) -> AgentTool:
     return AgentTool(
         name=name,
@@ -40,6 +46,7 @@ def _tool(
         input_schema=_schema(["value"]),
         execute=execute or (lambda args, _ctx: {"value": args["value"]}),
         execution_mode=execution_mode,  # type: ignore[arg-type]
+        parallel_safe=parallel_safe,
     )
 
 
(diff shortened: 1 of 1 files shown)


cloudenochcsis:
@greptile review

PR #3706, opened 2026-07-03
feat(shell): allow skipping first-launch GitHub sign-in

muddlebee:
## Summary

- Make the first-launch GitHub sign-in prompt skippable instead of blocking REPL startup
- Add a "Skip for now" menu and treat Escape/Ctrl-C during browser device-flow polling as skip
- Persist `interactive.github_login_deferred` in `~/.opensre/config.yml` so skipped users are not re-prompted every launch
- Clear the deferral flag on successful sign-in or when GitHub is removed via `/integrations remove github`

image

## Test plan

- [x] `uv run python -m pytest tests/interactive_shell/runtime/test_first_launch_github.py -q`
- [x] `uv run python -m pytest tests/config/test_repl_config.py -q`
- [x] `uv run ruff check` on touched files
- [ ] Manual: launch `opensre` without GitHub configured, press Escape at the menu or during browser wait, confirm REPL starts
- [ ] Manual: relaunch and confirm the gate stays skipped; run `/integrations remove github` and confirm the gate returns on next launch

### DIFF
--- a/config/repl_config.py
+++ b/config/repl_config.py
@@ -237,3 +237,33 @@ def read_prompt_log_settings() -> dict[str, Any]:
     interactive = _read_config_file()
     raw = interactive.get("prompt_log", {})
     return raw if isinstance(raw, dict) else {}
+
+
+def read_github_login_deferred() -> bool:
+    """Return True when the user skipped the first-launch GitHub sign-in prompt."""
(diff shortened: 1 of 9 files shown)


muddlebee:
@greptile review

Issue #3551, opened 2026-07-01
(T-18): Full T-1 vendor-first completion


0xpaulx:
## Task T-18

**Level:** Maintainer | **Complexity:** XL | **Track:** architecture

**Summary:** Complete the T-1 vendor-first layout — every vendor should have `{config, client, verifier, tools}` with no legacy flat modules remaining.

**Scope:** Builds on C-14, C-15, C-16a–e and all Q- refactors.

**This is a Maintainer-only task.** Do not pick as a community issue.

0xpaulx:
## ✅ Acceptance Criteria (added by @viktor)

**AC — Surfaces must remain functional after refactoring:**
- [ ] Telegram gateway works end-to-end
- [ ] CLI (`opensre` command) works end-to-end
- [ ] Interactive shell works end-to-end
- All three surfaces must be **tested** before the PR is merged.

> _Requirement raised by @vincentserpoul, confirmed by @yauhen-tarashenka._

davincios:
Closing as part of the open-source roadmap refresh: investigation, evidence-mapper, AI-SRE, and refactor-only tasks are no longer on the public roadmap. The roadmap now focuses on scheduled tasks, CI/CD skills, and parallelization.

PR #3719, opened 2026-07-04
(T-18): Wire jira/prefect verifiers, fix Elasticsearch tool source key

muddlebee:
Partial progress on #3551 (T-18) — see scope note below for what's intentionally not covered.

Partial progress on #3551 (T-18: full T-1 vendor-first completion).

## Summary

T-18 is genuinely XL (asks for `{config, client, verifier, tools}` completeness across ~73 `integrations/*` folders). Rather than a rushed sweep, I audited the whole tree first and found the raw file-presence signal ("no client.py", "no verifier.py") was mostly false positives — shared AWS/DB drivers, delivery-only integrations, non-vendor utility packages. This PR fixes the narrower set of **real, verified gaps**:

- **jira, prefect**: both had a full client + tools + config model but no `verifier.py`, so `opensre integrations verify` silently skipped them even though their documented setup path (`~/.opensre/integrations.json`) works. Added `integrations/jira/verifier.py` and `integrations/prefect/verifier.py` (config-presence checks, matching the existing `opensearch` pattern). Prefect was also missing its `classify()` function and `_CLASSIFIERS` registration entirely — its documented store-based config was being silently ignored. Both are now flagged `has_verifier=True` in `integrations/registry.py`.

- **Elasticsearch logs tool bug**: `integrations/elasticsearch/tools/ElasticsearchLogsTool` read from a `sources["elasticsearch"]` key that nothing in the codebase ever populates — no catalog wiring, no docs, ever existed for it as a standalone source. I initially assumed the fix was to wire up a new parallel "elasticsearch" catalog entry, but `docs/opensearch.mdx` already documents `query_elasticsearch_logs` as one of two tools that activate together once OpenSearch/Elasticsearch is configured, sharing the same credentials. So the actual bug is simpler: the tool should read the already-fully-wired `sources["opensearch"]` key, not a dead one. Fixed `is_available`/`extract_params` accordingly and updated the shared test fixture (`tests/tools/conftest.py`) and the two test files that asserted the old (dead) key.

- Added a **Verify** section to `jira.mdx` and `prefect.mdx`, and `tests/integrations/test_prefect_catalog.py` covering the new classify path.

## Scope note

Left for separate follow-up (each needs a product judgment call, not a mechanical fix):
- **`trello`** — only has a config model, unreferenced anywhere in `catalog.py`; needs a keep-and-build-out vs. delete-as-dead-code decision.
- **`notion`** — has a client but no `tools/` or `verifier.py`; needs new tool(s) built, not just wiring.
- **`airflow`** — only used for alert-source classification today; would need a real client + tools to become query-able.

## Test plan

- [x] `make lint`, `make format-check`, `make typecheck` — all pass
- [x] `pytest tests/integrations tests/tools` — 4744 passed, 3 skipped (pre-existing skips, unrelated)
- [x] `.github/ci/check_imports.py` (cycles + import-linter layers + forbidden direct imports) — all pass
- [x] `make verify-integrations` — `jira` and `prefect` now appear in the verify table (as `missing`, since neither is configured in this dev environment — same as every other unconfigured integration; confirmed this non-zero exit is pre-existing behavior on `main` too, not something this PR introduces)
- [x] Verified the ES fix by checking `tests/tools/test_opensearch_analytics_tool.py` and the renamed elasticsearch tests together — both tools now correctly share the `opensearch` source

### DIFF
--- a/docs/jira.mdx
+++ b/docs/jira.mdx
@@ -61,6 +61,22 @@ Add to `~/.opensre/integrations.json`:
 For Jira Data Center/Server, use a personal access token from **Profile** → **Personal Access Tokens** instead.
 </Info>
 
+## Verify
+
+```bash
+opensre integrations verify jira
+```
+
+Expected output on success:
+
+```
+Service: jira
+Status:  passed
+Detail:  Configured for Jira at https://your-org.atlassian.net.
+```
+
+This is a configuration-presence check (`base_url`, `email`, and `api_token` are all set) rather than a live API call.
+
 ## Troubleshooting
 
 | Symptom | Fix |
--- a/docs/prefect.mdx
+++ b/docs/prefect.mdx
@@ -73,12 +73,20 @@ For self-hosted Prefect Server, set `api_url` to your server's API endpoint (no 
 opensre integrations verify prefect
 ```
 
-Expected output on success:
+Expected output on success (self-hosted, `api_url` set):
 
 ```
 Service: prefect
 Status:  passed
 Detail:  Configured for Prefect at http://prefect-server:4200/api.
+```
+
+For Prefect Cloud (only `api_key` set, no `api_url`):
+
+```
+Service: prefect
+Status:  passed
+Detail:  Configured for Prefect at cloud.
 ```
 
 This is a configuration-presence check (`api_url` or `api_key` is set) rather than a live API call.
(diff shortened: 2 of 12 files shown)


muddlebee:
@greptile review

PR #3720, opened 2026-07-04
Fix silent failures for missing/invalid subcommands

### LINKED ISSUE #3345: [UX] Command errors lack usage guidance — fix all edge cases (not just `/integrations show`)
## Summary

The `/integrations show` command fails silently when called with sub-arguments or incorrect usage — it doesn't tell the user what valid input looks like.

## Current Behavior

- `/integrations show` runs without errors but silently does nothing in some code paths
- When a user types `/truecommand truesubcommand` but uses the wrong subcommand, the existing "unknown subcommand" handler is **not triggered** because the subcommand is syntactically valid but logically unrecognized
- The user gets no guidance on what valid input to use

## Expected Behavior

Every invalid or unclear command invocation should respond with a clear, actionable error message that shows the user what valid input looks like. For example:

```
❌ Unknown subcommand: 'show'
Usage:
  /integrations list           — list all configured integrations
  /integrations check <name>   — verify a specific integration
  /integrations help           — show this message
```

## Scope

This issue covers **all edge cases** of command parsing, not just `/integrations show`:

1. **`/fakecommand`** — entirely unknown top-level command → show top-level help
2. **`/truecommand fakesubcommand`** — known command, unknown sub-command → show subcommand help
3. **`/truecommand truesubcommand`** — known command AND known sub-command that yields no output / silently fails → show usage hint ← **the case `/integrations show` currently hits**
4. Edge: commands with required arguments called without them (e.g. `/integrations check` with no name)

The fix should be applied consistently across all command handlers, with `/integrations show` as the concrete example to start from.

## Why This Is Achievable for a First Contribution

- The command dispatch logic lives in a single, well-defined location
- No deep knowledge of the agent or integration layer required
- A fix for `/integrations show` is a clear reference to generalise from
- Good test coverage opportunity: one test per edge case above

## Steps to Reproduce

1. Run `opensre` interactively
2. Type `/integrations show`
3. Observe: no clear error or usage hint is printed

## Suggested Approach

1. Audit `cli/` command dispatch for all routes
2. Add a `_unknown_subcommand_handler` fallback that prints usage
3. Add a `_no_output_guard` that detects when a valid subcommand produces no output and suggests alternatives
4. Write tests for each edge case

aryan-pratik:
Fixes #3345

**Describe the changes you have made in this PR -**
This PR centralizes and fixes the missing/silent error messages for slash commands when they are called incorrectly or with unknown subcommands.

Previously, commands like `/integrations show invalid` or `/model bogus` would fail silently and return to the prompt without guiding the user. To fix this across all commands, this PR introduces:
1. **`unknown_subcommand_handler`**: A unified fallback router in `errors.py`. It is now the default path in command dispatch blocks, catching any syntactically valid but unrecognized subcommands and printing a standard `❌ Unknown subcommand:` error along with the full usage list.
2. **`@no_output_guard`**: A new decorator that captures console output during command execution. If a command runs but prints absolutely nothing (e.g., calling `/integrations list` when the store is empty), the guard automatically injects a fallback hint.
3. **Refactored Route Handlers**: Applied the fallback handler and output guard across `/integrations`, `/mcp`, `/model`, `/fleet`, `/rca`, and `/background`.
4. **Tests**: Added `test_errors.py` to cover the new guard and handler, and updated existing string-matching test assertions to reflect the new standardized error formatting.

**Demo for feature changes and bug fixes -**
```text
$ /integrations help                              
❌ Unknown subcommand: 'help'
Usage:
  /integrations list         — list all configured integrations
  /integrations ls           — alias for list
  /integrations verify       — run health checks on all integrations
  /integrations show         — show details for a single integration  

$ /model bogus                                    
❌ Unknown subcommand: 'bogus'
Usage:
  /model show                — show active provider and models
  /model set                 — switch provider  ·  /model set <provider> 
  /model restore             — restore the active provider's default reasoning model
  /model toolcall            — manage toolcall model for the active provider

### DIFF
--- a/config/webapp.py
+++ b/config/webapp.py
@@ -4,8 +4,12 @@ from fastapi import FastAPI, Response, status
 from pydantic import BaseModel, ValidationError
 
 from config.config import LLMSettings, get_environment
+from config.platform_bootstrap import ensure_project_platform_package
 from config.version import get_version
-from platform.observability.sentry_sdk import init_sentry
+
+ensure_project_platform_package()
+
+from platform.observability.sentry_sdk import init_sentry  # noqa: E402
 
 init_sentry(entrypoint="webapp")
 
--- a/gateway/agent/gateway_action_tools.py
+++ b/gateway/agent/gateway_action_tools.py
@@ -9,6 +9,7 @@ from typing import Any
 import config.constants.platform as _platform
 from core.agent_harness.session import ReplSession
 from core.context.state import InvestigationState
+from core.tool_framework.registered_tool import RegisteredTool
 from core.types import AgentToolContext
 from gateway.agent.gateway_output_sink import GatewayOutputSink
 from tools.interactive_shell.contracts import object_schema, string_property
@@ -16,7 +17,6 @@ from tools.interactive_shell.shell.display import format_shell_command_for_displ
 from tools.interactive_shell.shell.execution import ShellExecutionResult, execute_shell_command
 from tools.interactive_shell.shell.parsing import parse_shell_command
 from tools.interactive_shell.shell.policy import plan_shell_execution
-from core.tool_framework.registered_tool import RegisteredTool
 
 GATEWAY_RESOURCE_KEY = "gateway"
 _SHELL_COMMAND_TIMEOUT_SECONDS = 120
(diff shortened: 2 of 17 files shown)


muddlebee:
@aryan-pratik closing this PR, its heavily outdated and greptile review pending. feel free to open a fresh PR

PR #3745, opened 2026-07-05
Add auth login hint for credit exhausted errors

### LINKED ISSUE #3739: ux: add /auth login hint when credit-exhausted error is shown
## Background

When a user's LLM provider runs out of credits, the error message currently says something like 'go to billing' but gives no actionable path to switch to a different provider. Users are left in a dead end with no clear next step.

This was flagged by @Yauhen on Jul 4: the shell shows a billing error, but there is no hint that `/auth login` (or `opensre auth login <provider>`) can be used to switch to a different provider immediately.

## Task

In the credit-exhausted / authentication-failure error path, add a hint like:

The hint should appear:
- In the interactive shell REPL when a billing/auth error is returned by the LLM
- Ideally also in the CLI one-shot mode

## Files

Look for where LLM billing/credit errors are caught and shown to the user. The fix is additive — append the hint to the existing error message.

## Acceptance criteria

- A user who hits a credit-exhausted error sees a clear next-step hint pointing to `/auth login`
- No behaviour change for successful calls

## Verify

Run the interactive shell, trigger or simulate a credit-exhausted error, confirm the hint is shown.

⚠️ AC: Telegram, CLI, and interactive shell must all work and be tested before merging.

vivekpatil017:
## Summary

Adds a `/auth login` recovery hint when a credit-exhausted error is shown in the interactive shell.

## Changes

- Replaced the previous recovery hint with a `/auth login` hint.
- Users are now guided to authenticate or switch providers using:
  - `/auth login`
  - `opensre auth login <provider>`

## Testing

- Verified the interactive shell displays the new hint when a `LLMCreditExhaustedError` is [shown.[](url)](Fixes #3739)

### DIFF
--- a/surfaces/interactive_shell/runtime/agent_presentation.py
+++ b/surfaces/interactive_shell/runtime/agent_presentation.py
@@ -98,7 +98,7 @@ async def _render_agent_presentation_transition(
             if isinstance(exc, LLMCreditExhaustedError):
                 console.print(
                     f"[{DIM}]Run /auth login (or `opensre auth login <provider>`) "
-                    f"to switch to another provider.[/]"
+                    "to switch to another provider.[/]"
                 )
         case "turn_end":
             set_prompt_suppress_fn(None)
--- a/surfaces/interactive_shell/runtime/agent_presentation.py
+++ b/surfaces/interactive_shell/runtime/agent_presentation.py
@@ -96,7 +96,10 @@ async def _render_agent_presentation_transition(
             from core.llm.llm_retry import LLMCreditExhaustedError
 
             if isinstance(exc, LLMCreditExhaustedError):
-                console.print(f"[{DIM}]Run /model to switch to another provider.[/]")
+                console.print(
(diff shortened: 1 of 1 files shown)


cerencamkiran:
I can’t see the Telegram and CLI one-shot parts in this pr. Could you please point me to where they are tested?

vivekpatil017:
> I can’t see the Telegram and CLI one-shot parts in this pr. Could you please point me to where they are tested?

Thanks for pointing that out! This PR currently updates and verifies the interactive shell path only. While looking into the CLI and Telegram paths, I found that LLMCreditExhaustedError is only handled in surfaces/interactive_shell/runtime/agent_presentation.py, and the CLI investigation flow follows a different error-mapping path. I wasn't able to locate the Telegram or CLI one-shot handlers for this error. Could you point me to the code paths you expect this hint to be added to? I'll update the PR to cover those as well.

cerencamkiran:
Thanks for your work. I think this issue needs the full acceptance criteria covered, which another pr seems to do.

PR #3760, opened 2026-07-06
fix(core/llm): resolve provider name dynamically in OpenAIAgentClient errors

### LINKED ISSUE #3753: [BUG] Hardcoded "OpenAI" prefix in `OpenAIAgentClient` error messages when using other OpenAI-compatible LLM providers
### Summary

When using any non-OpenAI but OpenAI-compatible LLM provider (such as DeepSeek, Gemini, OpenRouter, Nvidia NIM, Groq, etc.), exceptions raised during agent/tool-calling loop invocations are incorrectly reported to the user as `"OpenAI"` failures.
This occurs because [OpenAIAgentClient.invoke](file:///home/saswatdev/OpenSRE/opensre/core/llm/sdk/agent_clients.py#L462) contains hardcoded error string prefixes (e.g. `OpenAI API failed`, `OpenAI authentication failed`, etc.) in its `except` blocks instead of dynamically resolving the active provider name.

## Expected Behavior
When an LLM provider such as **DeepSeek** is configured, any API errors (e.g., authentication, billing, or connection errors) should be reported under that provider's name (e.g., `Deepseek API failed: ...` or `Deepseek authentication failed`).

---

## Actual Behavior
When **DeepSeek** is configured and its API request fails (e.g., due to `Insufficient Balance` / HTTP 402), the error message is reported as:
`RuntimeError: OpenAI API failed: Error code: 402 - {'error': {'message': 'Insufficient Balance', ...}}`

### Steps to reproduce

1. Configure DeepSeek as the provider in your `.env` (or via `opensre config`):
   ```bash
   LLM_PROVIDER=deepseek
   DEEPSEEK_REASONING_MODEL=deepseek-v4-pro
   DEEPSEEK_MODEL=deepseek-v4-pro
   ```
2. Store a DeepSeek API key that has an insufficient balance in the system keychain (or environment).
3. Start the interactive shell:
   ```bash
   uv run opensre
   ```
4. Type `hi` and press Enter.
5. The shell displays:
   `assistant failed: OpenAI API failed: Error code: 402 - {'error': {'message': 'Insufficient Balance', 'type': 'unknown_error', 'param': None, 'code': 'invalid_request_error'}}`

Yes

### How often does it occur?

Every time

### Operating system

Linux

### Logs and error output

When invoking `get_llm_for_reasoning().invoke("hi")` directly with DeepSeek configured, the following traceback is generated:

```
ERROR encountered during invocation:
LLM API request failed after multiple retries. Try again in a few seconds.
  File "/home/saswatdev/OpenSRE/opensre/core/llm/sdk/llm_clients.py", line 752, in invoke
    response = client.chat.completions.create(**kwargs)
  File "/home/saswatdev/OpenSRE/opensre/.venv/lib/python3.14/site-packages/openai/_utils/_utils.py", line 298, in wrapper
    return func(*args, **kwargs)
  File "/home/saswatdev/OpenSRE/opensre/.venv/lib/python3.14/site-packages/openai/resources/chat/completions/completions.py", line 1251, in create
    return self._post(
        "/chat/completions",
        ...
  File "/home/saswatdev/OpenSRE/opensre/.venv/lib/python3.14/site-packages/openai/_base_client.py", line 1105, in request
    raise self._make_status_error_from_response(err.response) from None
openai.APIStatusError: Error code: 402 - {'error': {'message': 'Insufficient Balance', 'type': 'unknown_error', 'param': None, 'code': 'invalid_request_error'}}
```

### Additional context

The standard [OpenAILLMClient](file:///home/saswatdev/OpenSRE/opensre/core/llm/sdk/llm_clients.py#L632) already resolves provider names dynamically via `self._provider_label`. 

The fix is to implement the same logic in [OpenAIAgentClient](file:///home/saswatdev/OpenSRE/opensre/core/llm/sdk/agent_clients.py#L432):
1. In `__init__`, derive `self._provider_label` dynamically:
   ```python
   self._provider_label = api_key_env.removesuffix("_API_KEY").replace("_", " ").title()
   ```
2. Replace hardcoded `"OpenAI"` strings inside `invoke` with `self._provider_label`:
   ```python
   except AuthenticationError as err:
       raise RuntimeError(f"{self._provider_label} authentication failed.") from err
   except NotFoundError as err:
       raise RuntimeError(f"{self._provider_label} model '{self._model}' not found.") from err
   except Exception as err:
       ...
       raise RuntimeError(f"{self._provider_label} API failed: {err}") from err
   ```

muddlebee:
## Summary
- `OpenAIAgentClient.invoke` (`core/llm/sdk/agent_clients.py`) hardcoded the literal string `"OpenAI"` in every error message, so OpenAI-compatible providers (DeepSeek, Gemini, NVIDIA NIM, Groq, etc.) had their failures reported as `"OpenAI ..."` instead of under their own name.
- Added a `_provider_label` property that derives the display name from `_api_key_env`, matching the existing convention already used by `OpenAILLMClient` (`core/llm/sdk/llm_clients.py`) — reusing that precedent instead of adding a new provider-name mapping/object.
- Replaced every hardcoded `"OpenAI ..."` string in `invoke` (auth, not-found, bad-request, rate-limit, permission-denied, generic-failure, invocation-failed, unexpected-response) with the dynamic label.

## Test plan
- [x] `uv run python -m pytest tests/core/runtime/llm/test_agent_llm_client.py -q` — all tests pass except 4 pre-existing, unrelated failures (confirmed present on `main` before this change, tied to local `LiteLLMAgentClient` routing).
- [x] Added `test_openai_agent_client_reports_configured_provider_name_on_failure`, reproducing the DeepSeek 402 "Insufficient Balance" scenario from the issue and asserting the error now reports "Deepseek", not "OpenAI".
- [x] Updated 3 existing tests whose assertions relied on the old hardcoded "OpenAI" string.
- [x] `make lint` / `make format-check` / `make typecheck` on changed files.

Fixes #3753

### DIFF
--- a/core/llm/sdk/agent_clients.py
+++ b/core/llm/sdk/agent_clients.py
@@ -429,6 +429,17 @@ def _openai_max_token_kwarg(model: str) -> str:
     return "max_tokens"
 
 
+# .title() breaks brand names with an internal capital letter (e.g.
+# "OPENAI_API_KEY" -> "Openai" instead of "OpenAI"). Override just the
+# providers this class is actually constructed with (see
+# core/llm/openai_compat_providers.py) where that happens.
+_PROVIDER_LABEL_OVERRIDES = {
+    "OPENAI_API_KEY": "OpenAI",
+    "OPENROUTER_API_KEY": "OpenRouter",
+    "MINIMAX_API_KEY": "MiniMax",
+}
+
+
 class OpenAIAgentClient:
     """OpenAI-compatible client with tool-calling for the agent loop."""
 
@@ -459,8 +470,9 @@ class OpenAIAgentClient:
     @property
     def _provider_label(self) -> str:
         api_key_env = str(getattr(self, "_api_key_env", "OPENAI_API_KEY"))
-        if api_key_env == "OPENAI_API_KEY":
-            return "OpenAI"
+        override = _PROVIDER_LABEL_OVERRIDES.get(api_key_env)
+        if override:
+            return override
         return api_key_env.removesuffix("_API_KEY").replace("_", " ").title()
 
     def tool_schemas(self, tools: list[Any]) -> list[dict[str, Any]]:
--- a/core/llm/sdk/agent_clients.py
+++ b/core/llm/sdk/agent_clients.py
@@ -459,7 +459,9 @@ class OpenAIAgentClient:
     @property
     def _provider_label(self) -> str:
         api_key_env = str(getattr(self, "_api_key_env", "OPENAI_API_KEY"))
-        return api_key_env.removesuffix("_API_KEY").replace("_", " ").title()
+        label = api_key_env.removesuffix("_API_KEY").replace("_", " ").title()
+        # .title() mangles the canonical "OpenAI" casing to "Openai" — restore it.
+        return label.replace("Openai", "OpenAI")
 
     def tool_schemas(self, tools: list[Any]) -> list[dict[str, Any]]:
         return build_openai_tool_specs(tools)
--- a/core/llm/sdk/agent_clients.py
+++ b/core/llm/sdk/agent_clients.py
@@ -459,9 +459,9 @@ class OpenAIAgentClient:
     @property
     def _provider_label(self) -> str:
         api_key_env = str(getattr(self, "_api_key_env", "OPENAI_API_KEY"))
-        label = api_key_env.removesuffix("_API_KEY").replace("_", " ").title()
-        # .title() mangles the canonical "OpenAI" casing to "Openai" — restore it.
-        return label.replace("Openai", "OpenAI")
+        if api_key_env == "OPENAI_API_KEY":
+            return "OpenAI"
+        return api_key_env.removesuffix("_API_KEY").replace("_", " ").title()
 
     def tool_schemas(self, tools: list[Any]) -> list[dict[str, Any]]:
         return build_openai_tool_specs(tools)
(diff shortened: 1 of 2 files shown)


muddlebee:
While reviewing Greptile's finding on the `.title()` cosmetic regression for `"OpenAI"`, checked whether the same issue affects the other providers `OpenAIAgentClient` is actually constructed with (`OPENAI_COMPATIBLE_PROVIDERS` in `core/llm/openai_compat_providers.py`):

| provider | derived via `.title()` | canonical brand |
|---|---|---|
| openai | Openai ✗ | OpenAI |
| openrouter | Openrouter ✗ | OpenRouter |
| deepseek | Deepseek (accepted — matches the issue's own suggested output) | DeepSeek |
| gemini | Gemini ✓ | Google Gemini |
| nvidia | Nvidia (reasonable simplification) | NVIDIA NIM |
| minimax | Minimax ✗ | MiniMax |
| groq | Groq ✓ | Groq |
| ollama | Ollama ✓ | Ollama |

`.title()` breaks any brand name with an internal capital letter — not just OpenAI. OpenRouter and MiniMax have the identical bug. Extended the override (`_PROVIDER_LABEL_OVERRIDES`) to cover both, with parametrized test coverage across all providers this class supports.

Considered sourcing labels from `config/llm_auth/provider_catalog.py`'s `PROVIDER_SPECS` instead of a small override dict, since that's the canonical provider registry. Didn't go that route: that catalog is keyed for the auth/wizard UI, and for at least `ollama` its `api_key_env` field (`OLLAMA_HOST`) doesn't match what `openai_compat_providers.py` actually constructs the client with (`OLLAMA_API_KEY`, a synthetic default since Ollama needs no real key). Reconciling those two registries to fix a cosmetic label felt like a separate, riskier refactor rather than something to fold into this bug fix — happy to file that as a follow-up if useful.

PR #3771, opened 2026-07-06
fix(packaging): bundle LiteLLM data in PyInstaller binary

### LINKED ISSUE #3631: [FEATURE] Implementing new LLM provider for Azure OpenAI
### Problem statement

Hi, I would like to use OpenSRE with Azure OpenAI, but this LLM provider is currently not available in the documentation and it is currently not possible to configure a different URL for the existing LLM provider OpenAI as there is no dedicated environment variable.

If possible I would like to request the implementation of this new LLM provider.

### Proposed solution

The Azure OpenAI endpoint URL is usually in this format: `https://<resource-name>.openai.azure.com/`
If compatible, an option would be to allow to set a different URL for the existing LLM provider OpenAI so it can be used with Azure OpenAI endpoints instead of the default OpenAI URL.

muddlebee:
## Summary
- Fixes #3631 — curl-installed binaries crash on Azure/LiteLLM because PyInstaller did not bundle LiteLLM package data
- Add `--collect-data litellm` to the release binary build so `model_prices_and_context_window_backup.json` lands under `_internal/litellm/`
- Add a Unix release smoke check and packaging contract tests so this cannot regress silently

## Context
LiteLLM loads its bundled model cost/context catalog at `import litellm` time (even when fetching a remote map). OpenSRE imports LiteLLM for Azure OpenAI and `OPENSRE_LLM_TRANSPORT=litellm`. Source installs via `uv sync` work; only the frozen binary was missing the JSON.

## Test plan
- [x] `uv run python -m pytest tests/packaging/test_litellm_bundle_contract.py tests/github_ci/test_release_workflow.py`
- [ ] Release workflow smoke test passes on next main-build publish (checks `_internal/litellm/model_prices_and_context_window_backup.json`)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -273,6 +273,7 @@ jobs:
           --noconfirm
           --collect-data surfaces.cli
           --collect-data config
+          --collect-data litellm
           --collect-submodules integrations
           --collect-submodules surfaces.interactive_shell
           --collect-submodules tools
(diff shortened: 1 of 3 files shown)


muddlebee:
@greptile review

PR #3763, opened 2026-07-06
fix(core/llm): close remaining _opensre_* marker leak paths from #3762

muddlebee:
## Summary
- Strip internal `_opensre_*` markers in `LiteLLMAgentClient._build_request_kwargs` (`core/llm/litellm/clients.py`), mirroring the fix already applied to the native SDK agent clients in #3761. This path is reachable whenever `OPENSRE_LLM_TRANSPORT=litellm` routes to a strict provider (e.g. Anthropic), not just Azure OpenAI — `core/llm/litellm/routing.py` has a first-class `anthropic` route through LiteLLM.
- Strip markers in `MessageFormatter.to_provider_messages()` (`core/messages/message_formatter.py`) so a `RuntimeMessage` coerced from a marked provider dict can't replay `_opensre_*` verbatim if a future caller routes investigation transcripts through `MessageFormatter`/`ReactLoop` instead of the current raw-dict path. Currently latent (no live caller hits it today), but cheap to close now.

Addresses gaps #1 (LiteLLM transport) and #3 (MessageFormatter round-trip) from #3762. Gaps #2 (OpenAI SDK) was already fixed in #3761. Gaps #4-#6 are lower-severity behavioral/policy items tracked separately in #3762.

## Test plan
- [x] Added `test_litellm_agent_client_invoke_strips_internal_message_markers` (`tests/core/runtime/llm/test_litellm_compat.py`), mirroring the existing Anthropic/OpenAI marker-strip tests.
- [x] Added `test_normalize_then_to_provider_messages_strips_internal_markers` (`tests/core/messages/test_message_formatter.py`) reproducing the round-trip leak scenario.
- [x] `make lint`, `make format-check`, `make typecheck` all pass.
- [x] `uv run python -m pytest tests/core/messages/test_message_formatter.py tests/core/runtime/llm/test_litellm_compat.py tests/core/test_context_budget.py -q` passes.
- Note: `make test-scope` on this diff surfaces 43 failures in `tests/core/agent/test_turn_scenarios.py` and `tests/core/runtime/llm/test_llm_client*.py` (live-network Azure connection errors + provider-routing assertions). Verified these are **pre-existing test-order pollution**, reproducing identically on the unmodified `main` baseline before this change — not caused by this diff.

Fixes #3762 (partially — see remaining gaps above)

PR #3772, opened 2026-07-06
fix(core/domain): strip _opensre_* markers from persisted agent_messages

### LINKED ISSUE #3762: follow-up: harden _opensre_* marker handling after context-budget eviction (#2959)
## Context

#2959 added internal `_opensre_seed` and `_opensre_duplicate_result` keys on investigation message dicts for context-budget eviction heuristics. Those keys must stay on the in-memory transcript but must not reach provider APIs.

#3741 reported Anthropic HTTP 400 (`messages.N._opensre_seed: Extra inputs are not permitted`). #3761 strips markers in native `AnthropicAgentClient` and `BedrockConverseAgentClient` invoke paths.

This issue tracks **remaining gaps** from the #2959 design that are not covered by #3761.

### 1. LiteLLM transport (high)

When `OPENSRE_LLM_TRANSPORT=litellm` (required for Azure OpenAI; optional elsewhere), investigation uses `LiteLLMAgentClient`, which passes messages through without stripping internal keys (`core/llm/litellm/clients.py` → `_build_request_kwargs`).

Any LiteLLM route to Anthropic (direct `anthropic/...`, Azure-hosted Claude, etc.) can hit the same HTTP 400 as #3741.

**Suggested fix:** call `strip_internal_message_markers()` in `LiteLLMAgentClient._build_request_kwargs()` (or a shared pre-invoke helper used by all agent clients).

### 2. OpenAI SDK path (low)

`OpenAIAgentClient.invoke()` still forwards raw message dicts. OpenAI usually tolerates extra keys, but compat providers may not.

**Suggested fix:** apply the same shared strip helper for consistency.

### 3. MessageFormatter round-trip (latent)

`MessageFormatter._coerce_runtime_message()` copies the full provider dict into `provider_payload` **including** `_opensre_*` keys, while also extracting them into `metadata`. `to_provider_messages()` replays `provider_payload` verbatim, so markers leak on round-trip.

`ConnectedInvestigationAgent` bypasses this today (raw dicts → `llm.invoke`), but **`ReactLoop` / `Agent.run()` would reintroduce the bug** if investigation normalizes marked dicts through `MessageFormatter` and re-sends them.

**Suggested fix:** strip `_opensre_*` when building provider payloads from `provider_payload` / `provider_payloads`, or exclude those keys when populating `provider_payload` at coerce time.

### 4. Partial duplicate turns never tagged

Duplicate marking only runs when **all** tool calls in a turn are cached (`gather_evidence/agent.py`: `if duplicate_flags and all(duplicate_flags)`). Mixed turns (one fresh + one duplicate) get no `_opensre_duplicate_result` tag and are evicted like normal exchanges under budget pressure.

### 5. Pinned seed pairs under extreme budget pressure

Seed exchanges are never evicted. Truncation can shrink oversized seed content, but if system + tools + pinned seed still exceed the ceiling after trimming, `enforce_context_budget` logs a warning and proceeds — potential overflow on very large seed payloads + full tool registry.

### 6. Internal keys in persisted `agent_messages`

Investigation state retains `_opensre_*` on `agent_messages` (intentional for trimming). Harmless for diagnose/report (text-only), but leaks implementation details into persisted state, degraded-failure payloads, and telemetry consumers of raw message JSON.

## Test gaps

- No LiteLLM invoke test asserting stripped outbound payloads
- No `MessageFormatter` round-trip test with marked dicts
- Context-budget / agent tests assert markers remain in state but did not catch API payload leaks until #3761

## Acceptance criteria

- [ ] All investigation agent client invoke paths strip `_opensre_*` before provider submission (LiteLLM + OpenAI at minimum)
- [ ] `MessageFormatter.to_provider_messages()` cannot reintroduce internal markers from `provider_payload`
- [ ] Regression tests for LiteLLM and MessageFormatter round-trip
- [ ] (Optional) Decide policy for partial-duplicate marking and document or implement

## Related

- #3741 — Anthropic HTTP 400 from marker leak
- #3761 — strip in Anthropic + Bedrock Converse SDK clients
- #2959 — original context-budget eviction + marker introduction

muddlebee:
## Summary
- `result_to_state()` (`core/domain/diagnosis/result.py`) now strips `_opensre_*` internal markers from `InvestigationResult.agent_messages` before writing it into persisted investigation state.
- Traced the leak: `tools/investigation/stages/diagnose/node.py:68` does `result.agent_messages = messages` straight from the gather_evidence loop's marked state, and `result_to_state()` wrote that unstripped into the persisted state dict. Diagnose is the last stage to touch `agent_messages` — by the time it runs, the context-budget eviction loop is done and the `_opensre_seed`/`_opensre_duplicate_result` markers have served their purpose, so leaving them on only leaks implementation detail into session files, degraded-failure payloads, and any consumer of raw `agent_messages` JSON.

Addresses gap #6 from #3762 (the other two open gaps, #4 partial-duplicate tagging and #5 pinned-seed overflow under extreme pressure, are policy/design questions left for a separate discussion rather than mechanical fixes).

## Test plan
- [x] Added `test_result_to_state_strips_internal_markers_from_agent_messages` (`tests/agent/test_result.py`), asserting the persisted state is clean while the original `InvestigationResult.agent_messages` dict is left unmutated (markers preserved in case anything upstream still needs them before this boundary).
- [x] `make lint`, `make format-check`, `make typecheck` all pass.
- [x] `make test-cov` (escalated since `core/domain/` is shared code): 10292 passed, 43 pre-existing failures — identical set/error signatures to the pre-existing `tests/core/agent/test_turn_scenarios.py` and `tests/core/runtime/llm/test_llm_client*.py` test-order pollution documented in #3763, unrelated to this diff (reproduces on unmodified `main`).

Fixes #3762 (gap #6 of 3; #4 and #5 remain open as policy discussions)

### DIFF
--- a/core/domain/diagnosis/result.py
+++ b/core/domain/diagnosis/result.py
@@ -9,6 +9,7 @@ from typing import Any
 
 from pydantic import BaseModel, Field
 
+from core.context_budget import strip_internal_message_markers
 from core.domain.diagnosis.alignment import apply_category_alignment_adjustments
 from core.domain.types.root_cause_categories import (
     HERMES_ROOT_CAUSE_CATEGORIES,
@@ -89,7 +90,10 @@ def result_to_state(result: InvestigationResult) -> dict[str, Any]:
         "investigation_recommendations": result.investigation_recommendations,
         "evidence": result.evidence,
         "evidence_entries": result.evidence_entries,
-        "agent_messages": result.agent_messages,
+        # Diagnose is the last stage to read agent_messages — the context-budget
+        # eviction markers on it (_opensre_seed, _opensre_duplicate_result) have
+        # already served their purpose and must not leak into persisted state.
+        "agent_messages": strip_internal_message_markers(result.agent_messages),
     }
 
 
--- a/tests/agent/test_result.py
+++ b/tests/agent/test_result.py
@@ -63,7 +63,7 @@ def test_result_to_state_strips_internal_markers_from_agent_messages() -> None:
         root_cause_category="disk_pressure",
         agent_messages=[
             {"role": "user", "content": "alert", "_opensre_seed": True},
-            {"role": "assistant", "content": "ok"},
+            {"role": "assistant", "content": "ok", "_opensre_duplicate_result": True},
         ],
     )
 
@@ -74,3 +74,4 @@ def test_result_to_state_strips_internal_markers_from_agent_messages() -> None:
         {"role": "assistant", "content": "ok"},
     ]
     assert result.agent_messages[0]["_opensre_seed"] is True
+    assert result.agent_messages[1]["_opensre_duplicate_result"] is True
--- a/tests/agent/test_result.py
+++ b/tests/agent/test_result.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
 from core.domain.diagnosis import (
+    InvestigationResult,
     build_diagnosis_schema,
     extract_last_assistant_text,
+    result_to_state,
     taxonomy_categories_for_alert_source,
 )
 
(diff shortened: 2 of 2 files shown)


muddlebee:
@greptile review

PR #3782, opened 2026-07-06
Remove dead [LLM provider: X] error prefix

### LINKED ISSUE #3341: [BUG] Error message shows mismatched LLM provider label ("openai") when actual error is from Anthropic
## Description

When a user has no LLM configured ("NO LLM INTEGRATED"), the error response shown in the assistant output contains a contradictory provider label.

**Observed:** The provider is reported as `openai` in the prefix — `[LLM provider: openai]` — but the error body itself is an Anthropic API error:

```
[LLM provider: openai] Anthropic request rejected (HTTP 400): Error code: 400 - {
  'type': 'error',
  'error': {
    'type': 'invalid_request_error',
    'message': 'Your credit balance is too low to access the Anthropic API. Please go to Plans & Billing to upgrade or purchase credits.'
  },
  'request_id': 'req_011CcXb15EXT6fF4n4eJwPSd'
}
```

**Expected:** The provider label in the error prefix should match the actual provider that was invoked (Anthropic), or error messages should accurately reflect the resolved LLM backend rather than a hardcoded / default label.

## Steps to Reproduce

1. Configure an Anthropic API key with an empty or insufficient credit balance
2. Trigger a prompt (e.g. send the string "test")
3. Observe the error message in the assistant output

## Impact

- Confusing / misleading for users debugging LLM provider issues
- Error triaging is harder when the provider label contradicts the actual error source

## Proposed Investigation

- Find where the  prefix is injected and confirm whether it is resolved from the actual provider at call-time or defaulted/hardcoded
- Ensure the provider label always reflects the resolved backend (e.g. the one that threw the error)
- Consider normalising the error format so provider mismatch is impossible

## Notes

- UI shows "NO LLM INTEGRATED" badge alongside "GITHUB · CONFIGURED" integration
- The runtime config appears to route to Anthropic despite the provider label showing `openai`

MugemaneBertin2001:
Fixes #3341

The reported output — `[LLM provider: openai] Anthropic request rejected …` — was produced by `config.config.llm_provider_error_context()`, which prepended the **configured** provider from `resolve_llm_settings_verbose` rather than the **invoked** one. If OpenAI was configured but the runtime fell through to Anthropic, the prefix still said `openai` while the body was an Anthropic API error — exactly the mismatch the issue described.

The runtime path that used to attach that prefix (`llm_action_planner/llm_client.py` → `PlannerLLMError`) was replaced by the AgentTool-based planner (\"Route shell actions through AgentTools\", commit e9cea6a3), and its caller was removed. Nothing in the current tree calls `llm_provider_error_context()` — grep confirms:

```
$ rg -F "llm_provider_error_context"
config/config.py:def llm_provider_error_context(     # dead definition
tests/test_config.py: (3 orphaned unit tests)
```

Provider errors now propagate directly from the specific client with the actual provider label baked in — e.g.:

- `core/llm/transports/sdk/agent_clients.py:530` — `f"{self._provider_label} request rejected: {err}"`
- `core/llm/shared/llm_retry.py:133` — `maybe_raise_credit_exhausted(provider_name, err)` uses the client's `provider_name`

Both use the provider the failing client was constructed for, so the openai/anthropic mismatch is no longer reachable.

This PR:

1. Deletes `llm_provider_error_context` from `config/config.py`.
2. Removes its three orphaned unit tests from `tests/test_config.py`.
3. Fixes the misleading example in `docs/llm-providers.mdx` that still showed the old `[LLM provider: openai]` prefix — the current behavior is that the provider name is in the error text itself.

### Demo

Before/after for the docs example:

```diff
-- **Provider errors** are prefixed with the configured provider that served the request:
-
-  ```
-  [LLM provider: openai]
-  Missing credential for LLM provider 'openai'. Set OPENAI_API_KEY or run `opensre auth login openai`.
-  ```
+- **Provider errors** name the provider that served the request in the message itself, so you can tell at a glance which backend failed:
+
+  ```
+  Missing credential for LLM provider 'openai'. Set OPENAI_API_KEY or run `opensre auth login openai`.
+  ```
```

Verification:

```
$ rg -F "llm_provider_error_context"
$ rg -F "[LLM provider"
docs/pi-coding.mdx: [LLM providers](/llm-providers))         # normal link text
docs/interactive-shell-commands.mdx: [LLM providers](/llm-providers)
docs/interactive-shell-commands.mdx: [LLM providers](/llm-providers)
docs/interactive-shell-commands.mdx: - [LLM providers](/llm-providers)

$ uv run pytest tests/test_config.py -q
.....................................                                    [100%]
37 passed in 0.44s

$ uv run ruff check config/config.py tests/test_config.py
All checks passed!

$ uv run mypy config/config.py
```

---

Investigation was mostly grep-driven. `git log -S "llm_provider_error_context"` showed the introducing commit (d302abaf, `feat(llm): make provider fallback observable with accurate errors`) and its caller in `llm_action_planner/llm_client.py`. `rg -n "PlannerLLMError"` returned zero hits, and `rg -n "llm_action_planner"` — same. So the whole caller path was already gone, leaving `llm_provider_error_context` and its tests orphaned.

Two options existed: (a) re-wire the prefix at a new call site, or (b) delete the dead code. (a) is redundant — the current client-level error messages already name the provider that failed (I traced `_provider_label` and `provider_name` in `core/llm/transports/sdk/agent_clients.py` and `core/llm/shared/llm_retry.py`), so re-adding a top-level prefix would either duplicate the label or reintroduce the same drift the issue was about. (b) fits the AGENTS.md rule about not leaving compatibility-only forwarding modules after a refactor. Also updated the docs example, which still described the old behavior.

---

### DIFF
--- a/config/config.py
+++ b/config/config.py
@@ -564,22 +564,6 @@ def describe_llm_resolution(
     return "\n".join(lines)
 
 
-def llm_provider_error_context(
-    fallback_providers: Sequence[str] = (),
-) -> str:
-    """Return a short bracketed provider context for prefixing error messages.
-
-    Never raises — diagnostics must not mask the original error. Returns an
-    empty string when resolution itself fails so callers can fall back to the
-    raw provider error untouched.
-    """
-    try:
-        resolution = resolve_llm_settings_verbose(fallback_providers)
-    except Exception:
-        return ""
-    return f"[LLM provider: {resolution.resolved_provider}]"
-
-
 def has_credentials_for_active_llm_provider() -> bool:
     """Return prompt-safe auth availability for the configured LLM provider."""
     settings = resolve_llm_settings()
--- a/docs/llm-providers.mdx
+++ b/docs/llm-providers.mdx
@@ -512,10 +512,9 @@ OpenSRE does not silently switch LLM providers when the provider in `LLM_PROVIDE
 - **`opensre auth` and `/auth status`** show prompt-safe status from environment variables, provider metadata, CLI probes, or ambient/local config.
 - **`opensre auth verify <provider>`** intentionally checks request-time credentials and refreshes metadata.
 - **`opensre config llm` and `opensre doctor`** report the configured provider plus credential status without resolving secrets.
-- **Provider errors** are prefixed with the configured provider that served the request:
+- **Provider errors** name the provider that served the request in the message itself, so you can tell at a glance which backend failed:
 
   ```
-  [LLM provider: openai]
   Missing credential for LLM provider 'openai'. Set OPENAI_API_KEY or run `opensre auth login openai`.
   ```
 
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -9,7 +9,6 @@ from config.config import (
     LLMSettings,
     describe_llm_resolution,
     has_credentials_for_active_llm_provider,
-    llm_provider_error_context,
     resolve_llm_settings,
     resolve_llm_settings_verbose,
 )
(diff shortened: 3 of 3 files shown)


Davidson3556:
can you add relevant context like screenshots @MugemaneBertin2001

MugemaneBertin2001:
Thanks for the review, @Davidson3556 / @muddlebee — I opened #3807 as the follow-up with the demo output, grep evidence, and the runtime call site inline so the reviewer does not have to re-derive them. Same commit (6e18e230), no code change from what you saw here. Reopened as a fresh PR because the API does not let me reopen a maintainer-closed one; happy to close #3807 and reopen this one instead if that is preferred.

PR #3783, opened 2026-07-06
Add filter, search, and count subcommands to /tools

### LINKED ISSUE #3340: /tools REPL output truncated in terminal due to response length
## Description

When calling `/tools` in the REPL, the response is so long that it gets cut off in the terminal — making it impossible to see the full list of available tools.

## Steps to Reproduce

1. Open the OpenSRE REPL
2. Run `/tools`
3. Observe that the output is truncated before all tools are listed

## Expected Behavior

The full list of tools should be visible — either paginated, scrollable, or written to a file/pager so nothing is lost.

## Possible Fixes

- Pipe output through a pager (e.g. `less`)
- Truncate with a `--filter` or `--search` flag so users can narrow results
- Write output to a temp file and display the path
- Add a `--count` / summary mode

## Notes

This is a UX issue that affects discoverability of available tools in the REPL.

---

*This issue is open to the community — good first issue for anyone looking to improve the REPL experience!* 🙌

MugemaneBertin2001:
Fixes #3340

`/tools` printed the entire tool registry as a single Rich table. On a build with 60+ registered tools, the earliest rows scroll off the top before the user can read them — the terminal wins, the tool loses, and there is no way to narrow the view without external tooling.

Extend the `/tools` router in `surfaces/interactive_shell/command_registry/tools_cmds.py` with two named subcommands and a bare-query shortcut, without dropping any existing alias:

- `/tools`, `/tools list`, `/tools ls`, `/tools tool`, `/tools tools` — full catalog (unchanged)
- `/tools count` — total plus per-surface breakdown
- `/tools search <query>` — substring match on `name` + `description`, case-insensitive
- `/tools <query>` — bare query treated as an implicit filter (so `/tools slack` narrows to slack tools without needing a subcommand)

Empty search results print a hint pointing at `/tools` and `/tools count`, and mark the turn as OK so the shell does not stay in an error state.

Slash-catalog MCP metadata for `/tools` is updated so the LLM router description mentions the new verbs, and first-arg tab completions expose `count` and `search`.

### Demo

Command surface after the change:

```
/tools                        # full list (unchanged)
/tools list | ls | tool | tools    # aliases (unchanged)
/tools count                  # summary
/tools search <query>         # explicit filter
/tools <query>                # implicit filter, e.g. /tools slack
```

Sample output of `/tools count`:

```
12 registered tools
  chat: 4
  investigation: 10
```

Sample output of `/tools search slack`:

```
┏━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Tool                 ┃ Surf ┃ Description                    ┃
┡━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┩
│ slack_send_message   │ chat │ Send a Slack message.          │
└──────────────────────┴──────┴────────────────────────────────┘
```

Verification:

```
$ uv run pytest tests/interactive_shell/command_registry/ -q
..................................................                       [100%]
50 passed, 1 warning in 0.74s
```

`test_tools_cmds.py` adds 16 unit tests covering: substring match (case-insensitive, name and description), bare-command list, list aliases, `count` (plural / singular / empty catalog), `search` (missing query error, hit, miss), bare-query shortcut, and the SlashCommand + first-arg-completion metadata.

Ruff, ruff format, and mypy all pass on the touched files.

---

Three fixes were proposed on the issue — pager, filter/search flag, temp file, count/summary. Pager and temp file break the flow of the REPL and add surface area (temp file cleanup, external process). The two composable ones — filter and count — are the ones users can reach for from muscle memory. I combined them.

I replaced the `make_list_root_handler` factory with a small custom router because `make_list_root_handler` errors on unknown args, and I wanted `/tools slack` to act as a filter rather than print an "unknown subcommand" hint. The named subcommands (`list`, `ls`, `count`, `search`) are still dispatched first, so tab completion and the LLM slash-invoke schema stay consistent. Everything mutates through pure helpers (`_filter_catalog`, `_matches`, `_render_count`) so they can be unit-tested without a real Session.

I kept the double-counting behavior for `count`: a tool exposed on both `investigation` and `chat` is counted under each surface, matching the way the list view groups them (`format_tool_catalog_text` uses the same shape). If the maintainers prefer unique-tool counts, that's a one-line change.

---

### DIFF
--- a/surfaces/interactive_shell/command_registry/slash_catalog.py
+++ b/surfaces/interactive_shell/command_registry/slash_catalog.py
@@ -292,9 +292,12 @@ _MCP_BY_COMMAND: dict[str, _SlashMcpFields] = {
     ),
     "/tools": _mcp(
         "Explicit /tools command operation: list registered investigation/chat tools "
-        "wired into this OpenSRE build.",
+        "wired into this OpenSRE build. Subcommands: list (default), count "
+        "(per-surface totals), search <query> or a bare <query> to filter by "
+        "name/description substring.",
         "User explicitly types /tools or asks to run /tools",
         "User explicitly asks to list registered tools as a shell command",
+        "User asks to count how many tools are registered or filter them by keyword",
         anti_examples=(
             "User asks conversationally what tools or capabilities the assistant can use (assistant_handoff)",
         ),
--- a/surfaces/interactive_shell/command_registry/tools_cmds.py
+++ b/surfaces/interactive_shell/command_registry/tools_cmds.py
@@ -4,39 +4,108 @@ from __future__ import annotations
 
 from rich.console import Console
 
-from surfaces.interactive_shell.command_registry.types import (
-    SlashCommand,
-    make_list_root_handler,
-)
+from platform.terminal.theme import DIM, ERROR, HIGHLIGHT
+from surfaces.interactive_shell.command_registry.types import SlashCommand
 from surfaces.interactive_shell.runtime import Session
 from surfaces.interactive_shell.ui import render_tools_table
-from surfaces.interactive_shell.ui.tables.tool_catalog import build_tool_catalog
+from surfaces.interactive_shell.ui.components.rendering import repl_print
+from surfaces.interactive_shell.ui.tables.tool_catalog import (
+    ToolCatalogEntry,
+    build_tool_catalog,
+)
+
+_LIST_ALIASES: frozenset[str] = frozenset({"list", "ls", "tool", "tools"})
 
 
-def _list_tools(_session: Session, console: Console, _args: list[str]) -> bool:
-    render_tools_table(console, build_tool_catalog())
+def _matches(entry: ToolCatalogEntry, query: str) -> bool:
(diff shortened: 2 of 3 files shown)


Davidson3556:
kindly address greptile feedback @MugemaneBertin2001

muddlebee:
@MugemaneBertin2001 we dont need all such subcommands stick to only `/tools`

MugemaneBertin2001:
@muddlebee — reworked in #3808. New version drops **all** subcommands per your ask, keeping the command surface identical to `main` (`/tools`, `/tools list`, `/tools ls`, `/tools tool`, `/tools tools`). The only change is that when the tools table would overflow the terminal, the render is wrapped in `console.pager(styles=True)` so users can scroll with `less` instead of losing the top of the list. Non-TTY / piped output still renders inline, so downstream tooling and CI logs are unaffected.

Issue #2678, opened 2026-05-30
[BUG] Inbound messaging authorization is implemented but never enforced; no Slack request-signature verification


7vignesh:
### Summary

`app/integrations/messaging_security.py` implements a full inbound-auth layer (`authorize_inbound_message`, `complete_pairing`, allowlists, DM pairing), but it has **zero callers in `app/`** only tests reference it. There is no inbound webhook/bot loop that routes traffic through these checks, and no Slack Events API signature verification anywhere in the codebase. The security-critical enforcement is effectively dead code, so any inbound handler added later can trivially bypass it.

### Expected vs actual behavior

**Expected:** Inbound messages from Telegram/Slack/Discord are authenticated before processing — Slack requests verified via `signing_secret` HMAC, and senders checked against `authorize_inbound_message()` / DM pairing.

**Actual:** No inbound handler calls `authorize_inbound_message()` or `complete_pairing()` (grep across `app/` returns only the definitions). No code reads or verifies `SlackBotConfig.signing_secret`. The config model docstring itself warns this is unsafe:

> "Without it, the Slack Events API webhook handler cannot verify request authenticity and will accept forged requests from any source."
> `app/integrations/config_models.py` (`SlackBotConfig`)

The module is labeled "Prerequisite for issue #1482 (conversational loop)", confirming it is an unfinished foundation rather than a wired-up control.

## Steps to Reproduce

1. **Search the app source for callers of the enforcement helpers:**

   ```bash
   grep -rn "authorize_inbound_message|complete_pairing" app/
   ```

   **Result:** Only definitions found in `app/integrations/messaging_security.py`; no call sites exist.

2. **Search for Slack signature verification:**

   ```bash
   grep -rn "signing_secret|X-Slack-Signature" app/
   ```

   **Result:** `signing_secret` is only defined/described in `config_models.py`; it is never used to verify a request.

3. **Confirm there is no inbound update/webhook loop** (`getUpdates`, events handler, gateway) **that applies the policy.**

### Can you reproduce it consistently?

Yes

### How often does it occur?

Every time

### Operating system

Other

### Logs and error output

N/A — this is a code-path/security gap identified by review, not a runtime error. Platform-independent.

### Additional context

- Affected files: `app/integrations/messaging_security.py`, `app/integrations/config_models.py` (`SlackBotConfig`), `app/cli/commands/messaging.py` (CLI that creates policies/pairing codes).
[...]

PR #3790, opened 2026-07-07
Fix: warn when HMAC falls back to hostname-derived pairing key

### LINKED ISSUE #3788: [BUG] Hostname-derived HMAC fallback key in _get_hmac_key enables offline brute-force of pairing codes if hash leaks
### Summary

When `OPENSRE_PAIRING_SECRET` is not set, `_get_hmac_key` derives the HMAC key from the machine hostname. If the stored `pairing_secret_hash` leaks from the database, an attacker can reconstruct the key from the hostname and brute-force the entire 6-character code space offline, bypassing the 5-attempt online limit entirely.

**Expected:** The pairing HMAC key is always a strong, secret value. Without a configured secret, the system should warn loudly or refuse to generate pairing codes rather than silently fall back to a predictable key.

**Actual:** `_get_hmac_key` silently falls back to `f"opensre-pairing-{hostname}"` — a value that is not secret and is trivially knowable by anyone with access to the machine name. The 5-attempt brute-force limit only applies to online attempts; offline attacks against a leaked hash are unbounded.

### Steps to reproduce

1. Deploy without setting `OPENSRE_PAIRING_SECRET` in the environment.
2. Run `opensre messaging pair` to generate a pairing code — a hash is stored in the database.
3. Obtain the stored `pairing_secret_hash` (e.g. via a DB leak or backup exposure).
4. Knowing the hostname (or guessing it), reconstruct the HMAC key as `f"opensre-pairing-{hostname}".encode()`.
5. Brute-force all 36^6 (~2.2 billion) combinations of the 6-character uppercase alphanumeric code space offline — no rate limit applies.

Yes

### How often does it occur?

Under specific conditions

### Operating system

Linux

### Logs and error output

No runtime error — the fallback is silent. The only indication is the absence of OPENSRE_PAIRING_SECRET in the environment.

### Additional context

Relevant code in `integrations/messaging_security.py`:

```python
def _get_hmac_key() -> bytes:
    env_secret = os.environ.get("OPENSRE_PAIRING_SECRET", "")
    if env_secret:
        return env_secret.encode()
    # Fallback: derive from hostname
    import platform
    machine_id = platform.node() or "opensre-default"
    return f"opensre-pairing-{machine_id}".encode()
```

The existing docstring acknowledges this risk but the fallback remains silent in production. Suggested fix: emit a `logger.warning` when falling back, and consider raising an error or refusing to generate pairing codes in production mode without an explicit secret set.

Related: #2678, #2677 (same security area — messaging gateway auth).

bastitva0-blip:
Fixes #3788

Added a `logger.warning` to `_get_hmac_key` when `OPENSRE_PAIRING_SECRET` is not 
set and the system falls back to a hostname-derived key. The fallback itself is 
preserved for local dev compatibility, but operators are now explicitly warned that 
the fallback is insecure in production.

Without this warning, deployments missing the env var silently used a predictable 
key. If `pairing_secret_hash` leaked from the database, an attacker knowing the 
hostname could reconstruct the key and brute-force all 36^6 pairing codes offline, 
bypassing the 5-attempt online limit entirely.

Added a test that asserts the warning is emitted when `OPENSRE_PAIRING_SECRET` is absent.

Related: #2678, #2677

### DIFF
--- a/gateway/session/enforce_inbound_telegram_message_security.py
+++ b/gateway/session/enforce_inbound_telegram_message_security.py
@@ -109,7 +109,24 @@ def enforce_inbound_telegram_message_security(
             ),
         )
 
+    result: AuthorizationResult = authorize_inbound_message(
+        policy=policy,
+        user_id=user_id,
+        chat_id=chat_id,
+        message_text=text,
+    )
+
     if text.strip().lower() == "/new":
+        if not result:
+            audit_log_inbound_message(
+                platform=MessagingPlatform.TELEGRAM.value,
+                user_id=user_id,
+                chat_id=chat_id,
+                message_hash=_message_hash(text),
+                authorized=False,
+                reason=result.reason,
+            )
+            return InboundDecision(allowed=False, reply_text=result.reason)
         audit_log_inbound_message(
             platform=MessagingPlatform.TELEGRAM.value,
             user_id=user_id,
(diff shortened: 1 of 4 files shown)


bastitva0-blip:
@greptile review

bastitva0-blip:
@greptile review

PR #3816, opened 2026-07-07
fix(incident-window): unwrap nested alert fixture envelopes

### LINKED ISSUE #3813: Incident window falls back to default when investigating datadog_k8s_alert.json fixture
## Summary

`opensre investigate -i tests/e2e/kubernetes/fixtures/datadog_k8s_alert.json` does **not** anchor the incident window on the alert's `startsAt`. Instead it falls back to `source: "default"` (last 120 minutes from the agent's wall clock, `confidence: 0.0`).

This fails criterion: for incidents 2–4 hours old, the window must not be **smaller** than the actual incident window.

## Root cause

The fixture nests the real alert under an `"alert"` key (`_meta`, `alert`, `evidence` siblings). The intake stage passes the **whole file** to `resolve_incident_window()`, but anchor parsers only look for `alerts` / `startsAt` at the **top level** — not inside the nested `"alert"` envelope.

E2E tests pass `fixture["alert"]` directly, so they mask this bug.

## Reproduction

```bash
uv run python -c "
import json, logging
logging.basicConfig(level=logging.INFO)
from surfaces.cli.investigation.payload import load_file
from tools.investigation.state_factory import make_initial_state
from core.domain.types.incident_window import resolve_incident_window
state = make_initial_state(raw_alert=load_file('tests/e2e/kubernetes/fixtures/datadog_k8s_alert.json'))
print(json.dumps(resolve_incident_window(state['raw_alert']).to_dict(), indent=2))"
```

**CLI path (broken):** `"source": "default", "confidence": 0.0`

**Inner alert path (works):** `"source": "alert.startsAt", "confidence": 1.0`

## Aged-incident scenario (3h old)

When the incident occurred ~3 hours ago:

- **Inner alert path:** window covers log timestamps ✅
- **CLI path:** window is `now - 120min → now`, does **not** cover 3h-old logs ❌

## Impact

- Documented quickstart/README path uses `-i .../datadog_k8s_alert.json`
- Datadog/Grafana tool queries use a window that misses stale telemetry
- RCA may still read alert text but cannot corroborate with old logs/metrics

## Suggested fix

Unwrap `payload["alert"]` (when present) before `resolve_incident_window()` in intake, or teach `resolve_incident_window` / `_coerce_alert_dict` to probe nested `"alert"` envelopes.

## Files

- `tests/e2e/kubernetes/fixtures/datadog_k8s_alert.json`
- `tools/investigation/stages/intake/node.py`
- `core/domain/types/incident_window.py`
- `surfaces/cli/investigation/payload.py`

## Not a PR regression

Reproduces on `main` and `refactor/core-domain` (PR #3776).

larsspinetta12:
## Summary

Fixes #3813.

E2E fixtures such as `tests/e2e/kubernetes/fixtures/datadog_k8s_alert.json` nest the real webhook payload under an `"alert"` key with sibling `"evidence"` / `"_meta"` fields. `resolve_incident_window()` only probed top-level anchor fields, so the documented CLI path:

```bash
opensre investigate -i tests/e2e/kubernetes/fixtures/datadog_k8s_alert.json
```

fell back to `source: "default"` (last 120 minutes from wall clock) instead of anchoring on `alert.startsAt`.

This change unwraps the nested `"alert"` envelope when the outer dict has no anchor but the inner object does.

## Test plan

- [x] `uv run pytest tests/core/domain/types/test_incident_window_extract_wiring.py`
- [x] `make lint`
- [ ] `make typecheck` (unrelated pre-existing failure in `surfaces/interactive_shell/utils/error_handling/errors.py`)

## Verification

Before: `source: "default", confidence: 0.0` for whole fixture file.

After: `source: "alert.startsAt", confidence: 1.0` with `until` anchored on fixture `startsAt` + forward buffer.


larsspinetta12:
@greptile review again

Davidson3556:
looks good to me @larsspinetta12

PR #3820, opened 2026-07-07
fix(k8s): wait for transform job before Datadog verify (step 2)

larsspinetta12:
## Summary
- Wait for `etl-transform-error` to fail on EKS before polling Datadog (fixes race after API Gateway 504)
- Configure kubectl via `update_kubeconfig()` inside verify when `--wait-for-transform-job` is set
- Add 30s Datadog agent flush delay after job failure (matches `test_eks.py`)
- Anchor DD search on `trigger_epoch` (start) instead of `trigger_completed_epoch`
- Workflow verify step uses `--wait-for-transform-job` (~8 min budget)

Fixes the live E2E gap for Vaibhav's step 2 (trigger → logs in Datadog).

## Test plan
- [x] `pytest tests/e2e/kubernetes/test_trigger_alert_verify.py`
- [x] `make lint`
- [ ] Re-run `Trigger K8s Alert` workflow on this branch

### DIFF
--- a/.github/workflows/trigger-k8s-alert.yml
+++ b/.github/workflows/trigger-k8s-alert.yml
@@ -17,6 +17,7 @@ jobs:
   trigger-alert:
     if: github.repository == 'Tracer-Cloud/opensre'
     runs-on: ubuntu-latest
+    timeout-minutes: 20
     env:
       AWS_REGION: us-east-1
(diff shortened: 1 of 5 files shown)


VaibhavUpreti:
Closing this @larsspinetta12 - this was due to the account issue with DD, should be resolved now.

PR #3836, opened 2026-07-08
Fix: enable WAL mode for SQLite gateway DB to prevent concurrent write deadlocks

### LINKED ISSUE #3835: [BUG] SQLite gateway DB missing WAL mode — concurrent turns cause "database is locked" errors
### Summary

`connect_gateway_db` opens the SQLite connection without enabling WAL journal mode. With `max_concurrent_turns > 1`, multiple `ThreadPoolExecutor` workers can call `SessionBindingStore.bind()` simultaneously, causing `sqlite3.OperationalError: database is locked` under concurrent load.

**Expected:** Concurrent gateway turns from different users write their session bindings without contention. SQLite WAL mode allows concurrent readers and serializes only writers, which is the correct mode for a multi-threaded gateway process.

**Actual:** With the default SQLite journal mode (`DELETE`), any two threads attempting to write to `gateway_session_bindings` simultaneously will deadlock. One thread holds a table-level write lock while the other spins and eventually raises:
sqlite3.OperationalError: database is locked
This is silent in production — the `_run_telegram_gateway_thread` catches all exceptions with `logger.critical(...)` and exits the polling loop, dropping all in-flight messages.

### Steps to reproduce

1. Configure the Telegram gateway with `TELEGRAM_GATEWAY_MAX_CONCURRENT=4` (or any value > 1).
2. Have two or more users send messages to the bot simultaneously for the first time (so both trigger `SessionBindingStore.bind()` to create new bindings).
3. Observe `sqlite3.OperationalError: database is locked` in gateway logs, or observe the gateway thread silently dying.
Can also be reproduced in a unit test:
```python
import threading
from gateway.storage import connect_gateway_db, SessionBindingStore

conn = connect_gateway_db(tmp_path / "state.db")
store = SessionBindingStore(conn)

errors = []
def write(user_id):
    try:
        store.bind(platform="telegram", chat_id=user_id, session_id="s1")
    except Exception as e:
        errors.append(e)

threads = [threading.Thread(target=write, args=(str(i),)) for i in range(10)]
for t in threads: t.start()
for t in threads: t.join()
assert not errors  # fails without WAL mode
```

Sometimes

### How often does it occur?

Under specific conditions

### Operating system

Windows

### Logs and error output

[CRITICAL] Fatal error in Telegram gateway thread
  File ".../gateway/polling/telegram_gateway_background.py", line 94, in _run_telegram_gateway_thread
    asyncio.run(_poll_telegram_until_stopped(...))
  File ".../gateway/storage/session/bindings.py", line 28, in bind
    self._conn.execute(...)
sqlite3.OperationalError: database is locked

### Additional context

Root cause in `gateway/storage/db.py`:

```python
def connect_gateway_db(path: Path | None = None) -> sqlite3.Connection:
    db_path = path or _DEFAULT_DB_PATH
    db_path.parent.mkdir(parents=True, exist_ok=True)
    conn = sqlite3.connect(db_path, check_same_thread=False)  # ← no WAL
    conn.row_factory = sqlite3.Row
    conn.executescript(_SCHEMA)
    conn.commit()
    return conn
```

`check_same_thread=False` is set correctly for multi-threaded use, but the default journal mode (`DELETE`) still takes exclusive table locks on every write. The fix is a one-liner:

```python
conn.execute("PRAGMA journal_mode=WAL")
```

Added immediately after `sqlite3.connect(...)`. WAL mode allows concurrent reads alongside a single writer and is the standard recommendation for multi-threaded SQLite.

A test covering concurrent writes to `SessionBindingStore` should accompany the fix — none currently exists in `gateway/tests/test_bindings.py`.

bastitva0-blip:
Fixes #3835

Enabled WAL journal mode for the SQLite gateway database by adding
`conn.execute("PRAGMA journal_mode=WAL")` immediately after `sqlite3.connect()` 
in `connect_gateway_db`.

Without WAL mode, the default `DELETE` journal takes exclusive table locks on 
every write. With `max_concurrent_turns > 1`, multiple `ThreadPoolExecutor` 
workers calling `SessionBindingStore.bind()` simultaneously deadlock with 
`sqlite3.OperationalError: database is locked`, silently killing the gateway thread.

WAL mode allows concurrent reads alongside a single serialized writer — the 
standard recommendation for multi-threaded SQLite.

Added `test_concurrent_writes_do_not_deadlock` to `gateway/tests/test_bindings.py` 
— 10 threads writing simultaneously, all passing with the fix applied.
Capture2
Capture1
AI Usage: Yes, I used AI assistance

- [y ] I have reviewed every single line of the AI-generated code

- [y ] -I can explain the purpose and logic of each function/component I added

- [ y] I have tested edge cases and understand how the code handles them

- [ y] I have modified the AI output to follow this project's coding standards and conventions
implementation approach::
The root cause is that SQLite's default journal mode (DELETE) serializes all
writes with an exclusive lock. The fix is a single PRAGMA that switches to WAL
mode, which uses a write-ahead log to allow concurrent readers and serialize
only writers — exactly what a multi-threaded gateway needs. The test spawns 10
threads simultaneously calling bind() to confirm no deadlock occurs.
i affirm yes to all boxes namely:
I have added proper PR title and linked to the issue
I have performed a self-review of my code
I can explain the purpose of every function, class, and logic block I added
I understand why my changes work and have tested them thoroughly
I have considered potential edge cases and how my code handles them
If it is a core feature, I have added thorough tests
My code follows the project's style guidelines and conventions

### DIFF
--- a/gateway/storage/db.py
+++ b/gateway/storage/db.py
@@ -25,7 +25,9 @@ def connect_gateway_db(path: Path | None = None) -> sqlite3.Connection:
     db_path = path or _DEFAULT_DB_PATH
     db_path.parent.mkdir(parents=True, exist_ok=True)
     conn = sqlite3.connect(db_path, check_same_thread=False)
-    conn.execute("PRAGMA journal_mode=WAL")
+    row = conn.execute("PRAGMA journal_mode=WAL").fetchone()
+    if row is None or row[0].lower() != "wal":
+        raise RuntimeError(f"Failed to enable WAL journal mode (got {row[0] if row else None!r})")
     conn.row_factory = sqlite3.Row
     conn.executescript(_SCHEMA)
     conn.commit()
--- a/gateway/storage/db.py
+++ b/gateway/storage/db.py
@@ -25,6 +25,7 @@ def connect_gateway_db(path: Path | None = None) -> sqlite3.Connection:
     db_path = path or _DEFAULT_DB_PATH
     db_path.parent.mkdir(parents=True, exist_ok=True)
     conn = sqlite3.connect(db_path, check_same_thread=False)
+    conn.execute("PRAGMA journal_mode=WAL")
     conn.row_factory = sqlite3.Row
     conn.executescript(_SCHEMA)
     conn.commit()
--- a/gateway/tests/test_bindings.py
+++ b/gateway/tests/test_bindings.py
@@ -24,6 +24,8 @@ def test_rotate_assigns_new_session(binding_store: SessionBindingStore) -> None:
     new_id = binding_store.rotate(platform="telegram", chat_id="123")
     assert new_id != "uuid-1"
     assert binding_store.get_session_id(platform="telegram", chat_id="123") == new_id
+
+
 def test_concurrent_writes_do_not_deadlock(tmp_path) -> None:
     import threading
     from gateway.storage import connect_gateway_db, SessionBindingStore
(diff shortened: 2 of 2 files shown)


bastitva0-blip:
@greptile review

PR #3858, opened 2026-07-08
fix(agent): hand off instead of fabricating a delivery command when send tool is unavailable

### LINKED ISSUE #3857: [BUG] Action planner fabricates nonexistent `/messaging send slack` command when slack_send_message is unavailable
## Summary

When a user asks to send data to Slack in a compound/data-dependent turn but the `slack_send_message` tool is **not available** (Slack not configured), the interactive-shell action planner does not hand off — instead it fabricates a nonexistent CLI/slash command (`/messaging send slack ...`), which fails with `No such command 'send'`.

**Expected:** With Slack unconfigured and no delivery tool available, after the data-dependent lookup succeeds, the planner should emit `assistant_handoff` explaining Slack isn't configured (and suggest `/integrations setup slack` / `SLACK_WEBHOOK_URL`). It must never invent a slash/CLI subcommand that does not exist.

**Actual:** The lookup turn runs correctly, but the delivery turn invokes `/messaging send slack <text>`, which is not a real subcommand:

```
Usage: opensre messaging [OPTIONS] [COMMAND] [ARGS]...
Error: No such command 'send'.
```

## Steps to reproduce

1. Run `uv run opensre` in a session where **Slack is not configured** (no `slack` integration, `SLACK_WEBHOOK_URL` unset). Only GitHub configured in the repro.
2. Enter: `check the weather in Antarctica and then send it to slack`
3. Turn 1 correctly runs `shell_run` → `curl 'wttr.in/Antarctica?format=3'` → `Antarctica: -26°C`.
4. Turn 2 emits `/messaging send slack Antarctica: -26°C` → fails with `No such command 'send'`.

## Root cause

`slack_send_message` is availability-gated on a configured webhook:

```python
# tools/slack_send_message_tool/tool.py
def is_available(self, sources):
    slack = sources.get("slack") or {}
    configured_webhook = str(slack.get("webhook_url") or "").strip()
    env_webhook = os.getenv("SLACK_WEBHOOK_URL", "").strip()
    return bool(configured_webhook or env_webhook)
```

With Slack unconfigured the tool is not offered to the planner. The action prompt (`core/agent_harness/prompts/action_agent_system_prompt.py`) has no rule for "requested delivery tool is unavailable", so the model improvises a slash/CLI command. `/messaging` only supports `pair, allow, revoke, status` (see `surfaces/interactive_shell/command_registry/slash_catalog.py`), so `send` does not exist.

Note: the data-dependent compound-turn behavior itself works as designed (lookup first, deliver next turn) — the defect is the fabricated-command fallback when the delivery tool is missing.

## Proposed fix

- Add a guardrail to the action system prompt: if the user asks to send/post/notify a channel whose send tool is **not** in the available tool list, emit `assistant_handoff` explaining it isn't configured — never fabricate a slash/CLI subcommand for delivery.
- (Optional) Add a live turn scenario covering "lookup + send to <channel> with the delivery tool unavailable" so this fallback is regression-tested. There is currently no live-LLM scenario for the weather→slack chain.

Yes (with Slack unconfigured).

## How often does it occur?

Under specific conditions (delivery target requested but its tool is unavailable).

## Operating system

macOS

## Additional context

Discovered while testing the Slack data-dependent compound turn added in #3350. Model: gpt-5.4 / openai.

Gust-svg:
## Summary

Fixes #3857. When a user asks to send/post/notify a channel (Slack, Telegram, …) whose send tool is **not available** (channel unconfigured), the interactive-shell action planner used to fabricate a nonexistent slash/CLI subcommand — e.g. `/messaging send slack …` — which fails with `No such command 'send'`. This adds a guardrail so the planner responds with a real, non-erroring action instead.

- **Guardrail added** to `action_agent_system_prompt.py`: if the matching send tool (`slack_send_message`, `telegram_send_message`, …) is not in the available tools, the channel is not configured. The planner must NOT invent/guess a delivery subcommand (`/messaging send slack …` is not real) and must NOT substitute a different channel. Instead it does ONE of: emit `assistant_handoff` (reporting any looked-up value + that the channel isn't configured), OR route the user to enable it via `slash_invoke(command="/integrations", args=["setup", "<service>"])`. Covers the mid-chain case (data-dependent lookup already ran).
- **De-dup:** the `slack_send_message` entry's data-dependent guidance now points at rule (2) of the COMPOUND TURN RULE box instead of restating the whole mechanism (kept the unique single-channel guidance).
- **Test:** `test_system_prompt_hands_off_when_delivery_tool_unavailable` pins the guardrail wording.
- **Snapshot:** regenerated `prompt_characterization_snapshot.json` for the intentional prompt-text change.

## Root cause

`slack_send_message.is_available()` gates on a configured webhook, so with Slack unconfigured the tool isn't offered to the planner. The prompt had no rule for "requested delivery tool is missing", so the model improvised a command. `/messaging` only supports `pair, allow, revoke, status`. The data-dependent compound-turn behavior itself was correct (lookup first, deliver next turn) — only the fallback was wrong.

## Design note: setup routing is allowed, not just handoff

Live planning checks showed the model resists a pure `assistant_handoff` and prefers routing to the real `/integrations setup <service>` command when a channel is unconfigured. That is a valid, helpful, non-erroring response (it enables the channel), so the guardrail permits **either** a handoff **or** the real setup command — while still forbidding fabricated/nonexistent delivery subcommands.

## Repro (before)

```
❯ check the weather in Antarctica and then send it to slack
$ curl 'wttr.in/Antarctica?format=3'      → Antarctica: -26°C   (turn 1, correct)
$ /messaging send slack Antarctica: -26°C → Error: No such command 'send'  (turn 2, bug)
```

## Test plan

- [x] `uv run python -m pytest tests/core/agent/orchestration/test_action_prompt.py` (15 passed)
- [x] `uv run python -m pytest tests/core/agent/prompts/test_prompt_characterization.py` (snapshot regenerated, passes)
- [x] `ruff check` + `ruff format --check` on changed files
- [x] Live planning check (gpt-5.4, 2 runs each) with Slack **unconfigured** → `shell_run` + `/integrations setup slack`; no fabricated command
- [x] Live planning check with Slack **configured** → `shell_run` + `slack_send_message` carrying the looked-up value (`Antarctica: -26C`)
- [ ] Real end-to-end Slack delivery (requires a live webhook; planner selection verified, actual HTTP POST not yet exercised)
- [ ] (Follow-up) add a live turn scenario covering the weather→slack chain (configured + unconfigured); none exists today

> Note: the live checks above used an inert probe tool, so they validate planner tool selection and argument population, not real message delivery.


muddlebee:
@Gust-svg we would need proper demo for this? could you please add relevant screenshots like before/after scenarios? how your changes are more effective than before?

PR #3859, opened 2026-07-08
fix(cloudtrail): harden LookupEvents live payload parsing so events reach the agent (#3583)

### LINKED ISSUE #3583: Harden tool live payload parsing
| Field | Value |
|-------|-------|
| **Level** | Middle |
| **Complexity** | M |
| **Tier** | Tier 3 — Middle |
| **Category** | integrations |

**Summary:** Fixture test for real API response shapes (TOOL_INTEGRATION_CHECKLIST §1).

**Steps:** (1) Pick tool under `integrations/<vendor>/tools/`. (2) Add JSON fixture. (3) Test partial/weird payloads. (4) Fix parser only if test proves bug.

**Do not:** Touch `core/agent_harness/` or `tools/interactive_shell/`.

---

amsorrytola:
Fixes #3583

**Problem.** `lookup_cloudtrail_events` (the CloudTrail "who changed what, and when?" forensics tool) parsed the live `LookupEvents` response optimistically. Two classes of real-world payload broke it:

1. **Degenerate event shapes crashed the parser.** `_shape_event` assumed every `Resources[]` entry was a dict and that `CloudTrailEvent` (a JSON string) always decoded to an object. A `null` / bare-string / list `CloudTrailEvent`, invalid JSON, or the transport sanitizer's list-truncation marker (`"... (N more items truncated)"`) landing inside `Resources` raised `AttributeError` and failed the whole tool call.
2. **The success payload was silently discarded inside investigations.** The success return carried `"error": None`. The runtime tool loop (`core.execution._normalize_result`) treats the mere *presence* of an `"error"` key as a failure (`is_error = "error" in raw`) and replaces the payload with `{"error": ...}` before the agent sees it. Result: **every** investigation that called CloudTrail received zero events — even though the tool returned 50 when called standalone.

**Fix (minimal, test-proven).**
- Harden `_shape_event`: skip non-dict `Resources` entries, coerce a non-dict decoded `CloudTrailEvent` to `{}`, and normalise CloudTrail's string `ReadOnly` (`"true"` / `"True"`) into a real `bool`.
- Surface truncation instead of hiding it: each event now carries **`resources_truncated`**, and only dict events are shaped (`events = [_shape_event(e) for e in raw_events if isinstance(e, dict)]`) so a degraded page degrades rather than crashes.
- Remove `"error": None` from the success return so events actually reach the agent; the failure paths still set a real `"error"` message.
- Fixtures + tests: a faithful post-sanitizer `LookupEvents` page (`lookup_events_page.json`) and a degenerate-payload page (`lookup_events_weird.json`), with tests asserting exact event shaping, pagination surfacing (`truncated` / `next_token`), graceful survival of weird payloads, and a regression that a success payload never carries an `"error"` key.
- Docs: document the new `resources_truncated` signal on the CloudTrail page.

Scope is exactly what the issue asks: only `integrations/cloudtrail/tools/…`, its fixtures/tests, and its doc page — **no** changes to `core/agent_harness/` or `tools/interactive_shell/`.

**Validation.** `make lint`, `make format-check`, `make typecheck` all clean; `tests/tools/test_cloudtrail_events.py` + `tests/integrations/` → **2359 passed, 2 skipped**. Verified end to end against a live AWS account: `opensre investigate` now calls CloudTrail and produces an evidence-backed RCA (see below).

Real run in a real terminal against a live AWS account (no cache/mocks) — `opensre integrations verify aws` followed by a full `opensre investigate` that calls CloudTrail and lands an evidence-backed root cause.

**▶️ Full run (GIF):**

cloudtrail-demo

**1️⃣ AWS integration verified — `opensre integrations verify aws`**

1-verify-aws

**2️⃣ Investigation launched with the full CLI command**

2-investigate-command

**3️⃣ ReAct loop reviewing evidence after the CloudTrail lookup**

3-cloudtrail-evidence

**4️⃣ Root-cause diagnosis running (validity 100%)**

4-diagnose-root-cause

**5️⃣ Findings → Recommended Actions → Cited Evidence (`lookup cloudtrail events`) → Root cause**

5-findings-rca

**6️⃣ Final RCA + feedback saved**

6-feedback-saved

---

The bug surfaced while building the demo: the CloudTrail tool returned 50 events on its own but every investigation reported "no events / tool broken." Tracing the runtime tool loop showed `core.execution._normalize_result` flags a result as an error whenever an `"error"` key is *present*, regardless of value — so the success payload's `"error": None` was masking the entire event list. The smallest correct fix is to stop emitting that key on success (failure paths keep a real message). The parser hardening is driven by the actual `LookupEvents` shape: `ReadOnly` comes back as a string, `CloudTrailEvent` is a JSON string that can decode to a non-object under degraded conditions, and the transport sanitizer can drop a truncation-marker string into an oversized `Resources` list — each is covered by a fixture and a test that fails before the fix. I preferred `isinstance` guards + an explicit `resources_truncated` flag over broad `try/except` so the tool degrades transparently and the planner is told when a resource list understates blast radius.

---

---

### DIFF
--- a/docs/cloudtrail_events.mdx
+++ b/docs/cloudtrail_events.mdx
@@ -42,6 +42,10 @@ CloudTrail's `LookupEvents` API accepts **only one filter attribute per call**. 
 CloudTrail returns at most 50 events per page. When more matching events exist, the response sets **`truncated: true`** and returns a **`next_token`** — pass it back via the `next_token` parameter to fetch the next page, so a busy account or wide window never silently drops events.
 </Note>
 
+<Note>
+A single event can touch more resources than the transport layer returns inline (for example `CreateTags` across a fleet). When the affected-resources list for an event is trimmed, that event carries **`resources_truncated: true`**, signalling that the real blast radius is wider than the resources shown — so a large fan-out change is never silently understated.
+</Note>
+
 ### Use cases
 
 - Finding who modified an IAM policy, role, or security group just before an incident
--- a/integrations/cloudtrail/tools/cloudtrail_events_tool/__init__.py
+++ b/integrations/cloudtrail/tools/cloudtrail_events_tool/__init__.py
@@ -58,12 +58,17 @@ def _build_lookup_attribute(
 
 def _shape_event(raw: dict[str, Any]) -> dict[str, Any]:
     """Trim a raw CloudTrail event down to the fields RCA actually needs."""
+    # Skip non-dict Resources entries: the transport sanitizer replaces the tail
+    # of an oversized list with a "... (N more items truncated)" string (one
+    # event CAN reference >MAX_LIST_ITEMS resources, e.g. CreateTags across a
+    # fleet), and degraded payloads may carry nulls.
     resources = [
         {
             "type": resource.get("ResourceType"),
             "name": resource.get("ResourceName"),
         }
         for resource in (raw.get("Resources") or [])
+        if isinstance(resource, dict)
     ]
 
     # CloudTrailEvent is a JSON string carrying the full record; pull the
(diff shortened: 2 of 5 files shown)


Davidson3556:
LGTM. The parser hardening is clean, and the `error: None` fix is correct: I traced `_normalize_result` and confirmed its presence check (`"error" in raw`) swallows the whole success payload. Fixtures and tests are thorough. Merging.

Filed #3889 to fix the same bug at the root in `_normalize_result` and check the other tools (EKS/RDS/EC2/groundcover) that also return `error: None` on success.

PR #3868, opened 2026-07-08
fix(k8s): create etl-pipeline-sa in test_datadog before running jobs

### LINKED ISSUE #3866: fix(k8s): create etl-pipeline-sa in test_datadog before running jobs
## Problem

`make test-k8s-datadog` (`tests/e2e/kubernetes/test_datadog.py`) times out waiting for `etl-extract` because the job manifests require `serviceAccountName: etl-pipeline-sa`, but `test_datadog` never creates that service account.

`test_local.py` creates it in `setup_cluster()`; `test_datadog` was missing the equivalent step, so the extract job pod never completes.

## Expected

Extract and transform-error jobs schedule and finish on the local kind cluster.

## Repro

```bash
export DD_API_KEY=... DD_APP_KEY=... AWS_*=...
make test-k8s-datadog
```

First run failed with:
```
TimeoutError: Job 'etl-extract' did not finish within 120s
```

larsspinetta12:
## Summary
- Create `etl-pipeline-sa` in `test_datadog` before deploying jobs, matching `test_local.setup_cluster()`.
- Fixes `make test-k8s-datadog` timing out on `etl-extract` because the job pod never schedules without the service account.

Fixes #3866

## Test plan
- [x] `make test-k8s-datadog` completes extract → transform-error and finds `PIPELINE_ERROR` in Datadog
- [x] Local run: `TEST PASSED` after this change

### DIFF
--- a/tests/e2e/kubernetes/test_datadog.py
+++ b/tests/e2e/kubernetes/test_datadog.py
@@ -31,6 +31,7 @@ import urllib.request
 import uuid
 
 from tests.e2e.kubernetes.infrastructure_sdk.local import (
+    _run,
     apply_manifest,
     build_image,
     check_prerequisites,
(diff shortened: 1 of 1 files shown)


larsspinetta12:
@greptile review again

PR #3886, opened 2026-07-09
feat(observability): session-trace span port for stage, tool, and LLM metrics, observability package refactoring

YauhenBichel:
Fixes #3822 

#### Describe the changes you have made in this PR -

- Observability package refactoring
- Adds a session-trace span layer so a run's stages, tool calls, and LLM invokes are recorded with timing and outcome, and regroups the observability package into `render/`, `trace/`, and `errors/`.

Spans go through one `SessionTraceSink` Protocol. The default is a Noop sink, so every emit path returns after a single `isinstance` check — no clock, no sampling, no I/O in production. A JSONL adapter is wired only in the REPL. Core and community tool code never instrument anything: tool spans are emitted once at the universal choke point in `core.execution.execute_tool_calls`, so adding a new tool or integration needs no tracing code.

Call sites use readable semantic helpers — `component_span`, `stage_span`, `tool_span`, `llm_span`, `emit_route`, `traced_session` — instead of raw span plumbing.

`platform/observability/` flat files → `render/` (progress, debug, display, output_format, figlet), `trace/` (spans, hook, process_stats, redaction, prompts), `errors/` (boundary, sentry, service), plus `otlp_parser.py`. No forwarding shims left behind; all callers repointed to the canonical paths.

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [x] I have reviewed every single line of the AI-generated code
- [x] I can explain the purpose and logic of each function/component I added
- [x] I have tested edge cases and understand how the code handles them
- [x] I have modified the AI output to follow this project's coding standards and conventions

---

## Checklist before requesting a review
- [x] I have added proper PR title and linked to the issue
- [x] I have performed a self-review of my code
- [x] **I can explain the purpose of every function, class, and logic block I added**
- [x] I understand why my changes work and have tested them thoroughly
- [x] I have considered potential edge cases and how my code handles them
- [x] If it is a core feature, I have added thorough tests
- [x] My code follows the project's style guidelines and conventions

---

Note: Please check **Allow edits from maintainers** if you would like us to assist in the PR.

PR #3891, opened 2026-07-09
fix(observability): guard POSIX-only resource import for Windows

cloudenochcsis:

`platform/observability/trace/process_stats.py` (added in the #3886 observability span refactor) imports the **POSIX-only `resource` module at module scope** for a single `getrusage` RSS reading.

That module is not import-isolated: `platform/observability/trace/spans.py` imports `sample_turn_boundary_stats` from it at module scope, and `spans.py` is imported eagerly by `core/execution.py`, `core/agent/react_loop.py`, the `core/agent_harness/turns/*` drivers, and `gateway/turn_handler.py`. On Windows — a platform documented as supported in `SETUP.md` — this raises `ModuleNotFoundError: No module named 'resource'` **at import time**, crashing the CLI, REPL, and gateway before they can start.

This PR guards the import and skips only the RSS reading when `resource` is unavailable: `sample_resource_snapshot()` returns `rss_mb=None` on such platforms while the cross-platform GC generation counts are unaffected. POSIX behavior (and the macOS-vs-Linux `ru_maxrss` unit normalization) is unchanged.

`psutil` (used for process stats elsewhere) is intentionally **not** used here, to keep it confined to the fleet-monitoring probe module as enforced by `tests/fleet_monitoring/test_probe.py::test_psutil_is_not_imported_outside_probe_module`.

Simulating a platform without the `resource` module (`import resource` forced to raise, as on Windows):

**Before (current `main`)** — crashes at import:
```
CRASH at import: No module named 'resource'
  File ".../process_stats.py", line 6, in <module>
    import resource
ModuleNotFoundError: No module named 'resource'
```

**After (this PR)** — imports cleanly, degrades gracefully:
```
import OK on simulated Windows (no `resource`)
snapshot: {'rss_mb': None, 'gc_gen0': 963, 'gc_gen1': 7, 'gc_gen2': 0}
```

POSIX unchanged (macOS): `{'rss_mb': 17.06, 'gc_gen0': 362, 'gc_gen1': 7, 'gc_gen2': 0}`

Local CI: `make lint`, `make format-check`, `make typecheck` clean; `make test-cov` passes (10,724 passed; the one unrelated failure is an environment-dependent Copilot CLI adapter test that also fails on unmodified `main` because the `copilot` binary is installed on this machine).

---

The problem: a POSIX-only stdlib import (`resource`) sits at module scope in a file that is transitively imported by every entry point, so any platform lacking `resource` (Windows) fails to start. The fix has to keep the module importable everywhere while preserving the existing RSS metric where it is available.

I considered two alternatives: (1) reading RSS via `psutil`, which is already a dependency and cross-platform — rejected because an import-boundary test deliberately confines `psutil` to `tools/system/fleet_monitoring/probe.py`, and I did not want to widen that surface; (2) moving `import resource` into the function body — this fixes the import-time crash but still raises on every call on Windows. I chose a guarded top-level import with graceful degradation: `rss_mb` becomes `None` when `resource` is absent, and the GC counts (which are cross-platform) are always returned. This keeps the metric intact on POSIX, never raises on Windows, and touches nothing outside the one file.

---

### DIFF
--- a/platform/observability/trace/process_stats.py
+++ b/platform/observability/trace/process_stats.py
@@ -3,10 +3,14 @@
 from __future__ import annotations
 
 import gc
-import resource
 import sys
 import threading
 from typing import Any
+
+try:
+    import resource
+except ImportError:  # ``resource`` is POSIX-only; absent on Windows.
+    resource = None  # type: ignore[assignment]
 
 #: Cap thread rows embedded in a turn-boundary snapshot (ATM thread map).
 _MAX_THREADS_IN_SNAPSHOT = 40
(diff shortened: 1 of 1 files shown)


YauhenBichel:
Hi @cloudenochcsis thank you. The fix will be in https://github.com/Tracer-Cloud/opensre/pull/3890

PR #3892, opened 2026-07-09
ci: add importlinter check to quality job

Devesh36:
Wires up the existing `.importlinter` config as a CI gate so import boundary violations are caught before merge.


Devesh36:
@greptile review

PR #3904, opened 2026-07-09
fix(session): stop trace spans from rescanning the session file

cloudenochcsis:

The session-trace span port (#3886 / #3890) wraps every LLM invoke (`llm_span`) and every tool execution (`tool_span`) and writes a `trace_span` record through `JsonlSessionStorage`. Those helpers pass no `parent_id`, so each append falls into `_append_entry`'s fallback and calls `_current_leaf_id`, which reads and JSON-parses the **entire session file** to find the current leaf.

Because a span fires once per LLM call and once per tool call, this makes turn latency scale with session length: **O(n) per span, O(n^2) over a session's life** — and it's on by default for every interactive-shell session (`runtime/context.py` registers a JSONL trace sink unconditionally).

`trace_span` rows are side-channel telemetry (ATM / debug); they do not belong in the conversation parent chain. This PR:

1. Skips the leaf lookup when appending a `trace_span` (`_append_entry`), so a span never triggers a file read.
2. Skips `trace_span` rows in `_current_leaf_id`, so a span can never become the parent of the next real entry.

Change (2) also fixes a **latent correctness issue**: because spans are emitted mid-turn — before the turn's message is appended to the same file — the next real message was chaining its `parent_id` onto a telemetry span instead of the previous message.

Appending 50 `trace_span` records (a normal turn with many LLM/tool calls), counting full session-file reads, and checking the parent of the message that follows:

**Before (current `main`):**
```
50 trace_span appends -> 50 full-file reads
assistant message parent type = 'trace_span'   (conversation chain polluted)
```

**After (this PR):**
```
50 trace_span appends -> 0 full-file reads
assistant message parent type = 'message'      (chains to the previous real entry)
```

Adds two regression tests in `tests/platform/observability/test_session_trace.py`:
- `test_trace_span_append_skips_leaf_scan` — a `trace_span` append performs zero file reads.
- `test_trace_span_never_becomes_conversation_parent` — a message appended after a span chains to the prior real entry.

Local CI: `make lint`, `make format-check`, `make typecheck` clean; `make test-cov` passes (10,662 passed; the one unrelated failure is an environment-dependent Copilot CLI adapter test that also fails on unmodified `main` because the `copilot` binary is installed on this machine).

---

The problem: `trace_span` records are written into the same append-only JSONL file as conversation entries, and every append without an explicit `parent_id` resolves the leaf by re-reading the whole file. Spans are the highest-frequency writer (per LLM call, per tool call), so they dominate this cost, and they also shouldn't be part of the conversation tree at all.

I considered caching the last-appended entry id on the storage instance to make leaf resolution O(1). I rejected that as the primary fix: the leaf rule has to skip `leaf`/`session` (and now `trace_span`) record types, so a naive "last id" cache would be subtly wrong, and it adds mutable state to a class that is otherwise stateless per call. The narrower, correct fix is to recognize that a `trace_span` is side-channel telemetry: it neither needs a parent nor should it be one. So I skip the leaf lookup for `trace_span` appends and exclude `trace_span` from `_current_leaf_id`. This removes the per-span file read entirely and keeps the conversation chain pointing only at real entries. The pre-existing per-message leaf resolution (far lower frequency) is intentionally left unchanged to keep the diff scoped to the regression.

---

### DIFF
--- a/core/agent_harness/session/persistence/jsonl_storage.py
+++ b/core/agent_harness/session/persistence/jsonl_storage.py
@@ -289,7 +289,14 @@ class JsonlSessionStorage:
             records = self._read_records(path)
             if not records:
                 return
-            if records[-1].get("type") == "leaf":
+            # Ignore trailing side-channel trace_span rows when checking for an
+            # existing leaf, so a span written after a leaf can't slip a second
+            # leaf past this guard (consistent with _current_leaf_id's skip).
+            last_meaningful = next(
+                (rec for rec in reversed(records) if rec.get("type") != "trace_span"),
+                None,
+            )
+            if last_meaningful is not None and last_meaningful.get("type") == "leaf":
                 return
             if not self._has_turns(records):
                 path.unlink(missing_ok=True)
--- a/core/agent_harness/session/persistence/jsonl_storage.py
+++ b/core/agent_harness/session/persistence/jsonl_storage.py
@@ -354,7 +354,17 @@ class JsonlSessionStorage:
             if not path.exists():
                 return ""
             entry_id = _new_id()
-            parent = parent_id if parent_id is not None else self._current_leaf_id(path)
+            # ``trace_span`` rows are side-channel telemetry (ATM / debug) and
+            # never join the conversation parent chain, so they skip the leaf
+            # lookup every other append pays. They are emitted once per LLM and
+            # per tool call, so resolving the leaf here would re-read the whole
+            # session file on every span -- O(n) per span, O(n^2) per session.
+            if parent_id is not None:
+                parent = parent_id
+            elif entry_type == "trace_span":
+                parent = None
+            else:
+                parent = self._current_leaf_id(path)
             record = {
                 "id": entry_id,
                 "parent_id": parent,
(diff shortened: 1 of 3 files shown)


cloudenochcsis:
@greptile review

cloudenochcsis:
@greptile review

cloudenochcsis:
@greptile review

cloudenochcsis:
@greptile review

cloudenochcsis:
Status note for reviewers: Greptile's 4/5 summary above was computed on the first commit (`d90057f7`) and hasn't refreshed despite re-triggers. Both points it raised are addressed in the two follow-up commits, and there are no unresolved review threads:

- **Non-idiomatic `str.find(...) != -1`** → replaced with a plain `in` membership check (`6831e53c`).
- **`flush()` leaf-guard blind to trailing `trace_span`** → `flush()` now skips trailing `trace_span` rows when checking for an existing leaf, consistent with the `_current_leaf_id` skip, with a regression test that fails without the change (`7ae63f35`).

CI is green on `7ae63f35` (`make lint`/`format-check`/`typecheck` clean; full unit suite passes). Happy to re-trigger Greptile if a maintainer prefers a fresh 5/5 pass before merge.

PR #3906, opened 2026-07-09
fix(interactive-shell): don't mask REPL errors in alert listener

cloudenochcsis:

`_alert_listener` (`surfaces/interactive_shell/controller.py`) is a `@contextmanager` that wrapped its `yield inbox` **inside the same `try/except` that guards listener startup**:

```python
try:
    ...
    with _alert_listener_token(...):
        handle = serve_webapp_in_thread(...)
        try:
            yield inbox        # body runs here
        finally:
            handle.stop(); set_current_inbox(None)
except Exception as exc:       # <-- also catches exceptions from the body
    log.warning("Alert listener could not start: ...", exc)
    set_current_inbox(None)
    yield None                 # <-- second yield
```

When the REPL body inside `with _alert_listener(...)` raises, `@contextmanager` throws that exception back in at `yield inbox`. It then propagates to the outer `except`, which (a) logs a misleading *"Alert listener could not start"* even though it started fine, and (b) runs a **second `yield`**. A second yield after an exception makes contextlib raise `RuntimeError("generator didn't stop after throw()")` — so the REPL body's real error is masked and replaced with a confusing one.

Fix: restructure with `contextlib.ExitStack`. Acquire the inbox, token, and web server under the stack and catch only **startup** failures (which still degrade gracefully to `yield None`). Once startup succeeds, `yield` outside that `except` with the stack held open, so:
- a body exception propagates unchanged, and
- cleanup (stop server, clear inbox, restore token env) still runs via the stack.

The token env var must stay set across the yield because the webapp reads `OPENSRE_ALERT_LISTENER_TOKEN` per request (`gateway/webapp.py:97`), so the stack is what keeps it open for the serving lifetime.

New regression tests in `tests/interactive_shell/test_alert_listener.py`:
- `test_alert_listener_propagates_body_exception_without_masking` — raising inside the `with` block propagates the original exception, and cleanup still runs (server stopped, inbox cleared, token restored).
- `test_alert_listener_degrades_to_none_when_startup_fails` — a `serve_webapp_in_thread` failure yields `None` without crashing.

Verified the first fails before this change and passes after:
```
# before the fix
FAILED ...test_alert_listener_propagates_body_exception_without_masking
# (the RuntimeError from the double-yield replaces the body's exception)

# after the fix
4 passed
```

Local CI: `make lint`, `make format-check`, `make typecheck` clean; `make test-cov` passes (10,662 passed; the one unrelated failure is an environment-dependent Copilot CLI adapter test that also fails on unmodified `main` because the `copilot` binary is installed on this machine).

---

The root cause is a `@contextmanager` control-flow error: the `yield` sat inside the `except` that was meant to catch only startup failures, so exceptions from the managed body were misrouted into the startup handler and triggered an illegal second yield.

The constraint that shapes the fix: the token env var has to remain set for the whole serving window (the webapp reads it per request), so I can't simply exit the token context after startup. I considered manually calling `__enter__`/`__exit__` on the token context manager to span the yield, but that is exactly the error-prone bookkeeping `ExitStack` exists to remove. `ExitStack` lets me register the token context and the server/inbox teardown during the guarded startup, catch only startup exceptions there, and then yield with the stack still open so the body runs with everything held and unwinds cleanly on exit (including on a body exception). Teardown is registered so the server is stopped before the inbox is cleared. The two tests pin both paths — the previously-broken body-exception path and the already-working startup-failure path — so the behavior can't regress in either direction.

---

### DIFF
--- a/surfaces/interactive_shell/controller.py
+++ b/surfaces/interactive_shell/controller.py
@@ -6,7 +6,7 @@ import asyncio
 import logging
 import os
 from collections.abc import Iterator
-from contextlib import contextmanager
+from contextlib import ExitStack, contextmanager
 
 from prompt_toolkit import PromptSession
 from prompt_toolkit.patch_stdout import patch_stdout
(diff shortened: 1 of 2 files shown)


cloudenochcsis:
@greptile review

cloudenochcsis:
@greptile review

Devesh36:
looks good to me tagging @muddlebee & @Davidson3556  for final call

muddlebee:
@cloudenochcsis openinig a github issue first with relevant context might be a better idea? also a request to please not open such PRs without discussing before, it jsut leads to more spam.

cloudenochcsis:
Understood, and apologies for the noise — thanks for the steer. I've switched to the issue-first flow and opened a tracking issue (with context and file:line evidence) for each of the three, rather than re-opening PRs:

- #3938 — trace_span O(n^2) session-file rescan + parent-chain pollution (was this PR's sibling #3904)
- #3939 — parallel tool calls emit no trace spans (was #3905)
- #3940 — alert-listener masks REPL errors (this one, #3906)

Happy to be assigned any you'd like fixed and I'll open a linked PR then. Will hold off on PRs until then.

PR #3926, opened 2026-07-10
fix(gateway): restore POST /investigate REST API endpoint

### LINKED ISSUE #3842: [FEAT] Restore /investigate REST API endpoint
## Summary

The `/investigate` REST API endpoint was removed and needs to be added back.

## Context

The REST API endpoint for triggering investigations (equivalent to `opensre investigate`) was removed during a recent refactoring. It needs to be restored so that external systems (e.g. Telegram, Slack webhooks, CI pipelines) can trigger investigations programmatically via HTTP.

## Acceptance Criteria

- [ ] The `/investigate` REST API endpoint is restored and functional
- [ ] The endpoint accepts the same parameters as before (alert source, alert data, etc.)
- [ ] Integration tests cover the endpoint
- [ ] Telegram, CLI, and shell still work after the change (per channel AC)

## Notes

- Flagging @w3joe for visibility — please comment if you have context on where/why the endpoint was removed.

muddlebee:
Fixes #3842

The `/investigate` REST endpoint was lost when `infra/deployment/remote/server.py` — a standalone FastAPI app that also bundled unrelated Discord bot, Vercel deployment poller, and EC2 IMDS metadata code — was deleted in a prior "infra folder cleanup" commit.

This PR restores just the `/investigate` endpoint, not the legacy infrastructure it used to be bundled with:

- Adds `POST /investigate` to `gateway/webapp.py`, the app's single current FastAPI surface (alongside the existing `/health`, `/ok`, `/healthz`, `POST /alerts`). It runs an investigation synchronously and returns the RCA report.
- Reuses `tools.investigation.capability.run_investigation_payload` — the same headless runner the CLI already calls — so pipeline behavior is identical across surfaces, and the change respects the `gateway`/`surfaces` import-boundary rule (gateway must not import from `surfaces`, which is how the old code was previously wired and why it couldn't just be copy-pasted back).
- Accepts the same parameters as before: `raw_alert`, `alert_name`, `pipeline_name`, `severity`.
- Reuses the existing gateway auth boundary (loopback-only by default, or `OPENSRE_ALERT_LISTENER_TOKEN` bearer token for remote callers) — the same trust boundary `POST /alerts` already uses. Generalized the internal auth helper (`_alert_auth_error` → `_gateway_auth_error`) since it now guards two routes.
- Fixes `docs/deployment.mdx`, which already documented `/investigate` but with a stale `--host/--port/--api-key`/`X-API-Key` CLI contract that doesn't exist in the current `opensre gateway start` command — updated to match the real CLI and the actual bearer-token auth.
- **Follow-up commit addressing review feedback:** guarded `InvestigateResponse(**result)` construction inside the same `try/except` as the pipeline call (Greptile: a malformed pipeline result could raise an unguarded Pydantic `ValidationError`), and stopped returning raw exception text to HTTP callers — the response now carries only the exception type name, with full detail going to logs/Sentry server-side only (CodeQL: information exposure through an exception).

Out of scope (not required by the issue's acceptance criteria, and would reintroduce dead/legacy functionality): the Discord slash-command integration, the Vercel deployment poller, EC2 IMDS instance metadata, deep health checks, and `.md` report persistence (`/investigations` list/get) that lived in the old `infra/deployment/remote/server.py`. None of that infrastructure exists in the current codebase anymore.

Started a real `uvicorn` gateway process via `gateway.web_server.serve_webapp_in_thread` (not just `TestClient` mocks) and hit it with real HTTP requests:

```
$ curl http://127.0.0.1:39045/health
{"ok":true,"version":"0.1","llm_configured":true,"env":"development"}

$ curl -X POST http://127.0.0.1:39045/investigate   # missing raw_alert
422 {"detail":[{"type":"missing","loc":["body","raw_alert"],"msg":"Field required","input":{"alert_name":"x"}}]}
```

Full test run:

```
$ uv run pytest gateway/tests -q
======================== 117 passed, 1 warning in 7.05s =========================
```

---

The problem: `/investigate` was removed as collateral damage of an infra cleanup that deleted a much bigger legacy file it happened to live in. The fix isn't "restore the old file" — most of that file (Discord bot, Vercel poller, EC2 IMDS) is dead code with no current callers, and re-adding it would violate the current Tier-1 import-boundary rules (`gateway` and `surfaces` must not import each other) that didn't exist when the old server was written.

Alternative considered: resurrect `infra/deployment/remote/server.py` wholesale. Rejected — it imports directly from `surfaces.cli` and `surfaces.interactive_shell`, which is now a forbidden cross-surface import, and it reintroduces several unrelated, currently-unused integrations (Discord, Vercel, IMDS) the issue never asked for.

Chosen approach: add `POST /investigate` directly to `gateway/webapp.py`, the app's one current FastAPI surface, and have it call `tools.investigation.capability.run_investigation_payload` — a Tier-2 helper already explicitly documented as "the headless counterpart used by surfaces that do not render a live terminal stream (HTTP server, MCP, integration webhooks)." This keeps the endpoint's behavior byte-for-byte consistent with what the CLI already produces, without adding a new code path to maintain.

Key pieces:
- `InvestigateRequest`/`InvestigateResponse` — Pydantic models mirroring the old endpoint's request shape and the payload `run_investigation_payload` already returns.
- `_gateway_auth_error` (renamed from `_alert_auth_error`) — the same loopback-or-bearer-token check `/alerts` already used, now shared by both mutating routes since they sit behind the same trust boundary.
- Error handling — after review feedback, both the pipeline call and the response-model construction are wrapped in one `try/except`, exceptions are logged and sent to Sentry server-side, and the client only sees the exception type name (not the message), to avoid leaking internal detail through an HTTP error body.

Edge cases tested: missing `raw_alert` (422), pipeline exceptions (503, with exception text confirmed absent from the response body), a malformed pipeline result that fails response validation (503, not an unguarded 500), metadata resolution falling back to `raw_alert` fields when overrides are omitted, and the full 401/403/200 auth matrix (loopback vs. remote, with/without token).

---

---

### DIFF
--- a/docs/deployment.mdx
+++ b/docs/deployment.mdx
@@ -70,22 +70,24 @@ Use the built-in gateway command when you want a local HTTP server for investiga
 
 | Command | What it does |
 | --- | --- |
-| `uv run opensre gateway` | Starts the local gateway on `127.0.0.1:2024`. |
-| `uv run opensre gateway --host 0.0.0.0 --port 8080` | Binds the gateway to a custom host/port. |
-| `uv run opensre gateway --api-key local-dev-key` | Enables API key auth for non-exempt routes. |
-| `uv run opensre gateway --investigations-dir ./investigations` | Writes investigation reports to a custom directory. |
+| `uv run opensre gateway start` | Starts the gateway daemon in the background (web app, Telegram chat, task scheduler). |
+| `uv run opensre gateway start --foreground` | Runs the gateway attached to this terminal. |
+| `uv run opensre gateway stop` | Stops the background daemon. |
+| `uv run opensre gateway status` | Shows daemon and component status. |
+
+The web app binds `0.0.0.0` on the port from the `PORT` environment variable
+(default `8000`).
 
 Quick local checks:
 
 ```bash
-curl http://127.0.0.1:2024/ok
-curl http://127.0.0.1:2024/version
+curl http://127.0.0.1:8000/ok
 ```
 
 Example investigation request:
 
 ```bash
-curl -X POST http://127.0.0.1:2024/investigate \
+curl -X POST http://127.0.0.1:8000/investigate \
   -H "Content-Type: application/json" \
   -d '{
     "raw_alert": {
@@ -97,12 +99,13 @@ curl -X POST http://127.0.0.1:2024/investigate \
   }'
 ```
 
-When `--api-key` is set, include it for protected endpoints:
+By default, `/investigate` (like `/alerts`) accepts only loopback callers. Set
+`OPENSRE_ALERT_LISTENER_TOKEN` to allow remote callers with a bearer token:
 
 ```bash
-curl -X POST http://127.0.0.1:2024/investigate \
+curl -X POST http://127.0.0.1:8000/investigate \
   -H "Content-Type: application/json" \
-  -H "X-API-Key: local-dev-key" \
+  -H "Authorization: Bearer $OPENSRE_ALERT_LISTENER_TOKEN" \
   -d '{"raw_alert":{"alert_name":"test"}}'
 ```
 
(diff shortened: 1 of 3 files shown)


muddlebee:
@greptile review

PR #3919, opened 2026-07-09
feat(analytics): populate investigation_loop_count on lifecycle events

larsspinetta12:
Fixes #3920

## Summary

Upstream PR 1: make `investigation_loop_count` a canonical, always-populated field on every investigation lifecycle PostHog event.

- **Source of truth:** `ConnectedInvestigationAgent` ReAct loop counter (`for iteration in range(MAX_INVESTIGATION_LOOPS)`), written to state as `investigation_loop_count` + `investigation_iteration_cap`
- **Convention:** `investigation_started` uses `0` (no LLM loops completed yet)
- **New helper:** `platform/analytics/investigation_loop.py` merges loop metrics into event properties
- **Events updated:** `investigation_started`, `investigation_completed`, `investigation_failed`, `investigation_outcome`, `investigation_cancelled`, `investigation_first_hypothesis_rendered`, `investigation_abandoned`
- **Call sites:** lifecycle/capability pipeline bind, REPL `track_investigation` + `record_loop_metrics_from_state`, outcome/cancelled analytics bridge, stream renderer

Separate from #3915 (assistant 3-tier UX) and the `react_turn_completed` work (#3917).

## Test plan

- [x] `make lint` and `make typecheck`
- [x] `uv run python -m pytest tests/analytics/test_cli.py tests/analytics/test_investigation_loop.py tests/interactive_shell/utils/telemetry/test_investigation_analytics.py`

PR #3934, opened 2026-07-10
fix(investigation): bind stream error-handler state before deferred imports

cloudenochcsis:

`astream_investigation`'s background `_run_pipeline` (`tools/investigation/capability.py`) resolves loop metrics as the **first line** of its `except Exception` handler:

```python
except Exception as exc:
    loop_count, iteration_cap = loop_metrics_from_state(state)
    _capture_exception_once(exc, ...)
    # ... queue InvestigationPipelineStreamError ...
```

But both `state` and the `from platform.analytics.investigation_loop import loop_metrics_from_state` binding lived **inside the `try`**, after a block of eight deferred stage imports:

```python
try:
    from core.state.updates import apply_state_updates
    from platform.analytics.investigation_loop import loop_metrics_from_state
    ...
    state = initial
    ...
```

If any of those deferred imports raises — precisely the failure mode deferred imports exist to isolate (a stage transitively importing a missing/misconfigured provider SDK) — then `state` and/or `loop_metrics_from_state` are unbound, and the `except` handler **raises `UnboundLocalError`/`NameError` itself**. That pre-empts `_capture_exception_once` (no Sentry capture) and the `InvestigationPipelineStreamError` queue delivery (no error surfaced), so only the `finally` `None` sentinel reaches the consumer — whose loop treats `None` as a **clean end**. Net effect: a streaming investigation that should surface a hard error instead ends silently with empty results, and the real exception never reaches Sentry.

Introduced by #3919 (the loop-metrics line was added to the handler); pre-that, the `except` referenced neither symbol and always delivered the exception.

Fix: bind `state = initial` and import `loop_metrics_from_state` **before** the `try`. The remaining stage imports stay deferred inside the `try`, where their failure is now handled correctly.

New regression test `test_astream_surfaces_error_when_deferred_import_fails` deletes `core.state.updates.apply_state_updates` to force a deferred `from ... import ...` failure, then asserts the consumer still receives a wrapped `InvestigationPipelineStreamError`. Verified it fails before the change and passes after:

```
# before the fix
FAILED ...::test_astream_surfaces_error_when_deferred_import_fails
# (async for completes with no exception -- the error was swallowed)

# after the fix
1 passed  (InvestigationPipelineStreamError raised, cause is ImportError)
```

Local CI: `make lint`, `make format-check`, `make typecheck` clean; `make test-cov` passes (10,780 passed).

---

The bug is an ordering error in exception handling: the handler depends on two names (`state`, `loop_metrics_from_state`) that are only bound partway through the `try` it guards, so the one failure class that reaches the handler *before* those bindings — an import error in the deferred stage imports — makes the handler crash and swallow the real error.

I considered guarding the handler defensively (e.g. `locals().get("state")` or a `try/except` around the metrics read), but that hides the ordering problem rather than fixing it and would still report zeroed metrics without making clear why. The direct fix is to hoist exactly the two symbols the handler needs above the `try`: `state = initial` is a pure local binding with no import cost, and `loop_metrics_from_state` comes from a light analytics helper module (no heavy SDK), so hoisting it doesn't reintroduce the import-cost problem the deferred imports were avoiding. The seven stage imports stay deferred inside the `try`, so their failure now routes through a handler that can actually run. The test forces the exact previously-fatal path (a deferred import raising) rather than a mid-pipeline exception, which the existing tests already cover.

---

### DIFF
--- a/tests/tools/investigation/test_streaming_loop_metrics.py
+++ b/tests/tools/investigation/test_streaming_loop_metrics.py
@@ -89,8 +89,14 @@ async def test_astream_surfaces_error_when_deferred_import_fails(
     # runtime (e.g. a stage transitively importing a missing/broken SDK).
     monkeypatch.delattr(_updates, "apply_state_updates")
 
+    # The `async for` only terminates once the pipeline delivers its None
+    # sentinel; if the handler raised or skipped the finally, this would hang.
     with pytest.raises(InvestigationPipelineStreamError) as exc_info:
         async for _event in astream_investigation("alert text"):
             pass
 
     assert isinstance(exc_info.value.cause, ImportError)
+    # Metrics could not be read (import failed before any stage ran), so the
+    # handler falls back to the 0/0 defaults rather than raising.
+    assert exc_info.value.loop_count == 0
+    assert exc_info.value.iteration_cap == 0
--- a/tests/tools/investigation/test_streaming_loop_metrics.py
+++ b/tests/tools/investigation/test_streaming_loop_metrics.py
@@ -73,3 +73,24 @@ def test_main_thread_bridge_binds_metrics_from_wrapped_stream_failure() -> None:
         assert bound_loop_metrics() == (4, 20)
     finally:
         reset_investigation_loop_metrics(scope_token)
+
+
+@pytest.mark.anyio
+async def test_astream_surfaces_error_when_deferred_import_fails(
+    monkeypatch: pytest.MonkeyPatch,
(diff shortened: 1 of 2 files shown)


cloudenochcsis:
@greptile review

cloudenochcsis:
@greptile review

cloudenochcsis:
Closing in favor of the issue-first flow — tracking as #3941 with full context. Happy to reopen a linked PR once that issue is triaged/assigned.

PR #3936, opened 2026-07-10
fix(gateway): authenticate and cap body before parsing POST /investigate

cloudenochcsis:

The restored `POST /investigate` (#3926) bound the request body as a Pydantic parameter:

```python
def investigate(req: InvestigateRequest, request: Request) -> ...:
    if (auth_error := _gateway_auth_error(request)) is not None:
        return auth_error
```

Because `req` is a body-bound parameter, Starlette **buffers and JSON-parses the entire body during parameter resolution — before the handler runs**, so `_gateway_auth_error` executes only *after* the full parse, and there is no body-size limit anywhere on the route. In the non-loopback + `OPENSRE_ALERT_LISTENER_TOKEN` deployment this endpoint is built for, an **unauthenticated** caller can force full buffering/parsing of an arbitrarily large body ahead of the 401/403 — a pre-auth memory/CPU DoS.

The sibling `POST /alerts` already does this correctly: auth first, then a `Content-Length` check, then a 1 MiB (`MAX_ALERT_BODY_BYTES`) cap on the buffered body, then parse.

This PR makes `/investigate` mirror `/alerts`:
1. Take the raw `Request` (no Pydantic body parameter).
2. `_gateway_auth_error(request)` first.
3. `Content-Length` check, then `await request.body()` with the `MAX_ALERT_BODY_BYTES` cap.
4. Parse `InvestigateRequest` from the bytes (`ValidationError` -> 422).

The handler is now `async`; the blocking pipeline is offloaded via `anyio.to_thread.run_sync(functools.partial(run_investigation_payload, ...))`, so it still does not block the event loop or `/health`.

Two new tests in `gateway/tests/test_webapp_investigate.py`:
- `test_investigate_body_over_cap_returns_413` — with the cap lowered, an over-cap body returns **413** and the pipeline is never invoked.
- `test_investigate_auth_runs_before_body_parse` — a remote unauthenticated caller sending an **invalid** body gets **403** (auth), not 422 (body validation) — proving auth precedes the parse.

Verified both fail against the pre-fix handler and pass after:
```
# before the fix (Pydantic-param handler)
FAILED test_investigate_body_over_cap_returns_413            # no cap -> 200
FAILED test_investigate_auth_runs_before_body_parse          # parse-first -> 422, not 403

# after the fix
9 passed   (7 existing + 2 new)
```

Local CI: `make lint`, `make format-check`, `make typecheck` clean; `make test-cov` passes (10,781 passed).

---

The root cause is that a Pydantic body parameter moves body buffering + parsing *ahead* of the handler body, so no in-handler auth or size check can precede it. The fix therefore has to drop the body parameter and read the body manually, which is exactly what `/alerts` does — so I mirrored it rather than inventing a new shape, reusing `MAX_ALERT_BODY_BYTES` and the same Content-Length/`len(body)` two-step (the second guards chunked requests with no declared length).

Reading the body requires `await`, so the handler becomes `async`. A naive async handler would then run the long synchronous pipeline directly on the event loop and block `/health` — worse than before. I offload it with `anyio.to_thread.run_sync` (anyio is already a FastAPI dependency), preserving the original "runs off the event loop" property. I considered a Starlette body-size middleware instead, but that would cap size without moving the parse after auth (the Pydantic param would still parse first), so it doesn't close the actual ordering bug. A dedicated bounded executor to also isolate `/investigate` from `/health` under heavy concurrency is a reasonable follow-up but is a separate capacity-tuning decision, so I kept this PR to the security fix.

---

### DIFF
--- a/gateway/tests/test_webapp_investigate.py
+++ b/gateway/tests/test_webapp_investigate.py
@@ -158,3 +158,42 @@ def test_investigate_token_auth(monkeypatch: pytest.MonkeyPatch) -> None:
         ).status_code
         == 200
     )
+
+
+def test_investigate_body_over_cap_returns_413(
+    monkeypatch: pytest.MonkeyPatch, client: TestClient
+) -> None:
+    """A body larger than the cap is rejected before the pipeline runs."""
+    monkeypatch.setattr(webapp, "MAX_ALERT_BODY_BYTES", 100)
+    called = False
+
+    def _should_not_run(**_: Any) -> dict[str, Any]:
+        nonlocal called
+        called = True
+        return _fake_payload()
+
+    monkeypatch.setattr(webapp, "run_investigation_payload", _should_not_run)
+
+    resp = client.post("/investigate", json={"raw_alert": {"blob": "x" * 500}})
+
+    assert resp.status_code == 413
+    assert called is False
+
+
+def test_investigate_auth_runs_before_body_parse(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """An unauthenticated caller is rejected by auth *before* the body is parsed.
+
+    A remote caller sending an invalid body must get 403 (auth), not 422 (body
+    validation). With the body bound as a Pydantic parameter the framework
+    parsed and rejected it (422) before the handler's auth check ever ran --
+    the pre-auth parse this fix closes.
+    """
+    monkeypatch.setattr(webapp, "run_investigation_payload", lambda **_: _fake_payload())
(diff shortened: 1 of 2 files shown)


cloudenochcsis:
@greptile review

cloudenochcsis:
@greptile review

cloudenochcsis:
Closing this — it should be handled through the private security process per SECURITY.md rather than a public PR. Reporting it to support@opensre.com and will coordinate remediation there.

PR #3956, opened 2026-07-11
fix(wizard): persist Ollama host where the runtime reads it (#3291)

### LINKED ISSUE #3291: Custom Ollama host is saved to the keyring but read from the environment, so it is silently ignored

Ourbando:
Fixes #3291

A custom Ollama host entered in the onboarding wizard was written to the system
keyring, but every runtime consumer reads `OLLAMA_HOST` from the **environment** —
so the custom host was silently ignored (OpenSRE fell back to `localhost:11434`)
and trapped: the keyring entry made the wizard skip the re-prompt on the next run.

A `host` credential is non-secret runtime configuration, so this treats it as the
environment value it is, end to end:

1. **Persist to `.env` + `os.environ`, not the keyring** — a new
   `_persist_llm_credential` routes `credential_kind == "host"` through
   `sync_env_values`; secret credentials keep the keyring path unchanged.
2. **Survive the same-run provider sync** — `sync_provider_env` deliberately
   strips every provider's API-key env line (secrets live in the keyring), and
   Ollama's `api_key_env` *is* `OLLAMA_HOST`, so the host was collateral damage
   later in the same wizard run. It now keeps the active provider's `host` key as
   a non-secret active key (mirroring how the Azure endpoint is preserved), so
   the host is no longer stripped from `.env` and the environment.
3. **Env-based "already configured" check** — `_local_defaults` derives a host
   credential's presence from the environment, so a stale keyring-only entry no
   longer suppresses the re-prompt (which previously defaulted the host back to
   localhost).

The runtime read paths (`config/config.py`, `config/llm_auth/credentials.py`)
already read `OLLAMA_HOST` from the environment and are unchanged — this aligns
the wizard's persistence with the documented `export OLLAMA_HOST=...` behavior.

### Demo: end-to-end screencast

opensre#3291 custom Ollama host, before vs after

The real onboarding wizard, run on the pre-fix commit and then on this fix (same steps, same host):

- **Before:** enter a custom Ollama host. `opensre config` resolves `localhost:11434` (the host was written to the keyring, which the runtime never reads), and a real `opensre investigate` never reaches it (the stub host stays silent).
- **After:** same steps. `.env` keeps `OLLAMA_HOST`, `opensre config` resolves the custom host, and `opensre investigate` actually connects to it.

Reproduce: `opensre onboard` with a custom Ollama host, then check `opensre config`. Without this fix it resolves `localhost:11434`; with it, your host.

#### Regression test (red/green)

Without the `sync_provider_env` fix, the host is stripped in the same wizard run:

```
$ pytest -k preserves_host_credential
    assert "OLLAMA_HOST=http://10.0.0.5:11434\n" in content
E   AssertionError: assert 'OLLAMA_HOST=http://10.0.0.5:11434\n' in
                            'LLM_PROVIDER=ollama\nOLLAMA_MODEL=llama3.1\n'
1 failed
```

With the fix, the whole wizard + model-persistence surface passes:

```
$ pytest tests/cli/wizard/test_env_sync.py tests/cli/wizard/test_flow.py \
         tests/cli/test_model_persistence_scenarios.py
107 passed
```

`ruff check`, `ruff format --check`, and `mypy` are clean on all changed files.

---

- [x] Yes, I used AI assistance

🔏 Corroboration — a signed, re-runnable receipt for this contribution (measured PR state · CI · diff · green tests, then signed): https://gist.github.com/Ourbando/f2cc5e9315cc7c31be3df51fc0498171 — verify with stock OpenSSH (`ssh-keygen -Y verify`, zero installs); any tampered byte is refused.

The bug is a persistence/read mismatch: the wizard treated the Ollama host like an
API key (keyring), but the host is keyless runtime config the runtime only reads
from the environment. I considered teaching the runtime to also read the keyring,
but that spreads a secret-store dependency across every consumer for a non-secret
value and contradicts the documented env-based behavior — so instead the host is
routed to `.env`/`os.environ`, keeping one source of truth. The non-obvious part
is `sync_provider_env`: it intentionally removes every provider's API-key line
because secrets belong in the keyring, and since Ollama's `api_key_env` slot is
`OLLAMA_HOST`, a correct persist in step 1 was undone moments later in the same
run. Keeping the active `host` key as a non-secret active key fixes that, and a
regression test pins the behavior so it can't silently regress again.

---

### DIFF
--- a/surfaces/cli/wizard/_ui.py
+++ b/surfaces/cli/wizard/_ui.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import os
 from collections.abc import Mapping
 from dataclasses import dataclass
 from typing import cast
@@ -121,6 +122,7 @@ def _local_defaults() -> dict[str, str | bool | None]:
         local.get("api_key_env"), api_key_provider.api_key_env if api_key_provider else ""
     )
     is_cli = bool(raw_provider_option and raw_provider_option.credential_kind == "cli")
+    is_host = bool(api_key_provider and api_key_provider.credential_kind == "host")
     is_oauth_backend = bool(raw_provider_value and raw_provider_value != provider_value)
     raw_auth_method = local.get("auth_method")
     auth_method = (
@@ -136,7 +138,15 @@ def _local_defaults() -> dict[str, str | bool | None]:
         "auth_method": auth_method,
         "model": _string_value(local.get("model")),
         "api_key_env": api_key_env,
-        "has_api_key": True if is_cli else bool(api_key_env and has_llm_api_key(api_key_env)),
+        # A ``host`` credential (e.g. the Ollama host) is only real when the
+        # runtime can see it — the environment — never the keyring.
+        "has_api_key": True
+        if is_cli
+        else (
+            bool(api_key_env and os.getenv(api_key_env, "").strip())
+            if is_host
+            else bool(api_key_env and has_llm_api_key(api_key_env))
+        ),
         "legacy_api_key": _string_value(local.get("api_key")),
     }
 
--- a/surfaces/cli/wizard/env_sync.py
+++ b/surfaces/cli/wizard/env_sync.py
@@ -334,6 +334,11 @@ def sync_provider_env(
             active_non_secret.add(provider.endpoint_env)
         if provider.api_version_env:
             active_non_secret.add(provider.api_version_env)
+    # A ``host`` credential (e.g. the Ollama host) is non-secret runtime config
+    # that the wizard persists to ``.env`` — keep it as an active key so this
+    # sync does not strip it back out in the same wizard run (see #3291).
+    if provider.credential_kind == "host" and provider.api_key_env:
+        active_non_secret.add(provider.api_key_env)
     keys_to_remove -= active_non_secret
 
     lines = _remove_keys(existing, keys_to_remove)
--- a/surfaces/cli/wizard/flow.py
+++ b/surfaces/cli/wizard/flow.py
@@ -831,7 +831,11 @@ def run_wizard(_argv: list[str] | None = None) -> int:
             ):
                 has_api_key = bool(defaults["has_api_key"])
                 legacy_api_key = str(defaults["legacy_api_key"] or "").strip()
-                if not has_api_key and legacy_api_key:
+                # A ``host`` credential (e.g. the Ollama host) is not a secret api key: never
+                # migrate a stale legacy ``api_key`` value into it — that would leak a
+                # secret-shaped value into .env and point the runtime at a bogus host. Fall
+                # through to the host prompt instead (#3291).
+                if not has_api_key and legacy_api_key and provider.credential_kind != "host":
                     if not _persist_llm_credential(provider, legacy_api_key):
                         return 1
                     has_api_key = True
--- a/surfaces/cli/wizard/flow.py
+++ b/surfaces/cli/wizard/flow.py
@@ -224,6 +224,21 @@ def _credential_prompt_label(provider: ProviderOption) -> str:
     return provider.label
 
 
+def _persist_llm_credential(provider: ProviderOption, value: str) -> bool:
+    """Persist one prompted credential where the runtime will actually read it.
+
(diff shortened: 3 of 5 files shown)


Ourbando:
Addressed the P1 (legacy key becomes host): host-kind credentials such as Ollama now skip legacy api_key migration and fall through to the host-URL prompt, so a stale secret-shaped value can no longer be written into OLLAMA_HOST. Added a regression test that drives the saved-config path and asserts the legacy value is never persisted as the host.

@greptile review

muddlebee:
@Ourbando doesnt follow PR guidelines, a proper demo isn;t there

Ourbando:
Thanks for taking a look, and no problem closing it out.

For a wizard change I get wanting to see it work. There's a red/green `pytest` demo in there, but a live walkthrough is clearer. Happy to record a short screencast of the wizard saving a custom `OLLAMA_HOST` and the runtime reading it back if it's ever worth revisiting.

Appreciate you maintaining opensre either way. I'll lead with a screencast next time. 🤝

Ourbando:
@muddlebee following up on my note above. Here's the proper demo, a full screencast.

opensre#3291 demo

It runs the real onboarding wizard on the pre-fix code, then on this fix (same steps, same host):

- **Before:** the custom Ollama host lands in the keyring, so `opensre config` falls back to `localhost:11434`, and a real `opensre investigate` never reaches your host (the stub stays silent).
- **After:** it persists to `.env`, `opensre config` resolves your host, and `opensre investigate` actually connects to it.

Full before/after and repro steps are in the description. GitHub won't let me reopen from my side, so it's yours to flip back open if you're happy with it. Appreciate the push on the demo, it's a better PR for it. 🍻

muddlebee:
@Ourbando thank you, this looks valid. I will look into it deeper

muddlebee:
@Ourbando LGTM, thank you for clarification 🙏

PR #3957, opened 2026-07-11
fix(interactive_shell): stop masking REPL errors when alert listener is active (#3940)

### LINKED ISSUE #3940: [BUG] Interactive shell masks real REPL errors as "generator didn't stop after throw()" when alert listener is active
## Summary
In `surfaces/interactive_shell/controller.py`, the `_alert_listener` `@contextmanager` puts its `yield inbox` (`:96`) inside the same `try` (`:86`) whose `except Exception` (`:101`) is meant to guard only listener **startup**. Any exception raised in the REPL body under `with _alert_listener(...)` is thrown back at that `yield`, caught by the startup handler, mislogged as "Alert listener could not start", and followed by a second `yield None` (`:104`) — which makes `contextlib` raise `RuntimeError("generator didn't stop after throw()")`, replacing the user's real error.

**Expected:** An exception raised inside the REPL turn body while the alert listener is active propagates unchanged (the startup `except` should only ever catch failures bringing the listener up).

**Actual:** The real error is swallowed and masked. The startup `except` catches it, emits a misleading `Alert listener could not start: ... — continuing without it.` warning, then executes a second `yield`, so `contextlib.contextmanager` raises `RuntimeError: generator didn't stop after throw()`. The original error and its traceback are lost.

## Steps to reproduce
1. Configure the interactive shell with the alert listener enabled (`ReplConfig.alert_listener_enabled = True`, no pre-existing `existing` inbox) so `_alert_listener` enters the startup path and reaches `yield inbox`.
2. Start the interactive shell, which runs the REPL body inside `with _alert_listener(...)`.
3. Trigger any exception in the turn body (e.g. a dispatch/handler error during an input action).
4. Observe the masked `RuntimeError("generator didn't stop after throw()")` plus the misleading "could not start" warning, instead of the real error.

Yes

## How often does it occur?
Every time — any body error raised while the alert listener is active hits this path.

## Operating system
Other

## Logs and error output
```
WARNING  Alert listener could not start: <original error> — continuing without it.
...
RuntimeError: generator didn't stop after throw()
```
(The "could not start" line is misleading — the listener had already started; `<original error>` is the real REPL error being swallowed.)

## Additional context
Impact: genuine REPL errors are hidden and replaced with a confusing `RuntimeError` whenever the alert listener is running, making the interactive shell hard to debug. Root cause is scope — `yield inbox` must sit outside the startup `try/except` so only startup failures are caught.

A fix is already prepared (restructure with `contextlib.ExitStack`: catch only startup, yield outside the except with the stack open), previously in the now-closed PR #3906 (a reviewer had approved the code — "looks good to me"). Per CONTRIBUTING's issue-first flow, happy to open a linked PR once this is triaged/assigned.

```bash
uv run opensre gateway start -f      # so alerts arrive
uv run opensre                       # alert listener enabled in ~/.opensre/config.yml
> /theme [/]                         # any input that raises inside the turn
```

**Pass:** the real error and its traceback appear. **Fail (today):** "Alert
listener could not start … continuing without it", then a RuntimeError that
replaces the original error.

ishaand0314:
Fixes #3940

`_alert_listener` in `surfaces/interactive_shell/controller.py` is a `@contextmanager` that starts the alert-listener web server, then yields control to the REPL body, then tears it down. Its `try/except Exception` was meant to guard startup only, but `yield inbox` sat inside that same `try` block. When the REPL body raised, the exception resumed at the `yield`, so it was caught by the startup `except`, logged as "Alert listener could not start", and then a second `yield None` ran. That made `contextlib` raise `RuntimeError("generator didn't stop after throw()")`, replacing the real REPL error.

This PR restructures `_alert_listener` around `contextlib.ExitStack`. The `try/except` now wraps bring-up only (set current inbox, install alert token, start server thread), cleanup is registered as stack callbacks, and `yield inbox` moves outside the `try/except`. A REPL-body exception now propagates unchanged. Cleanup (`handle.stop()`, clearing the inbox, restoring the token) still runs via the `ExitStack`. A genuine bring-up failure still yields `None`.

Also removed an import (`WebAppServerHandle`) no longer needed after the restructure.

Before (main), a REPL-body exception is masked:

```
$ uv run python -m pytest tests/interactive_shell/test_alert_listener.py::test_repl_body_error_propagates_unchanged -v
...
E   RuntimeError: generator didn't stop after throw()
WARNING  surfaces.interactive_shell.controller:controller.py:102 Alert listener could not start: real repl error, continuing without it.
============================== 1 failed in 3.46s ==============================
```

After (this branch), the real exception propagates and the full suite passes:

```
$ uv run python -m pytest tests/interactive_shell/test_alert_listener.py -v
tests/interactive_shell/test_alert_listener.py::test_alert_listener_replaces_stale_process_token PASSED [ 25%]
tests/interactive_shell/test_alert_listener.py::test_alert_listener_clears_token_when_unconfigured PASSED [ 50%]
tests/interactive_shell/test_alert_listener.py::test_repl_body_error_propagates_unchanged PASSED [ 75%]
tests/interactive_shell/test_alert_listener.py::test_startup_failure_yields_none_without_masking PASSED [100%]
============================== 4 passed in 3.69s ==============================
```

`make lint`, `make format-check`, and `make typecheck` are clean on the changed files.

`make test-cov` (full suite) does not complete within 10 minutes in my local Windows environment; the run reached 98 percent with a single unrelated failure, `tests/fleet_monitoring/test_probe.py::test_probe_returns_nonzero_cpu_for_busy_process`. I confirmed this failure is pre-existing on `main` (reproduces identically with this branch's two changed files reverted) and is a CPU-sampling timing assertion unrelated to the interactive shell. The scoped, directly relevant suite (`tests/interactive_shell/`) passes in full: 4 passed.

---

The bug is a `@contextmanager` pitfall: code after `yield` inside a `try/except` catches exceptions raised from the `with` block's body, not only exceptions raised before the `yield`. The fix separates the two concerns. `ExitStack` was chosen over nested `try/finally` because it registers cleanup incrementally as each resource is acquired, so a failure partway through bring-up tears down only what was actually acquired, in reverse order. The `try/except` now wraps bring-up only; `yield inbox` is placed after it but still inside the `with ExitStack() as stack:` block, so cleanup still runs on any exit path through the stack's own `__exit__`.

Alternative considered: keep the nested `try/finally` and move `yield` out of the `except`-guarded region. This would work but scales poorly if a third bring-up step is added later. `ExitStack` keeps ordering and partial-failure cleanup correct without hand-nested blocks.

Edge cases tested: a REPL-body exception while the listener is running (propagates unchanged, cleanup still runs), a genuine startup failure (yields `None`, does not raise), and the two pre-existing token-management tests (stale token replaced or cleared) to confirm no regression.

---

---

### DIFF
--- a/surfaces/interactive_shell/controller.py
+++ b/surfaces/interactive_shell/controller.py
@@ -99,6 +99,7 @@ def _alert_listener(
             stack.callback(handle.stop)
         except Exception as exc:
             log.warning("Alert listener could not start: %s — continuing without it.", exc)
+            stack.close()
             yield None
             return
 
--- a/surfaces/interactive_shell/controller.py
+++ b/surfaces/interactive_shell/controller.py
@@ -6,7 +6,7 @@ import asyncio
 import logging
 import os
 from collections.abc import Iterator
-from contextlib import contextmanager
+from contextlib import ExitStack, contextmanager
 
 from prompt_toolkit import PromptSession
 from prompt_toolkit.patch_stdout import patch_stdout
@@ -79,29 +79,31 @@ def _alert_listener(
         yield None
         return
 
-    from gateway.web_server import WebAppServerHandle, serve_webapp_in_thread
+    from gateway.web_server import serve_webapp_in_thread
 
-    inbox: _alert_inbox.AlertInbox | None = None
-    handle: WebAppServerHandle | None = None
-    try:
-        inbox = _alert_inbox.AlertInbox()
-        _alert_inbox.set_current_inbox(inbox)
-        with _alert_listener_token(cfg.alert_listener_token):
+    inbox = _alert_inbox.AlertInbox()
+    with ExitStack() as stack:
+        try:
+            # Bring-up only. This ``except`` must not see exceptions from the
+            # REPL body that runs during ``yield inbox``: a body error caught
+            # here would be mislogged as "could not start", and the following
+            # ``yield None`` would make ``@contextmanager`` raise "generator
+            # didn't stop after throw()", masking the real error (#3940).
+            _alert_inbox.set_current_inbox(inbox)
+            stack.callback(_alert_inbox.set_current_inbox, None)
+            stack.enter_context(_alert_listener_token(cfg.alert_listener_token))
             handle = serve_webapp_in_thread(
                 host=cfg.alert_listener_host,
                 port=cfg.alert_listener_port,
             )
-            console.print(f"[{DIM}]listening for alerts on http://{handle.bound_address}/alerts[/]")
-            try:
-                yield inbox
-            finally:
-                if handle is not None:
-                    handle.stop()
-                _alert_inbox.set_current_inbox(None)
-    except Exception as exc:
-        log.warning("Alert listener could not start: %s — continuing without it.", exc)
-        _alert_inbox.set_current_inbox(None)
-        yield None
+            stack.callback(handle.stop)
+        except Exception as exc:
+            log.warning("Alert listener could not start: %s — continuing without it.", exc)
+            yield None
+            return
+
+        console.print(f"[{DIM}]listening for alerts on http://{handle.bound_address}/alerts[/]")
+        yield inbox
 
 
 def _resolve_runtime_context(
--- a/tests/interactive_shell/test_alert_listener.py
+++ b/tests/interactive_shell/test_alert_listener.py
@@ -9,6 +9,7 @@ import pytest
 from rich.console import Console
 
 from config.repl_config import ReplConfig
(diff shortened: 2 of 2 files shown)


ishaand0314:
@greptile review

ishaand0314:
@greptile review

Addressed the confidence-4/5 finding: the startup-failure fallback path now calls `stack.close()` before `yield None`, so the registered cleanup (clear current inbox, restore token) runs before the REPL body executes on the degraded path, instead of after. Added a regression test (`test_startup_failure_clears_global_inbox_before_body_runs`) that fails without the fix and passes with it.

PR #3958, opened 2026-07-11
fix(execution): propagate contextvars to parallel tool-call threads (#3939)

### LINKED ISSUE #3939: [BUG] Parallel tool calls emit no trace spans — ThreadPoolExecutor workers lose the session-trace contextvar
## Summary
On the parallel tool-execution path in `core/execution.py`, tool trace spans are silently dropped. `execute_tool_calls` submits each call to a `ThreadPoolExecutor`, but worker threads start with an empty `contextvars` context, so the session-trace binding set by `bind_session_trace` on the calling thread is invisible in the worker. `tool_span` -> `timed_span` -> `emit_span` short-circuits on the unset session id and emits nothing. Every tool that runs on the parallel path (the common 2+-tool turn) produces zero `tool` spans; only the single-tool / forced-sequential path (which runs on the calling thread) is traced, which is why it went unnoticed.

**Expected:** One `trace_span` of `span_kind="tool"` per tool call, including calls that run in parallel — matching the single-tool path.

**Actual:** Zero `tool` spans are emitted for any tool that runs on the parallel path. A turn dispatching 2+ parallel-safe tools produces no `tool` spans at all; a single-tool turn produces its span normally.

Root cause: `core/execution.py:143` does `pool.submit(_call, tc)` without wrapping in `contextvars.copy_context().run`. `_call` (`:122-123`) enters `tool_span(tc.name, tool_call_id=tc.id)`. In `platform/observability/trace/spans.py`, `tool_span` delegates to `timed_span`, which resolves `sid = session_id or _session_id.get()` and, when `sid` is falsy, yields the attrs dict and returns *without emitting*. The pool worker's context is empty, so `_session_id.get()` returns `None` and no span is recorded. The same subsystem already handles this correctly at `tools/investigation/capability.py:482` via `contextvars.copy_context().run` — exactly the pattern missing here.

## Steps to reproduce
1. Register a real (non-noop) session-trace sink and bind a session via `bind_session_trace(session_id)` (the REPL / JSONL / ATM path where tracing is active).
2. Run a turn where the LLM requests 2+ parallel-safe tools, so `execute_tool_calls` takes the `ThreadPoolExecutor` branch (not `len(tool_calls) == 1`, not `_requires_sequential_execution`).
3. Inspect the session JSONL / ATM output: there are **no** `trace_span` rows of `span_kind="tool"` for those calls.
4. Compare with a single-tool turn (or a turn forced sequential): the `tool` span is present. The delta confirms the parallel path is dropping spans.

Yes

## How often does it occur?
Every time — whenever 2+ tools run in parallel under an active trace session.

## Operating system
Other

## Logs and error output
No error is raised — silent gap. Relevant paths: `core/execution.py:143` (`pool.submit(_call, tc)`, no context copy), `core/execution.py:122-123` (`_call` enters `tool_span`), `platform/observability/trace/spans.py` (`timed_span`: `sid = session_id or _session_id.get()` then `if not sid: ... return`), `tools/investigation/capability.py:482` (correct `contextvars.copy_context().run` reference).

## Additional context
Impact is on observability, not correctness: tools still execute and return results, but the ATM / session trace loses `tool` spans for the most common multi-tool turn, undercounting tool activity and breaking per-tool timing/outcome data. Because the single-tool path is unaffected, spot checks look fine while real multi-tool traffic goes untraced.

A fix is already prepared (submit via `contextvars.copy_context().run`, mirroring `capability.py`), previously in the now-closed PR #3905. Per CONTRIBUTING's issue-first flow, happy to open a linked PR once this is triaged/assigned.

ishaand0314:
Fixes #3939

`execute_tool_calls` in `core/execution.py` dispatches 2+ parallel-safe tool calls to a `ThreadPoolExecutor` via `pool.submit(_call, tc)`. `_call` enters `tool_span(...)`, which resolves the active session id from a `contextvars.ContextVar` bound earlier on the calling thread by `bind_session_trace`. Pool worker threads are reused across submissions and start with an empty `contextvars` context, so that binding is invisible inside the worker. `tool_span`/`timed_span` read an unset session id and silently return without emitting anything. Every tool call dispatched on the parallel path produced zero `tool` trace spans; only the single-tool or forced-sequential path (which runs on the calling thread, context intact) traced correctly.

This PR copies the caller's context per submission and runs `_call` through it: `ctx = contextvars.copy_context(); pool.submit(ctx.run, _call, tc)`. This mirrors an existing, correct use of the same pattern elsewhere in the codebase (`tools/investigation/capability.py`, which uses `contextvars.copy_context().run` as a thread target for the same reason: propagating the session-trace binding into a new thread).

This supersedes the previously closed PR #3905, which addressed the same root cause; that attempt is referenced in the issue body.

Before (main), a 2-tool parallel batch emits zero `tool` spans:

```
$ uv run python -m pytest tests/core/runtime/test_tool_execution_contract.py::test_parallel_tool_calls_emit_tool_trace_spans -v
...
E       assert [] == ['first', 'second']
FAILED tests/core/runtime/test_tool_execution_contract.py::test_parallel_tool_calls_emit_tool_trace_spans
```

After (this branch), spans are emitted for both parallel calls and the full contract suite passes:

```
$ uv run python -m pytest tests/core/runtime/test_tool_execution_contract.py -v
...
25 passed in 2.30s
```

Broader scope also verified, per this repo's rule that `core/` changes require `make test-cov` and `make typecheck`:

```
$ uv run python -m pytest tests/core/ tests/platform/observability/ -q
1283 passed, 30 skipped in 36.63s
```

`make lint`, `make format-check`, and `make typecheck` are clean on the changed files.

`make test-cov` (full suite) does not complete within 10 minutes in my local Windows environment. The run reached 99 percent; the last event before the timeout was an xdist worker crash (`[gw5] node down: Not properly terminated`) on `tests/interactive_shell/ui/test_feedback.py::test_run_select_returns_highlighted_choice_on_enter`, a Unix terminal raw-mode test unrelated to this change. In isolation that single test hangs indefinitely on this machine and never completes, confirming it is a pre-existing, environment-specific issue rather than a regression from this diff. The scoped, directly relevant suites (`tests/core/`, `tests/platform/observability/`) pass in full: 1283 passed, 30 skipped.

---

The bug is a `contextvars` propagation gap: `ContextVar` state is thread-local by design, and a `ThreadPoolExecutor` worker thread does not inherit the submitting thread's context automatically. `contextvars.copy_context()` snapshots the calling thread's context at submission time; `ctx.run(fn, *args)` executes `fn` inside that snapshot on whichever thread picks up the task, giving the worker the same `ContextVar` values (in this case, the session-trace binding) the caller had.

The copy is taken once per submission, inside the loop, rather than once for the whole batch. Each `ToolCall` is independent, and this keeps the isolation semantics simple: a `ContextVar.set()` performed inside one tool's execution (if any tool ever does this) cannot leak into a sibling call's copy, since `Context.run` executes in an isolated copy and does not mutate the caller's context or other copies. A single shared copy would still work correctly for the read-only session-trace binding here, but per-submission copies keep the isolation guarantee general rather than relying on the current call graph never mutating context vars.

Alternative considered: wrapping the whole `with ThreadPoolExecutor(...) as pool:` block in a single `contextvars.copy_context().run(...)`, similar to the `capability.py` thread-target pattern. This does not fit here because `execute_tool_calls` submits multiple independent tasks to a pool rather than starting one dedicated thread, so the fix is applied at the `pool.submit` call site instead.

Edge cases tested: a 2-tool parallel batch emits a `tool` span per call, with the correct session id on each; the existing parallel-order test still passes unmodified.

---

---

### DIFF
--- a/core/execution.py
+++ b/core/execution.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import contextvars
 import json
 import logging
 from collections.abc import Callable, Sequence
@@ -140,7 +141,12 @@ def execute_tool_calls(
     try:
         with ThreadPoolExecutor(max_workers=min(_TOOL_EXECUTOR_WORKERS, len(tool_calls))) as pool:
             for i, tc in enumerate(tool_calls):
-                submitted[pool.submit(_call, tc)] = i
+                # Copy the caller's context so ContextVar bindings (session
+                # trace) reach the pool worker; ThreadPoolExecutor threads are
+                # reused across submissions and otherwise start with an empty
+                # context, which silently drops tool trace spans (#3939).
+                ctx = contextvars.copy_context()
+                submitted[pool.submit(ctx.run, _call, tc)] = i
             for fut in as_completed(submitted):
                 try:
                     results[submitted[fut]] = fut.result()
--- a/tests/core/runtime/test_tool_execution_contract.py
+++ b/tests/core/runtime/test_tool_execution_contract.py
@@ -21,6 +21,11 @@ from core.llm.types import AgentLLMResponse, ToolCall
 from core.provider import ProviderHooks, ProviderRequest
 from core.tool_framework.registered_tool import RegisteredTool
 from core.types import AgentTool, AgentToolContext
+from platform.observability.trace.spans import (
+    NoopSessionTraceSink,
+    bind_session_trace,
+    set_session_trace_sink,
+)
 
 
 def _schema(required: list[str] | None = None) -> dict[str, Any]:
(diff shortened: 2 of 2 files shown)


ishaand0314:
@greptile review

PR #2657, opened 2026-05-29
feat(cli) : background RCA email notifications via SMTP

muddlebee:
Fixes #2654

#### Describe the changes you have made in this PR -
- add REPL background investigation mode with `/background on|off|status|list|show|use|notify ...`
- route free-text alerts and `/investigate` launches into session-local background tasks when background mode is enabled
- collect completed RCA summaries without taking over the active follow-up context until `/background use <task_id>`
- add a new `smtp` integration with CLI setup, env bootstrap, strict config validation, and `opensre integrations verify smtp`
- send background RCA completion notifications through SMTP using a plain-text summary focused on root cause, top analysis, and next steps
- add docs for background investigations and SMTP setup, plus targeted CLI/integration/unit coverage

### Demo/Screenshot for feature changes and bug fixes - 
Terminal + local Mailpit verification:

```text
$ SMTP_HOST=127.0.0.1 SMTP_PORT=1025 SMTP_SECURITY=none SMTP_FROM_ADDRESS=opensre@example.com SMTP_DEFAULT_TO=team@example.com uv run opensre integrations verify smtp

  SERVICE    SOURCE       STATUS      DETAIL
  smtp      local env    passed      Connected to SMTP server successfully.
```

```text
$ uv run python ... send_smtp_report(...)
{'ok': True, 'error': '', 'subject': 'OpenSRE RCA complete: bg-local-1'}
```

```json
{"total":1,"messages":[{"Subject":"OpenSRE RCA complete: bg-local-1","To":[{"Address":"team@example.com"}],"From":{"Address":"opensre@example.com"}}]}
```

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [x] I have reviewed every single line of the AI-generated code
- [x] I can explain the purpose and logic of each function/component I added
- [x] I have tested edge cases and understand how the code handles them
- [x] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**
- This PR implements the main scoped issue only: session-local background investigations plus email-first RCA completion notifications.
- I kept background execution inside the existing interactive-shell session instead of adding durable workers/queues, because that is the smallest change that preserves current behavior while enabling async RCA runs.
[...]

Issue #2655, opened 2026-05-29
[FEATURE] Background investigation follow-up: additional notification channels and external RCA access


muddlebee:
## Problem statement

The initial background investigation slice can ship with **email-first notifications**, but the original request also calls out broader communication-channel support and the ability to access RCA results through preferred channels such as:

- Telegram
- WhatsApp
- other notification channels the user prefers

That broader scope is important, but it is larger than the first shippable background-mode slice because it overlaps with existing messaging architecture and inbound conversation work.

If we try to include all additional communication channels in the first background-mode issue, we risk delaying the core capability: run RCAs in the background and notify the user by default when they finish.

## Proposed solution

Track post-v1 background-mode communication expansion separately from the core email-first issue.

This follow-up issue should cover the next layer of delivery and access beyond SMTP:

### 1. Additional outbound background notification channels

Support background RCA completion notifications through configured messaging channels beyond email, including at minimum evaluating:

- Telegram
- WhatsApp
- Twilio SMS
- Slack / Discord where appropriate

This work should decide:

- how channel defaults are prioritized when multiple destinations are configured
- whether background notifications should fan out to multiple channels or choose one preferred default
- how delivery failures are surfaced when one channel fails and another succeeds

### 2. External access to completed background RCAs

Allow users to retrieve the results of completed background RCAs from communication channels rather than only from the local REPL.

Examples:

- list recent background jobs from Telegram
- request a completed RCA summary by task id
- resend the root-cause / findings / next-steps summary into a preferred destination

This should be explicitly separated from durable-job architecture: the first step can still operate against local/session-scoped or locally persisted completed-job metadata if that is what the core implementation provides.

### 3. Inbound conversational expansion

Evaluate whether any of the following should be supported for background RCA workflows:

- Telegram command/reply surfaces for completed-RCA lookup
- WhatsApp command/reply surfaces
- Slack / Discord background-RCA retrieval surfaces
- follow-up Q&A against a completed RCA from an external channel
- launching fresh investigations from external channels

This should build on, and stay aligned with:

- #1482 for inbound conversational loop work
- #1483 for shared messaging-provider / gateway architecture

### 4. Notification preference model expansion

The core issue can start with a simple local preference for `email`, but this follow-up should define the longer-term policy model for:

- preferred channel ordering
- multi-channel fanout vs first-success delivery
- per-channel recipient overrides
- user-specific vs machine-local defaults
- per-job override behavior

## Acceptance criteria

[...]

Issue #2654, opened 2026-05-29
[FEATURE] Background investigation mode with email-first RCA notifications


muddlebee:
## Problem statement

OpenSRE investigations are currently foreground-oriented: once a user starts an RCA from the interactive shell, the shell is effectively occupied until the run finishes. That blocks the workflow described in the background context for this feature:

- enable a background mode (for example `/background on`)
- push investigations/prompts into the background
- keep asking other questions in the same agent session while those RCAs continue running
- receive a structured RCA notification when the background run completes

There is also no email delivery path today for RCA completion notifications. Existing delivery paths are chat-oriented, and there is no SMTP integration for an operator to configure a mail server and receive a completion summary by default.

For this first version, the most important notification content is:

1. **What was the root cause**
2. **What were the top analysis findings**
3. **What you need to do next**

The request also calls for some lightweight internal run statistics in the notification (for example duration / investigation details), but those should remain secondary to the three items above.

## Proposed solution

Implement a **session-local background investigation mode** in the interactive shell, with **email as the default completion notification channel**.

### 1. Background mode in the interactive shell

Add a new `/background` command family to the REPL:

- `/background on`
- `/background off`
- `/background status`
- `/background list`
- `/background show <task_id>`
- `/background use <task_id>`

Behavior:

- When background mode is enabled, new investigations launched from free-text alert input or `/investigate ...` should run asynchronously as background REPL tasks instead of taking over the active terminal flow.
- The user must still be able to continue interacting with agents in the same shell while the RCA is running.
- Completed background RCAs should be visible through `/background list` and `/background show <task_id>`.
- `/background use <task_id>` should promote a completed RCA into the active follow-up context rather than auto-overwriting the current session context when a background task finishes.
- Cancellation should keep working through the existing task registry and `/cancel <task_id>` path.

### 2. Session-local background RCA runner

Add a non-interactive background investigation driver that:

- reuses the existing investigation pipeline
- collects a final investigation state without streaming the live renderer into the main prompt surface
- stores the final RCA state for later inspection and follow-up reuse
- tracks per-task summary metadata such as root cause, findings, remediation, timing, validity score, and notification outcome

For v1, this should remain **session-local**:

- background tasks live only as long as the current REPL process lives
- no durable queue, daemon, or cross-process persistence is required

### 3. SMTP integration for email notifications

Add a new `smtp` integration with setup, verification, env support, and docs.

Suggested config surface:

- SMTP host
- SMTP port
[...]

PR #3981, opened 2026-07-13
feat(notifications): add Telegram as a background-RCA completion channel

Harshvardhan86:
Refs #2655

Adds **Telegram as the second background-RCA completion notification channel**, following the channel seam the email-first v1 (#2654 / #2657) left in place:

- `surfaces/interactive_shell/runtime/background/notifications.py` — a `telegram` branch in `deliver_background_notifications`, mirroring the email branch: credentials resolve through the existing `load_credentials_from_env()` ladder (integration store → env → keyring), the RCA summary reuses the email body (`format_background_rca_email`), and delivery goes through the existing production transport `send_telegram_report` (which owns the 4096-char cap). Sent with `parse_mode=""` because the body is plain text and may contain `<`, `>`, `&` — HTML parse mode would 400 on it.
- `surfaces/interactive_shell/command_registry/background_cmds.py` — `"telegram"` added to `_ALLOWED_NOTIFY_CHANNELS`, so `/background notify set telegram` (or `email,telegram`) is accepted.
- No new config surface: users configure Telegram once via `opensre integrations setup telegram` (`TelegramBotConfig`: `bot_token` + `default_chat_id`), same as every other consumer of that integration.

Failure behavior: an unconfigured integration degrades to a per-cause result string (e.g. `missing telegram integration: TELEGRAM_BOT_TOKEN is not set.`) instead of raising — a notification problem can never mark a completed investigation as failed. Send failures record `failed: <error>` (e.g. `failed: Bad Request: chat not found`), visible in `/background show`. Telegram imports are lazy inside the branch so nothing telegram-related loads into the REPL boot path.

Deliberately out of scope (per the approved slice): WhatsApp/SMS channels, inbound RCA access (#1482/#3264), and any multi-channel preference policy — the existing simple fan-out is unchanged.

**Testing** — 22 new tests, all offline (transports mocked at their boundaries, same pattern as the existing email tests):
- `tests/interactive_shell/runtime/test_background_notifications.py` (+13): sent/failed/empty-error/unconfigured/blank-credentials paths, combined email+telegram in both orders, duplicate-channel dedupe, empty root-cause body, unicode/angle-bracket passthrough with `parse_mode=""`, lazy-import guards, no-raise safety.
- `tests/interactive_shell/test_commands.py` (+5): notify set/list/status acceptance, invalid-channel hint, command-layer dedupe, plus two `/background show` render tests driven through the real dispatcher.
- `tests/interactive_shell/runtime/test_background_notifications_e2e.py` (new, +2): end-to-end tests mocking only `httpx.post` — the real runner completion hook, dispatcher, env-credential resolution, and delivery stack all execute; assertions land on the exact JSON payload Telegram would receive (including the absence of `parse_mode`) and on the 400-failure path rendering in `/background show`.

Every new behavioral test was verified to FAIL against the unmodified code before the implementation existed. `make lint && make format-check && make typecheck && make test-cov` all green; import-linter contracts pass with the same commands CI uses.

Live demo against the real Telegram API with a real bot (video + screenshot below): `/background on` → `/background notify set telegram` → `/investigate generic` → completion → `/background show <id>` renders `telegram:sent`, and the RCA summary (root cause, top analysis, next steps, stats) arrives in the Telegram app as plain text.

https://github.com/user-attachments/assets/d78a7c3c-eb32-4e88-9f59-50286d16b0fb

telegram-rca-screenshot

### Update — `b81ad5c9` (docs + hardening)

Following @muddlebee's note to go through the docs, a second pass found the feature shipped **undocumented**, and that it made existing pages factually wrong. It also surfaced two real bugs in this branch.

**Docs**
- `docs/background-investigations.mdx` — claimed *"notification preferences are email-only in the current shipped path"*. Commands are now a quick-reference table, and the flow now states that channels are **empty by default** (`/background on` alone delivers nothing — the page never said so). Also replaces an invalid `/investigate kubernetes-high-cpu` template.
- `docs/messaging/telegram.mdx` — background RCA added to both delivery-surface enumerations, plus a worked end-to-end subsection.
- `docs/interactive-shell-commands.mdx` — `/background` was entirely absent from the "complete reference for every slash command".
- `docs/smtp.mdx` — Step 3 told users `/background on` was enough to receive mail.

**Fixes** (both in the telegram branch this PR owns, each with a regression test that fails without it)
- **Bot-token redaction.** The token travels in the request URL, and `post_telegram_message` surfaces a non-JSON error body verbatim, so a proxy error page echoing the URL would land a live token in the background record — which `/background show` renders. Now redacted at the call site.
- **4096-char budget.** The RCA body is built root-cause-first and ends with "What to do next" + the stats block, while `send_telegram_report` tail-truncates. An unbounded root cause therefore amputated exactly the actionable sections: a realistic record produced a **23,631-char body, ~83% of it discarded**. Sections are now budgeted so the tail always survives; email still carries the full report.

---

The problem: background-RCA completion notifications only supported email; #2655 asks for more channels. Rather than building new notification architecture, I extended the dispatcher loop that v1 deliberately left open — it iterates configured channels and returned "unsupported" for anything but email. The telegram branch mirrors the email branch's exact shape: guard (credentials), format (reuse the email body — Telegram has no subject concept, so the subject is discarded), send (existing hardened transport), record result string. I considered a new `format_background_rca_telegram` helper and rejected it: the email body already carries everything including the Task ID, and a second formatter would be duplicate surface to maintain. I also considered `parse_mode="HTML"` for nicer formatting and rejected it because the RCA text is untrusted plain text — unescaped `<`/`>` would make Telegram reject the message. The two production functions touched are `deliver_background_notifications` (channel loop; per-channel results dict consumed by `/background show`) and the `_ALLOWED_NOTIFY_CHANNELS` tuple (single validation point for `/background notify set`). Edge cases tested: unconfigured and half-configured credentials, blank-string credentials, transport failures with and without error detail, duplicate channels, empty root cause, >4096-char bodies (transport truncates), and channel-order independence.

---

### DIFF
--- a/surfaces/interactive_shell/runtime/background/notifications.py
+++ b/surfaces/interactive_shell/runtime/background/notifications.py
@@ -7,6 +7,39 @@ from surfaces.interactive_shell.session.background_investigations import (
     BackgroundInvestigationRecord,
 )
 
+# Telegram caps a message at 4096 characters and the transport tail-truncates to fit.
+# The RCA body ends with "What to do next" and the stats block, so an unbounded root
+# cause would push exactly the actionable sections off the end. Budget each section
+# instead: the worst case below stays under the cap, so the tail always survives.
+_TELEGRAM_COMMAND_CHARS = 200
+_TELEGRAM_ROOT_CAUSE_CHARS = 1000
+_TELEGRAM_ITEM_CHARS = 240
+_TELEGRAM_MAX_ITEMS = 5
+
+
+def _telegram_summary_sections(
+    record: BackgroundInvestigationRecord,
+) -> tuple[str, str, tuple[str, ...], tuple[str, ...]]:
+    """Return the RCA sections trimmed to fit one Telegram message.
+
+    Email keeps the full report; Telegram is a notification, so it carries a bounded
+    summary and points the reader at ``/background show`` for the rest.
+    """
+    from platform.common.truncation import truncate
+
+    def _items(values: tuple[str, ...]) -> tuple[str, ...]:
+        return tuple(
+            truncate(value, _TELEGRAM_ITEM_CHARS, suffix="…")
+            for value in values[:_TELEGRAM_MAX_ITEMS]
+        )
+
+    return (
+        truncate(record.command, _TELEGRAM_COMMAND_CHARS, suffix="…"),
+        truncate(record.root_cause, _TELEGRAM_ROOT_CAUSE_CHARS, suffix="…"),
+        _items(record.top_analysis),
+        _items(record.next_steps),
+    )
+
 
 def deliver_background_notifications(
     *,
@@ -30,6 +63,7 @@ def deliver_background_notifications(
             # REPL boot import path.
             from integrations.telegram.credentials import load_credentials_from_env
             from integrations.telegram.delivery import send_telegram_report
+            from platform.notifications.redaction import redact_token
 
             try:
                 creds = load_credentials_from_env()
@@ -37,12 +71,13 @@ def deliver_background_notifications(
                 results["telegram"] = f"missing telegram integration: {exc}"
                 continue
 
+            command, root_cause, top_analysis, next_steps = _telegram_summary_sections(record)
             _subject, body = format_background_rca_email(
                 task_id=record.task_id,
-                command=record.command,
-                root_cause=record.root_cause,
-                top_analysis=record.top_analysis,
-                next_steps=record.next_steps,
+                command=command,
+                root_cause=root_cause,
+                top_analysis=top_analysis,
+                next_steps=next_steps,
                 stats=record.stats,
             )
             ok, error = send_telegram_report(
@@ -50,7 +85,12 @@ def deliver_background_notifications(
                 {"bot_token": creds.bot_token, "chat_id": creds.chat_id},
                 parse_mode="",
             )
-            results["telegram"] = "sent" if ok else f"failed: {error}"
+            # The bot token travels in the request URL, and the transport surfaces a
+            # non-JSON error body verbatim, so redact before this reaches the record
+            # and `/background show`.
+            results["telegram"] = (
+                "sent" if ok else f"failed: {redact_token(error, creds.bot_token)}"
+            )
             continue
 
         if channel != "email":
--- a/surfaces/interactive_shell/runtime/background/notifications.py
+++ b/surfaces/interactive_shell/runtime/background/notifications.py
@@ -2,43 +2,9 @@
 
 from __future__ import annotations
 
-from platform.common.errors import OpenSREError
 from surfaces.interactive_shell.session.background_investigations import (
     BackgroundInvestigationRecord,
 )
-
-# Telegram caps a message at 4096 characters and the transport tail-truncates to fit.
-# The RCA body ends with "What to do next" and the stats block, so an unbounded root
-# cause would push exactly the actionable sections off the end. Budget each section
-# instead: the worst case below stays under the cap, so the tail always survives.
-_TELEGRAM_COMMAND_CHARS = 200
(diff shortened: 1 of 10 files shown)


muddlebee:
@Harshvardhan86 we will need a demo for this, pls go through opensre docs on how to use the opensre commands

Harshvardhan86:
Live demo evidence: terminal session (video) and the RCA summary arriving in Telegram (screenshot).

https://github.com/user-attachments/assets/d78a7c3c-eb32-4e88-9f59-50286d16b0fb

telegram-rca-screenshot

muddlebee:
@Harshvardhan86 great work :clap: kudos

Harshvardhan86:
@muddlebee demo attached above — recorded against the real Telegram Bot API, not a mock. The video shows `/background notify set telegram` → `/investigate` → completion, and the screenshot is the RCA landing in Telegram.

Your docs note sent me back through the pages, and it was a good call — the feature shipped undocumented, and it made the existing docs wrong. Fixed in `b81ad5c9`:

- **`background-investigations`** — said *"notification preferences are email-only in the current shipped path"*. Commands are now a quick-reference table, and the flow now states that channels are **empty by default**, so `/background on` alone delivers nothing (the page never said so). Also removed an invalid `/investigate kubernetes-high-cpu` template.
- **`messaging/telegram`** — background RCA was missing from both delivery-surface enumerations; added, with a worked end-to-end subsection.
- **`interactive-shell-commands`** — `/background` was entirely absent from the "complete reference for every slash command".
- **`smtp`** — Step 3 promised an RCA email that never arrives, because channels default to empty.

The same pass turned up two bugs in my own branch, both now fixed, each with a regression test that fails without the fix:

- **Bot token could reach `/background show` unredacted.** It rides in the request URL, and the transport surfaces a non-JSON error body verbatim — so a proxy error page echoing the URL would put a live token into the background record.
- **Telegram delivery silently discarded the actionable half of the RCA.** The body is built root-cause-first and ends with "What to do next" + the stats block, while `send_telegram_report` tail-truncates at 4096. A realistic record built a **23,631-character body, so ~83% of it was dropped** — the email recipient got the full report, the on-call reading Telegram got a truncated root cause and nothing to act on. Sections are now budgeted; email is unchanged.

`make lint && make format-check && make typecheck && make test-scope` all green (10,887 passed, 27 skipped).

@greptile review

surfaces/interactive_shell/runtime/background/notifications.py
```diff
@@ -23,6 +57,42 @@ def deliver_background_notifications(
     effective_integrations = resolve_effective_integrations()
 
     for channel in channels:
+        if channel == "telegram":
```
muddlebee, line 26: better we use DRY principle and create a separate class/file for telegram related code, could you check?

surfaces/interactive_shell/runtime/background/notifications.py
```diff
@@ -2,10 +2,44 @@
...
+# The RCA body ends with "What to do next" and the stats block, so an unbounded root
+# cause would push exactly the actionable sections off the end. Budget each section
+# instead: the worst case below stays under the cap, so the tail always survives.
+_TELEGRAM_COMMAND_CHARS = 200
+_TELEGRAM_ROOT_CAUSE_CHARS = 1000
+_TELEGRAM_ITEM_CHARS = 240
+_TELEGRAM_MAX_ITEMS = 5
+
+
+def _telegram_summary_sections(
```
muddlebee: same here, need refactor and use DRY

muddlebee:
@Harshvardhan86 a few nits, rest looks good

muddlebee:
@Harshvardhan86 im pushing some commits rn for the reviews above

muddlebee:
@Harshvardhan86 nice work :+1:

Issue #1359, opened 2026-05-06
CI Speed: Persist reusable uv environment cache across jobs


davincios:
## Objective
Contribute to the **50% collective CI speed improvement** goal with at least **15% speed improvement** by reducing repeated dependency setup time.

## Problem
Multiple jobs each perform `uv sync --frozen --extra dev`, repeating environment resolution/install work in separate jobs.

## Why this is high leverage
- Pydantic and FastAPI optimize with aggressive uv caching and `uv run --no-sync` in execution steps.
- Repeated dependency syncs across jobs are a recurring latency source.

## Proposed change
- Introduce a shared environment bootstrap strategy:
  - Cache `.venv` (or equivalent uv-managed environment artifacts) keyed by OS + Python + lock hash.
  - Use `uv run --no-sync` in downstream steps after validated bootstrap.
- Optionally build a reusable composite action for consistent setup.

## Acceptance criteria
- At least 2 core jobs reuse prebuilt env/cache without full re-sync.
- Cache hit ratio >=70% on warm runs.
- Job setup time reduction yields >=15% PR median CI improvement contribution.

## Measurement plan
- Capture per-job setup step timings before/after.
- Track cache hit/miss over 20+ runs.

## References
- Pydantic CI uv optimization (`setup-uv`, `uv run --no-sync`): https://github.com/pydantic/pydantic/blob/main/.github/workflows/ci.yml
- FastAPI uv cached setup pattern: https://github.com/fastapi/fastapi/blob/master/.github/workflows/test.yml

muddlebee:
@Davidson3556 u can pick this up

Davidson3556:
okay

davincios:
please redistribute this task to the community @devankitjuneja

mohamedaminehamdi:
@muddlebee can i work on this or is it already assigned

saurabhsingh121:
@Davidson3556 @muddlebee @davincios 
I'd like to work on this.

**Problem (confirmed on `main`):** `ci.yml` runs `uv sync --frozen --extra dev` independently in every job (quality + 5 test shards + coverage-report + test-kubernetes + test-thorough). `setup-uv` `enable-cache: true` caches uv's package cache, but each job still rebuilds/links `.venv` from scratch.

**Proposed approach:**

1. **Add a composite action** `.github/actions/setup-opensre-env/action.yml` that centralizes:
   - `actions/checkout` (optional — caller may already checkout)
[...]

PR #4005, opened 2026-07-14
Cache uv virtualenv across CI jobs to cut redundant dependency syncs

abuXsarkar:
-  then runs uv sync --frozen --extra dev once against that cache.
- Replaced the repeated three-step setup in all 6 jobs of ci.yml with a single uses: ./.github/actions/setup-python-env step.
- Switched the downstream uv run invocations in those jobs to uv run --no-sync, since the environment is already validated by the composite action and each step no longer needs to re-check the lock file.

Net effect: on a warm cache (unchanged uv.lock), every job restores its virtualenv from cache instead of reinstalling dependencies, and per-step uv run calls skip the sync check — addressing the ≥15% CI time reduction and ≥70% warm cache-hit target described in #1359.

Both changed YAML files were validated locally with python3 -c "import yaml; yaml.safe_load(...)" (parses cleanly). I have not yet observed a live GitHub Actions run of this workflow (no CI trigger available outside of opening this PR) — first-run cache-miss timing and warm-run cache-hit timing should be checked from the Actions tab once this is up, and I'm happy to iterate on the cache key or restore-keys if the hit rate doesn't land in range.
⸻
Code Understanding and AI Usage

Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?

If you used AI assistance:

I worked through this with Claude (Anthropic's coding assistant): I identified issue #1359 and directed the approach, and the diff was written and reviewed jointly — Claude drafted the composite action and workflow edits, and I reviewed the resulting diff line by line before it went up.

- Problem: every CI job independently re-ran uv's full dependency sync from scratch, which is the single biggest duplicated cost across the 6 jobs in ci.yml (they all install the identical uv.lock-pinned environment).
- Alternatives considered: relying only on setup-uv's built-in enable-cache (already present) — but that only caches uv's package/wheel cache, not the assembled .venv, so every job still pays the cost of rebuilding the virtualenv from those cached wheels. Caching .venv directly, keyed on lock-file hash, skips that rebuild entirely on a hit.
- Why this implementation: a composite action was the natural way to de-duplicate the identical 3-step setup block that was copy-pasted across all 6 jobs, and it's the natural place to add the actions/cache step alongside it. uv run --no-sync on the per-step commands avoids uv re-checking the lock file on every single invocation within a job, now that the composite action has already validated the env once.
- Testing status: YAML validity is confirmed locally; the actual cache hit-rate and timing improvement can only be observed once this runs in GitHub Actions, since there's no local equivalent of the hosted runner cache. I'll be watching the first few CI runs on this PR to confirm the cache behaves as expected and will adjust (e.g. cache key granularity, restore-keys fallback) if not.
⸻
- [x] I understand why my changes work; full validation is limited to YAML parsing locally, real cache behavior will be confirmed once Actions runs this PR (see note above)
⸻
Note: Please check Allow edits from maintainers if you would like us to assist in the PR.

### DIFF
--- a/.github/actions/setup-python-env/action.yml
+++ b/.github/actions/setup-python-env/action.yml
@@ -1,0 +1,32 @@
+name: Setup Python environment
+description: >-
+  Install uv, set up Python, and restore/create a cached uv-managed
+  virtualenv keyed by OS + Python version + uv.lock hash, so jobs stop
+  re-running a full ``uv sync`` from a cold environment on every run.
+
+runs:
+  using: composite
+  steps:
+    - name: Install uv
+      uses: astral-sh/setup-uv@v8.1.0
+      with:
+        enable-cache: true
+        cache-dependency-glob: |
+          pyproject.toml
+          uv.lock
+
+    - name: Set up Python
+      uses: actions/setup-python@v6
+      with:
+        python-version: "3.13"
+
+    - name: Cache virtualenv
+      id: venv-cache
+      uses: actions/cache@v4
+      with:
+        path: .venv
+        key: venv-${{ runner.os }}-py3.13-${{ hashFiles('uv.lock') }}
+
+    - name: Install dependencies
+      shell: bash
+      run: uv sync --frozen --extra dev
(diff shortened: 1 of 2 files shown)


muddlebee:
not planned

PR #3965, opened 2026-07-12
feat(tools): add model_spend_tool for spend and burn-rate estimation (3231-model-spend-tool)

SawhneySatvik:
Adds `tools/system/model_spend_tool` with two agent/chat tools that estimate AI cost from token/model/lap metadata, reusing the fleet-monitoring price table:

- **`estimate_model_spend`**: total USD from per-model token usage, broken down by model, by token bucket, and by cost-per-investigation-lap.
- **`summarize_model_burn_rate`**: projected `$/hr` from a per-minute sample, or from run totals plus elapsed seconds.

Pure calculators: estimates come only from tokens × the local price table, so there is no live billing-API call and unknown models are surfaced as **unpriced** (never silently costed at zero). Per-model price overrides are supported for custom or unlisted models.

Registers the package in `tools/system`, classifies both tools in the telemetry error-coverage allowlist (they let exceptions escape to the global wrapper), adds `docs/model_spend_tool.mdx` to the nav, and covers pricing, unknown-model fallback, bucket-sum, per-lap, burn-rate, overrides, and malformed-input handling in tests.

Fixes #3231

#### Describe the changes you have made in this PR -

New leaf tool package `tools/system/model_spend_tool/` (function-based, mirroring `sre_guidance_tool`):

- `estimation.py` — pure calculation: `estimate_spend`, `summarize_burn_rate`, and helpers.
- `__init__.py` — the two `@tool(source="knowledge", surfaces=("investigation", "chat"))` entrypoints.

| Tool | What it does |
|------|--------------|
| `estimate_model_spend` | Total USD from per-model token usage → `by_model`, per-`cost_by_bucket`, and `usd_per_lap`. |
| `summarize_model_burn_rate` | Projects `$/hr` from a per-minute sample, or from run totals + `elapsed_seconds`. |

Design notes:
- **Reuses the existing price table** (`tools/system/fleet_monitoring/pricing.py`: `usd_for_usage`, `normalize_model_name`, `PriceOverride`, `RATES_VERIFIED_AT`) and `meters.TokenUsage` — **no rate math or dependency duplication**.
- **Token buckets, not a scalar** — input/output/cache buckets are priced separately (output's higher rate is respected). Per-bucket USD is derived by isolating each bucket through the shared pricer, so the buckets provably **sum to the model total** (tested).
- **Unknown models → `priced: false` + `unpriced_models`**, never costed at `$0`.
- **`laps`** → `usd_per_lap = total_usd / laps` (guarded for `laps > 0`).
[...]

PR #4047, opened 2026-07-14
feat(fleet_monitoring): migrate model pricing to litellm's price table

### LINKED ISSUE #4035: Migrate model pricing to litellm's cost table, reintroduce model_spend_tool
## Problem

#3965 added `model_spend_tool` (estimate spend / burn-rate from token usage) on top of a
hand-vendored price table in `tools/system/fleet_monitoring/pricing.py` (`MODEL_PRICES`,
~80 entries, one per model). Review blocked it: maintaining accurate $$-per-token config for
every model/provider by hand doesn't scale and is easy to get subtly wrong across providers
(see discussion on #3965).

## Proposal

Migrate `pricing.py`'s internals to `litellm.cost_per_token()`'s pricing table instead of a
hand-vendored dict, then reintroduce `model_spend_tool` on top of the migrated module. Findings
from investigating the swap:

- **No new dependency** — `litellm>=1.90.0` is already in `pyproject.toml` and used in
  `core/llm/transports/litellm/clients.py`.
- litellm ships a bundled `model_prices_and_context_window_backup.json` with ~2,800 models,
  including current rates for `claude-sonnet-4-5-20250929`, `claude-opus-4-7`, `claude-fable-5`
  that match what's hand-vendored today.
- `litellm.cost_per_token()` **raises** on an unmapped model rather than silently returning `$0`
  — already satisfies this repo's "never silently cost at zero" invariant; we just catch it.
- **Coverage gap confirmed**: litellm's table is missing `gpt-5.6-{sol,terra,luna}` (GA'd
  2026-07-09, very recent) and only has bare `claude-sonnet-4`/`claude-opus-4` under
  Bedrock-prefixed keys, not the plain Anthropic-API id. litellm is not a strict superset of the
  vendored table — plan is a small local-override escape hatch for confirmed gaps, not a full
  per-model config file.
- **Import-time risk**: `litellm.model_cost` is populated via a live HTTP fetch at import unless
  `LITELLM_LOCAL_MODEL_COST_MAP=True` is set. `pricing.py` is imported unconditionally by the
  fleet-monitoring dashboard's always-on per-tick sampler (`sampler.py`), so this needs to be
  forced local-only to stay a pure offline dict lookup.
- **Packaging note**: `tests/packaging/test_litellm_bundle_contract.py` already documents that
  frozen release builds must bundle litellm's price JSON when litellm is imported. Migrating
  `pricing.py` makes that import unconditional (today it's gated behind
  `OPENSRE_LLM_TRANSPORT=litellm`/Azure) — needs re-verification against the frozen build.

## Scope

- Migrate `tools/system/fleet_monitoring/pricing.py` internals to litellm's price table, keeping
  its public API (`usd_for_usage`, `usd_per_hour_for_usage`, `PriceOverride`,
  `normalize_model_name`, `RATES_VERIFIED_AT`) stable so `sampler.py` needs no changes.
- Reintroduce `tools/system/model_spend_tool/` (`estimate_model_spend`,
  `summarize_model_burn_rate`) from #3965 on top of the migrated module.

Follow-up to #3231 and the review discussion on #3965.

muddlebee:
Fixes #4035

Migrates `tools/system/fleet_monitoring/pricing.py` off the hand-vendored `MODEL_PRICES` dict (~80 entries, one per model — the thing that got #3965 blocked in review) onto litellm's bundled community-maintained price table. `litellm>=1.90.0` is already a project dependency; no new dependency added.

- Reads litellm's *local* price snapshot directly (`GetModelCostMap.load_local_model_cost_map()`) instead of the shared `litellm.model_cost` global. That global is populated from a **live network fetch** on whichever import touches it first in the process — confirmed in practice during this change that a live fetch elsewhere had already picked up a same-week model release our local snapshot didn't have yet, which made pricing nondeterministic based on unrelated import order. This module is imported unconditionally by the always-on dashboard sampler, so pricing lookups need to stay a pure, deterministic offline dict read.
- A small local override table (`_LOCAL_MODEL_PRICES`) covers the two confirmed gaps in litellm's snapshot: the GPT-5.6 family (GA'd 2026-07-09, too new for the bundled snapshot) and the retired `claude-3-5-sonnet-20241022` direct-API id (litellm only keeps that generation under Bedrock-routed keys now).
- A narrow, last-resort fallback (`_COMPAT_PROVIDER_PREFIXES`) handles Groq and MiniMax specifically: `core/llm/providers/openai_compat_providers.py` reports these providers' models as **bare** names (no litellm-style `<provider>/` prefix — that's a litellm-only convention), but litellm keys them under that prefix regardless. Scoped to just these two providers deliberately — guessing a provider prefix for an arbitrary open-weight model would risk silently applying a different host's price for the same model.
- Case-insensitive lookup index, since litellm's own keys aren't uniformly lowercase (MiniMax entries are `minimax/MiniMax-M2.1`).

Public API of `pricing.py` (`usd_for_usage`, `usd_per_hour_for_usage`, `usd_per_token_blended`, `usd_per_hour`, `normalize_model_name`, `PriceOverride`, `RATES_VERIFIED_AT`) is unchanged, so `sampler.py` (the one production caller) needed zero changes.

**Compatibility verified against every provider in `config/llm_auth/provider_catalog.py`**, using each provider's actual wire-format model string (not a guessed one):

| Provider | Result |
|---|---|
| Anthropic, OpenAI, Azure OpenAI (deployment named after its model), Bedrock (cross-region Claude), DeepSeek, Gemini, OpenRouter, Groq, MiniMax | ✅ priced |
| Claude Code / Codex / Gemini / Copilot CLIs | ✅ priced (they report the underlying model directly) |
| NVIDIA NIM | ❌ unpriced — litellm has zero NIM coverage today; needs a `PriceOverride`/`agents.yaml` entry |
| Ollama (self-hosted) | ❌ unpriced — expected: no per-token API price exists to look up; this is the compute-basis ($/GPU-hr) case from the #3965 review discussion, not something a price table can answer |

Both unpriced cases correctly resolve to `None` (dashboard renders `-`), never a guessed rate.

```
$ uv run python -m pytest tests/fleet_monitoring/test_pricing.py -q
======================= 58 passed in 4.55s =======================

$ uv run python3 -c "
from tools.system.fleet_monitoring.pricing import usd_per_token_blended
for m in ('claude-sonnet-4-5', 'gpt-5-codex', 'gemini-2.5-pro', 'deepseek-chat',
          'llama-3.3-70b-versatile', 'MiniMax-M2.1', 'meta/llama-3.1-70b-instruct'):
    print(m, '->', usd_per_token_blended(m))
"
claude-sonnet-4-5 -> 6.6e-06
gpt-5-codex -> 3.775e-06
gemini-2.5-pro -> 3.5e-06
deepseek-chat -> 3.78e-07
llama-3.3-70b-versatile -> 3.99e-07
MiniMax-M2.1 -> 5.6e-07
meta/llama-3.1-70b-instruct -> None   # NVIDIA NIM: unpriced, not guessed
```

`make lint` · `make format-check` · `make typecheck` (1220 files) · `make test-scope` (755 passed, 3 skipped) all green locally.

`model_spend_tool` (the tool from #3965 that sits on top of this) is a deliberate follow-up PR once this lands, to keep this diff reviewable on its own.

---

The problem: `MODEL_PRICES` required a manual code change for every model, across every provider, forever — the exact thing #3965's review blocked on ("how are you ensuring the LLM cost and $$ is consistent and work across all models? this is a difficult task"). litellm already ships and maintains a ~2.8k-model price table as a project dependency, so the fix is to read from it instead of hand-duplicating it.

The one non-obvious design decision is bypassing `litellm.model_cost` (the module-level global litellm itself populates) in favor of reading `GetModelCostMap.load_local_model_cost_map()` directly. I initially wired to the global and two tests failed under pytest but passed in isolation — traced it to litellm doing a live network fetch on whichever import touches it first in the process (not necessarily this module), which had already picked up a newer upstream table than our local snapshot. Since the dashboard sampler imports this module unconditionally and needs deterministic offline behavior, reading the local snapshot directly (independent of what else in the process has imported litellm) was the correct fix, not a workaround.

The other non-obvious piece is the provider compatibility sweep: I didn't assume litellm's coverage was uniform — I checked `core/llm/providers/openai_compat_providers.py` to find the *actual* wire-format model string per provider (several route through a generic OpenAI-compatible client using bare model names, not litellm's `provider/model` convention) and tested each one for real. That's what caught the MiniMax case-sensitivity bug and the Groq/MiniMax bare-name mismatch, both now covered by regression tests, plus honestly documented the NVIDIA NIM gap rather than silently leaving it broken.

---

---

### DIFF
--- a/tests/fleet_monitoring/test_pricing.py
+++ b/tests/fleet_monitoring/test_pricing.py
@@ -70,6 +70,15 @@ class TestUsdPerTokenBlended:
         expected = 0.7 * entry["input_cost_per_token"] + 0.3 * entry["output_cost_per_token"]
         assert usd_per_token_blended("claude-sonnet-4-5") == pytest.approx(expected)
 
+    def test_local_snapshot_loads_a_substantial_table(self) -> None:
+        # Startup-time smoke guard: if litellm's packaged JSON file is ever
+        # missing/unreadable, _litellm_local_cost_map degrades to an empty
+        # dict rather than crashing the always-on sampler (see its
+        # docstring) — but that failure mode must be loud in CI, not silent.
+        from tools.system.fleet_monitoring.pricing import _litellm_local_cost_map
+
+        assert len(_litellm_local_cost_map()) > 1000
+
 
 class TestGpt56Family:
     """GPT-5.6 Sol / Terra / Luna (#3931) — the confirmed litellm gap.
--- a/tests/fleet_monitoring/test_pricing.py
+++ b/tests/fleet_monitoring/test_pricing.py
@@ -1,4 +1,4 @@
-"""Tests for the model pricing table and $/hr computation (#2023)."""
+"""Tests for the model pricing table and $/hr computation (#2023, #4035)."""
 
 from __future__ import annotations
 
@@ -8,7 +8,6 @@ import pytest
 
 from tools.system.fleet_monitoring.meters import TokenUsage
 from tools.system.fleet_monitoring.pricing import (
-    MODEL_PRICES,
     PriceOverride,
     normalize_model_name,
     usd_for_usage,
@@ -20,7 +19,8 @@ from tools.system.fleet_monitoring.pricing import (
 
 class TestUsdPerTokenBlended:
     def test_known_claude_model(self) -> None:
-        # claude-sonnet-4-5: 3 USD/M input, 15 USD/M output, 70/30 blend.
+        # claude-sonnet-4-5 (litellm's model_cost table): 3 USD/M input,
+        # 15 USD/M output, 70/30 blend.
         # Expected: 0.7 * 3e-6 + 0.3 * 15e-6 = 2.1e-6 + 4.5e-6 = 6.6e-6.
         rate = usd_per_token_blended("claude-sonnet-4-5")
         assert rate is not None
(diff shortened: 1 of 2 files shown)


muddlebee:
@greptile review

muddlebee:
@greptile review

PR #4055, opened 2026-07-15
feat(wizard): validate the LLM credential during onboarding, with failure recovery

### LINKED ISSUE #3591: Better LLM key failure UX during onboarding
| Field | Value |
|-------|-------|
| **Level** | Middle |
| **Complexity** | M |
| **Tier** | Tier 3 — Middle |
| **Category** | product |

**Summary:** Improve the product/user experience described in the linked GitHub issue: Better LLM key failure UX during onboarding.

**Steps:** (1) Read the linked issue. (2) Keep the UX copy/flow change narrow. (3) Add tests/snapshots if the surface supports it. (4) Run relevant CLI/onboarding tests.

**Done when:** [ ] UX improvement implemented [ ] Behavior is covered or manually verified [ ] CI green

**Files:** See linked GitHub issue.

---

Harshvardhan86:
Fixes #3591

The onboarding wizard was the one credential-entry surface in the product that saved its credential **without validating it**. Every other credential in the same wizard (Dagster, Slack, GitLab, OpenSearch, Telegram) already prompts → validates → retries on failure — the LLM key skipped that convention, so a typo'd key sailed through onboarding and only surfaced later, deep inside an investigation. If the keyring write failed, the wizard hard-exited with no way to recover.

This PR brings the LLM credential up to the same standard:

- **Validate against the model you actually picked.** Model selection is now hoisted above the key prompt in both the change-provider and saved-provider branches, so the live probe exercises the model that will be persisted — not the provider default. (Validating the default would, for example, tell every Ollama user to pull a model they never selected.)
- **A shared recovery menu.** On a validation failure the user gets one menu — **Re-enter the key** (default) / **Save anyway without validating** / **Pick a different provider** — and the same menu shape now handles a keyring-persist failure (**Retry saving** / **Continue for this session only** / **Pick a different provider**). One helper, both failure sites; the three previous bare `return 1` exits are gone.
- **No lockout.** Validation collapses a bad key, a dead network, a corporate proxy, a rate limit, and a capped account into the same failure, so a hard gate would lock out a user with a *valid* key. "Save anyway" is always available (with an explicit unverified warning), and "Continue for this session only" exports the key to the environment for the current run without ever writing it to `.env`.
- **An honest summary.** The final screen reports where the credential actually went: `system keychain`, `system keychain (unverified)`, `project .env`, or `not saved — re-enter next run`.

Two latent bugs that gating-on-validation would have weaponized are fixed in the same change: a valid **MiniMax** key was being sent to `api.openai.com` (its base URL was never wired in), and pressing **Enter** at the **Azure** key prompt silently persisted the placeholder endpoint URL as the API key. A typo'd Azure endpoint is now re-promptable on retry/repick instead of being an unrecoverable dead-end.

Tests: a new autouse socket guard makes any real network call in a wizard test a hard failure (the validator caches SDK classes in module globals, so a mis-patched test could otherwise hit `api.anthropic.com` with a live key and pass for the wrong reason). Every new behaviour ships with a test written to fail first. `make lint && make format-check && make typecheck && make test-scope` all green (857 passed, 3 skipped); the wizard suite is 224 passing.

**Known follow-up (not in this PR):** after 10 consecutive failed retries the wizard returns exit 1 instead of dropping back to the provider menu — bounded, since the escape hatches are offered every round. **Out of scope:** the two onboarding GIFs in `docs/quickstart.mdx` record the old key→model order and are now slightly stale (the prose is order-agnostic); validating an *already-stored* key that is never re-prompted; and a pre-existing double-Ctrl-C-exits-0 behaviour.

▶️ **A ~40s terminal recording of the wizard is attached below**, plus a screenshot of the recovery menu. It shows a fresh onboarding: the model is chosen first, a bad key is rejected live (`✗ Anthropic · claude-opus-4-7 · Failed`), the recovery menu appears, "Save anyway" completes onboarding, and the summary honestly reports `unverified`.

---

The problem is that the wizard persisted the LLM API key straight to the keyring with no live check, while every other integration in the same wizard already validated-then-retried — so a bad key only failed much later. My approach was to reuse that existing pattern (`validate_provider_credentials` already existed and is used by `opensre auth`) rather than invent anything new.

The one non-obvious decision was **ordering**: the key was prompted before the model was chosen, so a naive "validate before save" would have probed the provider's default model. That breaks Ollama (fails on a model the user never pulled), so I hoisted model selection above the credential block in both branches. I considered hard-gating onboarding on a passing validation, but rejected it because the validator can't distinguish a bad key from a dead network / proxy / rate-limit / capped account — a hard gate would lock out real users, so "Save anyway" is mandatory and "Re-enter" is the default.

Key pieces: `_prompt_validated_llm_credential` (the validate-then-retry loop), `_recovery_action` (one shared 3-verb menu used by both the validation-failure and keyring-persist-failure sites), and `_persist_llm_credential_with_recovery` (replaces the three bare hard-exits). The `continue_unsaved` path exports to `os.environ` only and is re-applied after `sync_provider_env` so the in-process shell handoff can read it, but is never written to `.env`.

---

---


Harshvardhan86:
opensre-3591-recovery-menu

https://github.com/user-attachments/assets/b62633b6-6ca1-4dc6-bd06-1bfe05c7aefe

Demo — a ~40s terminal recording of the wizard on this branch. Fresh onboarding: the model is chosen first, a typo'd key is rejected live (✗ Anthropic · claude-opus-4-7 · Failed — the failure names the model we validated against), the recovery menu appears, "Save anyway" completes onboarding, and the summary honestly reports unverified. Recorded against the real Anthropic API (a bad key returns a genuine 401).

Harshvardhan86:
@greptile review

Harshvardhan86:
@greptile review

Fixed the P1 in `61b8c86d`: the CLI-repick smoke test now uses the model-before-key order and the correct `could not be verified. What next?` prompt, matching `test_onboard_interactive_smoke`.

Devesh36:
@greptile review again

muddlebee:
@Harshvardhan86 nice work, will take a deeper look later :+1:

surfaces/cli/wizard/flow.py
```diff
@@ -239,6 +275,257 @@ def _persist_llm_credential(provider: ProviderOption, value: str) -> bool:
...
+#: onboarding menus so "pick a different LLM provider" reads the same everywhere, and
+#: it drives the existing ``force_repick`` mechanism in ``run_wizard``.
+_REPICK_CHOICE = Choice(
+    value="repick",
+    label="Pick a different LLM provider",
+    hint="Returns to the provider and model picks",
+)
+
+
+def _recovery_action(*, prompt: str, retry_label: str, retry_hint: str, escape: Choice) -> str:
```
muddlebee: all such logic should be extracted to a different file? could you check where it fits appropriately, create new file if necessary

surfaces/cli/wizard/flow.py
```diff
@@ -239,6 +275,257 @@ def _persist_llm_credential(provider: ProviderOption, value: str) -> bool:
...
+            ],
+            default="retry",
+        )
+    except KeyboardInterrupt:
+        _console.print(f"\n[{WARNING}]Setup cancelled.[/]")
+        return "cancel"
+
+
+def _render_credential_validation(label: str, model: str, result: ValidationResult) -> None:
+    """Render the live-probe outcome, naming the provider *and* the probed model.
```
muddlebee: same for all such functions in this class, cehck once

surfaces/cli/wizard/flow.py
```diff
@@ -892,6 +1143,72 @@ def run_wizard(_argv: list[str] | None = None) -> int:
...
 
+        if change_provider:
+            if auth_method == API_KEY_AUTH_METHOD and provider.credential_kind not in (
+                "cli",
+                "none",
+            ):
+                credential_outcome = _prompt_validated_llm_credential(
+                    provider, model=model, session_env_sink=session_env_sink
+                )
+                if credential_outcome == "cancel":
```
muddlebee: use constants pls for such keyworks like "cancel", "repick" and re-use acorss this PR

surfaces/cli/wizard/flow.py
```diff
@@ -892,6 +1143,72 @@ def run_wizard(_argv: list[str] | None = None) -> int:
...
+                legacy_api_key = str(defaults["legacy_api_key"] or "").strip()
+                # A ``host`` credential (e.g. the Ollama host) is not a secret api key: never
+                # migrate a stale legacy ``api_key`` value into it — that would leak a
+                # secret-shaped value into .env and point the runtime at a bogus host. Fall
+                # through to the host prompt instead (#3291).
+                if not has_api_key and legacy_api_key and provider.credential_kind != "host":
+                    migration_outcome = _persist_llm_credential_with_recovery(
+                        provider, legacy_api_key, session_env_sink=session_env_sink
+                    )
+                    if migration_outcome == "cancel":
```
muddlebee: same

cerencamkiran:
The recovery flow handles keyring failures, but I think there is still a gap for host-based credentials. _persist_llm_credential() calls sync_env_values() for values like OLLAMA_HOST without handling write errors. If writing to .env fails, the user may get an exception instead of the retry menu. Could this use the same recovery flow? It would also be good to add a test for a .env write failure.

surfaces/cli/wizard/flow.py
Harshvardhan86: Good call — moved the whole LLM-credential cluster out of `flow.py` into a new `surfaces/cli/wizard/llm_credential.py` in `8671b4df`: the validate-then-retry prompt, the recovery menu, the render/persist/summary helpers, and the Azure endpoint helpers. `run_wizard` just imports and calls them now, so `flow.py` drops ~460 lines and the credential logic lives in one place.

surfaces/cli/wizard/flow.py
Harshvardhan86: Same — this one moved too; it's all in `llm_credential.py` now (`8671b4df`).

surfaces/cli/wizard/flow.py
Harshvardhan86: Done in `8671b4df` — the outcome keywords are now named constants in `llm_credential.py` (`RETRY` / `REPICK` / `CANCEL` / `SAVE_ANYWAY` / `CONTINUE_UNSAVED`, plus the `OK` / `UNSAVED` / `UNVERIFIED` states) and reused across the PR instead of the string literals. The values are unchanged so the menu default (`retry`) and the outcome checks still line up.

surfaces/cli/wizard/flow.py
Harshvardhan86: Same — covered by the constants in `8671b4df`.

surfaces/cli/wizard/flow.py
```diff
@@ -892,6 +1143,72 @@ def run_wizard(_argv: list[str] | None = None) -> int:
...
+                if credential_outcome == "unverified":
+                    credential_state = "unverified"
+                elif credential_outcome == "unsaved":
+                    credential_state = "unsaved"
+                azure_env = _ensure_azure_openai_endpoint_settings(provider)
+                if azure_env is None:
+                    force_repick = True
+                    continue
+                provider_extra_env = azure_env
+                os.environ.update(azure_env)
```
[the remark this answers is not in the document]
Harshvardhan86: Intentional, but I added a comment in `8671b4df` to make it clear: the inner call inside `_prompt_validated_llm_credential` sets `AZURE_OPENAI_BASE_URL` so the live probe can read it, and this outer call repopulates `provider_extra_env` for `sync_provider_env` — it short-circuits on the already-set endpoint rather than re-prompting.

Harshvardhan86:
@cerencamkiran nice catch — you're right, the host path (`OLLAMA_HOST` → `.env`) could raise instead of showing the menu if the `.env` write failed. Fixed in `8671b4df`: the host branch now catches the write error, prints a `.env`-path warning, and returns into the same recovery menu (Retry saving / Continue for this session only / Pick a different provider) rather than letting the exception escape. Added a test that forces a `.env` write failure (`test_run_wizard_ollama_host_env_write_failure_offers_recovery_menu`). Thanks for spotting it!

Harshvardhan86:
@greptile review

Addressed the review in `8671b4df`: extracted the LLM-credential helpers into `surfaces/cli/wizard/llm_credential.py`, replaced the outcome-string literals with named constants, and fixed the host `.env`-write path to fall into the recovery menu (with a test). Details in the inline replies above.

cerencamkiran:
Could you please resolve the conflicts? Thanks!

Harshvardhan86:
@cerencamkiran done — merged latest `main` in and resolved the conflict (it was a single import line in `test_flow.py` after your #4069 configurators split; `flow.py` auto-merged). Wizard suite is green locally. Thanks for the nudge!

Harshvardhan86:
@muddlebee — addressed your review in `8671b4df` (pulled the credential helpers into `llm_credential.py`, and the outcome keywords are constants now), and just merged latest `main` to clear the conflict. Greptile's back to 5/5. Ready for another look whenever you have time — no rush 🙏

muddlebee:
@Harshvardhan86 thank you I will test this locally :)

Harshvardhan86:
Merged latest `main` and reconciled with #4117 — `llm_credential.py` now delegates Azure endpoint + model selection to the new `azure_openai.py` module (so deployment discovery is preserved), and I dropped my duplicated Azure helpers. For Azure the deployment pick happens after the endpoint + key are entered (discovery needs them) but still before validation, so the "validate the model you picked" behaviour holds. Wizard suite green locally; conflicts resolved.

Harshvardhan86:
@greptile review

Re-running on the reconcile merge (`d1b27b57`) — merged latest main and integrated #4117's new `azure_openai.py` (llm_credential.py now delegates Azure endpoint + deployment discovery to it).

Harshvardhan86:
@muddlebee — this one's been ready for a while (Greptile 5/5, all your review comments addressed) but it keeps conflicting with the ongoing wizard refactors — I've now rebased it three times, latest on #4167 (`llm_credential.py` delegates to your `azure_openai.py`, and I picked up the shared `.env`/keyring writer). Could you give it a merge when you get a window, or let me know if anything's blocking? Happy to adjust. 🙏

muddlebee:
@Harshvardhan86 thank you for your efforts, we would need reviews from core team on this as its a big PR.

VaibhavUpreti (approved):
LGTM — wizard now validates the chosen model’s credential with a proper recovery path (re-enter / save anyway / repick), and the llm_credential extraction addresses earlier review notes. CI green.

PR #4063, opened 2026-07-15
fix(slack-gateway): surface failed turns + close #4036 sub-tasks

### LINKED ISSUE #4036: Slack gateway: errored turn leaves placeholder stuck on "Digging in…" (no answer, no error message)
## Summary
When a Slack gateway turn raises, the user is left with a frozen `⚡ Digging in…` placeholder and no answer — only a small ❌ reaction. A failed turn is visually indistinguishable from one still running. Reproduced today in #devs (two `@opensre-dev-agent summarize …` requests each posted only the placeholder).

## Root cause
`gateway/slack/socket_mode_worker.py` → `_SlackTurnDispatcher._run_turn`: the `try/except` around `self._handler(...)` only reacts and re-raises:

```python
except Exception:
    mark_turn_failed(self._messaging, channel=inbound.channel_id, timestamp=inbound.ts)
    raise
```

`mark_turn_failed` merely swaps 👀→❌ on the user's original message and never touches the `SlackOutputSink`, so the placeholder it posted (`⚡ Digging in…`) is never updated or replaced. The happy path is fine because `GatewayTurnHandler.__call__` always calls `sink.finalize(...)` when `not answered`, but that guard is bypassed the moment the turn raises.

## Fix (primary)
In the `except` block, finalize the sink before re-raising, e.g.:

```python
except Exception:
    sink.render_error("Something went wrong handling that — please try again.")
    mark_turn_failed(self._messaging, channel=inbound.channel_id, timestamp=inbound.ts)
    raise
```

so the placeholder becomes a visible error. Add a gateway regression test asserting a raising turn ends with the placeholder edited to an error (not left on the status string).

## Related issues found in the same review (track separately or as sub-tasks)
2. **No turn watchdog/timeout** — placeholder only advances on streamed chunks (`output_sink.stream` → `_edit_preview`); a single non-streamed answer or a hung tool/LLM call has no heartbeat and no upper time bound. Add a max-turn timeout that finalizes with a timeout message.
3. **Two competing reply paths after #4026** — the sink owns the in-thread reply, but `slack_reply_message` is registered on the `chat` surface and has `requires_approval = True`. In a non-interactive Socket Mode chat there is no approver, so an answer routed through that tool can't be delivered (duplicate or none). Scope `slack_reply_message` to other channels/threads only (or drop from the `chat` surface) and never gate the in-thread reply on approval.
4. **Empty answer → `(no response)`** — `output_sink.stream` finalizes `text or "(no response)"` and ignores `label` / `suppress_if_starts_with`.
5. **Events with a `subtype` are silently dropped** — `gateway/slack/events.py` `parse_events_api_payload` returns `None` for any event carrying a subtype (e.g. `file_share`, `thread_broadcast`), so those mentions look unanswered.

## Acceptance criteria
- [ ] Errored Slack gateway turn replaces the placeholder with a visible error message.
- [ ] Regression test covering the raising-turn path.
- [ ] Telegram, CLI, and shell gateways verified working after the change (shared runtime).

YauhenBichel:
Fixes #4036 

Primary fix — failed turn no longer stuck on "Digging in…"

When a Slack turn raised, `_SlackTurnDispatcher._run_turn` only swapped the reaction (👀→❌) and re-raised, so the `⚡ Digging in…` placeholder was never replaced — a failed turn looked identical to a running one.

The except block now finalizes the sink before re-raising:

    except Exception:
        self._logger.exception("[slack-gateway] turn ERRORED …")
        if not timed_out.is_set():
            sink.render_error("Something went wrong on that request.")
            mark_turn_failed(…)
        raise

so the placeholder becomes a visible error. `render_error` routes through a shared redactor, so the user sees safe generic copy while the full detail is logged (no exception text leaks to the thread).

Regression test: `test_errored_turn_replaces_placeholder_with_error` — a raising handler must leave the placeholder edited to an error + ✗, not the status string.

Related sub-tasks from the issue (all addressed)

1. **No turn watchdog/timeout** — added a per-turn `threading.Timer` (default 240s, `SLACK_GATEWAY_TURN_TIMEOUT_SECONDS`). On expiry it finalizes the placeholder with a timeout message + ✗, cancelled on completion. A  blocking handler can't be force-cancelled, so the orphaned turn keeps running; if it later finishes, its own answer wins. Test: `test_turn_timeout_finalizes_placeholder_when_handler_hangs`.

2. **Two competing reply paths** — `slack_reply_message` is scoped off the `chat` surface (`surfaces=("investigation","action")`), so the approval-gated tool is never picked for the in-thread reply, which the sink owns. Extended here: `slack_send_message` and `telegram_send_message` are also dropped from `chat`, so a Slack turn can't mis-route through Telegram's send tool. Test: `test_cross_platform_send_tools_are_not_chat_surfaced`.

---

---

---


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4086, opened 2026-07-16
fix(scheduler): resolve Slack bot_token from Socket Mode setup for digest delivery

Avash-Banerjee:
Fixes #4019

Socket Mode setup stores Slack token under `bot_token`
(integrations/cli.py), but `resolve_slack_credentials` only checked
`access_token`, so digest schedule/delivery rejected valid Socket Mode
configs unless SLACK_BOT_TOKEN was also set in env.

Fix: fall back to store's `bot_token` when `access_token` lookup misses,
surfaced as `{"access_token": ...}` since delivery just needs bearer token.

Added regression test `test_from_store_bot_token_fallback` in
tests/scheduler/test_credentials.py.

### DIFF
--- a/platform/scheduler/credentials.py
+++ b/platform/scheduler/credentials.py
@@ -67,12 +67,24 @@ def resolve_slack_credentials(task_params: dict[str, str]) -> dict[str, str]:
     if webhook:
         return webhook
 
-    return _resolve_credentials(
+    access_token = _resolve_credentials(
         {},
         service="slack",
         credential_key="access_token",
         env_vars=("SLACK_BOT_TOKEN", "SLACK_ACCESS_TOKEN"),
     )
+    if access_token:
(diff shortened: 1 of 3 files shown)


cerencamkiran:
Could you also add a test to check that slack_delivery_ready() returns True when only bot_token is stored?

Avash-Banerjee:
> Could you also add a test to check that slack_delivery_ready() returns True when only bot_token is stored?

Added test_slack_ready_with_only_bot_token_stored in tests/integrations/sentry/test_digest_delivery.py, exercises slack_delivery_ready() end-to-end with only bot_token stored.

muddlebee:
@Avash-Banerjee thank you fr the PR, but needs more discussion. Please get assigned or discuss at discord before taking up PRs.

PR #4109, opened 2026-07-16
feat(deploy): tag ECR images by git SHA (+ latest) for reproducible, rollback-able deploys

YauhenBichel:

make build-image now produces an immutable sha-<git-sha> tag alongside a moving latest, from a single build — so prod deploys can pin an exact, reproducible source and roll back by redeploying a previous SHA.

Changes

- platform/deployment/aws/ecr.py — build_and_push gains an extra_tags param: one build, N tags. After building and pushing the primary tag it re-tags and pushes each extra (e.g. latest alongside the SHA tag) — no rebuild.

- platform/deployment/ecr_deploy/lifecycle.py — new _image_tag() derives sha-<short-sha> from git, marking -dirty when the working tree is uncommitted so a non-reproducible build can never masquerade as a release. build_image() now pushes the SHA tag as primary + latest as an extra, and saves the SHA-tagged URI locally for make deploy reuse.

- tests/platform/deployment/test_image_build.py (new) — covers clean-tree sha-<sha>, dirty-tree -dirty marking, and that build_image pushes the SHA tag with the latest alias.
Why: the deploy story needs a stable, source-identifiable image reference. :latest is fine for dev convenience but unsafe for prod (it moves); pinning sha-<git-sha> makes each deploy traceable to an exact commit and makes rollback a redeploy of the prior SHA. The -dirty suffix prevents an uncommitted local build from being mistaken for a reproducible one.

### DIFF
--- a/platform/deployment/aws/ecr.py
+++ b/platform/deployment/aws/ecr.py
@@ -81,8 +81,13 @@ def build_and_push(
     build_args: dict[str, str] | None = None,
     region: str = DEFAULT_REGION,
     context_dir: Path | None = None,
+    extra_tags: tuple[str, ...] = (),
 ) -> str:
-    """Build a Docker image and push it to ECR. Returns the full image URI."""
+    """Build a Docker image and push it to ECR. Returns the full image URI.
+
+    ``extra_tags`` are pushed as additional tags of the same image (one build,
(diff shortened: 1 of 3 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
I do not need that

PR #4110, opened 2026-07-16
fix(shell): record /mcp unknown subcommand as a failed turn

### LINKED ISSUE #3345: [UX] Command errors lack usage guidance — fix all edge cases (not just `/integrations show`)

AniketR10:
Fixes #3345 

`/mcp <unknown-subcommand>` printed the correct error but still recorded the turn as successful, while `/integrations <unknown-subcommand>` correctly recorded it as failed. That meant failed `/mcp` commands showed up wrong in session history and slash analytics. This change makes `/mcp` match `/integrations` — it prints through `repl_print` and marks the turn as failed — and adds a regression test covering both commands.

image

---

I first reproduce the issue by running them through the real dispatcher, and found they all already print proper guidance, only `/mcp` was inconsistent. Rather than add a new pattern, I mirrored what `_cmd_integrations` already does for the identical case: print through `repl_print` (which resets the TTY column and is safe after an inline menu) and call `session.mark_latest(ok=False, kind="slash")`. The test is parametrized over both `/integrations bogus` and `/mcp bogus` so the two handlers can't drift apart again. I confirmed the test actually catches the bug by stashing the fix and watching only the `/mcp` case fail, and verified the live output still renders correctly in a real TTY using ReplDriver.

---

---

### DIFF
--- a/surfaces/interactive_shell/command_registry/integrations.py
+++ b/surfaces/interactive_shell/command_registry/integrations.py
@@ -413,10 +413,12 @@ def _cmd_mcp(session: Session, console: Console, args: list[str]) -> bool:
     if sub == "disconnect":
         return _handle_remove(session, console, args[1] if len(args) > 1 else None)
 
-    console.print(
+    repl_print(
+        console,
         f"[{ERROR}]unknown subcommand:[/] {escape(sub)}  "
-        "(try [bold]/mcp list[/bold], [bold]/mcp connect[/bold], or [bold]/mcp disconnect[/bold])"
+        "(try [bold]/mcp list[/bold], [bold]/mcp connect[/bold], or [bold]/mcp disconnect[/bold])",
     )
+    session.mark_latest(ok=False, kind="slash")
     return True
 
 
--- a/tests/interactive_shell/command_registry/test_suggestions.py
+++ b/tests/interactive_shell/command_registry/test_suggestions.py
@@ -77,6 +77,22 @@ def test_dispatch_invalid_subcommand_is_handled_by_command_handler(
     assert resolve_literal_slash_typo("/integrations bogus", SLASH_COMMANDS) is None
 
 
+@pytest.mark.parametrize("command_line", ["/integrations bogus", "/mcp bogus"])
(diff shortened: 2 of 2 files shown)


cerencamkiran (approved):
Good work

PR #4116, opened 2026-07-17
fix(grafana): support TLS verification against internal/self-signed CAs

### LINKED ISSUE #3631: [FEATURE] Implementing new LLM provider for Azure OpenAI

muddlebee:
Fixes #3631

Grafana HTTP calls hardcoded strict TLS verification with no override. On-prem
Grafana instances signed by an internal/self-signed CA failed `opensre onboard`
with `SSLCertVerificationError: unable to get local issuer certificate`, even
when the CA was already installed in the OS trust store — Python's `requests`
never reads the OS store, only its bundled `certifi` CA list.

This mirrors the existing `verify_ssl` / `ca_bundle` pattern already used by
Splunk and Argo CD:

- `GrafanaIntegrationConfig` (stored/onboarding config) and `GrafanaAccountConfig`
  (runtime client config) gain `verify_ssl: bool = True` and `ca_bundle: str = ""`,
  plus an `ssl_verify` property (CA bundle path takes precedence, else the bool).
- Every `requests.get` call in `integrations/grafana/base.py`, `tempo.py`, and
  `verifier.py` now passes `verify=<ssl_verify>`.
- `opensre onboard` → Grafana step adds two prompts (mirroring Splunk's):
  "Verify SSL certificate?" and, if yes, "Path to CA bundle for SSL verification".
- New `GRAFANA_VERIFY_SSL` / `GRAFANA_CA_BUNDLE` env vars, wired through the
  env-based integration loader (`integrations/_catalog_impl.py`) and the
  `GRAFANA_INSTANCES` multi-instance JSON path (generic passthrough, so
  per-instance `verify_ssl`/`ca_bundle` works for free).
- Threaded through the 6 Grafana investigation tools
  (`integrations/grafana/tools/__init__.py`) via `_resolve_grafana_client`
  and `_grafana_creds`, so real investigations (not just onboarding) pick up
  the setting.

Default (`verify_ssl=True`, `ca_bundle=""`) reproduces today's exact behavior,
so this is purely additive/opt-in — existing configs are unaffected.

Reproduced the reported bug and verified the fix against a local self-signed
HTTPS server standing in for an on-prem Grafana instance, using the real
`integrations/grafana/verifier.py` code path:

```
1) default (no ca_bundle) -> Datasource discovery failed: could not reach https://localhost:8443
   (... SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify
   failed: self-signed certificate (_ssl.c:1081)')) ...)
2) ca_bundle set -> [{'uid': 'loki1', 'name': 'logs-prod', 'type': 'loki', 'isDefault': True}]
3) verify_ssl=False -> [{'uid': 'loki1', 'name': 'logs-prod', 'type': 'loki', 'isDefault': True}]
ALL OK
```

(1) reproduces the exact error class from the issue. (2) is the recommended
fix — full TLS verification stays on, trusting the org's CA. (3) is the
explicit bypass escape hatch for local/lab instances.

---

A user reported that on-prem Grafana behind a self-signed/internal CA fails
TLS verification during onboarding, with no way to configure trust. The root
cause is that `requests` never consults the OS trust store, and this repo's
Grafana client never passed a `verify=` override at all. Splunk and Argo CD
had already solved this exact problem with a `verify_ssl`/`ca_bundle` pair on
their config models, so I extended that same pattern to Grafana rather than
inventing a new mechanism — keeping the config surface, onboarding UX, and env
var naming consistent across integrations. The alternative (telling users to
set the global `REQUESTS_CA_BUNDLE` env var) was rejected because it would
also relax TLS checks for every other integration and the LLM provider calls
in the same process, not just Grafana.

---

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -193,6 +193,10 @@ GRAFANA_READ_TOKEN=
 GRAFANA_INSTANCE_URL=
 GRAFANA_LOKI_DATASOURCE_UID=
 GRAFANA_TEMPO_DATASOURCE_UID=
+# Set to false to skip TLS verification, or point GRAFANA_CA_BUNDLE at a PEM
+# file for a self-signed/internal CA (e.g. an on-prem Grafana instance).
+GRAFANA_VERIFY_SSL=true
+GRAFANA_CA_BUNDLE=
 
 # Optional multi-instance override. When set, this takes precedence over the
 # single-instance Grafana vars above.
--- a/docs/configuration/environment-variables.mdx
+++ b/docs/configuration/environment-variables.mdx
@@ -198,6 +198,8 @@ All configuration options for OpenSRE can be set via environment variables. This
 | `GRAFANA_READ_TOKEN` | | Grafana API token | `integrations/_catalog_impl` |
 | `GRAFANA_LOKI_DATASOURCE_UID` | | UID of Loki datasource in Grafana | `.env.example` |
 | `GRAFANA_TEMPO_DATASOURCE_UID` | | UID of Tempo datasource in Grafana | `.env.example` |
+| `GRAFANA_VERIFY_SSL` | `true` | Verify Grafana SSL certificate | `integrations/_catalog_impl` |
+| `GRAFANA_CA_BUNDLE` | | Path to a CA bundle for Grafana TLS verification (self-signed/internal CA) | `integrations/_catalog_impl` |
 | `DD_API_KEY` | | Datadog API key | `integrations/_catalog_impl` |
 | `DD_APP_KEY` | | Datadog application key | `integrations/_catalog_impl` |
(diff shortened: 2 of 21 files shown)


muddlebee:
@greptile review

PR #4137, opened 2026-07-18
fix(watchdog): treat denied sample as inaccessible, not target exit

### LINKED ISSUE #4136: bug(watchdog): a running process that cannot be introspected is reported as "target exited" (exit SUCCESS), so thresholds never run
**Summary.** `opensre watchdog` reports a running process as exited, and stops with exit code 0, whenever it cannot read that process's fields. The thresholds (`--max-cpu` / `--max-rss` / `--max-runtime`) then silently never run. Anyone watching a service that runs under its own account (postgres, nginx, mysqld, redis: the default deployment for all of them) hits this from a normal operator shell, and a single transient `AccessDenied` on a process you own ends an established watch the same way.

watchdog reports a live root-owned process as exited with exit code 0

**Steps to reproduce** (any macOS or Linux box, no Telegram needed)

1. Pick a process owned by another user. On macOS the system logger works: `pgrep -l syslogd`. On Linux use `postgres` or `nginx`.
2. Watch it: `opensre watchdog --name syslogd --max-rss 1G --verbose`
3. Output is `watchdog: target exited pid=128` and the exit code is 0, while `pgrep -l syslogd` still shows it running.

Scripted repro from the repo root:

```
python - <<'PY'
import psutil
from tools.system.watch_dog.config import WatchdogConfig
from tools.system.watch_dog.process_monitor import ProcessMonitor
cfg = WatchdogConfig.model_validate({"pid": None, "name": "syslogd", "pick_first": True, "max_rss": "1G"})
mon = ProcessMonitor(cfg)
print("is_running:", psutil.Process(mon._pid).is_running(), "| sample.alive:", mon.sample().alive)
PY
```

Prints `is_running: True | sample.alive: False`.

**Expected:** an explicit "cannot introspect pid 128 (permission denied)" error, or keep watching. Never a false "target exited" with a success code.

**Actual:** false "target exited", exit 0, monitoring silently stops at the first sample.

**Why it matters.** The watchdog's one job is to raise an alarm. This failure mode looks exactly like a clean success, so the operator believes monitoring is active when it ended at second zero; no alarm can ever fire. A crash would be safer than this. The same conflation ends self-monitoring watches on a single transient denial (for example the macOS hardened runtime denying one read).

**Scope.** Measured while filing this issue: 140 of 311 live processes (45 percent) are running but expose no readable sample to a normal user (repeat snapshots today ranged 40 to 45 percent). Every one of them produces the false "exited" if watched. `--name syslogd`, `configd`, `powerd`, and `launchd` all reproduce it here; postgres/nginx/mysqld/redis are the same mechanism wherever they run under a dedicated service user.

**Root cause** (main @ 0f07018)

- `tools/system/watch_dog/process_monitor.py:62-63`: `sample()` catches `psutil.AccessDenied` via `PROCESS_INACCESSIBLE_OR_GONE` and returns `_dead_sample(alive=False)`, conflating "unreadable" with "gone".
- `tools/system/watch_dog/runner.py:56-59`: `alive=False` maps to "target exited" and returns SUCCESS.
- `tools/system/watch_dog/monitor.py:61-65` (REPL `watch` loop): the same conflation through `probe()` returning `None` for both cases.

Two sibling paths already handle "cannot introspect" honestly, so the watchdog is the odd one out:

- `tools/system/fleet_monitoring/bus.py:156-159` (`_process_is_alive`): `PermissionError` from `os.kill(pid, 0)` returns True, commented "Process exists but we can't signal it. Treat as alive."
- `tools/system/fleet_monitoring/sampler.py:93-104`: a `None` probe clears only that tick's metric cell and retries next tick; it never declares the process gone.

**Environment.** macOS 13.7.8 (Intel), Python 3.13.7, psutil 7.2.2, opensre main @ 0f07018 (editable install). The mechanism is not macOS-specific: any cross-user watch on Linux behaves the same.

**Fix.** PR #4137: fail fast at construction with an actionable permission error when the target is never introspectable, retry transient denials, keep "target exited" only for a truly gone PID. One existing test pinned the old behavior (`tests/watch_dog/test_monitor.py:144`) and is updated there.

Ourbando:
## TL;DR

Point `opensre watchdog` at a process it lacks permission to read (postgres, nginx, redis: anything under its own service account) and it prints `target exited` and exits 0. The process is alive. Monitoring has silently stopped, and the exit code says success.

With this PR the watchdog tells the truth: a clear permission error up front, a retry on brief permission blips, and `target exited` only when the process is really gone.

Closes #4136.

## See it

before: false "target exited" exit 0 on a live process; after: honest permission error, and a real max-rss alarm firing on a runaway process

Real terminal output, root-owned `syslogd` standing in for any service-account process:

**Today (main):**

```
$ opensre watchdog --name syslogd --max-rss 1G --verbose
watchdog: target exited pid=128        <- syslogd is running; exit code 0
```

**With this PR, same command:**

```
$ opensre watchdog --name syslogd --max-rss 1G --verbose
✗  OpenSREError
   Process 128 (syslogd) is running but cannot be introspected (permission denied).
   Re-run with privileges matching the target process (for example via sudo)
   so CPU/RSS/runtime can be sampled.        <- exit code 1
```

And when the watchdog can read its target, thresholds now reliably protect: the demo's third act watches a real runaway process and the `--max-rss` alarm fires.

## Why it happens

Three spots treat "I was denied a read" as "the process died":

1. `ProcessMonitor.sample()` catches `psutil.AccessDenied` in the same `except` as `NoSuchProcess` and returns a dead sample (`process_monitor.py`).
2. The runner maps any dead sample to "target exited" and returns SUCCESS (`runner.py`).
3. The REPL `watch` loop does the same through `probe()` returning `None` for both cases (`monitor.py`).

## What changed

- `ProcessSample` gains `accessible: bool`. "Gone" and "unreadable" become different facts.
- Never introspectable at startup: fail fast with an actionable `OpenSREError`, like the existing "No process found for PID" errors. A watch that can never sample can never alarm, so pretending to monitor would be worse than refusing.
- Transient denial mid-watch: skip that tick and retry, matching how `fleet_monitoring/sampler.py` already handles a failed probe (clear one tick, retry, never declare the process gone). Zeroed metrics never reach the thresholds.
- REPL loop: `pid_exists()` now separates gone (still completes as "target process exited") from unreadable (fails the task loudly on the first tick, retries later ones).

## Every case, before and after

| Case | main | this PR |
|---|---|---|
| Watch another account's process | "target exited", exit 0 | permission error + sudo hint, exit 1 |
| One denied read mid-watch (own process) | watch ends as "exited" | tick retried, watch continues |
| Process really exits | "target exited", exit 0 | same, unchanged |
| Process vanishes mid-sample | dead sample | same, unchanged |
| REPL watch, denied on first tick | task COMPLETED "target process exited" | task FAILED "cannot introspect pid (permission denied)" |
| REPL watch, denied later | task COMPLETED "target process exited" | progress note + retry; completes on the real exit |

## Tests

Every row above is pinned by a test. The one test that pinned the old behavior (`test_process_monitor_returns_dead_sample_on_access_denied`) is split into the honest cases; eight new tests cover retry, no-thresholds-on-denied-metrics, and the REPL paths. `make lint`, `make format-check`, `make typecheck` (1310 files), watchdog suites 62/62 green.

```
python -m pytest tests/watch_dog tests/interactive_shell/test_watch_cmds.py

# live check (macOS: syslogd; Linux: postgres or nginx)
opensre watchdog --name syslogd --max-rss 1G --verbose; echo $?
```

## Environment

macOS 13.7.8 (Intel), Python 3.13.7, psutil 7.2.2. Reproduced against main @ 0f07018; scripted repro in #4136.

### DIFF
--- a/tests/interactive_shell/test_watch_cmds.py
+++ b/tests/interactive_shell/test_watch_cmds.py
@@ -231,3 +231,122 @@ def test_run_watchdog_once_without_thresholds_exits(monkeypatch: pytest.MonkeyPa
     assert task.status == TaskStatus.COMPLETED
     assert task.result == "single sample (once)"
     dispatcher.dispatch.assert_not_called()
+
+
+def test_run_watchdog_first_probe_inaccessible_marks_failed(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """A live PID this user cannot introspect must fail loudly, not report
+    'target process exited' (a permission failure is not an exit)."""
+    from platform.common.task_registry import TaskRegistry
+    from tools.system.watch_dog.monitor import run_watchdog
+
+    reg = TaskRegistry()
+    task = reg.create(TaskKind.WATCHDOG, command="watchdog pid=1")
+    task.mark_running()
+    dispatcher = MagicMock()
+    dispatcher.dispatch = MagicMock(return_value=True)
+
+    monkeypatch.setattr("tools.system.watch_dog.monitor.probe", lambda *_a, **_kw: None)
+    monkeypatch.setattr("tools.system.watch_dog.monitor.pid_exists", lambda _pid: True)
+
+    run_watchdog(
+        task=task,
+        watched_pid=1,
+        interval_seconds=0.01,
+        max_cpu=None,
+        max_runtime_seconds=None,
+        max_rss_mib=None,
+        once=False,
+        dispatcher=dispatcher,
+        on_alarm=None,
+    )
+    assert task.status == TaskStatus.FAILED
+    assert "permission denied" in (task.error or "")
+    dispatcher.dispatch.assert_not_called()
+
+
+def test_run_watchdog_gone_pid_completes_as_exited(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    from platform.common.task_registry import TaskRegistry
+    from tools.system.watch_dog.monitor import run_watchdog
+
+    reg = TaskRegistry()
+    task = reg.create(TaskKind.WATCHDOG, command="watchdog pid=1")
+    task.mark_running()
+    dispatcher = MagicMock()
+    dispatcher.dispatch = MagicMock(return_value=True)
+
+    monkeypatch.setattr("tools.system.watch_dog.monitor.probe", lambda *_a, **_kw: None)
+    monkeypatch.setattr("tools.system.watch_dog.monitor.pid_exists", lambda _pid: False)
+
+    run_watchdog(
+        task=task,
+        watched_pid=1,
+        interval_seconds=0.01,
+        max_cpu=None,
+        max_runtime_seconds=None,
+        max_rss_mib=None,
+        once=False,
+        dispatcher=dispatcher,
+        on_alarm=None,
+    )
+    assert task.status == TaskStatus.COMPLETED
+    assert task.result == "target process exited"
+
+
+def test_run_watchdog_transient_inaccessible_retries(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """A mid-watch AccessDenied tick (probe None while the PID exists) must
+    retry, then still report the real exit when the PID is truly gone."""
+    from datetime import UTC, datetime, timedelta
+
+    from platform.common.task_registry import TaskRegistry
+    from tools.system.fleet_monitoring.probe import ProcessSnapshot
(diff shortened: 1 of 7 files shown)


Ourbando:
@greptile review

Devesh36:
LGTM

VaibhavUpreti (approved):
Great @Ourbando!

PR #4144, opened 2026-07-18
fix: first_string returns literal 'None' for explicitly-None MCP config keys

adi-IL:
## Describe the changes you have made in this PR

\irst_string\ in \core/tool_framework/utils/mcp_params.py\ normalizes MCP integration source dicts (posthog_mcp, sentry_mcp, openclaw, x_mcp). It coerced each candidate value with \str(source.get(key, ''))\. The problem: when a key is *present* with value \None\ (a normal value for an unset optional field like \posthog_token\ or a URL), \.get\ returns \None\, \str(None)\ becomes the truthy string \'None'\, and the helper returns \'None'\ instead of falling through to the next alias key.

That means an unset optional MCP field can inject the literal \'None'\ as an auth token or URL. The fix skips \None\ values explicitly and only substitutes \''\ when the key is absent, so an unset field falls through to the next alias and \None\ is returned only when every key is genuinely unset.

### Demo / Screenshot for feature changes and bug fixes

Repro before the fix:

\\\
source = {'url': None, 'mode': 'stdio'}
first_string(source, 'url', 'mode')  # buggy: returns 'None'
\\\

After the fix it returns \'stdio'\, and \irst_string({'url': None, 'mode': None}, 'url', 'mode')\ returns \None\ as intended. New regression tests in \	ests/tools/test_mcp_params.py\ cover both cases (and fail against the pre-fix code).

---

The problem: \irst_string\ turned an explicitly-\None\ config key into the truthy string \'None'\, so an unset optional MCP field leaked \'None'\ into a token/URL. The fix is a small guard: skip a \None\ value and only default to \''\ when the key is missing, then strip and return the first non-empty string. The change is local to one helper and adds no imports or dependencies. I added two regression tests that fail on the old code (proving they actually guard the bug) and pass on the fix. This is a pure, deterministic function with no live infra or LLM, so the tests are fast and non-flaky.

- [x] I have added proper PR title and linked to the issue (no issue; this is a self-contained bug fix)

### DIFF
--- a/core/tool_framework/utils/mcp_params.py
+++ b/core/tool_framework/utils/mcp_params.py
@@ -20,9 +20,15 @@ def string_list(value: object) -> list[str]:
 
 def first_string(source: dict[str, object], *keys: str) -> str | None:
     for key in keys:
-        value = str(source.get(key, "")).strip()
-        if value:
-            return value
+        # An explicitly-None value (a normal value for an unset optional
+        # field) must fall through to the next alias key, not become the
+        # truthy string "None" that would be injected as a token/URL.
+        value = source.get(key, "")
+        if value is None:
+            continue
+        text = str(value).strip()
+        if text:
+            return text
     return None
 
 
--- a/core/tool_framework/utils/mcp_params.py
+++ b/core/tool_framework/utils/mcp_params.py
@@ -15,7 +15,10 @@ __all__ = ["first_list", "first_string", "string_list"]
 def string_list(value: object) -> list[str]:
     if not isinstance(value, list):
         return []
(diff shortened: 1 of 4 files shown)


adi-IL:
@greptile review

adi-IL:
@greptile review

cerencamkiran:
The pr looks good. But next time, please open an issue first and wait until it is assigned to you before starting the work. That's our workflow.

adi-IL:
Thanks for the review! Understood, I'll make sure to open an issue and wait for assignment next time.

PR #4152, opened 2026-07-19
fix(watchdog): treat an unreadable live process as a failure, not an exit

### LINKED ISSUE #4136: bug(watchdog): a running process that cannot be introspected is reported as "target exited" (exit SUCCESS), so thresholds never run

Dandiggas:
Fixes #4136

The watchdog conflated "process I'm not allowed to inspect" (psutil `AccessDenied`, e.g. any root-owned daemon) with "process exited", reporting **"target exited" / exit 0** while the target was still running — so thresholds silently never ran. This PR makes the sampler distinguish the two: a genuine `NoSuchProcess` (or a denial where the PID is gone) still yields a dead sample and a clean exit; a denial while the PID exists is retried briefly (transient denials), then fails **loudly** with an explicit `OpenSREError` (exit 1) telling the user the process is running but unreadable. The REPL watch daemon gets the same disambiguation (`mark_failed` instead of "target process exited"). A `PROCESS_ACCESS_DENIED` constant was added to `fleet_monitoring/probe.py` so callers don't import psutil directly (keeping psutil confined per #1489).

`runner.py` (cited in the issue) needed no change: once the sampler stops fabricating `alive=False` for denials, its exit-on-dead path only sees genuine exits — a new test pins that a permission-denied sampler propagates the error and never returns SUCCESS.

Before (unmodified `main`, macOS, `syslogd` PID 344 running, root-owned):

```
$ opensre watchdog --name syslogd --max-rss 1G --verbose
watchdog: target exited pid=344
$ echo $?
0
```

After (this branch, same live process):

```
$ opensre watchdog --name syslogd --max-rss 1G --verbose
✗  OpenSREError
   Process 344 (syslogd) is running but cannot be inspected (permission denied).
   Re-run with sufficient privileges (for example under sudo)...
$ echo $?
1
```

Healthy path unregressed (owned `sleep` process):

```
watchdog: status=ok pid=15218 name=sleep cpu=0.0% rss=1.2MiB runtime=0s
watchdog: status=ok pid=15218 name=sleep cpu=0.0% rss=1.2MiB runtime=2s
```

---

*(Draft until the unchecked boxes above are true — I'm completing my line-by-line review now and will mark ready once I can honestly check them.)*

The bug is a semantics collapse: three layers each translated "I couldn't read the process" into "the process is gone". The fix pushes the distinction to the lowest layer that can know the truth — the sampler — using `pid_exists()` as the tiebreaker after an `AccessDenied`: PID gone → real exit (dead sample, unchanged behavior); PID present → the process is alive but unreadable, which must never look like a clean exit. Transient denials get a bounded retry (3 × 0.5s) before failing, because a denial can be a momentary race. Alternatives considered: minting a dedicated exit code (rejected — `exit_codes.py` defines only SUCCESS/ERROR/USAGE_ERROR and a new code is a contract change beyond this bug) and making the runner catch-and-branch (rejected — fixing the sampler makes the runner correct transitively, smaller diff). Tests were written first and fail on `main` (5 failures pinning the bug).

---

### DIFF
--- a/tests/interactive_shell/test_watch_cmds.py
+++ b/tests/interactive_shell/test_watch_cmds.py
@@ -230,4 +230,67 @@ def test_run_watchdog_once_without_thresholds_exits(monkeypatch: pytest.MonkeyPa
     )
     assert task.status == TaskStatus.COMPLETED
     assert task.result == "single sample (once)"
+
+
+def test_run_watchdog_marks_completed_when_process_gone(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """An unreadable snapshot for a PID that no longer exists is a real exit."""
+    from platform.common.task_registry import TaskRegistry
+    from tools.system.watch_dog.monitor import run_watchdog
+
+    reg = TaskRegistry()
+    task = reg.create(TaskKind.WATCHDOG, command="watchdog pid=1")
+    task.mark_running()
+    dispatcher = MagicMock()
+    dispatcher.dispatch = MagicMock(return_value=True)
+
+    monkeypatch.setattr("tools.system.watch_dog.monitor.probe", lambda *_a, **_kw: None)
+    monkeypatch.setattr("tools.system.watch_dog.monitor.pid_exists", lambda _pid: False)
+
+    run_watchdog(
+        task=task,
+        watched_pid=1,
+        interval_seconds=0.01,
+        max_cpu=None,
+        max_runtime_seconds=None,
+        max_rss_mib=None,
+        once=False,
+        dispatcher=dispatcher,
+        on_alarm=None,
+    )
+    assert task.status == TaskStatus.COMPLETED
+    assert task.result == "target process exited"
+
+
+def test_run_watchdog_marks_failed_when_process_unreadable(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """A live PID whose snapshot is denied must fail, not report a clean exit."""
+    from platform.common.task_registry import TaskRegistry
+    from tools.system.watch_dog.monitor import run_watchdog
+
+    reg = TaskRegistry()
+    task = reg.create(TaskKind.WATCHDOG, command="watchdog pid=1")
+    task.mark_running()
+    dispatcher = MagicMock()
+    dispatcher.dispatch = MagicMock(return_value=True)
+
+    monkeypatch.setattr("tools.system.watch_dog.monitor.probe", lambda *_a, **_kw: None)
+    monkeypatch.setattr("tools.system.watch_dog.monitor.pid_exists", lambda _pid: True)
+
+    run_watchdog(
+        task=task,
+        watched_pid=1,
+        interval_seconds=0.01,
+        max_cpu=None,
+        max_runtime_seconds=None,
+        max_rss_mib=None,
+        once=False,
+        dispatcher=dispatcher,
+        on_alarm=None,
+    )
+    assert task.status == TaskStatus.FAILED
+    assert task.result != "target process exited"
+    assert "cannot be inspected" in (task.error or "")
     dispatcher.dispatch.assert_not_called()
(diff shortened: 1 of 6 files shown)


cerencamkiran:
Please don’t work on unassigned issues. Thanks

PR #4161, opened 2026-07-20
fix: scrub error values in redact_sensitive before LLM/evidence

Devesh36:
Integration clients often return raw failure text in tool payloads, for example:

```python
return {"success": False, "error": str(exc)}
# or
return {"success": False, "error": f"HTTP {status}: {response.text[:200]}"}
```

Those dicts flow into investigation evidence and LLM tool-result context via `merge_tool_evidence()` and the ReAct loop. Gateway Slack/Telegram sinks already redact turn-level errors, but tool payloads bypass that path.

`redact_sensitive()` already runs on every tool output before it is stored or traced, but it only redacted **dict keys** (`api_key`, `token`, …). The `"error"` **string value** was passed through unchanged, so secrets, hostnames, SQL fragments, and vendor response bodies could still reach:

- LLM context (and potentially be quoted in gateway replies)
- Investigation evidence and report formatting

This is a CWE-209 / information-exposure gap. Fixing it at the existing boundary matches AGENTS.md: redact at the sink/response boundary rather than touching dozens of integration call sites.

We considered migrating 60+ integration clients to a shared helper, but chose the boundary approach because:

- One place catches current and future integrations
- Much smaller, easier-to-review diff
- Same sanitization applies everywhere `redact_sensitive()` is already used

### What changed

1. **`platform/observability/trace/redaction.py`**
   - When a dict key is `"error"` and the value is a string:
     - Scrub embedded secrets (JWT, bearer tokens, GitHub tokens, AWS keys) → `[redacted]`
     - Truncate values longer than 120 characters → `...`

2. **`tests/utils/test_tool_trace.py`**
   - Added tests for secret scrubbing, truncation, and safe passthrough of short messages like `HTTP 403`

3. **Optional helpers (new, not load-bearing)**
   - `platform/observability/errors/client_errors.py` — `safe_integration_error_message()` / `integration_client_error_result()` for new integration code
   - `tests/platform/observability/test_integration_client_errors.py` — unit tests for those helpers

Integration client `except` blocks are **unchanged**; the boundary is the safety net.

### DIFF
--- a/platform/observability/errors/client_errors.py
+++ b/platform/observability/errors/client_errors.py
@@ -40,7 +40,9 @@ def integration_client_error_result(
         method=method,
         extras=extras,
     )
-    return {"success": False, "error": safe_integration_error_message(exc), **fields}
+    # Unpack fields first so a caller-supplied error= cannot replace the
+    # sanitized message (CWE-209).
+    return {"success": False, **fields, "error": safe_integration_error_message(exc)}
 
 
 __all__ = [
--- a/platform/observability/errors/client_errors.py
+++ b/platform/observability/errors/client_errors.py
@@ -1,0 +1,49 @@
+"""Safe error envelopes for integration service clients.
+
+Investigation tool payloads and LLM context consume client ``error`` strings.
+Those surfaces must not receive ``str(exc)``, response bodies, or other
+exception detail (CWE-209). Full detail is logged server-side via
+``capture_service_error``.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any
+
+import httpx
+
+from platform.observability.errors.service import capture_service_error
+
+
+def safe_integration_error_message(exc: BaseException) -> str:
+    """Return a safe, investigation-facing error string."""
+    if isinstance(exc, httpx.HTTPStatusError):
+        return f"HTTP {exc.response.status_code}"
+    return type(exc).__name__
+
+
+def integration_client_error_result(
+    exc: BaseException,
+    *,
+    integration: str,
+    method: str,
+    logger: logging.Logger,
+    extras: dict[str, Any] | None = None,
+    **fields: Any,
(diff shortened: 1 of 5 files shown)


Devesh36:
@greptile review

Devesh36:
@greptile review

YauhenBichel:
I am reviewing as well

YauhenBichel:
@Devesh36 I think we should close this PR, and create a GitHub issue with good understanding the  scope of work and what why and how to change

YauhenBichel (changes requested):
I think the PR should be closed, and described in GitHub issue

Potentional scope for work:

1) core/execution.py — _normalize_result
When error is truthy, set content (and the error field in details / compat_payload) to a safe message: HTTP {status} for httpx.HTTPStatusError, else type(exc).__name__ — or equivalent using safe_integration_error_message if you have the exception; if you only have a string, prefer a generic "tool failed" / status-only, not the raw vendor body. Log full detail server-side (existing capture_service_error at clients is fine).

2) Message path
After (1), OpenAI/Anthropic/CLI tool-result builders should automatically get safe content. Don’t rely on redact_sensitive alone for LLM context. Optional: redact once before serialize as belt-and-suspenders.

3) merge_tool_evidence
Store the same sanitized payload under evidence[tool_name] that you put in tool_outputs — no raw secret copy.

4) Keep redact_sensitive error scrubbing as a secondary belt for traces/reports if you want, but don’t market it as the LLM fix.

Tests (contract, not only unit)
Plant a JWT/bearer (or similar) in a fake tool {"error": "..."} and assert it is absent from:

provider_content() / execute_tools payload
build_tool_result_messages (OpenAI) and Anthropic build_tool_result_message
evidence[tool_name] after merge_tool_evidence
Existing helper unit tests for JWT scrub under key "error" can stay.

PR #4165, opened 2026-07-21
perf(investigation): cut context-budget hot-path token recomputation

Devesh36:
This PR optimizes the investigation context-budget hot path by removing repeated token estimation work inside the ReAct loop, while preserving existing behavior.

### What changed

- Refactored context token estimation in `core/context_budget.py`:
  - split message-token and system/tools-overhead estimation
  - precompute `system + tools` overhead once per enforcement pass
  - maintain per-message token estimates during trimming/truncation
  - update only affected message token counts after truncation
- Kept public behavior of budget enforcement/trimming intact.
- Improved evidence source lookup in `tools/investigation/stages/gather_evidence/agent.py` by using a prebuilt `tool_by_name` map instead of repeated linear scans.
- Added/updated tests in `tests/core/test_context_budget.py` to characterize behavior and validate overhead/token-estimation paths.

## Why we did this

Investigation runs a ReAct loop where `enforce_context_budget` is called repeatedly. Before this change, token estimation frequently re-walked message history and re-serialized large tool schema payloads (via `json.dumps`) across loop iterations and truncation attempts.

That created avoidable CPU overhead in a path that executes often, especially with:

- many registered tools / large schemas
- long investigation transcripts
- repeated budget checks near the context limit

This PR removes that repeated invariant work and narrows recomputation to only what changes.

## Why this was needed

- The context-budget path is on the per-iteration investigation hot path.
- Prior behavior paid repeated cost for mostly unchanged inputs.
- This impacted responsiveness and efficiency under heavier investigations.
- We needed a low-risk, behavior-preserving optimization that improves runtime cost without changing investigation output semantics.

## Behavior and risk

- No intended functional behavior change.
- Budget trimming/truncation semantics remain the same.
- Changes are internal to token accounting and lookup efficiency.
- Tests were added/updated to guard characterization and regressions.

## Validation

- `uv run pytest tests/core/test_context_budget.py tests/agent/test_investigation.py -q --tb=short`
- `make lint`
- `make format-check`
- `make typecheck`

### DIFF
--- a/core/context_budget.py
+++ b/core/context_budget.py
@@ -185,13 +185,22 @@ def _estimate_messages_tokens(messages: list[dict[str, Any]]) -> int:
     return sum(_message_token_estimate(message) for message in messages)
 
 
-def _system_and_tools_tokens(
+def _system_tokens(system: str | None) -> int:
+    return int(len(system) * _TOKENS_PER_CHAR) if system else 0
+
+
+def system_and_tools_overhead(
     system: str | None,
     tools: list[dict[str, Any]] | None,
 ) -> int:
-    total = 0
-    if system:
-        total += int(len(system) * _TOKENS_PER_CHAR)
+    """Fixed token overhead for the system prompt and tool schemas.
+
+    Callers on a hot path (e.g. the investigation ReAct loop) should call this
+    once and pass the result to ``enforce_context_budget`` via
+    ``fixed_overhead_tokens`` so tool schemas are not re-serialized every
+    iteration.
+    """
+    total = _system_tokens(system)
     if tools:
         for schema in tools:
             total += int(len(json.dumps(schema, default=str)) * _TOKENS_PER_CHAR)
@@ -211,7 +220,7 @@ def estimate_message_tokens(
     aggressively while system + tools (tens of thousands of tokens for
     opensre's 100+ tool registry) silently pushed us over the line.
     """
-    return _estimate_messages_tokens(messages) + _system_and_tools_tokens(system, tools)
+    return _estimate_messages_tokens(messages) + system_and_tools_overhead(system, tools)
 
 
 def trim_lowest_value_tool_pair(messages: list[dict[str, Any]]) -> bool:
@@ -341,10 +350,17 @@ def enforce_context_budget(
     *,
     system: str | None = None,
     tools: list[dict[str, Any]] | None = None,
+    fixed_overhead_tokens: int | None = None,
     ceiling: int = _TOKEN_BUDGET_CEILING,
 ) -> None:
-    """Trim low-value tool exchanges until the prompt fits under ``ceiling``."""
-    fixed_overhead_tokens = _system_and_tools_tokens(system, tools)
+    """Trim low-value tool exchanges until the prompt fits under ``ceiling``.
+
+    Pass ``fixed_overhead_tokens`` (from :func:`system_and_tools_overhead`) to
+    skip re-serializing tool schemas on every call.  When omitted the overhead
+    is computed from ``system`` and ``tools``.
+    """
+    if fixed_overhead_tokens is None:
+        fixed_overhead_tokens = system_and_tools_overhead(system, tools)
     message_tokens, total_message_tokens = _message_token_estimates(messages)
     while (total_message_tokens + fixed_overhead_tokens) > ceiling:
         if not trim_lowest_value_tool_pair(messages):
--- a/core/context_budget.py
+++ b/core/context_budget.py
@@ -162,11 +162,58 @@ def context_budget_ceiling_for_model(model: str | None) -> int:
     return max(window - _RESPONSE_HEADROOM_TOKENS, _RESPONSE_HEADROOM_TOKENS)
 
 
+def _message_token_estimate(message: dict[str, Any]) -> int:
+    total = 0
+    content = message.get("content", "")
+    if isinstance(content, str):
+        total += int(len(content) * _TOKENS_PER_CHAR)
+    elif isinstance(content, list):
+        for block in content:
+            if isinstance(block, dict):
+                total += int(len(json.dumps(block, default=str)) * _TOKENS_PER_CHAR)
+            elif isinstance(block, str):
+                total += int(len(block) * _TOKENS_PER_CHAR)
+    return total
+
+
+def _message_token_estimates(messages: list[dict[str, Any]]) -> tuple[list[int], int]:
+    tokens = [_message_token_estimate(message) for message in messages]
+    return tokens, sum(tokens)
+
+
+def _refresh_message_token_estimates(
+    messages: list[dict[str, Any]],
+    *,
+    current_estimates: list[int],
+) -> tuple[list[int], int]:
+    if len(current_estimates) != len(messages):
+        return _message_token_estimates(messages)
+    return current_estimates, sum(current_estimates)
+
+
+def _estimate_messages_tokens(messages: list[dict[str, Any]]) -> int:
+    return sum(_message_token_estimate(message) for message in messages)
+
+
+def _system_and_tools_tokens(
+    system: str | None,
+    tools: list[dict[str, Any]] | None,
+) -> int:
+    total = 0
+    if system:
+        total += int(len(system) * _TOKENS_PER_CHAR)
+    if tools:
+        for schema in tools:
+            total += int(len(json.dumps(schema, default=str)) * _TOKENS_PER_CHAR)
+    return total
+
+
 def estimate_message_tokens(
     messages: list[dict[str, Any]],
     *,
     system: str | None = None,
     tools: list[dict[str, Any]] | None = None,
+    system_tools_overhead_tokens: int | None = None,
 ) -> int:
     """Cheap upper-bound token estimate covering everything Anthropic sees.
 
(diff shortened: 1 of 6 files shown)


core/context_budget.py
```diff
@@ -162,11 +162,58 @@ def context_budget_ceiling_for_model(model: str | None) -> int:
...
+    return total
+
+
 def estimate_message_tokens(
     messages: list[dict[str, Any]],
     *,
     system: str | None = None,
     tools: list[dict[str, Any]] | None = None,
+    system_tools_overhead_tokens: int | None = None,
 ) -> int:
```
Davidson3556, line 215: Test-only parameter, and passing it alongside `tools=` silently ignores `tools`. Drop it:

```suggestion
def estimate_message_tokens(
    messages: list[dict[str, Any]],
    *,
    system: str | None = None,
    tools: list[dict[str, Any]] | None = None,
) -> int:
```

core/context_budget.py
```diff
@@ -175,23 +222,9 @@ def estimate_message_tokens(
...
-                    total += int(len(block) * _TOKENS_PER_CHAR)
-    if system:
-        total += int(len(system) * _TOKENS_PER_CHAR)
-    if tools:
-        for schema in tools:
-            total += int(len(json.dumps(schema, default=str)) * _TOKENS_PER_CHAR)
-    return total
+    if system_tools_overhead_tokens is None:
+        system_tools_overhead_tokens = _system_and_tools_tokens(system, tools)
+    return _estimate_messages_tokens(messages) + system_tools_overhead_tokens
```
Davidson3556: And the body collapses to one line once the override is gone:

```suggestion
    return _estimate_messages_tokens(messages) + _system_and_tools_tokens(system, tools)
```

core/context_budget.py
```diff
@@ -335,6 +379,10 @@ def enforce_context_budget(
                 "[agent] truncated oversized message to fit context budget (ceiling=%d)", ceiling
             )
             continue
+        message_tokens, total_message_tokens = _refresh_message_token_estimates(
+            messages,
+            current_estimates=message_tokens,
+        )
```
Davidson3556: `trim_lowest_value_tool_pair` ends in `del messages[start:end]` with `end > start` guaranteed, so length always shrinks and the equal-length branch of `_refresh_message_token_estimates` is dead. Call the estimator directly and delete the helper:

```suggestion
        message_tokens, total_message_tokens = _message_token_estimates(messages)
```

tools/investigation/stages/gather_evidence/agent.py
```diff
@@ -60,6 +59,13 @@ def _mark_messages(messages: list[dict[str, Any]], key: str) -> None:
         msg[key] = True
 
 
+def _runtime_tool_source(tool_by_name: dict[str, Any], tool_name: str) -> str:
```
Davidson3556: This duplicates `tool_source` in `core/execution.py`, which now has no callers left outside the lazy re-export in `core/__init__.py`. Rather than reimplement the `getattr(..., "source", "unknown")` fallback here, prefer changing the shared one to take the map you already build:

```python
def tool_source(tools: Mapping[str, RuntimeTool], tool_name: str) -> str:
    tool = tools.get(tool_name)
    return str(getattr(tool, "source", "unknown")) if tool else "unknown"
```

then call `tool_source(tool_by_name, tc.name)` and drop `_runtime_tool_source`. Keeps one canonical path per AGENTS.md.

Davidson3556 (commented):
Solid win on the hoist. Serializing 100+ tool schemas inside the trim loop was genuinely quadratic, and the subtraction in `_truncate_largest_message` checks out as equivalent to the old list comprehension. No behavior change that I can find.

Three small things before merge, all in the same direction (less new surface, not more):

1. `system_tools_overhead_tokens` has no production caller. `enforce_context_budget` uses the private helpers directly and `agent.py:148` doesn't pass it, so its only user is a test. It's also a footgun: pass it alongside `tools=` and `tools` gets silently ignored.
2. `tool_source` in `core/execution.py` is orphaned now that both call sites moved, and `_runtime_tool_source` reimplements it. AGENTS.md asks for one canonical path after a refactor.
3. `_refresh_message_token_estimates` can only ever take the recompute branch, since `trim_lowest_value_tool_pair` always deletes at least one message. Worth collapsing, and worth editing the PR description: the per-message cache doesn't help the trim path, the hoisted overhead is the whole win.

Also, the tests hardcode `0.50` in four places instead of importing `_TOKENS_PER_CHAR`, and most of them recompute the production formula to build the expected value. The dump-count test is the useful one, keep that.

YauhenBichel:
@Devesh36 I will take a look at the PR during a day today as well. I am interested to review

Devesh36:
> @Devesh36 I will take a look at the PR during a day today as well. I am interested to review

yeah i'll take a look at davidson review

Devesh36:
@greptile-apps review again

core/context_budget.py
```diff
@@ -162,6 +162,42 @@ def context_budget_ceiling_for_model(model: str | None) -> int:
...
+    if isinstance(content, str):
+        total += int(len(content) * _TOKENS_PER_CHAR)
+    elif isinstance(content, list):
+        for block in content:
+            if isinstance(block, dict):
+                total += int(len(json.dumps(block, default=str)) * _TOKENS_PER_CHAR)
+            elif isinstance(block, str):
+                total += int(len(block) * _TOKENS_PER_CHAR)
+    return total
+
```
YauhenBichel, line 177: I can see `json.dumps(block, default=str)) * _TOKENS_PER_CHAR` and I feel it should be improved

I would like to have this method like:
```
def _message_token_estimate(message: dict[str, Any]) -> int:
    content = message.get("content", "")
    if isinstance(content, str):
        chars = len(content)
    elif isinstance(content, list):
        chars = _sum_text_chars(content)
    else:
        chars = 0
    return int(chars * _TOKENS_PER_CHAR)

```

core/context_budget.py
```diff
@@ -162,6 +162,42 @@ def context_budget_ceiling_for_model(model: str | None) -> int:
...
+    tools: list[dict[str, Any]] | None,
+) -> int:
+    total = 0
+    if system:
+        total += int(len(system) * _TOKENS_PER_CHAR)
+    if tools:
+        for schema in tools:
+            total += int(len(json.dumps(schema, default=str)) * _TOKENS_PER_CHAR)
+    return total
+
```
YauhenBichel, line 208: `_system_and_tools_tokens` calls json.dumps for every tool schema on each `enforce_context_budget` call. We need to move it out of the loop

Prefer computing that overhead once at schema-build time and passing the int in:

```
tool_schemas = llm.tool_schemas(tools)
tools_tokens = sum(
    int(len(json.dumps(schema, default=str)) * _TOKENS_PER_CHAR)
    for schema in tool_schemas
)
enforce_context_budget(
    messages,
    system=system,
    tools_tokens=tools_tokens,  # precomputed once per investigation
    ceiling=ceiling,
)
```

Then the budget helper only needs system length:

```
def _system_tokens(system: str | None) -> int:
    return int(len(system) * _TOKENS_PER_CHAR) if system else 0

```

But please write tests before refactoring, do these changes, and after that that all tests are passed

Devesh36:
mmm @YauhenBichel anything else ?

YauhenBichel:
@Devesh36 please use TDD: before changes check that tests cover all functionality, which you are refactoring, after that do refactoring, and be sure that all tests are green after that

core/context_budget.py
Devesh36: Thanks for the suggestion — I looked at this closely, but I don’t think _sum_text_chars is a safe drop-in here.

_sum_text_chars only counts strings under content/text keys (it’s built for the truncation path). A tool_use block like {"type": "tool_use", "id": "t1", "name": "noop", "input": {}} has neither key, so _sum_text_chars returns 0, while the serialized block is ~60 chars that the provider actually bills for.

Even for text blocks, the provider sees the full JSON envelope, so json.dumps is the more faithful upper bound for token estimation.

core/context_budget.py
Devesh36: Agreed — this was the right call.

I added a public system_and_tools_overhead(system, tools) helper and a fixed_overhead_tokens param on enforce_context_budget. The investigation agent now precomputes full_overhead and system_only_overhead once before the ReAct loop and passes the right value each iteration (including the force-conclusion path where tools are dropped).

I also applied the same pattern in core/agent/react_loop.py, which had the same per-iteration re-serialization issue.

Added a test that asserts zero tool-schema json.dumps calls when fixed_overhead_tokens is passed. All existing tests still pass.

Devesh36:
> @Devesh36 please use TDD: before changes check that tests cover all functionality, which you are refactoring, after that do refactoring, and be sure that all tests are green after that

yes sirrr

core/context_budget.py
YauhenBichel: Could you please think how to avoid calling `json.dumps` each time in `for block in content:`

Devesh36:
Followed TDD for this refactor — existing characterization tests pinned the current behavior before the change (trimming, truncation, system/tools overhead), kept them green through the refactor, and added a new test for the precomputed overhead path. Ran uv run pytest tests/core/test_context_budget.py tests/agent/test_investigation.py — 43 passed.

Devesh36:
@greptile review

YauhenBichel:
@Devesh36 please be free to merge, also please check after merging manually that all is working. Good work, man

PR #4071, opened 2026-07-15
feat(interactive-shell): set up Telegram from a pasted token via a gated action tool (#3933)

Davidson3556:
Fixes #3933

#### Describe the changes you have made in this PR -

Follow-up to the docs-grounding PR (#4021), addressing muddlebee's point: when a user pastes a Telegram bot token in the REPL and asks to set it up, the agent should just do it.

Before this, the agent could only launch the interactive `/integrations setup telegram` wizard, which then re-prompts for the token — so pasting it in chat didn't help. This adds a `configure_telegram` action tool. Given a bot token (and optional chat id) the action agent pulled out of the message, it validates the token against the Telegram Bot API and, on success, saves the integration. The token is never printed back.

How it fits the existing patterns:

- Selection is the action agent's job through normal tool-calling — there is no keyword or regex detection of tokens (per `tools/interactive_shell/AGENTS.md`).
- The tool stays UI-agnostic: it validates and persists with no `surfaces`/UI imports, matching the other action tools (`sentry_fix`, `sample_alert`) after the #4065 runtime-ports refactor. It is a `tools -> integrations` dependency only.
- It validates before it persists, using the integration-layer verifier (`integrations.telegram.verifier.verify_telegram`), so a bad token fails with a reason instead of saving junk. The verifier's failure detail can embed the token (via the request URL), so the token is redacted before it's shown, on success or failure.
- A new action tool has to be registered in three places or a contract test fails: `registry_index.py` fallback descriptors, the `TOOL_MODULES` manifest, and `action_names.py`. It is also classified in the telemetry coverage allowlist.

### Demo/Screenshot for feature changes and bug fixes -

Live end-to-end run in the REPL. I pasted a fake bot token and asked the agent to set up Telegram; the action agent picked the new `configure_telegram` tool on its own, validated the token against the real Telegram Bot API, and reported the failure with a reason:

image
image

That is muddlebee's flow — paste token → agent verifies → error + reason — with a real LLM selecting the tool (not the old wizard) and a real Telegram API call. The success path (valid token → saved, token never echoed) is covered by the unit tests, since I don't have a real bot token to put in a public PR.

Verification: `make typecheck` clean (1249 files), `make lint` / `make format-check` clean, `make test-scope` 2648 passed / 1 skipped.

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
[...]

Issue #3933, opened 2026-07-10
[IMPROVEMENT] Test all integrations: verify OpenSRE provides complete setup instructions (including out-of-tool steps)


larsspinetta12:
### Current state

**Background**
During testing of the Telegram integration, a clear gap was identified: OpenSRE gives good guidance on what commands to run within the tool, but does not provide complete setup instructions for the steps that happen outside the tool (e.g. creating a Telegram bot via BotFather, getting a chat_id, etc.).

**Illustrative example — Telegram setup:**

When asked "how do I set up the telegram bot?", OpenSRE currently responds with:
"To set up Telegram messaging in OpenSRE, use the interactive shell command: /messaging pair — generates a one-time pairing code… /messaging status — checks current messaging security status"

This skips all the prerequisite steps. A complete guide (like our Cursor comparison) includes:

Creating the bot via @BotFather on Telegram
Choosing a name and username, copying the HTTP API token
Adding the bot to a group/channel/DM and getting the chat_id via getUpdates
Then running opensre integrations setup telegram / opensre onboard

_Also noted by Yauhen: /integrations setup telegram does not currently save the bot token to .env (bug)._

**Goal**
Systematically test every integration in OpenSRE and audit the setup instructions it provides. For each integration, verify that the instructions:

Cover every prerequisite step that happens outside OpenSRE (creating accounts, generating API keys, bot setup, OAuth flows, etc.)
Clearly direct the user to the right external tools/dashboards and explain how to navigate them
Correctly guide the user through the in-tool steps (/onboard, /integrations setup <name>, /integrations verify <name>)
Result in a working, verified integration

**Integrations to test**
Alerting / On-call

PagerDuty
OpsGenie
BetterStack
[Incident.io](http://incident.io/)

Metrics / APM / Tracing

Grafana
Datadog
Honeycomb
SignOz
Coralogix
OpenObserve
Victoria Logs
Groundcover

Logs

Elasticsearch / OpenSearch
Splunk
CloudWatch
CloudTrail

Databases

PostgreSQL
MySQL
MariaDB
MongoDB / MongoDB Atlas
ClickHouse
Redis
Snowflake
Azure SQL
Supabase
RDS

Cloud / Infrastructure

AWS (general)
AWS Lambda
EC2
EKS
Azure
S3
ELB
Helm
ArgoCD

Messaging / Comms

Telegram (reference case — known gap, also has bot token not saving to .env bug)
Slack
Discord
WhatsApp
[...]

PR #4167, opened 2026-07-21
refactor(config): move .env/keyring writing to config/ so every setup surface shares one writer

muddlebee:
Relates to #4071, #3933

Groundwork so that setting up an integration is written **once**, not once per surface. No user-visible behavior change.

**The problem.** Telegram (and every other integration) has three setup paths that drifted apart:

| Path | Saves to |
| --- | --- |
| `opensre onboard` → `surfaces/cli/wizard/configurators/chat_notifications.py` | store + keyring + `.env` |
| `opensre integrations setup telegram` → `integrations/cli.py` | store only |
| `configure_telegram` action tool (proposed in #4071) | store only |

All three look fine locally, because `integrations/telegram/credentials.py` resolves the store first. The divergence shows up at deploy time — `platform/deployment/ecr_deploy/prep.py:75` checks the **env var**, not the store:

```python
if not _env_set("TELEGRAM_BOT_TOKEN"):
    missing.append(DeployEnvIssue("telegram_bot", env_vars=("TELEGRAM_BOT_TOKEN",)))
```

So you can configure Telegram via the CLI, verify it, send messages — and then have `make deploy` tell you `TELEGRAM_BOT_TOKEN` is missing. Configure it via `onboard` instead and the same deploy succeeds.

**Why it drifted.** The generic `.env`/keyring writer lived in `surfaces/cli/wizard/env_sync.py`. `integrations/` (tier 2) and `tools/` (tier 2) cannot import `surfaces/` (tier 1), so the other two paths physically could not reach it and each wrote what they could.

**This PR** moves the generic half down to `config/` — the tier-4 floor everything may import — and leaves the wizard-specific half behind:

| Stays in `surfaces/cli/wizard/env_sync.py` | Moves to `config/env_file.py` |
| --- | --- |
| `sync_provider_env`, `sync_reasoning_model_env` (need `ProviderOption`) | `sync_env_values`, `sync_env_secret` |
| provider key ownership, stripping the old provider's keys on a switch | sensitive-key classification, `.env` line rewriting, `0o600` chmod, keyring persist |

`config/local_env.py` already owned *reading* the project env file (`get_project_env_path`, `_load_env_file`); writing now sits beside it. No import-linter exception was needed — `integrations → config` is an allowed edge, so the follow-up needs no layering debt entry.

Per AGENTS.md there is **no forwarding shim**: all 17 importers move to the canonical path in this change.

**Storage routing is decided by existing code, not by callers.** `is_sensitive_env_key()` classifies on the variable name, and the split is enforced rather than advised — `sync_env_values` refuses a sensitive key and `sync_env_secret` refuses a non-sensitive one. A mis-named credential fails loudly instead of landing in clear text.

Routing is name-driven, so a new integration gets correct storage without the author deciding anything:

```console
$ uv run python -c "from config.env_file import is_sensitive_env_key; ..."
TELEGRAM_BOT_TOKEN           -> keyring
TELEGRAM_DEFAULT_CHAT_ID     -> .env
DISCORD_PUBLIC_KEY           -> .env      # allowlisted: public despite ending in "key"
OPENAI_TOKEN_LIMIT           -> .env      # terminal word is "limit", not "token"
```

Layering holds, which is the whole point of the move:

```console
$ uv run lint-imports
Analyzed 1331 files, 3662 dependencies.
Config is independent KEPT
Contracts: 1 kept, 0 broken.
```

Full gate:

```console
$ make lint && make format-check
All checks passed!
2612 files already formatted

$ make typecheck
Success: no issues found in 1317 source files

$ make test-scope
2180 passed, 3 skipped, 2 warnings in 87.55s
```

---

This is a behavior-preserving refactor, so it is TDD-guarded per AGENTS.md. Both Telegram setup paths had **zero** test coverage, so the first commit characterizes them against the pre-refactor code — prompts asked, validate-before-persist ordering, the wizard's retry loop on a bad token, the exact credential shape written, the delivery advisory when no chat id is set, and the credential tiers each path touches. Those 35 tests passed before the move and stayed green through it.

I deliberately did **not** pin "the CLI path writes only the store". That is the gap being closed, and asserting it would lock the bug in.

The alternative I considered was injecting a writer callback into `integrations/` instead of moving the module — it avoids touching 17 files, but it pushes the decision of *where credentials go* back out to each caller, which is exactly the drift that caused this. Moving the writer to the floor means there is one answer and callers cannot disagree with it.

One real regression surfaced during the change, and it is worth calling out because it is the interesting part of the diff. `sync_reasoning_model_env` passed `env_path=None` and let `sync_env_values` fall back to its own module-level default. After the move that default resolved inside `config/env_file.py`, while the sibling writer `sync_provider_env` still resolved against `env_sync.PROJECT_ENV_PATH` — the name the `.env`-isolating test fixtures patch. The two writers disagreed about the target, so six tests escaped `tmp_path` and wrote to the real project `.env`. Fixed by resolving the default explicitly at the call site so both writers agree, and hardened by pinning the `config.env_file` default in both isolating fixtures (17 patch sites) so no test can reach the real file again.

**Follow-up (not in this PR):** with the writer reachable from `integrations/`, a shared `apply_setup(service, values)` can own merge → verify → persist-across-all-three-tiers, and the wizard, `integrations setup`, and the action tool from #4071 each keep only their own value collection. #4071 rebases onto that and shrinks to a thin adapter plus its execution gate.

---

### DIFF
--- a/config/env_file.py
+++ b/config/env_file.py
@@ -1,0 +1,227 @@
+"""Write credentials and settings to their correct storage tier.
+
+OpenSRE persists configuration in three places, and which one a value belongs in
+is decided by its **env var name**, not by the caller:
+
+* **system keyring** — anything :func:`is_sensitive_env_key` classifies as a
+  secret (``*_TOKEN``, ``*_KEY``, ``*_PASSWORD``, connection strings, …)
+* **project ``.env``** — everything else (URLs, ids, channels, model names)
+* the integration store — owned by ``integrations.store``, not this module
+
+The split is enforced rather than advised: :func:`sync_env_values` refuses a
+sensitive key and :func:`sync_env_secret` refuses a non-sensitive one, so a
+mis-classified credential fails loudly instead of landing in clear text on disk.
+Any ``.env`` rewrite also strips pre-existing secret assignments, so a file that
+predates the keyring does not keep leaking.
+
+This lives in ``config/`` — the layer floor — because every setup surface needs
+it: the onboarding wizard (``surfaces/``), ``opensre integrations setup``, and
+the interactive-shell action tools all persist the same credentials and must
+agree on where they go. ``config/local_env.py`` already owns *reading* the
+project env file; this owns writing it.
+"""
+
+from __future__ import annotations
+
+import os
+import re
+from contextlib import suppress
+from dataclasses import dataclass
+from pathlib import Path
+
+from config.llm_auth.credentials import delete as delete_provider_auth
+from config.llm_auth.credentials import save_api_key
+from config.llm_auth.provider_catalog import API_KEY_PROVIDER_ENVS
+from config.llm_credentials import delete_keyring_secret, save_keyring_secret
+from config.local_env import get_project_env_path
+
+PROJECT_ENV_PATH = get_project_env_path()
+
+_ENV_ASSIGNMENT = re.compile(r"^\s*([A-Za-z_][A-Za-z0-9_]*)\s*=")
+_NON_SECRET_ENV_KEYS: frozenset[str] = frozenset({"DISCORD_PUBLIC_KEY"})
+# Underscore-separated terminal tokens that mark an env var as sensitive.
+# Matching the terminal component (rather than a substring or a fixed suffix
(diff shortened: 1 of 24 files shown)


muddlebee:
@greptile review

Issue #4169, opened 2026-07-21
Agent-driven integration setup: one spec-driven action tool instead of one per integration


muddlebee:
## Problem

An integration can be configured from two surfaces today:

| Surface | Coverage |
| --- | --- |
| `opensre onboard` (wizard) | ~32 integrations |
| `opensre integrations setup <service>` | 45 integrations |
| Interactive shell / agent | **0** |

Asking the agent to "set up Telegram" does not work. #4071 proposes to fix that with a hand-written `configure_telegram` action tool — one action tool, for one integration. Repeating that shape 45 times is not a plausible path, and each copy would re-implement prompting, validation, and persistence.

## What makes a generic version possible

`integrations/setup_flow.py` introduced `IntegrationSetupSpec` — a declarative description of what an integration needs:

```python
TELEGRAM_SETUP = IntegrationSetupSpec(
    service="telegram",
    fields=(
        SetupField(name="bot_token", label="Telegram bot token",
                   env_var="TELEGRAM_BOT_TOKEN", secret=True),
        SetupField(name="default_chat_id", label="Default chat ID",
                   env_var="TELEGRAM_DEFAULT_CHAT_ID"),
    ),
    verify=verify_telegram,
    resolve=_resolve_default_chat_id,
)
```

That is exactly the metadata an agent needs to drive setup conversationally: which fields to ask for, which are required, which must be masked, and what to call them. So a single `configure_integration` action tool can serve every integration that has a spec, instead of one action tool per integration.

This depends on specs existing, which is #4168.

## Constraint to design around first: secrets in chat history

A credential typed into a chat message is persisted by that chat platform. In the local terminal REPL that is acceptable. Through the Slack or Telegram gateway it means a bot token sits in a channel's history indefinitely, outside OpenSRE's control and outside its masking layer.

`SetupField.secret` already marks which fields this applies to. Options, roughly in order of preference:

1. **Refuse secret fields on remote surfaces.** The agent collects non-secret fields, then tells the user to run `opensre integrations setup <service>` locally for the credential. Setup still gets easier without moving secrets through chat.
2. **Gate the whole tool to the local interactive shell.** Simplest to reason about; gives up remote setup entirely.
3. **Accept secrets everywhere with a warning.** Not recommended — the warning does not remove the message from the channel.

This should be decided before the tool is built, not retrofitted.

## Scope

- A `configure_integration` action tool driven by the spec, not per-integration copies.
[...]

Issue #4168, opened 2026-07-21
Migrate the remaining 44 integration setup handlers onto the shared setup flow


muddlebee:
## Problem

Every integration configured through `opensre integrations setup <service>` is written to the integration store **only**. The same integration configured through `opensre onboard` is written to the store, the system keyring, and `.env`.

Measured on `main`:

```
integrations/cli.py:  45 setup handlers
                       0 write the keyring or .env
surfaces/cli/wizard/configurators/: 10 of 14 files write the keyring
```

This is invisible during normal use, because runtime credential resolution checks the integration store first. It surfaces at deploy time — the preflight in `platform/deployment/ecr_deploy/prep.py` reads environment variables, not the store:

```python
if not _env_set("TELEGRAM_BOT_TOKEN"):
    missing.append(DeployEnvIssue("telegram_bot", env_vars=("TELEGRAM_BOT_TOKEN",)))
```

So an integration you configured, verified, and successfully sent messages with can still block `make deploy` as "missing". Whether it does depends only on which command you happened to use to set it up.

## Why it happened

The generic `.env`/keyring writer used to live in `surfaces/cli/wizard/env_sync.py`. `integrations/` is tier 2 and cannot import `surfaces/` (tier 1), so the CLI setup handlers physically could not reach it and each wrote what they could.

That part is fixed: #4167 moved the writer to `config/env_file.py` (tier 4, importable from anywhere), and the follow-up added `integrations/setup_flow.py` — an integration-agnostic `apply_setup(spec, values)` that owns validate → verify → resolve → persist-to-all-tiers.

Telegram is migrated onto it as the reference. The remaining 44 handlers are not.

## What is left

For each remaining integration:

1. Declare an `IntegrationSetupSpec` in the vendor package (`integrations/<vendor>/setup.py`), listing each credential, its `env_var` (or `None` for store-only), whether it is required, and whether it should be masked while typed.
2. Point the `integrations/cli.py` handler and the matching wizard configurator at `apply_setup(...)`.
3. Delete whatever duplicate validator the wizard was carrying — the registered verifier in `integrations/<vendor>/verifier.py` becomes the single prober. (Telegram had two `getMe` implementations, one on `httpx` and one on `requests`.)

Storage routing needs no per-integration decision: `config.env_file.is_sensitive_env_key()` classifies on the variable name and the split is enforced — `sync_env_values` refuses a sensitive key and `sync_env_secret` refuses a non-sensitive one.

[...]

PR #4170, opened 2026-07-21
feat(integrations): share one setup flow; require and resolve the Telegram chat id

muddlebee:
Stacked on #4167 — **base is `refactor/shared-integration-setup-flow`, not `main`.** Review #4167 first; this PR's own diff is the last commit.

Relates to #4071. Follow-ups filed as #4168 (remaining 44 integrations) and #4169 (agent-driven setup).

#4167 moved the `.env`/keyring writer down to `config/` so `integrations/` could reach it. Nothing used it yet — that PR changed no behavior. This one connects it.

**The problem.** Setting up an integration happens on three surfaces, and they persisted credentials differently:

| Surface | store | keyring | `.env` |
| --- | :-: | :-: | :-: |
| `opensre onboard` | ✅ | ✅ | ✅ |
| `opensre integrations setup telegram` | ✅ | ❌ | ❌ |
| `configure_telegram` action tool (proposed in #4071) | ✅ | ❌ | ❌ |

Runtime hides this — `integrations/telegram/credentials.py` resolves the store first — so the CLI path looks fine right up until deploy, where `platform/deployment/ecr_deploy/prep.py:75` checks the **env var**:

```python
if not _env_set("TELEGRAM_BOT_TOKEN"):
    missing.append(DeployEnvIssue("telegram_bot", env_vars=("TELEGRAM_BOT_TOKEN",)))
```

Configure Telegram with the CLI, verify it, send a message — then watch `make deploy` tell you `TELEGRAM_BOT_TOKEN` is missing. Configure the same integration with `onboard` and deploy succeeds. This is not Telegram-specific: **0 of the 45 setup handlers in `integrations/cli.py` write the keyring or `.env`.**

**The change.** `integrations/setup_flow.py` owns the whole lifecycle once — validate required fields → verify → resolve → persist to every tier — driven by a declarative spec:

```python
TELEGRAM_SETUP = IntegrationSetupSpec(
    service="telegram",
    fields=(
        SetupField(name="bot_token", label="Telegram bot token",
                   env_var="TELEGRAM_BOT_TOKEN", secret=True),
        SetupField(name="default_chat_id", label="Default chat ID",
                   env_var="TELEGRAM_DEFAULT_CHAT_ID"),
    ),
    verify=verify_telegram,
    resolve=_resolve_default_chat_id,
)
```

Callers keep only their own value collection. `_setup_telegram` and `_configure_telegram` both shrink to a prompt pair plus one `apply_setup(...)` call, and neither decides where anything is stored — `is_sensitive_env_key()` routes by variable name, and the split is enforced (`sync_env_values` refuses a secret, `sync_env_secret` refuses a non-secret).

**Two deliberate behavior changes**, both Telegram-specific:

*The chat id is now required.* A bot token alone passes `getMe`, so setup reported success — but `load_credentials_from_env()` raises without a chat id, so the watchdog, Hermes sinks, and the send-message tool all fail. The old wizard printed a warning and saved anyway; that just moved the failure to the first real alert, hours later, far from the setup that caused it.

*The chat id is resolved, not trusted.* `getChat` accepts a numeric id **or** an `@username`, so users can supply a public channel's `@name` — much easier to obtain than digging a numeric id out of `getUpdates` — and it is stored as the numeric id, which survives a channel rename. More importantly it answers the question `getMe` cannot: *can this bot see that chat?* The most common Telegram misconfiguration is forgetting to add the bot to the channel, and that now fails at setup.

One limit is documented rather than hidden: `getChat` proves reachability, not posting rights. A channel bot still needs the **Post Messages** admin permission, and only an actual send would prove it.

**Deleted:** `validate_telegram_bot`, the wizard's httpx copy of the `getMe` probe. The registered verifier in `integrations/telegram/verifier.py` is now the single prober; the deleted copy's test coverage — including its token-redaction case — moved to `tests/integrations/telegram/test_verifier.py` rather than being dropped.

image

Both Telegram setup paths had zero test coverage before #4167, so that PR characterized them first — prompts, validate-before-persist ordering, the wizard's retry loop, the exact credential shape, and the tiers each path wrote. Those tests are what made this change safe to make, and they are updated here only where behavior changed on purpose (chat id required, CLI path now writing all three tiers). Each such edit is called out in the test module docstrings, so a reviewer can tell an intended change from a regression.

The design question was where the "which tier does this value go to" decision lives. Three options:

1. **A writer callback injected into each caller.** Avoids moving code, but hands the decision back to every caller — precisely the drift that caused the bug.
2. **Make `upsert_integration` mirror to the keyring and `.env` itself.** Tempting, because it would fix all 45 handlers in one change. It does not work: the store key is `bot_token` while the env var is `TELEGRAM_BOT_TOKEN`, so a per-integration name mapping is required regardless. That mapping *is* the spec, so there is no way to avoid writing one per integration.
3. **A declarative spec consumed by one flow** — chosen. Callers declare what the integration needs; the flow decides where it goes. It also makes agent-driven setup tractable later (#4169), since a spec is exactly the metadata an agent needs to know what to ask for.

`verify` is wired onto the spec as a direct function reference rather than looked up in the verifier registry. The caller already knows which integration it is configuring, and an explicit reference keeps the flow independent of import-time registration ordering — which also means tests can swap it with `dataclasses.replace` and still exercise the real field definitions instead of a stand-in spec.

The `resolve` hook exists because verification and normalization are genuinely different steps: `verify` answers "do these credentials work" and returns a status, while Telegram additionally needs to *rewrite* what the user typed. Folding that into the verifier would have changed the `VerifierFn` contract for all 45 integrations to serve one.

**Not in this PR:** the other 44 integrations (#4168) and the agent-driven setup tool (#4169). #4071 rebases onto this and shrinks to a thin adapter plus the execution gate it is currently missing.

---

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,6 +24,40 @@
 - Do not keep compatibility-only forwarding modules after refactors. Once imports and tests
   are migrated, remove the old module path in the same change and use one canonical import path.
 
+### Docs under `docs/`
+
+`docs/` is **user-facing**. Write for someone configuring the product, not for
+someone who has read the diff. Ask of every sentence: *does this change what the
+reader does?* If not, cut it.
+
+Leave out:
+
+- **Vendor API endpoints and internal function names.** "The token is checked
+  against `getMe`" / "saved via `upsert_integration(...)`" — the reader calls
+  neither.
+- **Where a value is stored internally**, unless they must set it themselves.
+  Keyring vs `.env` vs the integration store is our problem, not theirs.
+- **The bug a change fixed**, and the mechanism behind it. That belongs in the PR
+  description and the module docstring.
+- **Restating what a code block already shows.**
+
+Keep:
+
+- **What is required vs optional**, and what happens when something is skipped.
+- **Shortcuts that save real work** ("a public channel's `@name` works, so you can
+  skip finding the numeric id").
+- **Gotchas that are invisible until they bite** ("the bot must be an admin to
+  post").
+- Exact commands, env-var names, and values the reader types or sets.
+
+Say it in the reader's terms: "a chat the bot was never added to fails during
+setup", not "`getChat` returns `ok: false`".
+
+Deep technical detail still gets written down — in the module docstring, the PR
+description, or a page under `docs/` explicitly aimed at contributors (e.g.
+[docs/ARCHITECTURE.md](docs/ARCHITECTURE.md)). It just does not belong on a
+setup page.
+
 ### Performance (algorithms & data structures)
 
 Apply on the **hot path** (per-request / per-iteration / per-tool-call); leave cold
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,37 +26,16 @@
 
 ### Docs under `docs/`
 
-`docs/` is **user-facing**. Write for someone configuring the product, not for
-someone who has read the diff. Ask of every sentence: *does this change what the
-reader does?* If not, cut it.
+`docs/` is user-facing. Test every sentence: *does this change what the reader
+does?* If not, cut it.
 
-Leave out:
-
-- **Vendor API endpoints and internal function names.** "The token is checked
-  against `getMe`" / "saved via `upsert_integration(...)`" — the reader calls
-  neither.
-- **Where a value is stored internally**, unless they must set it themselves.
-  Keyring vs `.env` vs the integration store is our problem, not theirs.
-- **The bug a change fixed**, and the mechanism behind it. That belongs in the PR
-  description and the module docstring.
-- **Restating what a code block already shows.**
-
-Keep:
-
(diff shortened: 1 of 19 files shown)


muddlebee:
@greptile review

muddlebee:
@greptile review

Devesh36 (commented):
quite a good approach for this ! Nice work 👍

Davidson3556:
Nice consolidation, this is great.
One thing you need to look at is the comment in apply_setup says "if the env write fails nothing is persisted at all," but _persist_env writes each keyring secret before the final sync_env_values, so a .env PermissionError leaves the token in the keyring. The guarantee that actually holds is that the store is never left holding credentials alone — worth wording it that way.

muddlebee:
demo updated

PR #4196, opened 2026-07-22
feat(integrations): migrate incident.io, Tracer, MongoDB Atlas, SigNoz, Jenkins, PagerDuty, Dagster, Temporal and Helm onto the shared setup flow

muddlebee:
Part of #4168. Follows #4191 (batch 1) and #4194 (batch 2, still open).

Third migration batch onto the shared setup flow: **incident.io, Tracer, MongoDB Atlas, SigNoz, Jenkins, PagerDuty, Dagster, Temporal, Helm**. All nine were "N fields + a registered verifier" with no branching prompts and — unlike batch 2 — no env-var name mismatches, so this batch needed no new spec features.

**Two more wizard configurators were silently dropping credentials**

| Configurator | Bug |
| --- | --- |
| `_configure_pagerduty` | called `sync_env_values({})` — neither `PAGERDUTY_API_KEY` nor `PAGERDUTY_BASE_URL` reached any tier but the store |
| `_configure_betterstack`* | same — `sync_env_values({})`, all four fields store-only |

\*Better Stack itself is **not** migrated in this batch (see below) — its wizard bug is left in place and flagged for whoever picks it up.

That makes seven such bugs across three batches (Datadog, Honeycomb, Coralogix, Sentry, Vercel, and now PagerDuty). `_configure_incident_io`, `_configure_jenkins` and `_configure_dagster` were already correct — migrated anyway so persistence has one implementation instead of two that happen to agree today.

**A false-positive secret classification, caught by the round-trip test**

`config.env_file.is_sensitive_env_key()` routes any env var whose name ends in `key` to the keyring. `MONGODB_ATLAS_PUBLIC_KEY` matched that even though it's a paired public identifier (used alongside a private key for Digest Auth), not a secret on its own — `tests/integrations/test_cli_spec_setup.py`'s keyring/env split assertion caught this immediately (it expected the public key in `.env`, got it in the keyring instead). Fixed the same way `DISCORD_PUBLIC_KEY` already was: added to the `_NON_SECRET_ENV_KEYS` allowlist in `config/env_file.py`, with a test case in `tests/cli/wizard/test_env_sync.py`.

**Tracer needed its own round-trip check**

`integrations._catalog_impl.load_env_integrations()` is the function every other spec's env round-trip goes through — but Tracer's env-only discovery isn't in it at all. It's a top-level fallback inside `resolve_effective_integrations()` instead (`JWT_TOKEN` checked directly, no per-vendor block in the env loader). `tests/integrations/test_setup_spec_env_round_trip.py` now special-cases Tracer to assert against `resolve_effective_integrations()` rather than `load_env_integrations()` — same guarantee, correct target.

**Helm's env-only discovery has a separate opt-in gate**

`load_env_integrations()` only surfaces Helm when `OSRE_HELM_INTEGRATION` is truthy — a deliberate manual switch (Helm shells out to a local binary) unrelated to any credential field. Not a bug, but the round-trip test needed to set that flag alongside the persisted values to exercise the discovery path at all; documented in `integrations/helm/setup.py`'s docstring so the next person doesn't mistake it for a missed `SetupField`.

**Betterstack and RDS were deliberately left out**

- `betterstack`'s `sources` field is a list downstream (`betterstack_extract_params` does `list(bs.get("sources", []))`), but every other `SetupField` is a plain string — persisting the field as-is would silently corrupt it (`list("a,b")` iterates characters). Needs a small fix to `integrations/betterstack/__init__.py` to normalize either shape, which is a real code change beyond wiring a spec, not a mechanical migration.
- `rds`'s current `_setup_rds` CLI handler prompts for `host`/`port`/`database`/`username`/`password`, but `integrations/rds/__init__.py`'s actual config (`classify`, `rds_is_available`, `rds_extract_params`) reads `db_instance_identifier`/`region` — fields the handler never asks for. `opensre integrations setup rds` has likely never produced a usable RDS integration. Worth its own PR rather than folding a behavior fix into a batch of otherwise-mechanical migrations.

Both are called out on #4168 for whoever picks them up next.

PagerDuty verify-before-persist, sandboxed `HOME`, bad key:

```
$ opensre integrations setup pagerduty

  Setting up pagerduty

❯ PagerDuty API key **********
❯ PagerDuty API base URL (press Enter to use default) https://api.pagerduty.com

  Validating pagerduty credentials...
  error: Connection failed: 401

$ cat $HOME/.opensre/integrations.json   →  (nothing saved)
$ cat $HOME/.env                          →  (nothing saved)
```

Before this change, `_configure_pagerduty` (the wizard path) would have saved the key to the store even after a failed probe elsewhere reported "ok" — the store and the failed check were never the same call.

---

Same problem as #4191/#4194 — three surfaces configured integrations and disagreed about where credentials land, so whether an integration blocked `make deploy` depended on which command you used. This batch moves nine more onto the one flow.

I picked these nine specifically because none of them have branching prompts, either/or field groups, or post-save side effects — the three gaps #4168 calls out as needing new spec features before they can move. Grafana, the databases, GitHub, AWS, Discord, Slack and Rocket.Chat are still blocked on those; Better Stack and RDS turned out to have their own vendor-specific issues once I looked closely (see above), so they're deferred rather than folded in here.

Key pieces:
- `integrations/<vendor>/setup.py` × 9 — declarative field lists; no logic.
- `config/constants/<vendor>.py` × 9 — new env-name modules, imported by both the writer (spec) and the reader (`load_env_integrations` / `resolve_effective_integrations`), so the two sides can't drift silently.
- Wizard configurators for incident.io, Jenkins, PagerDuty and Dagster collapse to one line each via `configure_from_spec`; Tracer, MongoDB Atlas, SigNoz, Temporal and Helm never had a wizard path (only `opensre integrations setup <service>`), so there was nothing to migrate there.

Edge cases covered by the extended parametrized suites: blank required field fails on that field rather than after every prompt (skipped only for Helm and Dagster, where every field now has a default — verified `next()` would raise `StopIteration` there before excluding them); defaulted fields (PagerDuty's base URL, Temporal's namespace, Helm's binary path) accept a bare enter; optional fields (incident.io's base URL, Dagster's API token, Temporal's API key, Helm's context/kubeconfig/namespace) may be left blank and clear every tier when blanked; failed verification persists nothing anywhere.

One mutation test to confirm the round-trip guard actually catches drift: pointed Temporal's namespace at a wrong env name and confirmed the suite fails with the exact mismatch, then reverted.

---


muddlebee:
Added a direct test: `test_sync_env_secret_with_blank_value_deletes_the_stored_secret` in `tests/cli/wizard/test_env_sync.py`, using the same `MemoryKeyring` backend as the existing `test_sync_env_values_routes_secrets_to_keyring`. It writes a real value via `sync_env_secret`, then writes blank, and asserts `resolve_env_credential` comes back empty — so the deletion path is now asserted directly rather than only implied through the mocked wizard test.

muddlebee:
@greptile review

PR #4198, opened 2026-07-22
perf(integrations): stop re-reading integrations.json on every startup

Devesh36:
### Problem

`load_integrations()` read and parsed `~/.opensre/integrations.json` from disk on **every call**, with no cache.

On REPL startup this happened multiple times before the first prompt:
- Session bootstrap loads integrations once
- Welcome banner loads them **twice more** (once for service names, once for health checks)

Same file, same data — parsed 3 times for no reason.

### Why fix it

Startup should stay fast, and this store is used in many places. Caching avoids repeat disk reads when the file has not changed. The banner path also had an easy win: load the store once instead of twice.

### What changed

- Cache store reads in memory; refresh when the file changes or is saved
- Writes still read fresh from disk (cache bypass)
- Banner health now loads the store once per call
- Added tests for cache behavior and single-load health

### DIFF
--- a/integrations/catalog.py
+++ b/integrations/catalog.py
@@ -169,6 +169,15 @@ def configured_integration_services() -> list[str]:
     login). Never raises; returns an empty list on any failure so callers can
     treat it as best-effort.
     """
+    try:
+        store_records = load_integrations()
+    except Exception:
+        store_records = []
+    return _configured_service_names(store_records=store_records)
+
+
+def _configured_service_names(*, store_records: list[dict[str, Any]]) -> list[str]:
+    """Merge env-visible and active store services into one deduplicated list."""
     services: list[str] = []
 
     try:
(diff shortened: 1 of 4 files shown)


Devesh36:
@greptile review

muddlebee:
@Devesh36 imo we dont need to cachify this, re-reading is OK as on every startup it takes up fresh values, caching might lead to stale values.

muddlebee:
Closing this in favor of a smaller follow-up: https://github.com/Tracer-Cloud/opensre/pull/4230

That PR only fixes the double read in the welcome-banner health path — load the file once, reuse it. No caching.

Why not cache:

- The file is tiny. Reading it again is fine.
- A cache can serve old data if the file changes while the process is still running (REPL / gateway stay up for a long time).
- Reviewers already flagged cases where the cache could keep wrong data around, or share objects that callers might accidentally mutate.
- Tests that swap the store path get messier with a global cache.

So keep the small cleanup, skip the cache. Thanks for the PR — the catalog half was the useful bit.

PR #4199, opened 2026-07-22
feat(integrations): migrate PostHog, Sentry and X MCP onto the shared setup flow

muddlebee:
Part of #4168. Follows #4191 (batch 1), #4194 (batch 2), #4196 (batch 3), and #4197 (batch 4).

Fifth migration batch onto the shared setup flow: **PostHog MCP, Sentry MCP, X MCP**.

These were the "constant / derived fields" gap from #4168 — their CLI handlers hard-code `mode`/`command`/`args` (and never prompt for them) with comments like "do NOT reintroduce a transport selection." Rather than inventing `SetupField.constant` for this batch, the specs only declare the fields a user is actually asked for. For these three hosted/streamable-HTTP servers, `PostHogMCPConfig` / `SentryMCPConfig` / `XMCPConfig` already default `mode` to `streamable-http` and leave `command`/`args` empty — the same values every setup path was hard-coding — so omitting them from the persisted credentials is equivalent to writing them.

**OpenClaw is intentionally not in this batch.** Its CLI hard-codes `mode=stdio`, but `OpenClawConfig`'s default is `streamable-http`. Omitting `mode` would silently flip transport. That one still needs `SetupField.constant` (or an explicit persist of the fixed stdio mode) before it can migrate.

**What this fixes for the three that did migrate**

Previously `opensre integrations setup posthog_mcp|sentry_mcp|x_mcp` only wrote the integration store. The matching wizard configurators already wrote store + keyring + `.env`. Same deploy-preflight footgun as the earlier batches. Both surfaces now go through `apply_setup`.

Env round-trip guard for the three MCP specs:

```
$ uv run python -m pytest tests/integrations/test_setup_spec_env_round_trip.py -q
.......                                                                  [100%]
7 passed
```

CLI spec-setup suite including the new handlers:

```
$ uv run python -m pytest tests/integrations/test_cli_spec_setup.py -q
.............................s                                           [100%]
29 passed, 1 skipped
```

(The skip is `x_mcp`: every required field has a default URL, so the "blank required field" case does not apply.)

---

#4168 listed these four as blocked on a new `SetupField.constant` feature. I checked whether that feature is actually required for each:

1. **posthog_mcp / sentry_mcp / x_mcp** — setup hard-codes the same defaults the pydantic config models already have for transport. Omitting those fields from the spec is safer than prompting for them (which `configure_from_spec` / `_run_spec_setup` would do for every field today) and avoids inventing a new spec feature for a case the config layer already covers.
2. **openclaw** — setup's fixed `stdio` conflicts with the config default (`streamable-http`). Omitting would change behavior. Left hand-rolled until `SetupField.constant` exists.

Key pieces:
- `integrations/{posthog_mcp,sentry_mcp,x_mcp}/setup.py` — prompted fields only; verifier wired explicitly.
- `config/constants/{posthog_mcp,sentry_mcp,x_mcp}.py` — shared env names for writer + catalog reader.
- Wizard MCP configurators collapse to `configure_from_spec`; REST PostHog/Sentry configurators left alone (batch 2 owns those).

Edge cases: optional `project_id` / `host` / `auth_token` clear every tier when blanked; URL defaults are offered as prompt prefills and applied by `apply_setup` so enter-to-accept works without a typed value.

---


muddlebee:
@greptile review

PR #4202, opened 2026-07-22
fix: address issue #4192 - [BUG] Grafana client cache reuses stale credentials and TLS configuration

### LINKED ISSUE #4192: [BUG] Grafana client cache reuses stale credentials and TLS configuration
### Summary

Expected:

Actual:

### Steps to reproduce

   - api_key=token-one
   - api_key=token-two

Observed:

Yes

### How often does it occur?

Under specific conditions

### Operating system

macOS

### Logs and error output

### Additional context

```
uv run opensre
> /verify grafana
> /verify grafana
```

`verify_ssl` or CA bundle.

kunalrawat425:
I have started working on this issue.

Closes #4192


kunalrawat425:
@greptile review

PR #4208, opened 2026-07-22
fix: address issue #3565 - GitHub Connect slow confirmation

### LINKED ISSUE #3565: ## GitHub Connect confirmation opens one MCP session per probe call (3 sessions per confirmation)
| Field | Value |
|-------|-------|
| **Level** | Junior→Middle |
| **Complexity** | S–M |
| **Tier** | Tier 2 — Junior→Middle |
| **Category** | bugfix |

**Symptom.** After entering a token in `opensre integrations setup`, the GitHub "connected" confirmation runs noticeably long with no feedback. :Each call pays a fresh connection + MCP handshake before doing any GitHub work.

**Root cause.** Every MCP helper opens its own session. `_list_tools_async` (`integrations/github/mcp.py:654`) and `_call_tool_async` (`:674`) each enter `_open_github_mcp_session` (`:494`), which builds a new transport connection to the hosted endpoint and runs the MCP `initialize()` handshake (`:570`) — and each sync wrapper spins up its own event loop. `validate_github_mcp_config` (`:1207`) chains 3+ such calls sequentially (list_tools → get_me → one or more repo probes), so connection setup is paid 3+ times for one confirmation.

**Measured (instrumented `validate_github_mcp_config`, healthy config, fast
network):**

    total wall time         :   7.9 s
    MCP sessions opened     : 3
      session 1: connect+initialize  0.6 s, tool work  1.2 s   (list_tools)
      session 2: connect+initialize  0.4 s, tool work  1.6 s   (get_me)
      session 3: connect+initialize  0.4 s, tool work  2.5 s   (search_repositories)
    handshake overhead      :   1.4 s, plus ~1.3 s per-call event-loop setup

On a fast network the redundant setup costs ~2 s of ~8 s. On high-latency networks the handshake is several round trips per session, which is where the earlier ~25 s measurement comes from — that environment is the main beneficiary. The ~5 s of GitHub-side tool latency is not affected by any client fix, so do not expect the confirmation to become instant.

**Fix (smallest scope, `integrations/github/mcp.py` only).** Run the whole validation inside one session: an async internal that opens `_open_github_mcp_session` once and performs list_tools, get_me, and the repo probes on that session; keep the existing sync `validate_github_mcp_config` signature as the facade. The steps stay sequential (list_tools gates the probe plan; get_me supplies the username for probe arguments) — do not parallelize. No change to the checks themselves, their messages, or `failure_category` values.

**Regression coverage.** A test with a fake session/transport that counts session opens: validation of a healthy config must open exactly one session. Assert connection counts, not wall-clock — timing assertions are flaky in CI.

**Done when:**
- [ ] One MCP session per validation run, pinned by the connection-count test
- [ ] Existing verifier tests green (semantics, messages, failure categories unchanged)
- [ ] Before/after wall time reported in the PR from the measurement script below
- [ ] CI green

**Out of scope (file separately if wanted):**
- Wizard progress feedback during the remaining seconds of validation  (a per-step status line — that is what makes the confirmation *feel* fixed on any network).
- Session reuse for investigation-time GitHub tool calls, which pay the same per-call setup on every agent step — needs session-lifecycle design, not a good first issue.

kunalrawat425:
I have started working on this issue.

Closes #3565


muddlebee:
@kunalrawat425 pls dont send PRs unless assigned.

PR #4223, opened 2026-07-23
feat(auth): org-scoped Clerk M2M tokens for silo → webapp calls

YauhenBichel:
Silo → webapp requests (credit metering and the integrations vault) now authenticate with a short-lived, org-scoped Clerk M2M token instead of a shared static secret.

The silo mints an `mt_…` token from its own `CLERK_MACHINE_SECRET_KEY` and sends it as the bearer credential. The webapp verifies the token against the org's bound machine, so a silo can only ever act as its own tenant. `AGENT_USAGE_SECRET` remains as the fallback credential and is still accepted for metering; the vault route accepts M2M only.

A shared `AGENT_USAGE_SECRET` cannot prove *which* tenant is calling. Any silo holding it could act as another organisation — the credential carried no tenant identity. Minting per-silo tokens bound to the org's machine closes that.

How

- `integrations/slack/clerk_m2m.py` — mints and caches the token in-process
  until shortly before expiry. The cache is keyed by the machine secret, so a
  rotated secret cannot serve a stale token, and is guarded by a lock: the mint
  runs under it deliberately, so concurrent turns coalesce into a single mint
  rather than stampeding Clerk. Fails closed — any error returns `""`.
- `integrations/slack/agent_auth.py` — owns the credential choice (M2M when a
  machine secret is set, shared secret otherwise), so the credits client and the
  vault cannot drift apart on that decision. Never raises: a mint failure
  degrades to the fallback rather than breaking the turn that triggered it.
- `config/constants/billing.py` keeps only env-var names and tunables. The
  selection logic lives in `integrations/` because minting reaches
  `clerk_m2m` — putting it in the constants leaf would make `config` depend
  upward on `integrations` and form an import cycle.
- `webapp_vault_configured()` checks that a credential is *configured* rather
  than resolving one, so the predicate stays a cheap env read instead of
  triggering a network mint.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4225, opened 2026-07-23
fix(deps): bump pyasn1 to 0.6.4 for the REAL float-conversion DoS

YauhenBichel:
Bumps the transitive dependency `pyasn1` from 0.6.3 to 0.6.4, closing the uncontrolled-resource-consumption advisory in `univ.Real` float conversion.

Lock-only change: 3 insertions, 3 deletions in `uv.lock` — the package's version and hashes. No other dependency moved, and `pyproject.toml` is untouched

`pyasn1` is purely transitive, reached via `google-auth` -> `pyasn1-modules` and `google-api-python-client`. Neither parent constrains it below 0.6.4, so `uv lock --upgrade-package pyasn1` resolved cleanly without adding a direct dependency on a package we do not import.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -2832,11 +2832,11 @@ wheels = [
 
 [[package]]
 name = "pyasn1"
-version = "0.6.3"
+version = "0.6.4"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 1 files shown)


YauhenBichel:
@greptile review

PR #4226, opened 2026-07-23
fix(deps): bump setuptools to 83.0.0 for the MANIFEST.in NFC/NFD exclusion bypass

YauhenBichel:
Bumps `setuptools` from 82.0.1 to 83.0.0, closing the MANIFEST.in exclusion bypass advisory (Unicode NFC/NFD normalization collision on macOS APFS/HFS+).

Lock-only change: 3 insertions, 3 deletions in `uv.lock` — the package's version and hashes. No other dependency moved.

Exposure here is nil, and worth stating plainly

The advisory is that `MANIFEST.in` exclude / global-exclude / prune directives silently fail to match when the on-disk name is NFD and the rule is NFC, so an intentionally-excluded file gets packed into the sdist.

This project has no `MANIFEST.in`. There are no exclusion rules, so there is nothing to bypass. The bump is alert hygiene, not a fix for a live risk.

Nor is setuptools broadly present at runtime. It enters only through `pyinstaller` / `pyinstaller-hooks-contrib`, which live in the `release` and `release-binary` optional extras and are not installed for dev or runtime. It is additionally the PEP 517 build backend via `[build-system] requires = ["setuptools>=68.0"]`, which resolves independently of the lock.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -3709,11 +3709,11 @@ wheels = [
 
 [[package]]
 name = "setuptools"
-version = "82.0.1"
+version = "83.0.0"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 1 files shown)


YauhenBichel:
@greptile review

PR #4242, opened 2026-07-25
fix(grafana): key the client cache on credentials, not just endpoint

### LINKED ISSUE #2836: Grafana integration: investigation runtime sends empty token → 401 on /api/datasources (Windows, v0.1 + nightly
### Summary

On Windows, every opensre investigate run fails to reach Grafana datasources. The runtime builds the integration account (account_id=user_integration) with the correct endpoint but appears to send an empty/missing token, so Grafana returns 401 Unauthorized on /api/datasources (and /api/ruler/...). As a result Loki/Mimir/Tempo queries never run and every investigation ends in "Unable to determine root cause" (validity 0.5), even though the configured service-account token is valid and the datasource exists.

This reproduces on both the released build and the --main nightly (both report opensre 0.1).

Environment
OS: Windows 11 Enterprise (10.0.26100), AMD64
OpenSRE: 0.1 — reproduced on both release and --main channels (native .exe, ~/.local/bin/opensre.exe)
Python (bundled): 3.13.13
Grafana: native Windows service at http://localhost:3035
Loki: grafana/loki:3.0.0 at http://localhost:3100, registered as a Grafana datasource (type=loki, uid=efp8mxnnpvr40d)
LLM: Ollama (LLM_PROVIDER=ollama, qwen2.5:3b) — works fine, not related to this bug

Configure a Grafana integration with a valid service-account token: opensre integrations setup grafana → endpoint http://localhost:3035, token glsa_… (opensre integrations list and opensre doctor both report it active / "1 configured: grafana".)
Ensure Grafana has a Loki datasource and some logs.
Run an investigation against a Grafana alert payload: opensre investigate -i sample_grafana_alert.json -o out.json

The runtime authenticates to Grafana with the configured token, discovers datasource UIDs, and queries Loki/Mimir/Tempo.

### Steps to reproduce

[grafana] Failed to discover datasource UIDs: 401 Client Error: Unauthorized for url: http://localhost:3035/api/datasources
[grafana] Failed to query alert rules: 401 Client Error: Unauthorized for url: http://localhost:3035/api/ruler/grafana/api/v1/rules
The investigation completes but query_grafana_logs/metrics/traces all return "<source> datasource not found", and root_cause = "Unable to determine root cause", validity_score = 0.5.
The exact token stored in ~/.config/opensre/integrations.json (credentials.api_key) works via curl:
GET /api/datasources              with Bearer <token>  -> HTTP 200  (returns the Loki datasource, uid efp8mxnnpvr40d)
GET /api/datasources              without auth          -> HTTP 401  (matches what the runtime gets)
GET /api/ruler/grafana/api/v1/rules with Bearer <token> -> HTTP 200
The same token even had enough permission to create the datasource via POST /api/datasources (HTTP 200). So the credential is valid and sufficiently scoped; the runtime simply isn't sending it.

Things tried that did NOT fix it
Valid token present in integrations.json under credentials.api_key (confirmed read by list/show/doctor).
Added the token under 8 alias keys in credentials (token, api_token, read_token, service_account_token, bearer_token, key, apiKey, readToken) — no change.
Set documented env vars: GRAFANA_INSTANCE_URL=http://localhost:3035, GRAFANA_READ_TOKEN=<token>, GRAFANA_LOKI_DATASOURCE_UID=efp8mxnnpvr40d — no change.
Verified env propagation from the launching shell to the native .exe works (cmd/powershell children see the vars), so the env vars do reach the process.
This strongly suggests the runtime Grafana client reads the endpoint from the integration store but resolves the token from a different field/path that ends up empty (a schema mismatch between the v2 integrations store and the runtime credential resolver).
--main nightly additionally probes a hardcoded http://localhost:3000/api/workspaces/1/logstore/all/search?... (note port 3000, not the configured 3035) and gets 404. Looks like a hardcoded/default host.
opensre integrations verify <svc> always reports missing — "Not configured in local store or env." even when list/show/doctor confirm the integration is active. (Likely the same credential-resolution path as the runtime, so probably the same root cause.)

Yes

### How often does it occur?

Every time

### Operating system

Windows

### Logs and error output

_No response_

### Additional context

The Grafana integration is effectively unusable for RCA on Windows: investigations can authenticate to nothing in Grafana and therefore can never produce a data-backed root cause.

### LINKED ISSUE #4192: [BUG] Grafana client cache reuses stale credentials and TLS configuration

SwapnilDey-git:
## Summary

`get_grafana_client_from_credentials` cached clients under `f"creds_{account_id}_{endpoint}"` alone. Rotating a service-account token, switching Basic Auth credentials, or changing `verify_ssl`/`ca_bundle` therefore kept silently returning the client built from the *previous* configuration until process restart.

- Fold a SHA-256 fingerprint of `api_key`/`username`/`password`/`verify_ssl`/`ca_bundle` into the cache key, so any credential or TLS change builds a fresh client. The fingerprint is hashed rather than stored raw so the cache key never carries a live secret.
- Evict the superseded cache entry for that `account_id`+`endpoint` instead of leaking it in memory forever.
- Add a `clear_grafana_client_cache()` test helper.

This is filed exactly as **#4192** describes it. It's also a plausible root cause of **#2836** ("empty token → 401 on every investigation, every time, on Windows"): all runtime call sites share the default `account_id="user_integration"`, so if any early call (e.g. a health-check/verify path) resolves with an empty/wrong token first, the poisoned client gets cached and every subsequent call — even with a correct token — reuses it for the rest of the process, matching the reported symptom of *persistent*, *every-time* 401s despite a valid, correctly-scoped token.

Fixes #4192. Likely fixes #2836.

## Test plan
- [x] Added regression tests: token rotation, Basic Auth/TLS change, and same-config reuse all produce the expected cache behavior; a separate test asserts the stale entry is evicted, not merely shadowed.
- [x] `pytest tests/integrations/grafana/ tests/tools/test_grafana_*` — 182 passed, no regressions.

### DIFF
--- a/integrations/grafana/client.py
+++ b/integrations/grafana/client.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import hashlib
 import logging
 
 from integrations.grafana.base import GrafanaClientBase
@@ -13,6 +14,28 @@ from integrations.grafana.tempo import TempoMixin
 logger = logging.getLogger(__name__)
 
 _grafana_client_cache: dict[str, GrafanaClient] = {}
+
+
+def _credential_fingerprint(
+    *,
+    api_key: str,
+    username: str,
+    password: str,
+    verify_ssl: bool,
+    ca_bundle: str,
+) -> str:
+    """Hash everything that changes how requests authenticate or verify TLS.
+
+    Used as part of the client cache key so rotating a token, switching auth
+    mode, or changing TLS trust invalidates the cached client instead of
+    silently reusing one built from the previous credentials (#4192). Hashed
+    rather than stored raw so the cache key never carries a live secret in
+    memory/logs.
+    """
+    fingerprint_input = "\x1f".join(
+        (api_key, username, password, str(verify_ssl), ca_bundle)
+    ).encode("utf-8")
+    return hashlib.sha256(fingerprint_input).hexdigest()
 
 
 class GrafanaClient(LokiMixin, TempoMixin, MimirMixin, GrafanaClientBase):
@@ -46,9 +69,26 @@ def get_grafana_client_from_credentials(
     ca_bundle: str = "",
 ) -> GrafanaClient:
     """Create a Grafana client from integration credentials."""
-    cache_key = f"creds_{account_id}_{endpoint}"
+    fingerprint = _credential_fingerprint(
+        api_key=api_key,
+        username=username,
+        password=password,
+        verify_ssl=verify_ssl,
+        ca_bundle=ca_bundle,
+    )
+    cache_key_prefix = f"creds_{account_id}_{endpoint}_"
+    cache_key = f"{cache_key_prefix}{fingerprint}"
     if cache_key in _grafana_client_cache:
         return _grafana_client_cache[cache_key]
+
+    # Drop any client cached under the old credentials for this account_id +
+    # endpoint — it is superseded and would otherwise linger indefinitely.
+    for stale_key in [
+        key
+        for key in _grafana_client_cache
+        if key.startswith(cache_key_prefix) and key != cache_key
+    ]:
+        del _grafana_client_cache[stale_key]
 
     config = GrafanaAccountConfig(
         account_id=account_id,
@@ -91,3 +131,8 @@ def get_grafana_client_from_credentials(
 
     _grafana_client_cache[cache_key] = client
     return client
+
+
+def clear_grafana_client_cache() -> None:
+    """Drop every cached client; test-only, real processes never need this."""
+    _grafana_client_cache.clear()
--- a/tests/integrations/grafana/test_client.py
+++ b/tests/integrations/grafana/test_client.py
@@ -2,7 +2,20 @@ from __future__ import annotations
 
 from unittest.mock import patch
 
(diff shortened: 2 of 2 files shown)


SwapnilDey-git:
Addressed both P2 findings in cf2410c:

1. **Normalize endpoint cache identity** — the cache key now uses the same `.rstrip("/")`-normalized endpoint that `GrafanaAccountConfig.instance_url` already uses, so a trailing slash no longer creates a second, never-evicted cache entry.
2. **Synchronize cache initialization** — the whole check-build-insert sequence (including the blocking `discover_datasource_uids()` call) is now guarded by a lock, so concurrent callers with the same new credentials can't both miss the cache and both pay for datasource discovery.

Added regression tests for both (trailing-slash reuse, and a multi-threaded test asserting discovery is invoked exactly once under concurrent calls).

@greptile review

Issue #4212, opened 2026-07-22
explore(env): harden is_sensitive_env_key for missed secret name shapes


muddlebee:
## Summary

`config.env_file.is_sensitive_env_key` decides keyring vs `.env` by heuristic:

- last `_`-segment in `{token, secret, password, passwd, key, apikey, credential(s)}`
- or substring `connection_string`
- with an allowlist for known public keys (`DISCORD_PUBLIC_KEY`, `MONGODB_ATLAS_PUBLIC_KEY`)

That works for conventional `*_PASSWORD` / `*_TOKEN` / `*_KEY` names used by setup specs, but the classifier can miss (or over-classify) other real-world shapes.

## Problem

**False negatives (leak risk):** a secret-ish name whose terminal token isn’t in the set goes to plain `.env`. Examples to explore:

- `*_PASSPHRASE`, `*_PASS` (terminal `pass` / `passphrase`)
- `*_BEARER`, `*_AUTH`, `*_JWT`
- `*_CLIENT_SECRET` is fine (terminal `secret`); what about `*_APP_SECRET_KEY` variants already covered by `key`?
- multi-segment endings like `*_API_TOKEN_V2` (terminal `v2`)
- vendor-specific jargon (`*_DSN` with embedded credentials? usually URL — unclear)

**False positives (annoying):** public or non-secret values forced into keyring / refused by `sync_env_values`, requiring `_NON_SECRET_ENV_KEYS` one-offs.

## Goal

Audit and optionally expand the classifier so product env names we care about don’t silently land secrets on disk — without turning every `*_KEY` public id into a keyring write.

## Explore / decide

1. Inventory env var names already used in `config/constants/**` and setup specs; classify which are secrets vs config.
2. List gaps against `_SENSITIVE_TERMINAL_TOKENS` / `_SENSITIVE_SUBSTRINGS`.
3. Decide expansions that are safe (e.g. `passphrase`, `bearer` if we use those names) vs ones that create false positives.
4. Document naming convention for new integrations (“last segment must be a sensitive token”) so heuristics stay the source of truth.
5. Explicitly defer (unless needed later): per-`SetupField` storage flag — keep tier derived from the name.

## Non-goals

- One-shot migration/backfill of store-only legacy installs into keyring.
- Treating this as “just add a test” — tests can follow once the classifier policy is chosen; the work item is the heuristic itself.

## Context

[...]

Issue #3348, opened 2026-06-30
bug(cli): onboard fails on macOS when system Keychain is locked/inaccessible


0xpaulx:
## Summary

`opensre onboard` fails on macOS when the user enters their LLM API key. The wizard cannot save the key to the macOS system Keychain and exits with a non-fatal-looking error that is actually fatal — the onboarding does not complete.

## Environment

- **OS:** macOS (Apple Silicon / Intel — reproduces on both)
- **Keyring backend detected:** `keyring.backends.macOS.Keyring`
- **Command:** `opensre onboard`

## Steps to Reproduce

1. Clone the repo (up-to-date `main`)
2. Run `opensre onboard`
3. Choose **OpenAI** as LLM provider, choose **API key** auth method
4. Enter a valid `OPENAI_API_KEY` when prompted

## Expected

Onboarding completes successfully and the key is persisted (either in the macOS Keychain or a supported fallback).

## Actual

The wizard exits with:

```
✗  Secure local credential storage is unavailable on this machine.
▲  OpenSRE could not save your API key to the local system keychain.
   Current keyring backend: keyring.backends.macOS.Keyring.
   Make sure your system keychain service is installed and unlocked, then rerun `opensre onboard`.
   For deeper diagnostics run `python -m keyring diagnose`.
```

The key is **not saved** and onboarding does not finish.

## Why this happens

On macOS the `keyring` library correctly detects the native Keychain backend, but the Keychain access may fail if:
- The login Keychain is locked (e.g. after waking from sleep without re-authenticating)
- The app has not been granted Keychain access (first-run permission dialog was dismissed)
- The user is running in a context where macOS Keychain is unavailable (e.g. certain CI or sandboxed environments)

The error is treated as fatal: there is no fallback (e.g. project-level `.env` or `~/.config/opensre/.env`) offered to the user.

## Possible Workarounds (before a fix)

- Unlock the macOS login Keychain in **Keychain Access.app**, then rerun `opensre onboard`
- Set the env var in the shell before onboarding: `export OPENAI_API_KEY=sk-...`
- Run `python -m keyring diagnose` for deeper diagnostics

## Related

- #1403 — same root cause on headless Linux/EC2 (no keyring backend at all)
- A `.env` fallback when keyring save fails would fix both issues

---
*Reported via Slack by @Gust Vandecandelaere on a cloned, up-to-date macOS repo.*

Davidson3556:
[...]

Issue #1403, opened 2026-05-06
bug(cli): onboarding exits on fresh EC2/headless Ubuntu when system keychain is unavailable


muddlebee:
## Summary

`opensre onboard` fails on a **fresh Ubuntu 22.04 LTS EC2** instance after entering an LLM API key (e.g. Anthropic). The wizard aborts because OpenSRE cannot persist the key to a system keychain.

## Environment

- OS: Ubuntu 22.04 LTS
- Host: New / minimal EC2 instance (typical SSH session, no desktop keyring)
- Command: `opensre onboard`

## Steps to reproduce

1. Run `opensre onboard`
2. Choose **Quickstart**
3. Choose **Anthropic** (or another API-key provider)
4. Enter a valid API key when prompted

## Expected

Onboarding should complete (or offer a supported fallback for server/headless setups, e.g. project `.env`).

## Actual

The process stops with an error (red text), for example:

> Secure local credential storage is unavailable. Set ANTHROPIC_API_KEY in your shell or configure a working system keychain.

The user is returned to the shell and Quickstart does not finish.

## Why this happens

On headless Linux, Python `keyring` often has **no working backend** (e.g. `keyring.backends.fail.Keyring`). There is no GNOME Keyring / D-Bus session as on a desktop, so saving the API key to the “secure local credential store” fails and the wizard treats that as fatal.

## Possible workarounds (before a code fix)

- Export the key in the shell: `export ANTHROPIC_API_KEY=...` before running onboarding, if supported by the version in use
- Install and configure a system keychain stack (e.g. gnome-keyring + D-Bus); heavy for a minimal server

## Related

- Fix in progress: PR adding `.env` fallback when keyring save fails (e.g. **#1399**).

---
*Reported from: fresh EC2 Ubuntu 22.04 Quickstart + Anthropic API key flow.*

muddlebee:
Possible fix: https://github.com/Tracer-Cloud/opensre/pull/1399 — add `.env` fallback when keyring is unavailable.

Yajush-afk:
@muddlebee I’m going to follow the approach in draft PR [#1399](https://github.com/Tracer-Cloud/opensre/pull/1399): keep keychain as the default, but if it’s unavailable on headless Linux, fall back to saving the active LLM key in project `.env`. I’ll also make sure `.env` is locked down to `0600` and add coverage for the TTY prompt path.

Please **assign** me this issue.

Yajush-afk:
[...]

PR #4243, opened 2026-07-25
fix: fall back to local storage when the OS keyring is unavailable

### LINKED ISSUE #3295: bug(cli): macOS prompts keyring authorization 3 times on every app launch
## Summary

On macOS, the system keychain authorization dialog is shown **3 times** on every app launch, instead of once (or not at all after the first approval).

## Expected Behavior

The keyring should request authorization once. After the user grants access, subsequent runs should not re-prompt (or at most prompt once per session).

## Actual Behavior

macOS asks for keyring/keychain access authorization **3 times** in succession every time the app is used.

## Impact

- Poor user experience on macOS
- Confusing for new users who may not understand why they're being asked repeatedly
- Likely indicates the app is creating/accessing multiple separate keyring entries or using separate keyring calls where one consolidated call would suffice

## Possible Root Cause

The app may be making 3 separate `keyring.get_password()` / `keyring.set_password()` calls for different credentials on startup rather than batching or caching the unlock. Each individual call triggers macOS Keychain's authorization prompt.

## Steps to Reproduce

1. Install opensre on macOS
2. Run the app
3. Observe 3 keychain authorization dialogs appear sequentially

## Environment

- macOS (all versions)

## Fix Direction

- Audit all keyring accesses at startup and consolidate where possible
- Consider caching the keyring unlock so subsequent reads in the same session don't re-prompt
- Related: #2534 (credential sync gaps), #3291 (keyring read/write inconsistency)

SwapnilDey-git:
## Summary

`save_keyring_secret`/`sync_env_secret` had no fallback: any OS keychain failure raised a `RuntimeError` that onboarding treated as fatal, with no path to a working state. One root cause, four symptoms:

- **#3348** — `opensre onboard` fails outright on macOS when the login Keychain is locked (e.g. after waking from sleep).
- **#1403** — `opensre onboard` fails outright on headless Linux/EC2, where `keyring` has no working backend at all (`keyring.backends.fail.Keyring`).
- **#3591** — "better LLM key failure UX during onboarding" *is* this fix, once a fallback exists.
- **#3295** — macOS shows 3 separate Keychain authorization prompts per run, because each read/write of the same credential (status check, save, verify) hit the OS backend independently with no per-process memoization.

### What changed

`config/llm_keyring.py`:
- `save_secret_with_fallback` / `resolve_secret_with_fallback`: try the OS keyring first, and only fall back to a local JSON store (owner-only permissions, under `OPENSRE_HOME_DIR`, distinct from the project `.env` that `config/env_file.py` deliberately keeps secret-free) when the keyring backend is unavailable. Raises only when *both* tiers fail.
- `fallback_store_path()` honors an `OPENSRE_FALLBACK_SECRETS_PATH` override, mirroring the existing `get_project_env_path()` pattern, so tests (and anyone else) can redirect it off `~/.opensre`.
- A per-process `_KeyringSession` memoizes successful reads and flips a sticky `backend_unavailable` flag on the first hard failure, so a run that touches the same credential more than once (status check → save → verify) hits the OS backend at most once instead of re-prompting macOS Keychain authorization on every call.

Wired through `config/env_file.py` (`sync_env_secret`, used by every integration's setup flow, not just LLM keys), `config/llm_auth/credentials.py` (`save_api_key`, `resolve_for_request`), and the wizard's `_persist_llm_api_key`, which now prints a clear (non-fatal) warning when the fallback tier was used instead of aborting onboarding.

Also hardens `is_sensitive_env_key` (**#4212**) with `passphrase`/`pass`/`bearer`/`jwt`/`auth` terminal tokens, so those credential shapes route to secure storage instead of landing in plaintext `.env`. Checked against `config/constants/**` for false positives (e.g. `LLM_AUTH_METHOD`'s terminal segment is `method`, not `auth`).

Fixes #3295, #3348, #1403, #3591, #4212.

## Test plan
- [x] New tests: fallback save/resolve round-trip when the keyring is disabled/unavailable, "both tiers fail" still raises, and the expanded `is_sensitive_env_key` classifications (including non-regressions like `LLM_AUTH_METHOD`, `PROXY_PASSTHROUGH`).
- [x] Updated `tests/cli/wizard/test_flow.py`: removed now-redundant `save_keyring_secret` stubs (every test already runs with `OPENSRE_DISABLE_KEYRING=1` and an isolated fallback path via the root conftest) and updated the "both tiers unavailable" regression test's expected message.
- [x] Added a `reset_keyring_session()` test hook and wired it into the existing `_disable_system_keyring` autouse fixture in `tests/conftest.py` — the new sticky `backend_unavailable` flag is process-global, so a test that intentionally provokes a backend failure would otherwise leak that state into unrelated later tests in the same worker.
- [x] `pytest tests/config/ tests/cli/wizard/ tests/cli/test_auth_command.py tests/integrations/test_*credential_resolve.py tests/interactive_shell/test_model_credential_entry.py` — 377 passed (5 pre-existing, unrelated failures confirmed present on `main` too: Windows-only perf-threshold/encoding/`os.fpathconf` issues, nothing touching credential storage).

### DIFF
--- a/config/env_file.py
+++ b/config/env_file.py
@@ -32,7 +32,8 @@ from pathlib import Path
 from config.llm_auth.credentials import delete as delete_provider_auth
 from config.llm_auth.credentials import save_api_key
 from config.llm_auth.provider_catalog import API_KEY_PROVIDER_ENVS
-from config.llm_credentials import delete_keyring_secret, save_keyring_secret
+from config.llm_credentials import delete_fallback_secret, delete_keyring_secret
+from config.llm_credentials import save_secret_with_fallback
 from config.local_env import get_project_env_path
 
 PROJECT_ENV_PATH = get_project_env_path()
@@ -53,10 +54,19 @@ _SENSITIVE_TERMINAL_TOKENS: frozenset[str] = frozenset(
         "secret",
         "password",
         "passwd",
+        "pass",
+        "passphrase",
         "key",
         "apikey",
         "credential",
         "credentials",
+        # Bearer/JWT values authenticate a request on their own, same risk
+        # class as a token. Bare "auth" is included for names like
+        # `*_AUTH` that store a credential rather than a method/flag (see
+        # `LLM_AUTH_METHOD`, whose terminal segment is "method", not "auth").
+        "bearer",
+        "jwt",
+        "auth",
     }
 )
 _SENSITIVE_SUBSTRINGS: tuple[str, ...] = (
@@ -122,8 +132,11 @@ def _ensure_no_sensitive_env_lines(lines: list[str]) -> None:
             )
 
 
-def _persist_env_secret(key: str, value: str) -> bool:
-    """Store a secret in the keyring. Returns False when keyring is unavailable."""
+def _persist_env_secret(key: str, value: str) -> str | None:
+    """Store a secret in the keyring, falling back to local storage if it's
+    unavailable (see #1403, #3348). Returns the storage tier used
+    (``"keyring"`` or ``"fallback"``), or ``None`` if both tiers failed.
+    """
     normalized = value.strip()
     provider = next(
         (name for name, env_var in API_KEY_PROVIDER_ENVS.items() if env_var == key),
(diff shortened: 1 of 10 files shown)


SwapnilDey-git:
Addressed all three P1 findings in 6b96682:

1. **Logout leaves fallback credential active** — `delete()` now clears the fallback-store copy alongside the keyring entry, not just the keyring one.
2. **Backend exceptions bypass fallback** — `resolve_for_request` only caught `keyring.errors.KeyringError`; widened to also catch the bare `RuntimeError`/`OSError` that SecretService/other backends can raise on the same "unreachable" condition, so it actually reaches the fallback store instead of crashing request-time resolution.
3. **Fallback file permissions are best-effort** — rewrote `_write_fallback_store` to create the file via `os.open(..., 0o600)` up front instead of `write_text()` + suppressed `chmod()`, so there's no window where it's readable beyond the owner, and a permission failure on a *new* secret now fails the save instead of silently persisting with broader access. (Cleanup-only deletes of an already-secured, now-stale entry still swallow a permission error, since that's a stale-file nuisance, not a new secret exposed — noted in the docstring.)

Added regression tests for all three (`tests/config/test_llm_credentials.py`).

@greptile review

muddlebee:
Pls open a github issue first, then we'll see.

PR #4246, opened 2026-07-25
fix(agent): compact conversation overflow into a running summary instead of dropping turns

### LINKED ISSUE #4085: [BUG] Agent forgets earlier conversation: 24-message hard cap drops turns before compaction fires
## Summary

In a long interactive-shell conversation the agent forgets earlier turns. The conversation history is hard-capped at 24 messages (12 turns) at the end of every turn, and the compaction/summary safety net does not engage for ordinary short-turn conversations, so anything older than ~12 turns is dropped with no summary.

**Expected:** When the conversation grows past the window, older turns get rolled into a running summary the agent can still see, so a fact stated early in the session stays available later.

**Actual:** Older turns are deleted outright once the history exceeds 24 messages. Verified below: after 14 turns, a fact stated in turn 1 is no longer present in the history the model receives.

## How history actually reaches the model (corrected)

Worth being precise, because it makes the impact clear-cut: the provider is **not** sent a growing message array. Each turn, `runtime_messages()` returns only the current user turn (`core/agent_harness/turns/turn_snapshot.py:200-204`: `return [UserRuntimeMessage(content=self.text)]`), which flows to the provider via `core/agent/run_io.py:58`.

All multi-turn memory in the answer path is a **rendered block inside the system prompt**, built by `format_recent_conversation(...)` from `turn_snapshot.conversation_messages` (`core/agent_harness/prompts/assistant.py:236`). That list is the 24-message-capped `session.agent.messages`. So the 24-message cap is the *entire* conversational memory — there is no full-transcript backstop. If a fact is not in the last 24 messages or in a compaction summary, the model never sees it again.

## Root cause

Two history mechanisms operate on the same list (`session.agent.messages`, aliased as `cli_agent_messages` in `core/agent_harness/session/session_core.py:146-153`) but are calibrated so one defeats the other:

1. Hard 24-message cap at the end of every turn, `core/agent_harness/turns/orchestrator.py:129-133`:
   ```python
   session.cli_agent_messages.append(("user", message))
   session.cli_agent_messages.append(("assistant", assistant_text))
   if len(session.cli_agent_messages) > MAX_CONVERSATION_MESSAGES:  # 24
       session.cli_agent_messages[:] = session.cli_agent_messages[-MAX_CONVERSATION_MESSAGES:]
   ```
   Dropped turns are gone with no summary. `MAX_CONVERSATION_MESSAGES = 24` (`core/agent_harness/prompts/conversation_memory.py:11-12`).

2. Char-threshold compaction at the start of every turn via `auto_compact_if_needed(session)` (`orchestrator.py:333`), which only fires when the same list exceeds 48,000 chars (`DEFAULT_AUTO_COMPACTION_CHARS`, `core/agent_harness/turns/transcript_compaction.py:9`).

Because the list is already capped at 24 messages, it rarely reaches 48,000 chars: that needs the 24 messages to average ~2,000 chars each. Ordinary short turns are far below that, so compaction does not fire and the 24-message cap silently deletes context first. (It *can* fire in sessions dominated by long assistant answers, so this is threshold-dependent, not literally never.)

## Verified reproduction

Driving the real production functions (no LLM, no mocks of the code under test) — `_record_answer_turn`, `format_recent_conversation`, and `should_compact`:

```
=== PART A: within-session forgetting ===
MAX_CONVERSATION_MESSAGES = 24
messages retained in session : 24
turn-1 fact ('prod-eu-42') still in the model's history block? False
first retained line: ('user', "Turn 3: what's the status of service-3?")

=== PART B: compaction safety net ===
compaction threshold          : 48000 chars
24 short messages total        : 564 chars
would compaction fire here?    : False
```

Turn 1 states `My cluster is named prod-eu-42.`, followed by 13 ordinary short turns. The block that would be rendered into the assistant system prompt no longer contains the fact; retained history starts at turn 3. Compaction does not engage (564 chars vs the 48,000 threshold).

Manual repro in the shell (expected to match; not run live here — no provider configured in this environment):
1. Start `uv run opensre`.
2. Turn 1: "my cluster is named `prod-eu-42`".
3. Have a normal short back-and-forth for 13+ turns.
4. Ask something depending on the turn-1 fact — the agent no longer has it.

## Secondary issues found in the same area

- Compaction is deterministic-only. `auto_compact_if_needed` never passes an LLM summary, so it always falls back to `deterministic_summary` (first-4 + last-4 message excerpts, 700 chars each) — the middle of the conversation is lost even when it does fire (`transcript_compaction.py:82-104`).
- `MAX_CONVERSATION_TURNS = 12` is duplicated in `core/agent_harness/prompts/conversation_memory.py:11` and `core/state/agent_state.py:15`.
- The persistence layer records `compaction`, `investigation_result`, and `accumulated_context` entries (`core/agent_harness/session/persistence/memory.py`) but nothing reloads them into a live window mid-session.

## Proposed direction

Make the turn-end trim compact instead of delete: when history passes the message-count window, roll the trimmed turns into a running summary that stays in the window, rather than dropping them. This aligns the two mechanisms (a message-count-aware trigger) instead of leaving a char threshold that rarely fires.

## Scope note

This is within-session conversation forgetting. It is distinct from the cross-session persistent memory store being built on `feat/memory_management` (agent memory tool + REPL commands), though both live under "agent memory".

YauhenBichel:
Fixes #4085 

In a long interactive-shell conversation the agent forgot earlier turns: history is hard-capped at 24 messages (12 turns) at the end of every turn, while the compaction safety net only fires past 48,000 chars — a threshold a 24-message list of ordinary turns cannot reach. The cap always won, so anything older than ~12 turns was deleted with no summary. Since the provider receives only the current turn plus a rendered history block built from this capped list, a fact stated in turn 1 was simply gone by turn 14 (reproduced exactly as described in the issue; credit to @Davidson3556 for the diagnosis and the running-summary direction in #4091).

The fix makes the message-count window compact instead of delete, with one policy point:

- `core/state/transcript_window.py` (new leaf, stdlib-only) — `compact_messages_to_window`: within the window the list is unchanged; on overflow the most recent turns are kept verbatim (even count, user/assistant pairing preserved) and everything older is summarized into a single leading `Session summary:` assistant message. An existing summary is merged, never stacked; merged text truncates tail-first so the earliest facts are the last to go. `SESSION_SUMMARY_PREFIX`, `SUMMARY_MAX_CHARS`, and the line formatter are shared with the char-threshold compactor, so both mechanisms produce one summary format and cannot stack.
- Write paths (two writers, one policy): `MutableAgentState.record_turn`/setter compact via the leaf, and the new `core/agent_harness/turns/conversation_recording.py::record_conversation_turn` does the same for port-based sessions — `orchestrator._record_answer_turn` and `action_driver._persist_tool_calling_error` are one-line delegations to it. (`SessionStore` does not expose `.agent`, so a shared helper rather than routing through the state object.)
- Read side: `TurnSnapshot.from_session` compacts the snapshot through the same function instead of a raw `[-24:]` slice — belt-and-suspenders; a write path that ever bypassed compaction would produce a summarized snapshot instead of a silent drop.
- `_latest_want_me_to_offer` skips summary messages, so a bare "yes" can never expand into a stale `Want me to:` offer quoted inside compacted history (pinned by a test that plants a marker offer in a summary and asserts it does not leak).
- Constants dedup (secondary issue finding): `MAX_CONVERSATION_TURNS`/`MAX_CONVERSATION_MESSAGES` now live only in `core/state/agent_state.py`; the duplicate in `conversation_memory.py` is removed and all importers use the canonical path.

Unchanged by design: the char-threshold auto-compaction stays as the safety valve for few-but-huge messages; `format_recent_conversation` needs no changes (the summary is an ordinary assistant message inside the window); the evidence-gather 3-turn slice is pre-existing and out of scope; the deterministic extractive summary is kept — swapping in an LLM-generated summary is the agreed follow-up.

Tests (all written red-first against the defect they pin): 14 new/updated across `tests/core/state/test_transcript_window.py` (window spec: no-op within window, overflow compaction, turn pairing, single-summary merge, early-fact survival over 40 turns, tail-not-head truncation, line truncation), `tests/core/state/test_agent_state.py`, `tests/core/agent/orchestration/test_conversation_window.py` (end-to-end issue repro through the real recording paths, long-session single-summary invariant, tool-error path), `tests/core/agent/test_turn_snapshot_runtime_request.py`, and `tests/core/agent_harness/prompts/test_conversation_memory_follow_up.py` (stale-offer guard).

Repro driving the real production functions (no mocks of code under test), before vs after:

    === before ===
    turn-1 fact ('prod-eu-42') in the model's history block? False
    first retained line: ('user', "Turn 3: what's the status of service-3?")
    would compaction fire here? False (975 chars vs 48000 threshold)

    === after ===
    turn-1 fact ('prod-eu-42') in the model's history block? True
    first retained line: ('assistant', "Session summary:\n- user: my cluster is named prod-eu-42\n- assistant: ...")

---

---

---

### DIFF
--- a/core/agent_harness/prompts/conversation_memory.py
+++ b/core/agent_harness/prompts/conversation_memory.py
@@ -8,8 +8,8 @@ from __future__ import annotations
 
 import re
 
-MAX_CONVERSATION_TURNS = 12
-MAX_CONVERSATION_MESSAGES = MAX_CONVERSATION_TURNS * 2
+from core.state import MAX_CONVERSATION_TURNS
+from core.state.transcript_window import SESSION_SUMMARY_PREFIX
 
 NO_HISTORY_PLACEHOLDER = "(no prior messages in this CLI thread)"
 _ACTION_FACT_MARKERS = (
@@ -87,6 +87,10 @@ def _latest_want_me_to_offer(
         except (TypeError, ValueError):
             continue
         if role != "assistant" or not isinstance(content, str):
+            continue
+        # A compacted session summary can quote an old offer; only live
+        # assistant messages carry an actionable "Want me to:".
+        if content.startswith(SESSION_SUMMARY_PREFIX):
             continue
         match = _WANT_ME_TO_RE.search(content)
         if not match:
--- a/core/agent_harness/prompts/conversation_memory.py
+++ b/core/agent_harness/prompts/conversation_memory.py
@@ -148,6 +148,12 @@ def format_prior_action_facts(
             continue
         if role != "assistant" or not isinstance(content, str):
             continue
+        # A compacted session summary quotes old tool output wholesale; letting
+        # it in as one giant "fact" can spend the whole budget and starve
+        # newer live results. The summary reaches the model via the history
+        # block instead.
+        if content.startswith(SESSION_SUMMARY_PREFIX):
+            continue
         text = content.strip()
         if not text:
             continue
--- a/core/agent_harness/turns/action_driver.py
+++ b/core/agent_harness/turns/action_driver.py
@@ -29,8 +29,8 @@ from core.agent_harness.ports import (
     ToolProvider,
 )
 from core.agent_harness.prompts import build_action_system_prompt, build_action_user_message
-from core.agent_harness.prompts.conversation_memory import MAX_CONVERSATION_MESSAGES
 from core.agent_harness.session.integration_resolution import resolve_and_cache_integrations
+from core.agent_harness.turns.conversation_recording import record_conversation_turn
 from core.agent_harness.turns.turn_plan import TurnPlan
 from core.agent_harness.turns.turn_results import ToolCallingTurnResult
 from core.agent_harness.turns.turn_snapshot import TurnSnapshot
@@ -303,10 +303,7 @@ def _turn_resolved_integrations(
 
 
 def _persist_tool_calling_error(session: SessionStore, user_text: str, error_text: str) -> None:
-    session.cli_agent_messages.append(("user", user_text))
-    session.cli_agent_messages.append(("assistant", error_text))
-    if len(session.cli_agent_messages) > MAX_CONVERSATION_MESSAGES:
-        session.cli_agent_messages[:] = session.cli_agent_messages[-MAX_CONVERSATION_MESSAGES:]
+    record_conversation_turn(session, user_text, error_text)
 
 
 def _render_tool_calling_error(output: OutputSink, message: str) -> None:
--- a/core/agent_harness/turns/conversation_recording.py
+++ b/core/agent_harness/turns/conversation_recording.py
@@ -1,0 +1,25 @@
+"""Record a conversation turn on a session, keeping it within the window."""
+
+from __future__ import annotations
+
+from core.agent_harness.ports import SessionStore
+from core.state import MAX_CONVERSATION_MESSAGES
+from core.state.transcript_window import compact_messages_to_window
+
+
+def record_conversation_turn(session: SessionStore, user_text: str, assistant_text: str) -> None:
+    """Append one user/assistant exchange and compact to the message window.
+
+    The single write path for ``session.cli_agent_messages`` in the turn
+    engine; sessions backed by ``MutableAgentState`` apply the same window
+    through ``record_turn``.
+    """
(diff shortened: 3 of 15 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4247, opened 2026-07-25
fix(grafana): prevent datasource discovery stampede on parallel tool calls (#4245)

### LINKED ISSUE #4245: [BUG] Grafana client cache stampede: parallel tools each run datasource discovery against the same instance (10s timeout each)
## Summary
`get_grafana_client_from_credentials` caches the constructed client per `(account_id, endpoint)`, but the cache is a plain check-then-set dict with no synchronization. When an investigation plans several Grafana tools in one batch (Mimir, alert rules, Loki, Tempo, annotations, service names), they all miss the cache before the first caller finishes construction, so each one runs its own `discover_datasource_uids()` HTTP call. Against an unreachable instance every one of those pays the full 10-second connect timeout and logs the same failure pair.

## Root cause
`integrations/grafana/client.py`: `_grafana_client_cache` is read at the top of `get_grafana_client_from_credentials` and written at the bottom, with `discover_datasource_uids()` (a network call, up to 10s) in between. Any callers arriving during that window miss the cache and duplicate the discovery. The failed-discovery client IS cached afterwards, so this is purely a concurrency race, not a missing negative cache.

## Proposed direction
Single-flight the client construction: a per-cache-key lock (or one module lock) so the first caller constructs and discovers while concurrent callers block briefly and then reuse the cached client. This bounds discovery to one attempt per account/endpoint per process regardless of tool-batch parallelism.

## Additional context
Observed during manual smoke testing of the interactive shell on 2026-07-25. With a reachable Grafana the duplicate discoveries are cheap and invisible; the cost is only obvious when the endpoint is slow or unreachable, where it multiplies the connect timeout by the number of parallel Grafana tools and floods the terminal with repeated tracebacks.

Expected: datasource discovery runs once per `(account_id, endpoint)` per process; concurrent callers wait for (or reuse) the in-flight construction.

Actual: one `/investigate` turn produced six back-to-back discovery attempts for `account_id=user_integration` against the same endpoint before any query ran, each timing out after 10s:

    [grafana] Failed to discover datasource UIDs: HTTPConnectionPool(host='...', port=3001): Max retries exceeded with url: /api/datasources (... connect timeout=10)
    (repeated six times)

A seventh discovery for `account_id=investigation_sink` is a separate cache key and expected.

### Steps to reproduce

1. Configure the Grafana integration with an unreachable instance URL (e.g. a private address off VPN).
2. Run `/investigate grafana` in the interactive shell so the planner batches multiple Grafana queries.
3. Observe one discovery-failure log pair per planned Grafana tool instead of one total, each after a 10s connect timeout.

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

_No response_

### Additional context

_No response_

HarshSharma1112:
## Summary

Fixes #4245.

`get_grafana_client_from_credentials` caches the built client per `(account_id, endpoint)` but used a plain read-then-write dict with no synchronisation around `discover_datasource_uids()` — a network call that takes up to 10 seconds. When an investigation batches multiple Grafana tools in one turn (Mimir, Loki, Tempo, alert rules, annotations, service names), all callers race past the cache check before any of them finishes, so every one pays the full discovery cost independently. Against an unreachable instance this multiplies the 10 s connect timeout by the number of parallel tools and floods the terminal with repeated failure pairs.

## Root cause

```
cache_key = f"creds_{account_id}_{endpoint}"
if cache_key in _grafana_client_cache:   # ← all threads see miss
    return _grafana_client_cache[cache_key]

client = GrafanaClient(config=config)
discovered = client.discover_datasource_uids()  # ← every thread calls this
...
_grafana_client_cache[cache_key] = client       # ← last writer wins
```

## Fix

Add a `defaultdict(threading.Lock)` keyed by `cache_key`. The first caller acquires the key lock, builds the client, runs discovery, and writes the cache. Concurrent callers for the **same key** block on the lock and read the populated entry when it becomes available — without calling `discover_datasource_uids` again. Callers for **different keys** get independent locks and run in parallel.

A fast-path read before the lock acquisition keeps steady-state overhead at zero — no locking at all once the entry is cached.

```python
# Fast path — already cached, no lock needed.
if cache_key in _grafana_client_cache:
    return _grafana_client_cache[cache_key]

with _grafana_client_locks[cache_key]:
    # Re-check inside the lock: a concurrent caller may have built and
    # cached the client while we were waiting.
    if cache_key in _grafana_client_cache:
        return _grafana_client_cache[cache_key]

    # ... build client, run discovery, write cache ...
```

## Changes

- **`integrations/grafana/client.py`**: add `_grafana_client_locks` (`defaultdict(threading.Lock)`), wrap the construction block in a per-key lock with double-checked read, update docstring to explain the concurrency contract.
- **`tests/integrations/grafana/test_client.py`**: add `TestClientCaching` (second call returns same object; discover runs once; different endpoints independent) and `TestConcurrentCacheStampede` (8 threads share a single discovery; failed-discovery client is also cached).

## Testing

```
pytest tests/integrations/grafana/test_client.py -v
# 6 passed
```

Lint and format: `ruff check` and `ruff format --check` both pass.

## Pre-PR Checklist

- [x] `ruff check` — no issues
- [x] `ruff format --check` — clean
- [x] 6 tests pass, 0 regressions

### DIFF
--- a/integrations/grafana/client.py
+++ b/integrations/grafana/client.py
@@ -3,6 +3,8 @@
 from __future__ import annotations
 
 import logging
+import threading
+from collections import defaultdict
 
 from integrations.grafana.base import GrafanaClientBase
 from integrations.grafana.config import GrafanaAccountConfig
@@ -13,6 +15,12 @@ from integrations.grafana.tempo import TempoMixin
 logger = logging.getLogger(__name__)
 
 _grafana_client_cache: dict[str, GrafanaClient] = {}
+
+# One lock per cache key so only concurrent callers for the *same*
+# (account_id, endpoint) pair serialise — callers for different pairs
+# proceed in parallel.  The outer dict itself is only ever mutated while
+# the corresponding per-key lock is held, so no second lock is needed.
+_grafana_client_locks: defaultdict[str, threading.Lock] = defaultdict(threading.Lock)
 
 
 class GrafanaClient(LokiMixin, TempoMixin, MimirMixin, GrafanaClientBase):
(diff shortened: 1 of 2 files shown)


muddlebee:
@HarshSharma1112 please add a proper demo

muddlebee:
feel free to re-open with a proper demo, closing this for now.

PR #4250, opened 2026-07-25
fix(correlation): score periodicity by threshold crossings, not elevation

### LINKED ISSUE #4249: [BUG] Upstream periodicity scoring counts elevated samples, not threshold crossings
### Summary

score_periodic_spikes counts every sample at or above spike_threshold instead of the number of threshold crossings, so a single sustained spike is mis-scored as a repeated/periodic pattern.

**Expected:** A signal that crosses the threshold once and stays elevated for several samples is **one** spike — a sustained event, not a repeated pattern — so `repeated_spikes == 1` and `score == 0.0`.

**Actual:** Each elevated sample is counted as its own spike, so the same input returns `repeated_spikes == 3` and `score == 1.0`.

**Example:**

```python
# Values cross 80 once, then stay high for three samples — a single sustained spike.
score_periodic_spikes(
    signal_name="upstream_cpu",
    values=(20.0, 90.0, 90.0, 90.0, 20.0),
    spike_threshold=80.0,
)
# expected: repeated_spikes=1, score=0.0   (one crossing = one spike)
# actual:   repeated_spikes=3, score=1.0   (counted three elevated samples)
```

### Steps to reproduce

```python
from core.domain.correlation.scoring import score_periodic_spikes

# One spike: crosses 80 once, stays high for three samples, then falls back.
score_periodic_spikes(
    signal_name="upstream_cpu",
    values=(20.0, 90.0, 90.0, 90.0, 20.0),
    spike_threshold=80.0,
)
```

Yes

### How often does it occur?

Every time

### Operating system

Linux

### Logs and error output

```text
Actual return:
  PeriodicityScore(signal_name='upstream_cpu', repeated_spikes=3, score=1.0,
                   rationale='Detected repeated threshold crossings for upstream_cpu.')

Expected:
  repeated_spikes == 1, score == 0.0   (one crossing = one spike)
```

### Additional context

**Problematic code** — `core/domain/correlation/scoring.py`, `score_periodic_spikes`:

```python
repeated_spikes = sum(1 for value in values if value >= spike_threshold)
```

This counts elevated *samples*, but the field is named `repeated_spikes`, the docstring says "repeated threshold **crossings**", and the existing test is named `test_periodic_spikes_score_repeated_threshold_crossings` — all describe crossings. They only agree when each spike is one sample wide (the shape the current fixtures use), so the divergence is never caught.

**Impact on the Planner:** this produces the **periodicity** evidence axis (weight `0.10`) for upstream root-cause correlation. Its score feeds `score_candidate_correlation` → `final_confidence` → `most_likely_causal_drivers`, the ranked suspect list the investigation Planner reasons over. A sustained-high upstream service (the most common real-metric shape) wrongly gets a flat **+0.10** confidence — enough to re-rank it above a better-correlated suspect or flip its label across a cutoff (`medium ≥ 0.4`, `high ≥ 0.75`) — so the Planner can surface the wrong root cause with unearned confidence.

**Proposed fix** — count below→above crossings so only genuinely recurring spikes score `1.0`:

```python
repeated_spikes = sum(
        1
        for previous, current in zip(values, values[1:], strict=False)
        if previous < spike_threshold <= current
    )
```

Sustained highs that are causally relevant still register on the time-window axis (weight `0.45`); a pre-existing flat-high is correctly de-emphasized. Both existing tests stay green.

---

Vidit-lab:
Fixes #4249 

**The problem**

`score_periodic_spikes` (in `core/domain/correlation/scoring.py`) is supposed to detect whether an upstream signal spikes *repeatedly* — a recurring pattern like a cron job, retry storm, or connection-pool thrash. Instead, it counts every sample at or above `spike_threshold` rather than the number of times the signal actually **crosses** the threshold. So a single *sustained* spike — one crossing that stays high for several samples — is mis-read as a repeated/periodic pattern.

**Example**

```python
from core.domain.correlation.scoring import score_periodic_spikes

# One spike: crosses 80 once, stays high for three samples, then falls back.
score_periodic_spikes(
    signal_name="upstream_cpu",
    values=(20.0, 90.0, 90.0, 90.0, 20.0),
    spike_threshold=80.0,
)
# actual  : PeriodicityScore(repeated_spikes=3, score=1.0, ...)   # counted 3 elevated samples
# expected: PeriodicityScore(repeated_spikes=1, score=0.0, ...)   # one crossing = one spike
```

**The problematic code**

```python
repeated_spikes = sum(1 for value in values if value >= spike_threshold)
```

This counts elevated *samples*. But the field is named `repeated_spikes`, the docstring says "repeated threshold **crossings**", and the existing test is named `test_periodic_spikes_score_repeated_threshold_crossings` — all three describe crossings. They only agree when every spike is exactly one sample wide, which is precisely the shape the existing fixtures use, so the divergence was never caught.

**The fix**

Count below→above threshold crossings, so only genuinely recurring spikes score `1.0`:

```python
repeated_spikes = sum(
        1
        for previous, current in zip(values, values[1:], strict=False)
        if previous < spike_threshold <= current
    )
```

Sustained highs that are causally relevant are not lost: a *new* sustained rise still registers on the **time-window** axis (weight `0.45`), which is the axis designed to reward change that co-moves with the symptom; a *pre-existing* flat-high is correctly de-emphasized. Both existing tests stay green (their fixtures use one-sample spikes, where crossings == samples), and a new regression test pins the sustained-spike case.

The periodicity axis is meant to reward *recurring* spikes, so the count must reflect distinct spike events, not how long a spike lasts. The original `sum(...)` counts elevated samples, which conflates one sustained high with a repeating pattern. I considered adding a separate "sustained elevation" axis, but that changes the model and the weight budget; the sustained case is already handled by the time-window axis, so the correct, minimal fix is to count below→above crossings within the existing axis. The loop walks the series once (O(n)), incrementing only on a fresh crossing (previous sample below, current at/above the threshold), which keeps the `repeated_spikes` field honest to its name and preserves the existing behavior for one-sample spikes.

---

### DIFF
--- a/core/domain/correlation/scoring.py
+++ b/core/domain/correlation/scoring.py
@@ -144,7 +144,11 @@ def score_periodic_spikes(
     values: tuple[float, ...],
     spike_threshold: float,
 ) -> PeriodicityScore:
-    repeated_spikes = sum(1 for value in values if value >= spike_threshold)
+    repeated_spikes = sum(
+        1
+        for previous, current in zip(values, values[1:], strict=False)
+        if previous < spike_threshold <= current
+    )
 
     if repeated_spikes <= 1:
         score = 0.0
--- a/tests/core/domain/correlation/test_scoring.py
+++ b/tests/core/domain/correlation/test_scoring.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from core.domain.correlation.scoring import (
     rank_upstream_candidates,
+    score_periodic_spikes,
     score_time_window_correlation,
     score_topology_adjacency,
 )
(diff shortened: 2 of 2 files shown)


Vidit-lab:
@greptile review

PR #4253, opened 2026-07-26
fix(correlation): initialize spike state from first sample to avoid false crossings

vinu4545:
## Summary

This PR fixes an edge case in periodicity scoring where query windows beginning with an already-elevated value were incorrectly treated as containing an additional upward threshold crossing.

Instead of assuming every window starts below the threshold, the scorer now initializes its state from the first observed sample and counts only genuine upward transitions within the observed data.

## What Changed

- Initialize the threshold state from the first sample instead of a synthetic "below threshold" state.
- Count only real upward threshold transitions between consecutive samples.
- Return zero cleanly for empty input windows.
- Added regression tests covering all identified edge cases.

## Regression Tests

Added coverage for:

- Sustained plateau counts as a single spike.
- Window starting already above the threshold does not invent a crossing.
- Always-elevated windows produce zero crossings.
- Multiple genuine threshold crossings are counted correctly.
- Empty input returns zero spikes.

## Validation

```bash
uv run pytest tests/synthetic/rds_postgres/correlation/test_periodicity.py
```

**Result:** `7 passed`

### DIFF
--- a/core/domain/correlation/scoring.py
+++ b/core/domain/correlation/scoring.py
@@ -145,8 +145,18 @@ def score_periodic_spikes(
     spike_threshold: float,
 ) -> PeriodicityScore:
     repeated_spikes = 0
-    was_above_threshold = False
-    for value in values:
+    if not values:
+        score = 0.0
+        rationale = "No repeated spike pattern detected."
+        return PeriodicityScore(
+            signal_name=signal_name,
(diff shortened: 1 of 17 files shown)


muddlebee:
@vinu4545 please dont open such PRs without prior discussion, please go through contributor guidelines and try not to violate.

PR #4259, opened 2026-07-26
fix(deps): bump httplib2 to 0.32.0 for the gzip decompression-bomb advisory

YauhenBichel:
Bumps the locked `httplib2` from 0.31.2 to 0.32.0, closing the unbounded
gzip/deflate decompression advisory (Dependabot alert #31). Lockfile-only —
`0.32.0` already satisfies the existing `httplib2<1.0.0,>=0.19.0` constraint
from `google-api-python-client`, so `pyproject.toml` is unchanged.

`httplib2` is a transitive dependency: `google-api-python-client` (direct dep)
→ `httplib2`. The only consumer in this repo is
`integrations/google_docs/client.py`, which calls Google's own HTTPS endpoints
rather than untrusted or attacker-controlled servers, so real exposure was low.
Patching anyway: the amplification is unbounded, and a compromised or
MITM'd endpoint is exactly the scenario the advisory describes.

The upstream fix is structural rather than a version-only change — the
unbounded `gzip.GzipFile(...).read()` and `zlib.decompress()` calls named in
the advisory are replaced by a `LimitDecoder` (ratio 100, 10 MB safe limit,
10 GB hard limit), and `_decompressContent` now requires an explicit limits
argument.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1174,14 +1174,14 @@ wheels = [
 
 [[package]]
 name = "httplib2"
-version = "0.31.2"
+version = "0.32.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "pyparsing" },
 ]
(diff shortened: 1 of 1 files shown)


YauhenBichel:
@greptile review

PR #2899, opened 2026-06-21
feat(groundcover): add groundcover observability integration (minimal)

AvitalTamir:
Fixes #2850

#### Describe the changes you have made in this PR -

Adds **groundcover** as a first-class, **read-only** observability provider over its public **MCP** endpoint (`https://mcp.groundcover.com/api/mcp`), queried with **gcQL**. This is an intentionally **minimal first PR** (per the discussion on #2850): the integration registers and verifies cleanly, plus a few basic tool calls. The rest follows in subsequent PRs.

**Included:**
- **Config + catalog + verification + CLI setup** — `GROUNDCOVER_API_KEY` (read-only service-account token; `GROUNDCOVER_MCP_TOKEN` alias), `GROUNDCOVER_MCP_URL` (default `https://mcp.groundcover.com/api/mcp`), optional `GROUNDCOVER_TENANT_UUID` / `GROUNDCOVER_BACKEND_ID` routing, `GROUNDCOVER_TIMEZONE`, and `GROUNDCOVER_INSTANCES` multi-instance support. `opensre integrations setup/verify groundcover` work and it shows in `integrations list`.
- **MCP service client** (`app/services/groundcover/`) — streamable-HTTP JSON-RPC transport, bearer auth + tenant/backend/timezone headers, token redaction, bounded connect retries, cached gcQL reference, and a fail-closed verification probe (confirms the read-only tool surface and validates workspace/backend routing with actionable errors on ambiguous/mistyped routing).
- **Three read-only tools** (`source="groundcover"`): `get_groundcover_query_reference`, `query_groundcover_logs`, `query_groundcover_traces`. Credentials are bound into a runtime client (never model-facing) and tool schemas set `additionalProperties: false`, so prompt-injected `mcp_url`/routing keys are rejected. Tool descriptions carry the gcQL query-guidance contract (lead with the filter; `| filter` is post-aggregation only; project with `| fields …` or aggregate with `| stats …` rather than raw select-all; narrow time windows; mandatory `| limit N`; `status_code>=500` / `status:error` for traces).
- **Docs** — `docs/groundcover.mdx` + nav, plus `.env.example` and environment-variable reference entries.
- **Tests** — config/env/verify, the MCP client (JSON/SSE parsing, token redaction, cached reference, probe routing), and the three tools.

**Intentionally deferred to follow-up PRs (kept out to keep this one minimal):**
- The remaining tools — metrics, APM, Kubernetes events, entities, monitors, and monitor issues.
[...]

Issue #2850, opened 2026-06-18
[FEATURE] Add groundcover integration for logs, traces, metrics, APM, and monitors (MCP/gcQL)


AvitalTamir:
## Problem statement

OpenSRE currently requires Grafana, Datadog, CloudWatch, or one of the other
configured providers before the agent can pull evidence. Teams running
[groundcover](https://groundcover.com) — an eBPF-based, OpenTelemetry-native
observability platform — have no native integration path.

groundcover exposes a public, read-only **MCP** endpoint
(`https://mcp.groundcover.com/api/mcp`) and a single query language (**gcQL**)
that spans all of these signals. That makes it a high-value integration: one
transport + one query layer covers logs, traces, metrics, APM, events,
entities, and monitors without per-signal adapter complexity. The catch is that
efficient gcQL needs discipline (narrow time windows, bounded row limits,
aggregation over raw scans), so the provider must ship that query skill rather
than leaving it to chance.

## Proposed solution

Add a first-class, **read-only** groundcover integration built on the public MCP
surface (no write/remediation actions in v1):

- **Config and onboarding** — `GROUNDCOVER_API_KEY` (read-only service-account
  token; `GROUNDCOVER_MCP_TOKEN` alias), `GROUNDCOVER_MCP_URL` (default
  `https://mcp.groundcover.com/api/mcp`), optional `GROUNDCOVER_TENANT_UUID` /
  `GROUNDCOVER_BACKEND_ID` routing, `GROUNDCOVER_TIMEZONE`; `opensre onboard`
  prompts + `opensre integrations setup/verify groundcover`, with
  `GROUNDCOVER_INSTANCES` multi-instance support.
- **MCP service client** — streamable-HTTP JSON-RPC transport (reusing the
  existing `mcp` client stack), bearer auth + routing/timezone headers, token
  redaction, bounded connect retries, and a fail-closed verification probe
  (lists tools + validates workspace/backend routing).
- **gcQL signal tools** — one tool per evidence family with built-in query
  guidance and examples:

```python
query_groundcover_logs(query, start, end, period)
query_groundcover_traces(query, start, end, period)
query_groundcover_metrics(mode, promql, metric_name, filter, step, ...)
query_groundcover_apm(query, start, end, period)
query_groundcover_events(query, start, end, period)
query_groundcover_entities(query)            # live K8s state
query_groundcover_monitors(query)            # monitor definitions + health
query_groundcover_issues(query, start, end, period)
[...]

PR #4261, opened 2026-07-27
docs: list groundcover in the observability integrations table

dannyshmueli-gc:
Adds groundcover to the Observability row of the integrations table. Docs only, one line plus an icon.

The integration itself landed in #2899 (provider, validation, gcQL tools) from #2850, and has been kept current since, most recently in #4194 when it moved onto the shared setup flow. It ships with `integrations/groundcover/`, tests, and `docs/groundcover.mdx`, but the README table never picked it up, so it is the one supported observability provider the list does not name.

Icon added as `docs/assets/icons/groundcover.webp` (40x40, 554 bytes) and rendered at width 16, matching Grafana and Datadog in the same row.

No code changes, so the code-understanding section of the template does not apply.


dannyshmueli-gc:
@greptile review

PR #4395, opened 2026-07-27
test(fleet_monitoring): use wait_until in test_lifecycle.py

### LINKED ISSUE #4362: [IMPROVEMENT] C-41 — Use wait_until in test_lifecycle.py
**Task ID:** C-41 · **Level:** Junior · **Size:** S · **Category:** tests · **Good first issue** (after C-38)

### Current state

`tests/fleet_monitoring/test_lifecycle.py` uses bare `sleep` waits that are slower and flakier than a condition-based helper.

### Desired state

Replace the bare sleeps with the shared `wait_until` helper added in C-38. One file only. Unit-test style only — no live API / e2e as the task itself.

### Steps

1. Confirm C-38 has landed and note where `wait_until` lives.
2. Replace bare `sleep` waits in this file with `wait_until`.
3. Run: `uv run pytest tests/fleet_monitoring/test_lifecycle.py -q --tb=line`

### Done when

- [ ] Bare sleeps in scope are replaced with `wait_until`
- [ ] Tests pass

**Do not:** edit a second file.

**File:** `tests/fleet_monitoring/test_lifecycle.py`

---

To claim this task, comment here (or in Discord `#contribute`): *"I'd like to take C-41"*. One task per pull request.

arnavpopley:
Fixes #4362

**Depends on #4381 (C-38's `wait_until` helper).** This PR is opened against the `test/wait-until-polling-helper` branch rather than `main` since `tests/utils/polling.py` doesn't exist on `main` yet. Once #4381 merges, this should be retargeted to `main` (or rebased) before merging.

`tests/fleet_monitoring/test_lifecycle.py` had two bare sleeps: `_spawn_sleep()` and `_spawn_unkillable()` each did a fixed `time.sleep(0.2)` after spawning a child process, to "give the child a moment to register its signal handler" before the tests send it SIGTERM.

This sleep guards a real race, not just a slow test: if the parent sends SIGTERM before the child's handler is installed, the default disposition kills the child outright. For `_spawn_unkillable()` specifically, that would make `test_force_kill_after_grace_period` pass for the wrong reason (child dies on first SIGTERM instead of exercising the SIGKILL escalation path).

Rather than convert this into a `wait_until` on a proxy condition, I made the wait meaningful: both child scripts now `pathlib.Path(ready_path).touch()` immediately after registering their signal handler, and the parent calls `wait_until(lambda: os.path.exists(ready_path), timeout=2.0, interval=0.01)` before returning the process — polling the actual readiness condition instead of guessing a fixed delay is long enough (and never longer than it needs to be on a fast machine). Only this one file was touched, per the task's scope.

```
$ uv run pytest tests/fleet_monitoring/test_lifecycle.py -q --tb=line
collected 5 items
tests/fleet_monitoring/test_lifecycle.py .....                           [100%]
5 passed in 21.17s
```

---

`_make_ready_path()` reserves a filesystem path via `tempfile.mkstemp()` and immediately deletes it, so its later existence is an unambiguous signal (rather than reusing an already-existing temp file, which would be true from the start). Both child one-liners now `pathlib.Path(<path>).touch()` right after `signal.signal(...)`, so the touch happens exactly when the handler is live. `_wait_for_ready()` wraps the `wait_until` call in a `try/finally` that removes the marker file regardless of outcome, so a timeout doesn't leak a stray temp file. I kept the childrens' inline "import signal, sys" style but added `import pathlib` (and moved the `import time` for the child's own 60s sleep to a second statement, since it's only needed after the touch) — the child scripts themselves are otherwise unchanged (same handler behavior, same eventual `time.sleep(60)` placeholder workload).

---

### DIFF
--- a/tests/fleet_monitoring/test_lifecycle.py
+++ b/tests/fleet_monitoring/test_lifecycle.py
@@ -5,10 +5,11 @@ from __future__ import annotations
 import os
 import subprocess
 import sys
-import time
+import tempfile
 
 import pytest
 
+from tests.utils.polling import wait_until
 from tools.system.fleet_monitoring.lifecycle import TerminateResult, terminate
 
 # Windows ``os.kill`` / ``signal.SIGTERM`` delivery to a Python ``Popen`` child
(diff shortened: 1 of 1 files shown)


arnavpopley:
Closing to reopen after fixing the greptile P1 (readiness-timeout child leak) via a fresh PR.

PR #4428, opened 2026-07-28
docs(kubernetes): clarify get_resource vs describe_pod tool boundaries

### LINKED ISSUE #4368: [IMPROVEMENT] Q-41 — Improve one tool’s description / use_cases
**Task ID:** Q-41 · **Level:** Junior · **Size:** S · **Good first issue**

### Current state

Some tool `description` / `use_cases` / input-field text do not tell the agent when **not** to call the tool.

### Desired state

Tighten `description` / `use_cases` / input-field text so the agent knows when **not** to call the tool (see Slack/GitHub “Do NOT use…” style). **One tool package only.**

### Steps

1. Pick one tool package and say which one in your claim comment.
2. Improve description / use_cases / input help text only.
3. `make lint` — no behavior change to execution logic.

### Done when

- [ ] One tool’s agent-facing text is clearer about when not to use it
- [ ] Only that tool package is touched

**Files:** one tool package

---

To claim this task, comment here (or in Discord `#contribute`): *"I'd like to take Q-41"*. One task per pull request.

a-simptoad:
Fixes #4368 

Tightens the `description`, `use_cases`, and adds `anti_examples` for two tools in the **Kubernetes** tool package:

- `KubernetesGetResourceTool` (`kubernetes_get_resource`)
- `KubernetesDescribePodTool` (`kubernetes_describe_pod`)

Both tools can return data for the same resource (a pod), but `describe_pod` redacts environment variable *values* (keys only), while `get_resource` returns the raw resource object, which can include unredacted env values set directly on the pod spec. Nothing in either tool's prior description told the agent about this overlap or the redaction difference, so an agent investigating a pod could pick either tool and get a plausible-looking result — with very different exposure of potentially sensitive data depending on which one it picked.

The fix adds explicit "prefer X for pods, use Y for everything else" language to both tools' descriptions, plus `anti_examples` entries pointing each tool at its sibling for the overlapping case. No other tools in this file were changed — the remaining ten cover distinct, non-overlapping resource types (deployments, statefulsets, services, ingress, etc.) where there's no realistic tool-selection ambiguity to resolve.

No execution logic was changed — description/use_cases/anti_examples text only.

`make lint` passes with no errors.

Ran the scoped test subset for the touched file:
`uv run pytest tests/ -k kubernetes -v`

---

- [X] Yes, I used AI assistance (continue below)

- [X] I have reviewed every single line of the AI-generated code
- [X] I can explain the purpose and logic of each function/component I added
- [X] I have tested edge cases and understand how the code handles them
- [X] I have modified the AI output to follow this project's coding standards and conventions

The issue (Q-41) asks for one tool package's agent-facing text to be tightened so the agent knows when *not* to call a tool. I went through all twelve tools in the Kubernetes tools package and, for each one, asked whether calling the wrong sibling tool could produce a plausible-but-wrong result — i.e. a case where an agent might reasonably pick either tool and get back something that looks correct but isn't. Ten of the twelve tools map to distinct resource types (deployments, statefulsets, services, ingress, nodes, etc.) with no realistic overlap, so they were left untouched. The one genuine collision is `kubernetes_get_resource` and `kubernetes_describe_pod`, which can both return data for the same pod, but differ in a way that actually matters:
`describe_pod` redacts environment variable values (only keys are shown), while `get_resource` returns the raw resource object, which can expose unredacted env values set directly on the pod spec. That's a real, consequential difference rather than a stylistic one, so I added explicit "prefer X for pods, use Y for everything else" language to both tools' descriptions and matching `anti_examples` entries pointing each tool at its sibling. I scoped the PR to just these two tools to keep the diff small and reviewable, matching the "Size: S" / "Junior" label on the issue.

---

- [X] I have added proper PR title and linked to the issue
- [X] **I can explain the purpose of every function, class, and logic block I added**
- [X] I have considered potential edge cases and how my code handles them
- [X]  My code follows the project's style guidelines and conventions

---

### DIFF
--- a/integrations/kubernetes/tools/__init__.py
+++ b/integrations/kubernetes/tools/__init__.py
@@ -458,9 +458,10 @@ class KubernetesDescribePodTool(BaseTool):
         "requests/limits, environment variables (values redacted, keys only), volume "
         "mounts, conditions, container states, and owner references. Use when list_pods "
         "shows a problem and you need deeper detail on one pod. Preferred over "
-        "kubernetes_get_resource for pods specifically, since that tool returns the raw "
-        "resource including unredacted env values. For any resource type other than "
-        "pod, use kubernetes_get_resource instead."
+        "kubernetes_get_resource for pods specifically — both redact env values "
+        "identically, but this tool returns a curated, investigation-shaped view "
+        "(container states, owner references) rather than the raw API object. For any "
+        "resource type other than pod, use kubernetes_get_resource instead."
     )
     use_cases = [
         "Inspecting container image versions and resource limits on a specific pod",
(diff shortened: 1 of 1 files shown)


muddlebee:
@greptile review

PR #4432, opened 2026-07-28
refactor(gateway): replace the SQLite binding index with a JSON file, drop the install catalog

YauhenBichel:

Session bindings were a SQLite database on the S3 Files volume. SQLite relies on POSIX advisory locks, and S3 Files is NFS-backed, where those locks are not reliable — so the index could fail or corrupt rather than fail cleanly. It also forced a deployment constraint: `desired_count = 1`, no rolling deploys.

Bindings are now a JSON document beside the rest of a principal's context, written temp-then-`os.replace`. That rename is atomic on POSIX and NFS, so a reader never sees a partial file. Same layout on a laptop and on the mounted volume.

The install catalog is deleted rather than ported:
`slack_installs` answered "which organization owns this Slack team?" — but that is control-plane data the webapp already owns (`organizations.workspace_provider` plus the Slack team id on the `organizations` row). Keeping a second copy with no writer meant guaranteed drift. A deployment is now simply told which organization it serves via `OPENSRE_ORGANIZATION_ID`, restricted to named workspaces by `OPENSRE_SILO_TEAM_IDS`. Resolution fails closed: unknown team, or no configured organization, refuses the turn.

---

The problem is a storage engine chosen for a filesystem it does not support. SQLite on NFS is a correctness hazard, not a style preference.

Postgres was considered and rejected: the silo already provisions `DATABASE_URL` for investigations, so it was the cheapest option, but it adds an operational dependency for an index that one process writes. A JSON file with atomic rename gives the durability that was actually missing — surviving task replacement — without one.

The single-writer constraint remains, but storage is no longer what imposes it: Socket Mode is single-consumer, so the gateway is `desired_count = 1` regardless.

The swap was drop-in because `BindingStore` is a Protocol with one construction point, `open_binding_store()`. No call site changed, and the border suite asserts isolation properties rather than backend details, so it ran unchanged across the
migration and is the evidence that behavior held.

---

---


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4434, opened 2026-07-28
chore(gateway): remove unused Telegram gateway startup helper

### LINKED ISSUE #4429: Dead code: try_load_gateway_settings_for_startup in gateway/telegram/settings.py is never called
## Summary
`try_load_gateway_settings_for_startup` in `gateway/telegram/settings.py:151` appears to be dead code — it has no call sites anywhere in the codebase. Its only references are its own definition and the test file written for it (`gateway/tests/telegram/test_get_gateway_settings.py`).

Verified via code search: the symbol appears only in those two files.

## Task
Either:
1. **Remove** the function and its dedicated tests, if it is truly unused; or
2. **Wire it in** where it was originally intended (optional background startup of the Telegram gateway), if that integration was simply never completed.

Please check git history / the original PR to decide which option applies before deleting.

## Acceptance Criteria
- No unused/dead code paths remain in `gateway/telegram/settings.py`.
- **Telegram gateway must be tested and working after the change** (per #devs refactoring AC: Telegram, CLI, and shell must work and be tested after any change touching this area).
- Test suite updated accordingly (remove orphaned tests, or add tests covering the newly wired-in behavior).

_Reported by @YauhenBichel in Slack #devs._

M-Talha-Atif:
## Summary

Fixes [#4429](https://github.com/Tracer-Cloud/opensre/issues/4429)

`try_load_gateway_settings_for_startup` in `gateway/telegram/settings.py` had no call sites left anywhere in the codebase, only its own definition and the test file written for it. I traced it back through git history to figure out which of the two options in the issue applied (remove it, or wire it back in).

Turns out its only caller used to live in `surfaces/interactive_shell/main.py`, auto-starting a background Telegram poller when the interactive shell launched. That caller was deleted in an earlier refactor (commit `1fdb2b61`, "Feature/gateway code"), and this helper function got left behind by accident. Since then, the Telegram gateway has moved to its own separate daemon process (`opensre gateway start`, wired through `gateway/runtime/manager.py` and `gateway/telegram/wiring.py`), which already loads settings and starts polling on its own. Wiring the old helper back into the shell would mean two separate processes polling the same Telegram bot at once, which Telegram's API does not allow, so removing it was the right call, not rewiring it.

## Changes

- `gateway/telegram/settings.py`: removed the unused `try_load_gateway_settings_for_startup` function.
- `gateway/tests/telegram/test_get_gateway_settings.py`: removed the 4 tests written only for that function, dropped it from the import list, and removed the now unused `MagicMock` import that only those tests needed.

## Testing

Ran lint, format check, and typecheck, all clean, no issues.

For tests, I ran the full `gateway/tests/` suite since the acceptance criteria specifically call out that the Telegram gateway needs to keep working after this change: 388 passed, 4 failed, 2 errors.

All 4 failures and 2 errors are unrelated to this change. They come from `os.WNOHANG`, a POSIX-only constant that does not exist on Windows, used by daemon process-management tests (`test_gateway_daemon.py`) and one Slack test that depends on the same daemon path. I confirmed this by stashing my changes and running those same tests against a clean `main`, same failures, same errors, with zero code changes present. These only fail locally on Windows; CI runs on Linux where this is not an issue.

### DIFF
--- a/gateway/telegram/settings.py
+++ b/gateway/telegram/settings.py
@@ -146,26 +146,3 @@ def load_gateway_settings() -> GatewaySettings:
         )
     except ValidationError as exc:
         raise GatewayConfigurationError("Invalid Telegram gateway configuration") from exc
-
-
-def try_load_gateway_settings_for_startup(
-    *,
-    logger: logging.Logger,
-    respect_auto_start: bool = True,
-) -> GatewaySettings | None:
-    """Load gateway settings for optional background startup; return None when skipped."""
-    try:
-        settings = load_gateway_settings()
-    except GatewayConfigurationError as exc:
-        logger.debug("[telegram-gateway] startup skipped: %s", exc)
-        return None
-
-    if respect_auto_start and not settings.auto_start_enabled:
-        logger.debug("[telegram-gateway] auto-start disabled in config")
-        return None
-
-    if not settings.bot_token:
-        logger.warning("[telegram-gateway] no bot token configured")
-        return None
-
-    return settings
--- a/gateway/tests/telegram/test_get_gateway_settings.py
+++ b/gateway/tests/telegram/test_get_gateway_settings.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import logging
 import os
 from collections.abc import Iterator
-from unittest.mock import MagicMock, patch
+from unittest.mock import patch
 
 import pytest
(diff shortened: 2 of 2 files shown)


Davidson3556 (approved):
Clean deletion, LGTM.

PR #4452, opened 2026-07-28
Signal log truncation and retry with larger tail on step miss

### LINKED ISSUE #4425: fix(github-actions): step log filtering runs on a 500-line tail with no truncation signal
### Current behaviour

`github_actions_job_log` calls `get_job_logs` with `tail_lines` defaulting to `500` (`integrations/github/tools/actions.py`). For jobs whose logs exceed 500 lines, `extract_step_log` then runs over only that tail. If the requested step's `##[group]` block was cut off, `step_name`/`step_number` silently miss and the tool returns `match_strategy="full-log"` with the truncated tail — the agent receives partial log content with no indication that anything was dropped.

### Desired behaviour

- Signal truncation in the payload (e.g. `truncated: true` plus returned/total line counts) so the agent can reason about it.
- When a specific `step_name`/`step_number` was requested and no group matched, re-fetch with a larger window (the `get_job_logs` response exposes the total log length) before falling back to `full-log`.
- Minor: with an ungrouped-only log, `step_number=1` currently returns the whole log but reports `match_strategy="step_number"`; it should report `full-log`.

### Scope

`integrations/github/tools/actions.py` only. Supersedes the log-fetch optimization item from #2719 (see https://github.com/Tracer-Cloud/opensre/issues/2719#issuecomment-4630005541); the rest of #2719 was completed in #2755.

sahilnyk:
Fixes #4425

`integrations/github/tools/actions.py` only, per the issue's scope:

1. **Truncation signal**: returns `truncated`, `returned_lines`, `original_lines`, `retry_attempted`, `retry_error`. Error payloads carry the same keys.

2. **Retry on step miss**: when a requested step falls back to `full-log` on a truncated log, re-fetch once with `tail_lines` set to the log's total line count. The bigger body is adopted only if it matched the step, so a failed retry can't push a huge log into context.

3. **Fixed the ungrouped-only log bug**: a log with zero `##[group]` markers faked a "full-log" section counted as step 1, so `step_number=1` wrongly reported `match_strategy="step_number"`. `_extract_log_sections` now returns `[]`.

`make lint`, `make format-check`, `make typecheck`, `make verify-integrations-smoke` pass. Full suite 13,568 passed. `test_github_actions_tool.py`: 44 passed.

Verified against six real failed jobs via the live GitHub MCP. `original_lines` exact in all six (559, 559, 8781, 68, 508, 53), `truncated` correctly `false` on the two that fit in the tail. On the 8781-line job, keeping the first response returns 57,683 chars instead of 696,693.

### Known limitation

Step matching doesn't work on real logs today, and this PR doesn't change that: every log line starts with an RFC3339 timestamp, so `line.startswith("##[group]")` never matches (0 sections found on the 8781-line job despite 18 markers present). The retry therefore fires and misses. Truncation signalling and the ungrouped fix are unaffected. Filing that parsing bug separately with a follow-up PR.

Traced `extract_step_log` and `_extract_log_sections` to confirm the zero-groups gap the issue describes, which PR 2755 didn't cover.

`original_length` is a line count, not chars: `github-mcp-server` returns it as `totalLines` from `downloadLogContent`, and a live call reported 8781 for a log of exactly 8781 lines.

Retry is sized from `original_lines` with no client-side ceiling, since `get_job_logs` alrontent window (`min(tailLines, contentWindowSize)`, 5000 by default).

### DIFF
--- a/integrations/github/tools/actions.py
+++ b/integrations/github/tools/actions.py
@@ -707,8 +707,12 @@ def get_github_actions_step_log(
         and wants_step
         and extracted["match_strategy"] == "full-log"
     ):
+        # original_length is characters (same units as len(log_text)); tail_lines
+        # is a line count. Estimate total lines from the tail's own chars-per-line.
+        chars_per_line = max(len(log_text) // max(log_text.count("\n"), 1), 1)
+        estimated_lines = max(original_length // chars_per_line, 1)
         retry_tail_lines = min(
-            max(original_length, _STEP_LOG_RETRY_TAIL_LINES), _STEP_LOG_MAX_RETRY_TAIL_LINES
+            max(estimated_lines, _STEP_LOG_RETRY_TAIL_LINES), _STEP_LOG_MAX_RETRY_TAIL_LINES
         )
         if retry_tail_lines > tail_lines:
             retry_text, retry_original_length, retry_error = _fetch_job_log(
--- a/integrations/github/tools/actions.py
+++ b/integrations/github/tools/actions.py
@@ -701,6 +701,8 @@ def get_github_actions_step_log(
     # length, once before accepting that fallback.
     wants_step = bool(step_name or failed_step or step_number is not None)
     truncated = original_length is not None and original_length > len(log_text)
+    retry_attempted = False
+    retry_error: str | None = None
     if (
         original_length is not None
         and truncated
@@ -715,6 +717,7 @@ def get_github_actions_step_log(
             max(estimated_lines, _STEP_LOG_RETRY_TAIL_LINES), _STEP_LOG_MAX_RETRY_TAIL_LINES
         )
         if retry_tail_lines > tail_lines:
+            retry_attempted = True
             retry_text, retry_original_length, retry_error = _fetch_job_log(
                 config, owner, repo, job_id, retry_tail_lines
             )
@@ -737,8 +740,10 @@ def get_github_actions_step_log(
             "job_conclusion": job.get("conclusion", ""),
             "job_steps": normalized_steps,
             "truncated": truncated,
-            "returned_length": len(log_text),
-            "original_length": original_length,
+            "returned_chars": len(log_text),
+            "original_chars": original_length,
+            "retry_attempted": retry_attempted,
+            "retry_error": retry_error,
         }
     )
     return extracted
--- a/integrations/github/tools/actions.py
+++ b/integrations/github/tools/actions.py
@@ -55,17 +55,20 @@ def _extract_workflow_jobs(result: dict[str, Any]) -> list[dict[str, Any]]:
 
 
 def _extract_log_text(result: dict[str, Any]) -> tuple[str, int | None]:
-    """Extract log text and its original (pre-tail) length from an MCP tool result.
+    """Extract log text and its original (pre-tail) line count from an MCP tool result.
 
-    ``original_length`` is the full log's length before ``tail_lines`` truncated
-    it, in the same units as the returned text's length; ``None`` when the MCP
-    response doesn't report it.
+    ``original_lines`` is the job's total log line count before ``tail_lines``
+    truncated it (github-mcp-server's ``get_job_logs`` reports this as
+    ``original_length``, despite the name it's a line count, not a character
+    count: it's the ring buffer's total line tally, from the same downstream
+    fetch that also produces the tailed ``logs_content``). ``None`` when the
+    MCP response doesn't report it.
     """
     json_result = _extract_json_text(result)
     if isinstance(json_result, dict) and "logs_content" in json_result:
         text = str(json_result["logs_content"] or "").strip()
-        original_length = json_result.get("original_length")
-        return text, original_length if isinstance(original_length, int) else None
+        original_lines = json_result.get("original_length")
+        return text, original_lines if isinstance(original_lines, int) else None
     return str(result.get("text") or "").strip(), None
 
 
@@ -560,7 +563,7 @@ _STEP_LOG_MAX_RETRY_TAIL_LINES = 20000
 def _fetch_job_log(
     config: Any, owner: str, repo: str, job_id: int, tail_lines: int
 ) -> tuple[str, int | None, str | None]:
-    """Fetch job log text via get_job_logs. Returns (text, original_length, error)."""
+    """Fetch job log text via get_job_logs. Returns (text, original_lines, error)."""
     log_result = call_github_mcp_tool(
         config,
         "get_job_logs",
@@ -574,8 +577,8 @@ def _fetch_job_log(
     )
     if log_result.get("is_error"):
         return "", None, str(log_result.get("text") or "unknown error")
-    text, original_length = _extract_log_text(log_result)
-    return text, original_length, None
+    text, original_lines = _extract_log_text(log_result)
+    return text, original_lines, None
 
 
 @tool(
@@ -665,7 +668,7 @@ def get_github_actions_step_log(
         ) | {"error": "Unexpected job metadata format"}
 
     # Fetch job logs
-    log_text, original_length, log_error = _fetch_job_log(config, owner, repo, job_id, tail_lines)
+    log_text, original_lines, log_error = _fetch_job_log(config, owner, repo, job_id, tail_lines)
     if log_error is not None:
         return code_host_unavailable_payload(
             source="github",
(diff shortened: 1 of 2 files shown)


muddlebee:
@larsspinetta12 pls review

larsspinetta12:
@greptile review again

larsspinetta12:
@sahilnyk 
Looks good overall:

- truncation metadata in the payload
- retry with a larger tail before accepting full-log
- ungrouped-only step_number=1 now correctly reports full-log

Two small things to fix below:
**1. Inline issue 1 — length semantics** --> On returned_length / original_length:
Problem: the issue asked for returned/total line counts, but these are character lengths (len(log_text) + MCP original_length). That’s fine if MCP reports chars, but please either:
- rename to something like returned_chars / original_chars, or
- document in the payload/tool description that these are characters, not lines,
so agents don’t treat them as line counts when sizing a follow-up fetch.

**Inline issue 2 — silent retry failure** --> On the if retry_error is None: block:
Problem: if the retry fetch fails, we silently keep the truncated first response. Consider adding retry_attempted: true and/or retry_error so the agent can tell “we tried a larger window and it failed” vs “step really wasn’t there.”

@ me in a comment here when those are done!

sahilnyk:
@larsspinetta12 added those, please review!

Davidson3556 (changes requested):
Nice fix — returning `[]` for the ungrouped case is the right root cause, and surfacing `retry_error` is a good touch.

One real issue: `truncated` will false-positive. `_extract_log_text` strips the tail but `original_length` is unstripped, so any complete log ending in a newline reports truncated=True — and when the step isn't found, that fires a wasted second fetch. Compare against the unstripped length instead.

Also: `original_length` isn't referenced anywhere else in the repo — can you confirm the pinned github-mcp-server version actually returns it? Otherwise the retry silently no-ops. A test for the missing-field case would pin that down.

lastly fix merge conflict

sahilnyk:
@Davidson3556 Addressed your suggestions and fixed them, please review again.

integrations/github/tools/actions.py
```diff
@@ -178,7 +188,9 @@ def _extract_log_sections(log_text: str) -> list[dict[str, str]]:
         _append_log_section(sections, current_name or UNGROUPED_SECTION_NAME, current_lines)
 
     if not saw_group:
-        return [{"name": "full-log", "text": log_text.strip()}]
+        # No ##[group] markers at all: nothing to select by name/number.
+        # extract_step_log's own full-log fallback handles this case.
+        return []
```
Davidson3556, line 197: This is the right fix for the third bug. Returning `[]` kills the fake section at the source instead of teaching `extract_step_log` to ignore it, and `extract_step_log` already had a working full-log fallback to land on. One caller, no blast radius.

integrations/github/tools/actions.py
```diff
@@ -544,6 +556,31 @@ def list_github_actions_run_jobs(
     return payload
 
 
+_STEP_LOG_RETRY_TAIL_LINES = 5000
```
Davidson3556: `UNGROUPED_SECTION_NAME` lives at the top of the module with the other constants. These two are stranded mid-file. Move them up to match.

integrations/github/tools/actions.py
```diff
@@ -669,10 +693,44 @@ def get_github_actions_step_log(
...
         step_number=step_number,
     )
+
+    # A requested step that missed because the tail cut off its ##[group] block
+    # looks identical to "no such step" (both fall back to match_strategy
+    # "full-log"). Re-fetch a bigger tail, sized from the log's own reported
+    # total line count, once before accepting that fallback.
+    wants_step = bool(step_name or failed_step or step_number is not None)
+    returned_lines = len(log_text.splitlines())
+    truncated = original_lines is not None and original_lines > returned_lines
```
Davidson3556, line 710: `_extract_log_text` strips the body but `original_length` counts it unstripped, so a complete log with blank lines at either end reports `truncated: True` and burns a second `get_job_logs`. Repro: 6-line log with trailing blanks returns `original: 6 | returned: 4`. Count lines before stripping.

Davidson3556 (commented):
Ran it: 40 tests pass, ruff/format/mypy clean. Also confirmed `original_length` is `totalLines` in github-mcp-server, so your docstring is right.

Fix #3 is the best part — returning `[]` kills the fake section at the source rather than special-casing downstream.

One blocking comment on the retry path (the fallback payload can hit ~340KB straight into the agent's context), two smaller bugs I reproduced, and some nits. Also: the description says `returned_length`/`original_length`, the code says `returned_lines`/`original_lines`.

integrations/github/tools/actions.py
```diff
@@ -669,10 +693,44 @@ def get_github_actions_step_log(
...
+    retry_attempted = False
+    retry_error: str | None = None
+    if (
+        original_lines is not None
+        and truncated
+        and wants_step
+        and extracted["match_strategy"] == "full-log"
+    ):
+        retry_tail_lines = min(
+            max(original_lines, _STEP_LOG_RETRY_TAIL_LINES), _STEP_LOG_MAX_RETRY_TAIL_LINES
```
Davidson3556: The 5000 floor pulls against your own description of sizing the retry from `original_length`. For a 603-line log this asks for 5000 lines. `min(original_lines, _STEP_LOG_MAX_RETRY_TAIL_LINES)` is exact and lets you delete a constant.

integrations/github/tools/actions.py
```diff
@@ -669,10 +693,44 @@ def get_github_actions_step_log(
...
+            retry_text, retry_original_lines, retry_error = _fetch_job_log(
+                config, owner, repo, job_id, retry_tail_lines
+            )
+            if retry_error is None:
+                log_text, original_lines = retry_text, retry_original_lines
+                extracted = extract_step_log(
+                    log_text,
+                    step_name=step_name or failed_step,
+                    step_number=step_number,
+                )
```
Davidson3556, line 731: Blocking. When the retry still misses, `log_text` becomes the 20000-line retry body and goes straight into the agent's context — nothing caps it before `context_budget` fires. Repro: 60k-line log, `step_name="Run tests"` -> 20000 lines / 340KB, vs ~500 lines / 8.5KB before this PR. Only promote the retry body if it matched:

```python
if retry_error is None:
    retry_extracted = extract_step_log(retry_text, step_name=step_name or failed_step, step_number=step_number)
    if retry_extracted["match_strategy"] != "full-log":
        extracted = retry_extracted
        log_text, original_lines = retry_text, retry_original_lines
        returned_lines = len(log_text.splitlines())
        truncated = original_lines is not None and original_lines > returned_lines
```

integrations/github/tools/actions.py
```diff
@@ -669,10 +693,44 @@ def get_github_actions_step_log(
...
+                returned_lines = len(log_text.splitlines())
+                truncated = original_lines is not None and original_lines > returned_lines
```
Davidson3556: Line 725 overwrites `original_lines`, so a retry response without `original_length` turns a known count into `None` and flips `truncated` to `False`. Repro: 100k-line log, retry omits the field -> `truncated: False, returned: 20000`. Guard it:

```python
if retry_original_lines is not None:
    original_lines = retry_original_lines
```

integrations/github/tools/actions.py
```diff
@@ -682,6 +740,11 @@ def get_github_actions_step_log(
             "job_name": job.get("name", ""),
             "job_conclusion": job.get("conclusion", ""),
             "job_steps": normalized_steps,
+            "truncated": truncated,
```
Davidson3556, line 755: These five keys only exist on the success path — every `code_host_unavailable_payload` return gives back `['available', 'error', 'log_text', 'source']`. Default them there too so the agent doesn't have to key-check.

integrations/github/tools/actions.py
sahilnyk: Fixed, but at the source. `_extract_log_text` doesn't strip at all now, since `extract_step_log` strips every branch it returns anyway, so `log_text` comes out identical.

Test uses your repro shape: blank lines at both ends gives `returned_lines: 5, truncated: false`, no retry. Fails on the old code with `assert 3 == 5`.

integrations/github/tools/actions.py
sahilnyk: Fixed, close to your sketch. Extracts from the retry text first, swaps in the bigger body only if `match_strategy != "full-log"`.

Your 340KB was conservative. On a real 8781-line job it was 696,693 chars. It's 57,683 now.

integrations/github/tools/actions.py
sahilnyk: All four error paths go through one `_step_log_unavailable_payload` helper now. Also killed four copies of the same call.

integrations/github/tools/actions.py
sahilnyk: Deleted both instead of moving them, so nothing is stranded now.

Davidson3556 (commented):
LGTM looks good to me . well done

Davidson3556 (approved):
Verified the fixes by re-running the probes rather than reading the diff — all five findings are closed, with a regression test each. CI green.

Needs an update from main before merge.

PR #4456, opened 2026-07-28
fix(config): add a local fallback credential store when no OS keychain works

### LINKED ISSUE #4244: [IMPROVEMENT] Add a local fallback credential store for when the OS keyring is unavailable
### Current state

`save_keyring_secret` (`config/llm_keyring.py`) and `sync_env_secret` (`config/env_file.py`) have no fallback: any OS keychain failure raises a `RuntimeError` that onboarding currently treats as fatal, with no path to a working state. That single gap is the root cause of several separately-filed issues:

- #3348 — `opensre onboard` fails outright on macOS when the login Keychain is locked.
- #1403 — `opensre onboard` fails outright on headless Linux/EC2, where `keyring` has no working backend (`keyring.backends.fail.Keyring`).
- #3591 — "better LLM key failure UX during onboarding" is asking for exactly this fallback.
- #3295 — macOS shows 3 separate Keychain authorization prompts per run, because each read/write of the same credential (status check, save, verify) hits the OS backend independently with no per-process memoization.
- #4212 — same file area; the `is_sensitive_env_key` classifier's terminal-token set has real-world gaps (`passphrase`, `bearer`, `jwt`, bare `pass`/`auth`).

### Desired state

Introduce a small local fallback tier, tried only when the OS keyring backend is genuinely unavailable:

- A local JSON store under `OPENSRE_HOME_DIR`, owner-only permissions enforced at creation (not a best-effort `chmod` after the fact), distinct from the project `.env` (which stays secret-free by design).
- Save path: try the OS keyring first; fall back to the local store only on a keyring failure; raise only if both tiers fail.
- Read path: env → keyring → fallback store, so a credential saved via the fallback during onboarding is actually usable at request time, not just accepted and then silently unresolvable.
- Logout/delete clears both tiers, not just the keyring entry.
- A per-process cache of keyring reads (plus a sticky "backend unavailable" flag after the first failure) so a single run touches the OS backend at most once per credential, instead of re-prompting macOS Keychain authorization on every read/write of the same key.
- Expand `is_sensitive_env_key`'s terminal-token set to cover `passphrase`, `pass`, `bearer`, `jwt`, `auth`, checked against `config/constants/**` for false positives (e.g. `LLM_AUTH_METHOD`'s terminal segment is `method`, not `auth`).

A candidate implementation with tests (including regression coverage from a Greptile review pass — request-time exception handling, fallback-file permission enforcement, logout cleanup) was put up as #4243 and closed pending this issue, per @muddlebee's note. Happy to reopen that branch once there's agreement on the approach, or split it into smaller per-issue PRs if that's preferred over one cross-cutting change.

---

### macOS — login Keychain locked or inaccessible (#3348)

**Environment:** macOS (Apple Silicon and Intel); `keyring.backends.macOS.Keyring`

1. Clone the repo (up-to-date `main`).
2. Run `opensre onboard`.
3. Choose **OpenAI** as LLM provider; choose **API key** auth method.
4. Enter a valid `OPENAI_API_KEY` when prompted.
5. If the macOS login Keychain is locked (e.g. after waking from sleep without re-authenticating), or if the first-run Keychain permission dialog was dismissed, the wizard exits with:
   ```
   ✗  Secure local credential storage is unavailable on this machine.
   ▲  OpenSRE could not save your API key to the local system keychain.
      Current keyring backend: keyring.backends.macOS.Keyring.
      Make sure your system keychain service is installed and unlocked, then rerun `opensre onboard`.
      For deeper diagnostics run `python -m keyring diagnose`.
   ```
6. The key is **not saved** and onboarding does not complete.

---

### Headless Linux / EC2 — no working keyring backend (#1403)

**Environment:** Fresh Ubuntu 22.04 LTS EC2 instance (minimal SSH session, no desktop keyring); `keyring.backends.fail.Keyring`

1. Run `opensre onboard`.
2. Choose **Quickstart**.
3. Choose **Anthropic** (or any API-key provider).
4. Enter a valid API key when prompted.
5. The wizard aborts with:
   ```
   Secure local credential storage is unavailable. Set ANTHROPIC_API_KEY in your shell or configure a working system keychain.
   ```
6. Quickstart does not finish and the key is not persisted.

---

### macOS — triple Keychain authorization prompt per run (#3295)

**Environment:** macOS (all versions)

1. Install opensre on macOS.
2. Run the app (any command that reads credentials on startup).
3. Observe **3 separate** macOS Keychain authorization dialogs appearing sequentially.

**Root cause:** `keyring.get_password()` / `keyring.set_password()` are called three times independently per credential (status check, save, verify) with no per-process memoization, causing macOS Keychain to re-prompt on each call.

---

## Acceptance criteria (consolidated)

> The items below augment the "Desired state" spec above.

**Keyring fallback (covers #3348 and #1403)**
- [ ] `opensre onboard` completes successfully on macOS when the login Keychain is locked or inaccessible.
- [ ] `opensre onboard` completes successfully on headless Linux / EC2 where no working keyring backend exists (`keyring.backends.fail.Keyring`).
- [ ] A credential saved via the fallback during onboarding is resolvable at request time (read path: env → keyring → fallback store).
- [ ] Both the OS keyring entry and the fallback store entry are cleared on logout/delete.

**Per-process keyring cache (covers #3295)**
- [ ] A single process run triggers the OS Keychain authorization dialog **at most once** per unique credential, regardless of how many internal reads/writes occur.
- [ ] A sticky "backend unavailable" flag prevents repeated OS keyring calls after the first failure.

**`is_sensitive_env_key` classifier (covers #4212)**
- [ ] `passphrase`, `pass`, `bearer`, `jwt`, and `auth` added as sensitive terminal tokens (verified against `config/constants/**` for false positives — e.g. `LLM_AUTH_METHOD`'s terminal segment is `method`, not `auth`).
- [ ] A naming convention for new integrations ("last name-segment must be a sensitive terminal token") is documented so the heuristic stays the source of truth.
- [ ] Non-goals: per-`SetupField` storage flag and one-shot migration/backfill remain explicitly out of scope.

**UX copy (covers #3591)**
- [ ] Error messaging when the keyring is unavailable is user-friendly and actionable (guides toward the fallback path rather than asking the user to configure a system keychain).
- [ ] CI green on relevant CLI/onboarding tests.

---

*Issues #3348, #1403, #3591, #3295, and #4212 are superseded by this issue and have been closed. Their unique reproduction steps and acceptance criteria have been folded in above.*

muddlebee:
Fixes #4244

Steps 1 and 2 of #4244. The per-process read cache (#3295) and the `is_sensitive_env_key` token expansion (#4212) are deliberately **not** here — see "Left for follow-ups" below.

**The bug.** `save_keyring_secret` raised, `_persist_env_secret` returned `False`, and `sync_env_secret` re-raised as fatal. There was no path to a working state when the OS keyring could not store a secret — every headless Linux/EC2 box (`keyring.backends.fail.Keyring`) and any macOS run with a locked login Keychain.

**Step 1 — unify (no behavior change).** Five call sites each reimplemented "ask the keyring", with their own error handling, which is why previous attempts kept leaking a tier at one of them. They now delegate to one policy object:

| File | Function |
| --- | --- |
| `config/llm_credentials.py` | `resolve_env_credential` |
| `config/llm_auth/credentials.py` | `resolve_for_request` |
| `config/llm_auth/credentials.py` | `has_llm_api_key` |
| `config/llm_auth/credentials.py` | `llm_api_key_source` |
| `config/env_file.py` | `_persist_env_secret` |

New package, one purpose per file:

```
config/secrets/
  backend.py     SecretBackend protocol + KeyringUnavailableError/Reason
  os_keyring.py  the keyring tier; normalizes every backend failure
  local_file.py  owner-only JSON store
  store.py       the tier policy
  guidance.py    user-facing "storage didn't work" copy
  records.py     record: JSON blobs, now inheriting the same policy
```

`config/llm_keyring.py` is deleted, not left forwarding. `config/llm_credentials.py` stays as the façade — it already was one, and 30+ integrations import `resolve_env_credential` from it, so rewriting those imports would be churn without a canonical-path benefit.

**Step 2 — the fallback tier.**

```
read   env  ->  OS keyring  ->  ~/.opensre/credentials.json
write  keyring, falling back to the file only when the keyring cannot store it
delete both tiers, always
```

Four things worth reviewer attention:

1. **Error normalization is the whole trick.** `keyring` backends disagree about how they fail — macOS raises `KeyringLocked`, SecretService raises a bare `RuntimeError` when D-Bus is unset, a read-only home raises `OSError`, no backend raises `NoKeyringError`. `os_keyring.py` collapses all of them into one `KeyringUnavailableError` so the fallback policy has exactly one condition to test. (For the same reason, `keyring`'s own `ChainerBackend` is not usable here: its `set_password` catches only `NotImplementedError`, so a locked keychain propagates and never reaches a lower tier.)

2. **Falling back is reported, not silent.** With no backend at all there is nothing to downgrade *from*, so it just works — same as `~/.aws/credentials`, `~/.config/gh/hosts.yml`, `~/.docker/config.json`. A *locked* keychain is different: a working secure store exists, so `_persist_llm_api_key` prints which file the credential went to and how to move it back. One `if` on `KeyringUnavailableReason`.

3. **File permissions are established at creation, not narrowed after.** `mkstemp` (0600) → write → `os.replace`, mirroring `config/llm_auth/records.py:_write_unlocked`. There is no window where the plaintext is world-readable. Not encrypted on purpose: a passphrase prompt defeats the headless case this exists for, and a key stored beside the ciphertext protects nothing.

4. **Path is `host_home()`, not `opensre_home()`.** The org-scoped root would put one organization's secrets on another's mounted volume in a deployed silo. Deployed silos get credentials from the environment anyway.

Two switches, distinct meanings: `OPENSRE_DISABLE_KEYRING` keeps its current fail-closed contract (no local persistence at all — every existing test relying on it is unchanged); `OPENSRE_DISABLE_CREDENTIAL_FALLBACK` is new, for operators who want the keychain but would rather fail than have a secret reach disk.

Simulating a headless box (`keyring.backends.fail.Keyring`) — before, this aborted onboarding:

```
$ pytest tests/config/test_secret_store.py -q
..............                                                   [100%]
14 passed
```

```
test_save_falls_back_to_local_file_when_no_backend_exists  #1403 headless Linux/EC2
test_save_falls_back_when_the_keychain_is_locked           #3348 locked macOS Keychain
test_fallback_credential_is_resolvable_at_request_time     accepted at setup == works later
test_provider_logout_clears_the_fallback_copy              logout actually logs out
test_fallback_file_is_never_readable_beyond_its_owner      mode == 0o600
test_save_raises_when_both_tiers_refuse                    still fails loudly when it should
test_fallback_can_be_refused_so_storage_stays_fail_closed  opt-out works
test_one_backend_probe_per_process_after_a_failure         5 reads -> 1 backend call
```

The stored file, after a save on a machine with no keychain:

```
$ ls -l ~/.opensre/credentials.json
-rw------- 1 user user 68 ...
$ cat ~/.opensre/credentials.json
{"secrets": {"ANTHROPIC_API_KEY": "sk-..."}, "version": 1}
```

### Left for follow-ups (deliberately not in this PR)

- **#3295** — the sticky "backend unavailable" flag *is* here (a broken-D-Bus box must not pay a timeout per read now that reads are more frequent), but the read memoization that collapses the three macOS Keychain prompts is not. It belongs with its own tests.
- **#4212** — the `is_sensitive_env_key` token expansion is independent and six lines. Worth noting for whoever picks it up: I grepped `config/ integrations/ tools/ platform/ surfaces/ gateway/` and **no** env var currently ends in `_PASS`, `_AUTH`, `_JWT`, `_BEARER`, or `_PASSPHRASE`, so it fixes nothing today and breaks nothing today — it is forward-looking hardening plus the naming-convention doc, and should be described that way rather than as a bug fix.

---

The problem is not "we need a fallback store" — that part is easy. It is that "how do I ask about a secret" was implemented five times with five different error-handling behaviors, so adding a tier meant adding it five times and missing one. Hence step 1 before step 2: with one resolver, the fallback is a change in one file and the classes of bug that reviews kept catching (logout leaving a live copy, a backend exception escaping before the fallback is reached) become structurally impossible rather than individually patched.

Alternative considered and rejected: leaning on `keyring`'s own backend chain by installing `keyrings.alt`. It reads elegantly — the chain would be SecretService → encrypted file → plaintext file with no OpenSRE changes at all — but reading the source kills it twice. `ChainerBackend.set_password` catches only `NotImplementedError`, so a locked macOS Keychain propagates and never reaches a lower tier (#3348 stays broken); and `keyrings.alt.file_base._ensure_file_path` creates the file with `open(...,'w')` and *then* chmods 0600, which is the exposure window we are specifically trying not to have. So the error handling has to be ours regardless, and the file store is better written against the pattern already proven in `config/llm_auth/records.py`.

Also considered: making `OPENSRE_DISABLE_KEYRING` engage the fallback. Rejected — that flag currently means fail-closed, silently turning it into "write plaintext instead" would surprise exactly the users who set it, and it would have changed the meaning of every test in the suite that relies on it.

Key components: `os_keyring.py` owns the messy vendor reality and exposes three clean methods; `local_file.py` owns atomic owner-only persistence; `store.py` owns the policy and is the only file that knows there is more than one tier; everything else delegates.

---

**Local verification:** `make lint`, `make format-check`, `make typecheck`, `make verify-integrations-smoke` all clean. `make test-cov` — 12700 passed; the residual failures are environment-specific on my box, not from this change: two `test_env_sync` D-Bus cases (`LiveNetworkCallBlocked` — reproduced on unmodified `main`), a `test_runtime_metadata_perf` timing threshold under `-n auto`, and one `live_llm` scenario needing `OPENAI_API_KEY`. `tests/config/` and `tests/cli/wizard/` are 446/446 green.


muddlebee:
@greptile review

muddlebee:
@greptile review

Davidson3556:
this is good. well done @muddlebee 
 please fix the merge conflict, also add a screenshot.

muddlebee:
@greptile review

muddlebee:
@greptile review

PR #4457, opened 2026-07-28
Fix/sm 79 s3 client comments

sarthak-cc:
Issue: SM-79
Closes SM-79

Removes the "# Get both versions" comment in
integrations/aws/s3_client.py's compare_versions function, which only
restated the two get_object calls directly below it.

No behavior change. Lint (ruff check) and existing unit tests (28 passed) pass.
image


muddlebee:
missing PR template

PR #4462, opened 2026-07-28
fix(github-mcp): single MCP session in validate_github_mcp_config (#3565)

### LINKED ISSUE #3565: ## GitHub Connect confirmation opens one MCP session per probe call (3 sessions per confirmation)

deveshpandee:
Fixes #3565 

**Before this change:** validate_github_mcp_config opened a separate
_open_github_mcp_session context for every internal probe call
(list_tools, get_me, and each repo-access probe), paying a full
TCP connect + MCP initialize handshake each time — typically 3×
the necessary latency.

**Changes** 
- Add _validate_github_mcp_config_async(session, ...) — all three
  validation steps (list_tools -> get_me -> repo probe) run on the
  single ClientSession passed in by the caller.
- Rewrite validate_github_mcp_config to open _open_github_mcp_session
  exactly once, drive the async inner via _run_async, and keep the
  existing synchronous public signature unchanged.

---

**Before:** 
 - Every time a github MCP config is validated , it opened a  new connection to the MCP server for each check it needed to do.
 - It checked 3 things: list of tools, user's identity (get_me), and repo access - each one started its own separate connection.
 - That means 3 full TCP handshakes + 3 MCP "hello" handshakes just to validate one config.
 - Wasteful and slow -like hanging up and calling someone back 3 times in a row to ask 3 questions.

**After:**
 - Now it opens the connection once, asks all 3 questions on the same connection, then closes it.
 - A new internal helper (_validate_github_mcp_config_async) handles all 3 checks in sequence using that single shared connection.
 - The function we call from the outside looks and works exactly the same: no breaking changes.
 - End result: ~3× faster validation, way fewer wasted resources

---


integrations/github/mcp.py
```diff
@@ -1229,79 +1229,159 @@ def validate_github_mcp_config(
...
+) -> GitHubMCPValidationResult:
+    """Inner async validation — One MCP session per validation run."""
+
+    list_tools_result = await session.list_tools()
+    raw_tools = list_tools_result.tools
+    tools: list[dict[str, Any]] = [
+        {
+            "name": t.name,
+            "description": t.description or "",
+            "input_schema": getattr(t, "input_schema", None),
```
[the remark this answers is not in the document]
deveshpandee: the attribute that `list_github_mcp_tools` uses is `input_schema` not `inputSchema`. i have verified 
image

Davidson3556:
kindly address greptile feedback

deveshpandee:
> kindly address greptile feedback

Will implement the change in evening

muddlebee:
@deveshpandee demo is missing, pls add a synthetic scenario which will call this affected MCP

deveshpandee:
@greptile review

deveshpandee:
@YauhenBichel and @cerencamkiran please review

cerencamkiran:
I couldn't find the before/after results. Could you add them to the pr, please?

Davidson3556 (approved):
kindly fix the merge conflict.  great work

integrations/github/mcp.py
```diff
@@ -1230,79 +1230,177 @@ def validate_github_mcp_config(
...
-            structured = raw_structured
-        profile_pub, profile_priv = _repo_visibility_counts_from_get_me_profile(
-            structured, me_result.get("text", "")
+async def _validate_github_mcp_config_async(
+    session: ClientSession,
+    *,
+    repo_view: GitHubMcpRepoView = "auto",
+    repo_visibility: GitHubMcpRepoVisibilityFilter = "any",
+) -> GitHubMCPValidationResult:
+    """Inner async validation — One MCP session per validation run."""
```
Davidson3556, line 1263: Lowercase the `One`, and a comma reads better than the dash: `"Run every validation step on a single MCP session."`

integrations/github/mcp.py
```diff
@@ -1230,79 +1230,177 @@ def validate_github_mcp_config(
...
+    raw_tools = list_tools_result.tools
+    tools: list[dict[str, Any]] = [
+        {
+            "name": t.name,
+            "description": t.description or "",
+            "input_schema": getattr(t, "inputSchema", None),
+        }
+        for t in raw_tools
+    ]
+    tool_names = tuple(sorted(t["name"] for t in tools))
```
Davidson3556, line 1275: Same comprehension as `list_github_mcp_tools` (~line 662), `getattr(t, "inputSchema", None)` included. Extract `_tool_defs(tools)` and call it from both.

integrations/github/mcp.py
```diff
@@ -1230,79 +1230,177 @@ def validate_github_mcp_config(
...
+    except Exception as err:
+        logger.debug("GitHub MCP tool call failed: get_me", exc_info=True)
+        me_result = {
+            "is_error": True,
+            "text": _root_cause_message(err),
+            "content": [],
+            "structured_content": None,
+        }
+    me_result["tool"] = "get_me"
+    me_result["arguments"] = {}
```
Davidson3556, line 1312: This is `_call_tool_async` + `call_github_mcp_tool`'s except clause inlined, and it appears again in the probe loop below. Extract the session-scoped version and put the other two on top of it:

```python
async def _call_tool_on_session(session, tool_name, arguments=None):
    try:
        payload = _tool_result_to_dict(await session.call_tool(tool_name, arguments or {}))
    except Exception as err:
        logger.debug("GitHub MCP tool call failed: %s", tool_name, exc_info=True)
        payload = {"is_error": True, "text": _root_cause_message(err), "content": [], "structured_content": None}
    payload["tool"] = tool_name
    payload["arguments"] = arguments or {}
    return payload
```

integrations/github/mcp.py
```diff
@@ -1230,79 +1230,177 @@ def validate_github_mcp_config(
...
+        except Exception as err:
+            logger.debug("GitHub MCP tool call failed: %s", repo_tool, exc_info=True)
+            list_result = {
+                "is_error": True,
+                "text": _root_cause_message(err),
+                "content": [],
+                "structured_content": None,
+            }
+        list_result["tool"] = repo_tool
+        list_result["arguments"] = repo_args
```
Davidson3556, line 1383: Second copy — `list_result = await _call_tool_on_session(session, repo_tool, repo_args)` once that helper exists.

Aside: one shared session means a transport failure on the first probe leaves the rest running on a dead connection. `_is_recoverable_repo_probe_error` only continues for search 422s, so nothing regresses, but the retry loop is weaker than it looks.

tests/synthetic/test_github_mcp_scenario.py
```diff
@@ -0,0 +1,215 @@
...
+        "input_schema": {
+            "type": "object",
+            "properties": {"query": {"type": "string"}},
+            "required": ["query"],
+        },
+    }
+]
+
+
+def _make_session_patcher(
```
Davidson3556, line 63: This factory and `_make_fake_session_patcher` in `test_github_mcp_validation.py` are the same 35 lines. Pick one home.

Bigger question: both success tests here duplicate existing ones (`..._success_includes_repo_samples`, `..._uses_search_repositories_when_no_list_tool`). The only new assertion is `session_open_count[0] == 1` — thread a counter through the existing fake and drop the other ~180 lines.

Davidson3556 (commented):
Verified rather than eyeballed. Loaded the pre-PR `mcp.py` from the merge-base and ran both versions through 11 scenarios covering every branch — every field of `GitHubMCPValidationResult` matched, 0 diffs. Session count with real helpers against a counting fake: 3 -> 1. Messages and `failure_category` values untouched. 32 tests pass, ruff and mypy clean.

Two things before merge:

**Measurement is missing.** #3565 lists before/after wall time as a done-when item, and the Demo section plus the whole checklist are commented out. "~3x faster" has nothing behind it, and the issue says ~5s of GitHub-side latency won't move, so the real number is smaller.

**Should land net-negative.** It's +486/-143 and a lot of it is copy-pasted from helpers a few hundred lines up. Comments inline.

Also 41 commits behind main. Nothing else touched these files since your merge-base, so the rebase is clean.

Davidson3556:
kindly address the suggested comment  thank you

Davidson3556:
@deveshpandee I pushed the review fixes straight to your branch so this isn't waiting on a round trip

Davidson3556:
@greptile review

Davidson3556:
image

PR #4475, opened 2026-07-28
fix(memory): break the files ↔ index import cycle

YauhenBichel:
CodeQL flagged a cycle between `core/domain/memory/files.py` and `core/domain/memory/index.py`. One direction is correct:

- `index.py` → `files.py` for `write_text_atomically`. Index is generated  output; it sits above the file primitives.

The other was a **back-edge from the low layer up to the high one**:

- `files.py` → `index.py` for `write_index`, inside `ensure_memory_store()`.

The import was already function-local, which is why it needed a structural fix rather than a lazier one. `AGENTS.md` records this exact footgun: CodeQL counts function-local and `TYPE_CHECKING` imports as part of a cycle, so hiding an import inside a function never clears the alert.

### DIFF
--- a/core/domain/memory/files.py
+++ b/core/domain/memory/files.py
@@ -1,4 +1,10 @@
-"""Filesystem path and write primitives for long-term memory."""
+"""Filesystem path and write primitives for long-term memory.
+
+The lowest layer of the memory package: it knows where files live and how to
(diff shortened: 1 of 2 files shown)


YauhenBichel:
@greptile review

PR #4486, opened 2026-07-29
refactor(google-docs): use HTTPStatus in the API error handler (SM-29)

### LINKED ISSUE #4287: [IMPROVEMENT] SM-29 — Use HTTPStatus in Google Docs client
**Task ID:** SM-29 · **Level:** Junior · **Size:** S (1 file) · **Good first issue**

### Current state

`integrations/google_docs/client.py` compares raw HTTP status numbers. The project rule is to use `http.HTTPStatus` instead of bare numeric literals.

### Desired state

Every status comparison / `status_code=` assignment uses the named enum (`HTTPStatus.OK`, `HTTPStatus.UNAUTHORIZED`, `HTTPStatus.NOT_FOUND`, …). Behavior stays identical.

### Steps

1. `from http import HTTPStatus`
2. Replace bare status numbers with the matching `HTTPStatus` members (or named constants for codes not in the enum).
3. `make lint format-check typecheck`
4. Run the nearest unit tests for this file if present.

### Done when

- [ ] No bare HTTP status numbers left that should be `HTTPStatus` (except documented non-enum codes)
- [ ] lint / format / typecheck pass

**Do not:** edit a second file. One ID = one file = one PR. Do not change behavior or user-facing messages.

**File:** `integrations/google_docs/client.py`

---

To claim this task, comment here (or in Discord `#contribute`): *"I'd like to take SM-29"*. One task per pull request.

SawhneySatvik:
Replace the bare HTTP status literals in `_handle_google_api_error` with the matching `http.HTTPStatus` members, per the project rule against numeric status literals.

- `403` -> `HTTPStatus.FORBIDDEN`
- `404` -> `HTTPStatus.NOT_FOUND`
- `429` -> `HTTPStatus.TOO_MANY_REQUESTS`
- `400` -> `HTTPStatus.BAD_REQUEST`

`HTTPStatus` is an `IntEnum`, so comparing it against the integer status the Google client puts on `exc.resp.status` behaves identically. The generic `elif status_code:` fallback and every error message are unchanged.

Fixes #4287

SM-29. `integrations/google_docs/client.py` was the only file touched.

Every status comparison in that file lives in one place, `_handle_google_api_error`, which maps a `googleapiclient` exception onto an operator-readable message. `status_code` there comes from `getattr(exc.resp, "status", None)`, so it is an `int` or `None`; `HTTPStatus` being an `IntEnum` means both the match and the `None` fall-through behave exactly as before.

Left alone: the trailing `elif status_code:` catch-all, which is a truthiness check on "some code we have no specific message for" rather than a comparison against a particular status, so there is nothing to name.

```
$ uv run ruff check integrations/google_docs/client.py
All checks passed!

$ uv run ruff format --check integrations/google_docs/client.py
1 file already formatted

$ uv run mypy integrations/google_docs/client.py

$ uv run pytest tests/integrations/google_docs/ tests/tools/test_google_docs_create_report_tool.py -q
============================== 47 passed in 0.63s ==============================
```

`tests/integrations/google_docs/test_integration.py` already exercises this function directly `test_handle_google_api_error_with_403`, `_with_404`, `_with_429`, so the existing suite pins the exact branches this PR touches.

image

Full harness also run green on this change:

```
$ make lint                        -> All checks passed!
$ make format-check                -> 2850 files already formatted
$ make typecheck                   -> Success: no issues found in 1484 source files
$ make verify-integrations-smoke   -> 31 passed in 2.10s
$ make test-cov                    -> 12638 passed, 60 skipped in 126.66s
```

---

`_handle_google_api_error` is the single place where a raw Google API exception becomes something an operator can act on "insufficient permissions to create document, check the service account has access" instead of a stack-trace string. Every other method in `GoogleDocsClient` funnels its `except` branch through it. Since that mapping *is* the function, the status codes are exactly the part a reader needs to follow, and naming them removes the step where you translate `429` in your head.

The detail I checked rather than assumed: this is not an httpx-style `response.status_code`. It comes from `getattr(exc.resp, "status", None)` on a `googleapiclient` `HttpError`, and it can legitimately be `None` when the exception has no `resp` shaped the way we expect. `None == HTTPStatus.FORBIDDEN` evaluates `False` exactly as `None == 403` did, so the `None` case still falls past all four branches, past the `elif status_code:` guard, and out with the generic `str(exc)` message. Unchanged.

I left that final `elif status_code:` as-is on purpose, swapping it would have meant inventing a comparison that isn't there.

The alternative I considered was collapsing the `if/elif` chain into a dict keyed by `HTTPStatus` with the message templates as values. It reads more cleanly and would scale better if more codes get added. I rejected it because it changes control flow rather than just naming constants: it makes the diff harder to check against the existing tests, and the issue scopes this to replacing literals. If maintainers would prefer the dict form, it is a small follow-up.

---

- [ ] If it is a core feature, I have added thorough tests — n/a, behavior-preserving refactor already covered by direct unit tests on this function

### DIFF
--- a/integrations/google_docs/client.py
+++ b/integrations/google_docs/client.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import importlib
 import logging
+from http import HTTPStatus
 from pathlib import Path
 from typing import Any, cast
 
@@ -28,15 +29,15 @@ def _handle_google_api_error(exc: Exception, operation: str) -> dict[str, Any]:
     # Try to extract HTTP error details
     if hasattr(exc, "resp"):
         status_code = getattr(exc.resp, "status", None)
-        if status_code == 403:
+        if status_code == HTTPStatus.FORBIDDEN:
             error_msg = (
                 f"Insufficient permissions to {operation}. Check service account has access."
             )
-        elif status_code == 404:
+        elif status_code == HTTPStatus.NOT_FOUND:
             error_msg = f"Resource not found while {operation}. Verify folder/document ID."
-        elif status_code == 429:
+        elif status_code == HTTPStatus.TOO_MANY_REQUESTS:
             error_msg = f"Rate limit exceeded while {operation}. Please retry later."
-        elif status_code == 400:
+        elif status_code == HTTPStatus.BAD_REQUEST:
             error_msg = f"Invalid request while {operation}. Check parameters."
         elif status_code:
             error_msg = f"HTTP {status_code} error while {operation}: {exc}"


SawhneySatvik:
@greptile review

PR #4496, opened 2026-07-29
feat: task management v1

VaibhavUpreti:


VaibhavUpreti:
@greptile review

Addressed the three P1s on `d28eb5cec`:
1. **Obsolete reminder schedule** — `_schedule_item_reminder` now disables prior enabled `WORK_ITEM_REMINDER` tasks for the same work item before creating a new one.
2. **Read failures erase durable tasks** — corrupt/unreadable/`non-list` store files raise `WorkItemStoreError` instead of returning `[]`, so mutations cannot overwrite with empty state.
3. **Reminder state precedes delivery** — `last_reminded_at` is recorded in the executor only after successful delivery, not during message build.

PR #4514, opened 2026-07-29
refactor(protocols): docstring-only Protocol bodies (CodeQL py/ineffectual-statement)

YauhenBichel:
Two CodeQL alerts, plus the same defect wherever else it appeared.

**`py/ineffectual-statement` (#2049)** — five `Protocol` methods in `platform/deployment_contracts/ports.py` had a bare `...` body. Python needs at least one statement in a function body, and a Protocol method has no implementation, so it needs a placeholder. `...` is the idiom inherited from `.pyi` stub files; at runtime it evaluates the `Ellipsis` singleton and discards it, which is exactly what the rule detects — a discarded expression value is usually a typo (`x.show` for `x.show()`).

A docstring is the same kind of placeholder — a string-literal statement — but CodeQL explicitly exempts it ("will not flag a statement consisting solely of a string") and it documents the contract. Both bodies return `None`; there is no behavioural difference.

Sweeping for the same shape found two more files the scanner had not reported:

- `integrations/llm_cli/base.py` — three methods had a docstring **and** a
  trailing `pass`. Dropped the `pass`.
- `core/agent/loop_host.py` — nine `LoopHost` methods had a bare `pass` and no
  docstring. Replaced with one-line docstrings naming each seam.
- `core/agent_harness/ports.py` — removed a dead `if TYPE_CHECKING: pass` block.

`pass` never trips CodeQL, so these were invisible to CI. Worth knowing: mypy
exempts Protocol bodies from "missing return", so `pass` under a
`-> list[RuntimeToolT]` signature passes `make typecheck` too. Neither tool
enforces this, which is why it is now written down.

**`py/unused-global-variable` (#1975)** — `TurnCallbackFactory` in
`gateway/runtime/concurrency.py` was defined, referenced nowhere, and not in
`__all__`. Removed, along with the now-unused `Callable` import.


YauhenBichel:
@greptile review

PR #4515, opened 2026-07-29
perf: O(1) session appends, single-pass masking, and prompt caching with a cache-stable prefix

YauhenBichel:
Hot-path performance pass over the agent runtime, driven by a fresh algorithms/data-structures audit and grounded in Anthropic's and OpenAI's published engineering guidance on prompt caching and token accounting. Every behavioral refactor is pinned by characterization tests written against the pre-change code.

**Session persistence: appends are O(1).** Every auto-parented JSONL append used to re-read and JSON-parse the entire session file to find the last record's id — cumulative Θ(E²) per session. The storage now caches the conversation tip in memory per session, and a cold resume scans the file backwards, parsing only the tail. Pinned by `tests/interactive_shell/sessions/test_jsonl_storage_leaf.py`: five characterization tests on the parent-chain semantics (trace spans stay out of the chain, appends after a `leaf` marker continue from the marked tip, fresh instances resume correctly), plus cost pins — 50 appends do at most one file read, and a cold resume of a 200-entry file parses fewer than 10 records.

**Prompt caching with a cache-stable prefix.** The agent previously used no prompt caching, and its prompt assembly would have defeated it: live runtime facts (`now_iso`, uptime, disk, memory) were rendered into the system prompt every turn — one changed byte at the front of the prefix invalidates everything after it. Now:

- runtime facts are split static vs live; the live block renders immediately  before the user message, so the tools → system prefix is byte-stable
- the Anthropic client sets `cache_control: ephemeral` on the last tool, the  system block, and the newest message's last content block (3 of the 4  allowed breakpoints), so ReAct iterations reuse both the static prefix and  the growing conversation history
- cache hit rate is observable: `LLMResponse` carries  `cache_read_tokens`/`cache_creation_tokens` (`None` distinct from a  measured 0), extracted from Anthropic's flat and OpenAI's nested usage  shapes, with one greppable `[llm-usage]` debug line per call
- tool-schema serialization is pinned byte-deterministic by test, since a  varying tools byte-stream would silently invalidate the cache

Field grounding: moving dynamic content out of the system prompt is the change ProjectDiscovery reported taking cache hit rate from 7% to 84% and total LLM cost down 59%; Anthropic prices cache reads at ~0.1× input.

**Context budget: incremental token ledger.** Per-message token counts are computed once and summed by slice instead of re-estimated per candidate exchange.

**Masking: single-pass everywhere.** Mask application is one forward pass + join (was reverse-order string splicing, quadratic); unmask is a single regex scan with dict lookup (was longest-first `str.replace`, which could also rewrite text inside already-restored originals — the new scan fixes that latent bug and the docstring carries the correctness argument). Redaction is computed once per tool call and reused across tracker, evidence, and events (was 3–4 passes).

**Bounded investigation tool cache.** The per-investigation result cache is now an LRU bounded by entry count and approximate payload size; lookup stays O(1). Limits live in `config/constants/investigation.py`.

**Slack stream sink.** Per-chunk string `+=` became a part list joined at flush.

**Smaller items:** Discord's cancelled-task reap moved into a `_reap_cancelled_task` helper that binds the await result (CodeQL `py/ineffectual-statement` does not model `await`); the AGENTS.md footgun entry was updated to match. Two gateway tests that had landed under `tests/gateway/` moved to the co-located `gateway/tests/` tree, where gateway-scoped CI actually runs them.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4544, opened 2026-07-30
perf(runtime_metadata): cache session-static probes and PATH tool walk

iagogfe:
Fixes #

`build_runtime_metadata()` was failing its own `<5ms` regression guard on WSL (~15ms median). The dominant cost was `installed_tools()`: six separate `shutil.which` walks that re-scanned slow `/mnt/c/...` PATH entries on every call, plus repeated importlib/gitdir work for version and build markers.

This change:

- Probes all tools in **one left-to-right PATH walk** (same first-hit / executable / `PATHEXT` semantics as `shutil.which`) and caches the result while `PATH` is unchanged
- Memoizes the other session-static facts (package version, build marker, tz, hostname, kubeconfig, python version, `runtime_env`, scratchpad) with keys that invalidate when env vars or test-rebound paths change
- Adds a composite-key cache for the full metadata dict when nothing relevant changed

Callers that need live facts still go through `capture_runtime_facts()` (time, uptime, disk, memory) — those are not cached.

```
build_runtime_metadata median (warm, n=200): 0.0038 ms

============================== 7 passed in 0.81s ===============================
exit 0
```

Baseline (same machine, pre-change): ~14.9 ms warm median. After: ~0.004 ms warm median. Cold first call still pays the PATH walk once per process (~15–25 ms on WSL when `docker` lives under `/mnt/c`).

All 38 tests in `tests/config/test_runtime_metadata*.py` pass, including the perf suite that was red before.

---

Problem: session metadata bootstrap and any repeated `build_runtime_metadata` call paid full PATH/tool discovery cost, and the existing perf test pinned a 5ms warm median that was failing badly on WSL.

Alternatives considered:
1. Raise the threshold for WSL — hides a real cost, does not help repeated calls
2. Drop slow tools (e.g. `docker`) from the probe list — loses facts the agent uses
3. Cache only at the session layer — still leaves the function expensive if called more than once

Chose (3) layered with algorithmic fix: single multi-tool PATH walk + process caches keyed so tests that monkeypatch `KUBECONFIG`, `_HOSTNAME_FILE`, or `_LOCALTIME_LINK` still invalidate correctly. Full-dict composite cache is a final layer for the common steady-state case.

Key pieces:
- `_probe_installed_tools` / `installed_tools` in `probes.py` — multi-tool which + PATH key
- path/env-keyed caches for tz, hostname, kubeconfig
- process caches in `version.py` and `build_info.py`
- composite full-dict cache in `assembly.py`

---

---

### DIFF
--- a/config/runtime_metadata/assembly.py
+++ b/config/runtime_metadata/assembly.py
@@ -32,6 +32,11 @@ from config.version import get_opensre_version
 _PROCESS_START_MONOTONIC = _time.monotonic()
 
 
+_scratchpad_dir_cache: str | None = None
+_runtime_env_cache: tuple[str, str] | None = None
+_full_metadata_cache: tuple[tuple[Any, ...], dict[str, Any]] | None = None
+
+
 def build_runtime_metadata() -> dict[str, Any]:
     """Session-lifetime read-only runtime facts.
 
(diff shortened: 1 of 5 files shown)


iagogfe:
Addressed both Greptile findings on `e8cee746`:

1. **P1 PATHEXT** — when `PATHEXT` is unset/empty on Windows, fall back to the same default as `shutil.which` (`.COM;.EXE;.BAT;…`). Covered by `test_probe_installed_tools_uses_windows_pathext_default`.
2. **P2 nested tools share** — full-metadata cache now stores/returns via `_copy_metadata`, which copies the nested `tools` dict. Covered by `test_build_runtime_metadata_nested_tools_copy_is_isolated`.

@greptile review

Devesh36:
Can you link the issue for this pr

PR #4569, opened 2026-07-30
Report bytes transferred alongside file counts in SyncReport

### LINKED ISSUE #4551: [Good first issue] Report bytes moved, not just file counts
### Complexity

### Context
`SyncReport` counts files, so completion says "2 uploaded" with no size.

### Done when
- `SyncReport` carries bytes transferred (up + down).
- Both surfaces show a human-readable total.
- Existing count assertions still pass (extend them; don't drop them).
- Pairs naturally with issues #2 (`--dry-run`) and #3 (progress reporting) above.

---

Zburgers:
Fixes #4551

**Problem:** `SyncReport` only counted files (e.g. "2 uploaded") with no indication of how much data was transferred. Users had no visibility into the volume of bytes moved during a remote sync.

**Changes:**

- **`platform/filestorage/engine.py`** — Added `uploaded_bytes` and `downloaded_bytes` fields to the `SyncReport` dataclass, plus a `total_bytes` computed property. `push()` now accumulates `result.uploaded_bytes += len(data)` per uploaded file. `pull()` captures the return of `store.get_object()` to track `downloaded_bytes` before the atomic write.

- **`platform/filestorage/messages.py`** — Added `_human_size()` to format byte counts with IEC labels (`B`/`KiB`/`MiB`/`GiB`) for the existing 1024-based scaling. Updated `format_report_lines()` to show per-direction and total transfer sizes inline:  
  `Sync complete — 2 downloaded (1.2 KiB), 3 uploaded (4.4 KiB), 1 already current (5.6 KiB total).`  
  When no bytes were transferred, size suffixes are omitted so existing zero-byte output remains unchanged.

- **`platform/filestorage/operations.py`** — Updated `_owned_report()` to preserve the new byte fields for concurrency-safe snapshots.

- **`tests/filestorage/test_service_and_registry.py`** — Added coverage for nonzero transfer sizes, zero-byte behavior, one-direction transfers, byte-sized transfers, and MiB formatting.

**Design decisions:**

- Byte fields default to `0`, so existing `SyncReport` construction remains backward-compatible.
- CLI, `/remote-sync`, and gateway surfaces all share `format_report_lines()`; no duplicated surface logic was added.
- Binary scaling uses the repository-consistent IEC unit labels.

**Verification:**

- `ruff check`: clean
- `ruff format`: clean
- `mypy`: clean
- focused filestorage tests: pass
- Greptile: 5/5 with no actionable defects

The previous repository-wide `tools-runtime` failure was unrelated to this PR: the workflow selected Anthropic with an empty `ANTHROPIC_API_KEY`, causing three LLM credential tests to fail while 4,736 tests passed.

### DIFF
--- a/platform/filestorage/engine.py
+++ b/platform/filestorage/engine.py
@@ -40,10 +40,16 @@ class SyncReport:
     #: resolves these by pulling first; a push-only run cannot.
     kept_remote: list[str] = field(default_factory=list)
     skipped: int = 0
+    uploaded_bytes: int = 0
+    downloaded_bytes: int = 0
 
     @property
     def changed(self) -> int:
         return len(self.uploaded) + len(self.downloaded)
+
+    @property
+    def total_bytes(self) -> int:
+        return self.uploaded_bytes + self.downloaded_bytes
 
 
 def content_tag(data: bytes) -> str:
@@ -144,6 +150,7 @@ def push(
                 continue
         store.put_object(key, data)
         result.uploaded.append(key)
+        result.uploaded_bytes += len(data)
     return result
 
 
@@ -166,7 +173,9 @@ def pull(
         if not _should_download(obj, target):
             result.skipped += 1
             continue
-        _write_atomically(target, store.get_object(obj.key))
+        data = store.get_object(obj.key)
+        result.downloaded_bytes += len(data)
+        _write_atomically(target, data)
         result.downloaded.append(obj.key)
     return result
 
--- a/platform/filestorage/messages.py
+++ b/platform/filestorage/messages.py
@@ -19,6 +19,18 @@ from config.constants.filestorage import (
 from platform.filestorage.engine import SyncReport
 from platform.filestorage.operations import SyncStatus
 
+
+def _human_size(n: int) -> str:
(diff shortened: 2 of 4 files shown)


Zburgers (commented):
Addressed the remaining review concern in commit `a6c6e60`:

- kept the existing 1024-based scaling and changed the labels to the correct IEC units (`KiB`/`MiB`/`GiB`)
- updated the formatter tests accordingly
- resolved the earlier total-bytes and non-zero formatting review threads

The previous CI failure is unrelated to this filestorage change. The only failed shard was `test (tools-runtime)`, where the runner had `LLM_PROVIDER=anthropic` but an empty `ANTHROPIC_API_KEY`:

```text
RuntimeError: Missing credential for LLM provider 'anthropic'. Set ANTHROPIC_API_KEY or run `opensre auth login anthropic`.
3 failed, 4736 passed, 2 skipped
```

Quality checks, CodeQL, synthetic tests, and the remaining test shards passed. The new commit has triggered a fresh CI run.

cerencamkiran (commented):
In platform/filestorage/messages.py, the issue says to show the total size. Here it is hidden when the total is 0. Could we always show 0 B total and update the test too? Thanks!

PR #4571, opened 2026-07-30
fix: cache the classification prompt prefix, resolve prose root-cause categories, and add installer/quickstart coverage

YauhenBichel:
Follow-up to the prompt-caching wave now in `main`. Three independent threads: closing the remaining prompt-cache gap, fixing a category-parsing defect, and adding installer + quickstart regression coverage.

**Prompt caching now covers every Anthropic path.** The previous change added `cache_control` to the agent client only. Console telemetry then showed the classification model (`claude-sonnet-4-6`) at a flat **0.0% cache-read ratio across 12.7M input tokens** — roughly a fifth of all input paying full price — because classification and other single-shot calls go through `LLMClient` / `BedrockLLMClient`, whose request kwargs carried no marker at all.

- the breakpoint helpers move to a new leaf module `core/llm/transports/sdk/anthropic_cache.py`, so both clients share them with no  import cycle (CodeQL counts `TYPE_CHECKING` and function-local imports as part of  a cycle, so the symbols had to move rather than be imported sideways)
- `agent_clients.py` imports them instead of defining them — 288 existing LLM tests  stayed green through the extraction, which is what shows the move was  behavior-neutral
- `LLMClient._build_request_kwargs` and `BedrockLLMClient._invoke_anthropic` wrap  `system` in a cache-marked text block

Only `system` is marked. These are single-shot calls: a large stable prefix (the 112-entry root-cause taxonomy among others) plus a small varying question. Marking the varying tail would write a cache entry nothing ever reads. Marking a prefix below the model's minimum cacheable length is safe — the request succeeds and simply is not cached — so no size check is needed.

**Root-cause categories written in prose are no longer discarded.**
`ROOT_CAUSE_CATEGORY: resource exhaustion` resolved to `unknown` because the parser matched only underscore-joined tokens. The harm is bigger than a test: the narrative report confidently names the cause while the stored category reads `unknown`, so reports, dashboards, and incident memory all see nothing.

| Model writes | Before | After |
|---|---|---|
| `resource exhaustion` | `unknown` | `resource_exhaustion` |
| `pod oomkilled` | `unknown` | `pod_oomkilled` |
| `resource-exhaustion` | `unknown` | `resource_exhaustion` |
| `dual resource exhaustion` | `unknown` | `dual_resource_exhaustion` (longest wins) |
| `gremlins in the rack` | `unknown` | `unknown` (no guessing) |

Reproducible with no LLM. Fixed via a `_category_candidates` generator yielding normalized and adjacent-word-joined forms, longest first, accepting **only** exact taxonomy matches — no fuzzy inference.

**Synthetic suites can explain their own failures.** A scored failure carries the expected category, what the agent answered, missing keywords, failing gates, and the trajectory divergence — but pytest keeps only the first line of the assertion, so the reason had to be re-derived by hand. `tests/synthetic/score_artifacts.py` writes one JSONL record per scored attempt, gated behind `OPENSRE_SYNTHETIC_ARTIFACT_DIR` so normal and CI runs pay nothing. Records are appended incrementally, so a run killed part-way still yields usable data.

Also fixed: `equivalent_root_cause_categories` is validated by the shared synthetic schema and used by 12 RDS fixtures, but the EKS loader never parsed it and the EKS scorer never read it — it hardcoded a single DNS special case instead. A fixture could declare the field, pass validation, and still fail scoring.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4575, opened 2026-07-30
fix: keep marker-rejecting models usable by degrading prompt caching to uncached

YauhenBichel:
Prompt caching landed in `main` as an unconditional request shape: every Anthropic-shaped call marks its system prefix with `cache_control`. Models that do not support the markers reject the request with HTTP 400 — older Bedrock
Claude model IDs are the realistic case — which made caching a hard dependency rather than an optimisation. A user on such a model would see every call fail.

Caching is now best-effort:

- `anthropic_cache.py` gains `is_cache_unsupported_error()` and `strip_cache_markers()`. The stripper removes markers from exactly the three  spots this module writes them — system blocks, top-level tool schemas, and  message content blocks — and deliberately does not recurse into tool  `input_schema` bodies, where `cache_control` could be a legitimate
  user-defined property name.
- `strip_cache_markers` exactly inverts `cached_system`: a lone text block  collapses back to the plain string a non-caching client would have sent. This
  matters because a model old enough to reject `cache_control` may not accept
  block-form `system` either, so the retry has to look like the request we would
  have made if caching had never existed.
- Every Anthropic-shaped path degrades on a marker rejection: `LLMClient.invoke`,
  `LLMClient.invoke_stream`, `BedrockLLMClient._invoke_anthropic`, and
  `BedrockAgentClient` via `AnthropicAgentClient`. Each retries once without
  markers, logs a warning naming the model, and disables marking for that
  client's lifetime so the probe costs one request, not one per call.
- Marking is gated on an instance flag with a `True` class default, so a fresh
  client starts optimistic and only the rejecting one stays uncached.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -258,5 +258,17 @@ Steps:
   when appending to an existing file: **use the import style the file already
   established** (an existing `import core.context_budget as budget` means new
   code calls `budget.name`, not `from core.context_budget import name`).
+- Shared client state under concurrent turns: LLM clients are cached per role
+  (`get_llm`) and the gateway runs turns in parallel, so **one client instance
+  serves several in-flight requests**. An instance flag mutated inside an error
+  handler is then read by requests that were already built under the old value.
+  Branch on **what the current request carried**, not on the flag's present
+  value — capture the fact locally where the request is built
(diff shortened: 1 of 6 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4579, opened 2026-07-30
fix(bench): report.json records whether the run completed

### LINKED ISSUE #4567: [BUG] report.json does not record whether the run completed
### Summary

A halted run publishes a `report.json` that is key-identical to a complete run's.
Nothing in the file says which one you are holding.

`RunOutcome` carries `aborted` and `abort_reason`, and the CLI uses them to refuse a zero exit code on a halted run. `_report_to_dict` carries neither.

So the exit code is the only thing that tells the two apart, and the exit code is not what gets published. `entrypoint.sh` turns off `set -e` specifically so a non-zero bench exit does not skip the S3 sync, and the README lists the uploaded set as `report.json`, `provenance.json` and the per-case files. The exit code is not in that set.

Anchors:
- `RunOutcome` fields: `tests/benchmarks/_framework/runner.py:98-99`
- `_report_to_dict`: `tests/benchmarks/_framework/runner.py:827`
- uploaded set: `tests/benchmarks/cloudopsbench/README.md:109`
- upload survives a non-zero exit: `tests/benchmarks/cloudopsbench/infra/entrypoint.sh:76-78`

**Expected:** a published report says whether the run it describes completed.

**Actual:** it does not;

Same two-arm config, run twice through the production `run()`:

```
complete run: aborted=False, 36 cells, report_validation PASSED
halted run: aborted=True, 18 cells, report_validation PASSED
abort_reason: LLM dispatch failed: 'gpt-5' requires env var OPENAI_API_KEY ...

report.json key sets identical: True
report.json (halted) contains abort/partial/halted: False
RunOutcome.aborted in memory: True
written to report.json: False
```

The pre-registration commits to the distinction the artifact does not carry: "Numbers from a partial run are NEVER promoted to a baseline; only complete runs are baselines."

Anchors:
- `stopping_rules.partial`: `tests/benchmarks/cloudopsbench/configs/preregistrations/cloudopsbench_v1.yml:243-246`

### Steps to reproduce

No API key is needed for the halted arm. An absent key is what triggers it.

1. Configure a two-arm run where the second arm's provider key is absent.
2. Run it through the production `run()` path.
3. Compare `report.json` against a complete run's. The key sets are identical.

`RunOutcome.aborted` is `True` in memory throughout. `_report_to_dict` has 13 keys and none of
them is a completion status.

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Logs and error output

```
abort_reason (in memory): LLM dispatch failed: 'gpt-5' requires env var OPENAI_API_KEY ...
report.json: no abort/partial/incomplete/halted key
report.md: 0 lines mentioning abort or halt
cost line the reader sees: $0.00 of $10.00 budget
```

### Additional context

Two more mechanisms follow from the same cause. Both are context, not part of the fix.

On a cost-ceiling halt part-way through the second arm, `cases/` publishes 24 per-case
artifacts while `per_stratum` aggregates 18. Six published artifacts are counted by no
stratum.

`report.md` reads `cases/` off disk while `report.json` uses the in-memory cell list. A
dropped arm still renders, printing `gpt-5 | opensre+llm | n=2` above that model's `n=452`
paper row.

The principle is already in the tree: "a halted run that exits 0 is silently lost." A PR
follows that carries the same fact into the artifact. Two keyword-only parameters with
defaults, additive, no existing consumer changes.

Anchors:
- exit-code principle: `tests/benchmarks/_framework/cli.py:194-198`

Ourbando:
Fixes #4567

**The bug:** a halted run's `report.json` is byte-identical to a complete run's. No signal in the uploaded artifact tells you which one you have.

The exit code can't fix this alone:

- `RunOutcome` already carries `aborted` and `abort_reason`, and the CLI rejects zero exit on a halt
- the entrypoint runs with `set -e` off so failed benches still upload (`tests/benchmarks/cloudopsbench/infra/entrypoint.sh:76-78`)
- the upload is just artifacts: `report.json`, `provenance.json`, cases (`tests/benchmarks/cloudopsbench/README.md:109`), and no exit code travels with them

So today, whoever reads the numbers has no way to know.

- `runner.py`: `_report_to_dict` gains both as keyword-only parameters with defaults, passed from locals already in scope.
- `test_runner_budget.py`: two tests, one per field. They run the real `run()` via `run_without_integrity()` (which skips the pre-flight gate that `run()` adds). Both funnel into `_run_inner`, and the `report.json` write sits before the `dev_mode` branch, so it's the production write under test.
- `test_reporting.py`: the fixture docstring says it mirrors `_report_to_dict`, so it gains the same keys.

Backward compatible: existing callers don't change, `report.json` readers pull named keys only (never the whole dict), so the new fields don't appear in rendered output, and no schema validates against these keys.

opensre_demo_4567

Two runs of the same two-arm grid through the production `run()`, using the repo's own tiny-adapter harness. `complete/` had every provider key present and ran all 36 cells. `halted/` had the second provider's key absent and stopped at 18. `report_validation` passed for both, and their `report.json` key sets are identical.

Put the same question to both runs. **Before this PR:**

```console
$ jq '.aborted' complete/report.json halted/report.json
null
null
```

**After this PR**, same command, same two runs:

```console
$ jq '.aborted' complete/report.json halted/report.json
false
true

$ jq -r '.abort_reason' halted/report.json
LLM dispatch failed: 'gpt-5' requires env var OPENAI_API_KEY to be set. Run with `set -a && source .env && set +a` or export the key first.
```

The keys are record-only. The rendered `report.md` does not surface them.

Why the exit code cannot carry this on its own:

```mermaid
flowchart LR
    B["bench run<br>(knows it halted)"] -->|"exit code"| C["container status<br>(set +e; not a file)"]
    B -->|"run directory"| S["S3: report.json<br>provenance.json, cases/"]
    C -.->|"not uploaded"| S
    S --> R["whoever reads<br>the numbers"]
```

The tests, against unpatched and patched code:

```
vs unpatched : 2 failed, KeyError: 'aborted'   (4 existing pass)
with the fix : 6 passed
tests/benchmarks -n auto : 496 passed, 18 skipped   (baseline 494)
```

Reproduce on this branch: `pytest tests/benchmarks/_framework/tests/test_runner_budget.py -k report_json -q` drives the same sequence (2 passed, 4 deselected in 1.92s).

---

**Why.** The pre-registration's `stopping_rules.partial` bars promoting partial-run numbers to a baseline. Enforcing that needs the artifact to say what kind of run it describes, and today only the process exit code knows.

**Alternatives.**

- Fields on `BenchmarkReport`: no. That is the integrity contract `report_validation` checks; the serialiser is where run-level facts (`cost`, `opensre_sha`) already live.
- Fixing the upstream cause, where an abort drops the in-flight batch: not in this PR. It changes the numbers a halted run reports, and that call is yours. Recording the status changes none.
- Inferring completion from cell counts at read time: needs an expected count the artifact does not carry, and guesses where the runner has the fact.

**How.** Two keyword-only parameters with defaults go on `_report_to_dict`, two keys go out. The write site (`tests/benchmarks/_framework/runner.py:308`) passes them from the abort handlers, and the names follow `RunOutcome`'s exactly, so there is one vocabulary for the concept. `host` at `tests/benchmarks/_framework/runner.py:844` is the precedent for a record-only field.

The tests force the abort by omitting a provider key, then verify both fields reach `report.json`:

- complete run: `aborted: false`, reason `null`
- halted run: `aborted: true`, reason is the exception message

**On the reason string.** `abort_reason` is not provider text. It comes from four exceptions defined in this repo: `CostBudgetExceeded`, `UnknownLLM`, `ModelVersionMismatch`, `MissingAPIKey`. Each has a message authored here, and `MissingAPIKey` embeds the env var name, never its value. Both handlers already print this string today, so the report isn't leaking anything new.

If you want it out of the uploaded set, drop it and I'll keep just the boolean. That's the field the pre-registration rule actually needs.

---

### DIFF
--- a/tests/benchmarks/_framework/runner.py
+++ b/tests/benchmarks/_framework/runner.py
@@ -306,7 +306,12 @@ class BenchmarkRunner:
 
         # Persist a JSON sidecar to output_dir/report.json regardless of validation
         (output_dir / "report.json").write_text(
-            json.dumps(_report_to_dict(report, self.cost), ensure_ascii=False, indent=2) + "\n",
+            json.dumps(
+                _report_to_dict(report, self.cost, aborted=aborted, abort_reason=abort_reason),
+                ensure_ascii=False,
+                indent=2,
+            )
+            + "\n",
             encoding="utf-8",
         )
 
@@ -824,10 +829,26 @@ def _cell_to_dict(case: BenchmarkCase, run: RunResult, score: CaseScore) -> dict
     }
 
 
-def _report_to_dict(report: BenchmarkReport, cost: CostTracker) -> dict[str, Any]:
-    """Serializable shape for report.json."""
+def _report_to_dict(
+    report: BenchmarkReport,
+    cost: CostTracker,
+    *,
+    aborted: bool = False,
+    abort_reason: str | None = None,
+) -> dict[str, Any]:
+    """Serializable shape for report.json.
+
+    ``aborted`` / ``abort_reason`` mirror the same fields on ``RunOutcome``.
+    The CLI already refuses to let a halted run exit 0, but that exit code
+    is not part of the run directory the bench container uploads to S3, so
+    the artifact carries the fact too. The pre-registration's
+    ``stopping_rules.partial`` clause depends on telling a partial run from
+    a complete one after the fact.
+    """
     return {
         "run_id": report.run_id,
+        "aborted": aborted,
+        "abort_reason": abort_reason,
         "config_hash": report.config_hash,
         "started_at": report.started_at,
         "ended_at": report.ended_at,
--- a/tests/benchmarks/_framework/tests/test_reporting.py
+++ b/tests/benchmarks/_framework/tests/test_reporting.py
@@ -24,6 +24,8 @@ def _write_report_json(run_dir: Path) -> dict:
     cases_dir.mkdir(parents=True, exist_ok=True)
     report = {
         "run_id": "dev-2026-01-01T00-00-00Z_cloudopsbench",
+        "aborted": False,
+        "abort_reason": None,
         "config_hash": "abc123",
         "started_at": "2026-01-01T00:00:00+00:00",
         "ended_at": "2026-01-01T00:05:00+00:00",
--- a/tests/benchmarks/_framework/tests/test_runner_budget.py
+++ b/tests/benchmarks/_framework/tests/test_runner_budget.py
@@ -16,6 +16,7 @@ See Greptile P1 review on the bench-image PR for the original report.
 
 from __future__ import annotations
 
+import json
 from collections.abc import Iterator
 from pathlib import Path
 from typing import Any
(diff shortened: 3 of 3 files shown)


muddlebee:
could we pls discuss this issue first before PRs?

Ourbando:
hi, I wasn't sure and went ahead anyway, my bad. I'll reach out when I'm unsure next time.

Eager to talk about why I picked this one, I've been looking at the benchmark suite ahead of the August research. There's a couple of things I'd like your read on before I touch any code. no rush 🤙

PR #4582, opened 2026-07-30
fix(remote-sync): validate stored settings on read

### LINKED ISSUE #4559: [Intermediate] Validate the remote_sync settings section on read
### Complexity
Intermediate — needs more product judgment or multi-file design. Still largely testable with fakes; live cloud optional unless noted.

### Context
`config/local_settings.py` returns whatever `yaml.safe_load` produces. A typo like `bucket: [my-bucket]` becomes a confusing late S3 failure.

### Done when
- The `remote_sync` section is validated on read.
- A bad value fails early naming the key.
- Still raises as `RemoteSyncError` so surfaces catch it.
- Env-precedence behavior from issue #5 (soft-fail optional settings) above is preserved.

---

santhiprakash:
Fixes #4559

Validate stored `remote_sync` values before they are consumed so malformed YAML fails early with the offending key name. The change preserves env precedence and only touches the remote-sync config path plus a focused regression test.

Verified with a direct Python check in this checkout:
- env-only config still loads when `config.yml` is malformed
- a stored `remote_sync.bucket` list now raises `RemoteSyncConfigError: invalid remote_sync.bucket: expected a string, got list`

---

I validated the stored `remote_sync` section at the config boundary instead of letting malformed YAML flow into later S3/provider code. The validator checks the scalar fields that this feature reads, raises `RemoteSyncConfigError` with the specific offending key, and keeps the environment-variable precedence rule intact. I chose the narrow config-layer fix because it keeps the surfaces and the engine unchanged while making the failure mode deterministic and easier to understand.

---

### DIFF
--- a/platform/filestorage/config.py
+++ b/platform/filestorage/config.py
@@ -26,6 +26,7 @@ from config.constants.filestorage import (
 from platform.filestorage.errors import RemoteSyncConfigError
 
 _TRUTHY = frozenset({"1", "true", "yes", "on"})
+_REMOTE_SYNC_STRING_KEYS = frozenset({"bucket", "provider", "prefix", "region", "profile"})
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


santhiprakash:
Updated to preserve env precedence for remote_sync: env-supplied fields now short-circuit before storage is read, and malformed stored values only matter when that field actually falls back to disk. Added a regression for the mixed env/stored case.

santhiprakash:
@greptile review

cerencamkiran:
I still see the old whole section validation in the current diff. Could you please push the update? Thanks!

cerencamkiran:
For your first prs, please start with good first issues and wait until one pr is merged before opening another one.

cerencamkiran:
It looks like the stored values are still validated even when they are overridden by environment variables, so the per field environment precedence may not work correctly.

PR #4590, opened 2026-07-30
ci: add a single CI Gate check and remove two superseded infrastructure files

YauhenBichel:
**A single required check for branch protection.** `quality` and `test` are matrix jobs, so protecting the branch today means listing every matrix leg by name and updating that list whenever the matrix changes. `ci-gate` aggregates them into one check that branch protection can point at instead.

The job runs under `if: always()`, which is what lets it execute when an upstream job failed or was skipped. Without it the gate would itself be skipped — and a skipped required check reads as **passing** in branch protection, which is the failure mode this is meant to prevent.

`coverage-report` is treated differently from the other two, deliberately:

```yaml
if: always() && github.event_name == 'push' && github.ref == 'refs/heads/main'


YauhenBichel:
@greptile review

PR #4594, opened 2026-07-30
refactor: strategic refactoring milestones — shared runtime, goal-directed loop, capability warnings

YauhenBichel:
Milestones from *Strategic Refactoring: Shared Runtime, Autonomous Agents & Parallel Execution* (PMF gameplan priorities 2 and 3). Status was checked against `main` before starting, which showed two milestones already complete — those are untouched here.

**A readable entry point (M3).** `main.py` at the repo root, delegating to the same `main()` the `opensre` console script uses. It returns the CLI's exit status: previously `python main.py <bad command>` exited **0** while `opensre <bad command>` exited **2**, so the documented entry point silently passed failures to CI steps and shell `&&` chains.

**Goal-directed stop decisions (M4 / M5 #1D).** `core/agent/goals.py` lets the ReAct loop refuse to conclude while a structured goal is unmet, so the agent pursues the objective instead of stopping at the first plausible answer. A hard iteration ceiling always wins, so the loop cannot spin forever — that ceiling is tested explicitly, not just the happy path.

**Capability warnings at boot (M5 #1C).** `platform/sandbox/capabilities.py` probes what the environment actually *grants* — Python execution, shell, file reading, network — surfaced as an `agent_capabilities` check in `opensre doctor`. This is a different question from the existing PATH probe: a binary can be present and still unusable, which is the bug class the milestone describes ("the agent is too limited to execute normal actions").

Two constraints, both pinned by tests: the probe **never raises** (a diagnostic that kills startup is worse than no diagnostic), and the network probe **opens no socket** — it inspects the local policy that would refuse the call, so startup never depends on an external host or hangs on a firewalled network.

**Separation of concerns in the gateway turn path (M1 / M2).** `GatewayTurnHandler` was building agents, building sinks, and dispatching. Extracted `SessionAgentPool` (`gateway/runtime/session_agents.py`) and `LiveOutputSink` (`gateway/runtime/live_sink.py`), so one agent is constructed per logical session and reused across turns — Goal B of M2. `LiveOutputSink` now declares the `OutputSink`/`GatewaySink` methods explicitly instead of forwarding through `__getattr__` with a `type: ignore`, which restores a real interface for callers and type checkers.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4607, opened 2026-07-31
refactor: make the documented entry path the easy one — AgentHarness.start(), grouped commands, honest landing page

YauhenBichel:
`main.py` landed last week as the file newcomers are meant to read first. Reading it as a newcomer showed the problem: it documented the wiring rather than an API worth copying. Driving the agent from Python meant assembling a config, calling `startup()`, constructing a `Console`, a logger and an output sink, calling the factory with four keyword arguments, then attaching the result. Every one of those was defaultable.

**`AgentHarness.start()`** does them, so the documented path is two lines:

```python
from core.agent_harness.harness import AgentHarness

harness = AgentHarness.start()
result = harness.dispatch_message("why is checkout-api slow?")

```

This is the shape the refactoring brief asked for (agent = agent_harness(...)
then agent.dispatch_message(...)). Surfaces that need their own ports — a live
gateway sink, a REPL console — still build the agent with
build_default_headless_agent and call attach_agent; start() is the default,
not a replacement. AgentHarness.agent is exposed so a caller can reach the
attached agent without touching a private field.

The GitHub PR sweep and Sentry morning digest now use the same factory instead of
each hand-rolling a Console(file=StringIO()) and sink, and
core.agent_harness re-exports build_default_headless_agent and
BufferOutputSink so callers have one import path.

Landing page: a formatting bug on the first screen a user sees. The
quick-start table padded to a fixed 44-column gutter, but the longest command is
50 characters, so f"{label:<44}" emitted it unpadded and the description ran
straight into it:

before:  opensre "investigate high latency in checkout-api"Start the interactive
after:   opensre "investigate high latency in checkout-api"  Start the interactive
width is now a preferred minimum rather than a cap, so the column always fits
its longest label. That also immunises the command table, which is one long
command name away from the same bug.

Release stage said two different things. The README badge and callout say
"Public Alpha"; the CLI banner said "Public Beta". Both are user-facing and a new
user sees them within a minute. config/constants/product.py now holds the
single statement, the banner reads it, and the test asserts against the constant
rather than a literal so the two cannot drift apart again.

Commands are grouped instead of one flat alphabetical list. onboard — the
command a new user must run first — sat between misses and remote-sync,
weighted identically to debug, hermes and tests. Now:
Getting started / Everyday / More commands. Anything not explicitly listed
falls through to More commands, so a newly added command is never silently
hidden — the existing test that loops over every registered command still proves
it.

Quick start trimmed from 13 rows to 5 genuine first-run actions, ending with
opensre --help for everything else. A landing page listing every command reads
as "here is everything" rather than "start here".

$ uv run python -c "from core.agent_harness.harness import AgentHarness; \
                    h = AgentHarness.start(); print(type(h.agent).__name__)"
HeadlessAgent

$ uv run opensre | head -1
🚧 OpenSRE is in Public Alpha — core workflows are usable, and APIs and integrations may still change.

$ uv run opensre --help | grep -E "Getting started|Everyday|More commands"
  Getting started:
  Everyday:
  More commands:

$ make lint format-check typecheck check-imports
All checks passed! · 3052 files already formatted
Success: no issues found in 1539 source files
All 3 import checks passed.

$ uv run pytest tests/core/agent_harness tests/cli tests/interactive_shell/ui -q
1316 passed, 4 skipped
Code Understanding and AI Usage
Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?

 No, I wrote all the code myself
 Yes, I used AI assistance (continue below)
If you used AI assistance:

 I have reviewed every single line of the AI-generated code
 I can explain the purpose and logic of each function/component I added
 I have tested edge cases and understand how the code handles them
 I have modified the AI output to follow this project's coding standards and conventions

The starting point was walking the chain a new user walks — README, install,
opensre --help, the no-argument landing page, main.py — and reading what it
actually prints rather than what it intends to. Two of the four findings were
only visible that way: the column collision does not show up in any test, and the
Alpha/Beta contradiction only appears if you read both surfaces in one sitting.

start() was written test-first: three tests describing the call I wanted before
it existed, including one that dispatches a message end to end. That ordering
matters here because the point of the change is the shape of the API, not its
internals.

I also executed the docstring example verbatim rather than eyeballing it, which
caught that an earlier draft referenced startup.output — an attribute
HarnessStartupResult does not have. That example would have failed for the
first person who copied it out of the file the project points newcomers at.

For the command grouping, the risk is hiding a command by forgetting to list it.
The partition therefore routes everything unclaimed into More commands, and the
pre-existing test that asserts every registered command appears in the output was
kept as the guard rather than replaced.

Three assertions pinned the old Commands: header (two tests plus a smoke
scenario) and were updated to the new headings — the header text changed, the
behaviour they were protecting did not.

 I have added proper PR title and linked to the issue
 I have performed a self-review of my code
 I can explain the purpose of every function, class, and logic block I added
 I understand why my changes work and have tested them thoroughly
 I have considered potential edge cases and how my code handles them
 If it is a core feature, I have added thorough tests
 My code follows the project's style guidelines and conventions

Two caveats before you open it. **Everything is uncommitted** — 14 modified plus 2 untracked (`config/constants/product.py`, `tests/core/agent_harness/test_harness_start.py`), and the branch is `issue/clean-code` with no upstream yet.

And the **full regression is still running**; the numbers above are the touched suites only. I'll report the full result — I won't call this merge-ready on a partial run, particularly since the command-grouping change touches output that several suites assert on.

### DIFF
--- a/config/constants/__init__.py
+++ b/config/constants/__init__.py
@@ -226,6 +226,7 @@ from config.constants.posthog_mcp import (
     POSTHOG_MCP_PROJECT_ID_ENV,
     POSTHOG_MCP_URL_ENV,
 )
+from config.constants.product import RELEASE_STAGE, RELEASE_STAGE_BANNER
 from config.constants.rds import RDS_DB_INSTANCE_IDENTIFIER_ENV, RDS_REGION_ENV
 from config.constants.redis import (
     REDIS_DATABASE_ENV,
@@ -311,6 +312,8 @@ from config.constants.vercel import VERCEL_API_TOKEN_ENV, VERCEL_TEAM_ID_ENV
 from config.constants.x_mcp import X_MCP_AUTH_TOKEN_ENV, X_MCP_URL_ENV
 
 __all__ = [
+    "RELEASE_STAGE",
+    "RELEASE_STAGE_BANNER",
     "ALERT_TEMPLATE_CHOICES",
     "ALERTMANAGER_BEARER_TOKEN_ENV",
     "ALERTMANAGER_PASSWORD_ENV",
--- a/config/constants/product.py
+++ b/config/constants/product.py
@@ -1,0 +1,21 @@
+"""Product-level statements shown to users in more than one place.
+
+Release maturity appears on the README badge, in the README callout, and in the
+CLI banner. Held here so those cannot drift apart — they did, with the README
+saying "Public Alpha" while the CLI said "Public Beta".
+"""
+
+from __future__ import annotations
+
+from typing import Final
+
+#: Release maturity, as users see it. Keep in step with the README badge.
+RELEASE_STAGE: Final[str] = "Public Alpha"
+
+#: The one-line maturity banner printed before the landing page.
+RELEASE_STAGE_BANNER: Final[str] = (
+    f"🚧 OpenSRE is in {RELEASE_STAGE} — core workflows are usable, "
+    "and APIs and integrations may still change."
+)
+
+__all__ = ["RELEASE_STAGE", "RELEASE_STAGE_BANNER"]
--- a/core/agent_harness/__init__.py
+++ b/core/agent_harness/__init__.py
@@ -21,8 +21,10 @@ if TYPE_CHECKING:
     from core.agent_harness.turns.action_driver import (
         run_action_agent_turn as execute_action_agent_turn,
     )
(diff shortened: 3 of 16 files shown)


YauhenBichel:
@greptile review

PR #4159, opened 2026-07-20
refactor(config): centralize credit + Azure OpenAI env-var constants under config/

YauhenBichel:
#### Describe the changes you have made in this PR -

Removes duplicated / scattered env-var-name constants by giving each a single home under `config/`, and adds smoke coverage for credit metering. No behavior change — same env vars, same values.

**Credit-metering constants → `config/constants/billing.py` (new)**
- `WEBAPP_URL_ENV`, `USAGE_SECRET_ENV`, `ORGANIZATION_ID_ENV`, and   `CREDITS_HTTP_TIMEOUT_SECONDS` moved out of `gateway/billing/credits_client.py into a domain constants module, re-exported via `config/constants/__init__.py` (matching the existing `sentry` / `posthog` pattern). `credits_client.py` and the two gateway tests now import them from `config.constants.billing`.

**Azure OpenAI connection constants → `config/config.py`**
- `AZURE_OPENAI_BASE_URL_ENV` / `AZURE_OPENAI_API_VERSION_ENV` / `AZURE_OPENAI_API_KEY_ENV` were defined in `core/llm/providers/azure_openai.py` **and** read as raw string literals in `config/config.py`. Centralized the names in `config/config.py` (which its own `_llm_settings_env_payload` resolver consumes) and had the provider import them, removing the duplicate literals and the stale `__all__` re-exports.

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [x] I have reviewed every single line of the AI-generated code
- [x] I can explain the purpose and logic of each function/component I added
- [x] I have tested edge cases and understand how the code handles them
- [x] I have modified the AI output to follow this project's coding standards and conventions

---

## Checklist before requesting a review
- [x] I have added proper PR title and linked to the issue
- [x] I have performed a self-review of my code
- [x] **I can explain the purpose of every function, class, and logic block I added**
- [x] I understand why my changes work and have tested them thoroughly
- [x] I have considered potential edge cases and how my code handles them
- [x] If it is a core feature, I have added thorough tests
- [x] My code follows the project's style guidelines and conventions

---

Note: Please check **Allow edits from maintainers** if you would like us to assist in the PR.

PR #4614, opened 2026-07-31
Separate the tenant organization id from the billing organization id, and extract CLI process startup

YauhenBichel:
Two unrelated env vars both named an "organization" and were being confused for each other. This gives them distinct, purpose-revealing names and pins the separation with tests. Also extracts process-level setup out of the CLI
entrypoint.

`ORGANIZATION_ID` and `OPENSRE_ORGANIZATION_ID` look like two spellings of one thing. They are not:

| | `OPENSRE_ORGANIZATION_ID` | `ORGANIZATION_ID` |
| --- | --- | --- |
| Added by | #4159 (credit/env centralization) | #4476 (multi-tenant Fargate MVP) |
| Set by | the product's own config | the control plane's ECS task definition |
| Read for | usage attribution, and the fail-closed mount-ownership check in `config/constants/paths.py` that refuses a turn when a context volume belongs to another org | selecting which tenant's credentials to hydrate, and whose runs the Neon worker polls |

`STARTUP_FARGATE_MULTI_TENANT_MVP.md` states the contract directly: ECS supplies `ORGANIZATION_ID`.

Treating them as interchangeable is not cosmetic. Pointing credential hydration at the billing name makes a deployed silo see the bootstrap secret ARN set but no org id, treat that as partial configuration, and raise at gateway startup; the remote-run worker separately gets an empty id and silently never starts. In the other direction, letting the ownership check fall back to the control plane's value would weaken a fail-closed data-isolation boundary.

The three env names for the credential-hydration family were also inline string literals in a feature module, against the `config/` rule in AGENTS.md.

### DIFF
--- a/config/constants/__init__.py
+++ b/config/constants/__init__.py
@@ -298,6 +298,11 @@ from config.constants.temporal import (
     TEMPORAL_BASE_URL_ENV,
     TEMPORAL_NAMESPACE_ENV,
 )
+from config.constants.tenancy import (
+    CREDENTIALS_API_URL_ENV,
+    CREDENTIALS_BOOTSTRAP_SECRET_ARN_ENV,
+    TENANT_ORGANIZATION_ID_ENV,
+)
 from config.constants.tracer import TRACER_BASE_URL_ENV, TRACER_JWT_TOKEN_ENV
 from config.constants.twilio import (
     TWILIO_ACCOUNT_SID_ENV,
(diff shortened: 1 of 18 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4621, opened 2026-07-31
Hydrate a silo's integrations from its Secrets Manager blob, on an ephemeral store path

YauhenBichel:
Adds the second credential route: a silo can hydrate its integration store from the tenant's IntegrationStore v2 secret instead of calling the web app. Writes the hydrated store to ephemeral container storage, and removes the one name that could bypass that path.

The deployed route is the per-tenant Secrets Manager blob. The task role already reads exactly that ARN, and the control plane rolls the service after every integrations write, so a boot-time read is the whole refresh mechanism — no polling, no inbound path to the silo.

Two properties are load-bearing:

- **Credentials must not land on the persistent mount.** The workspace volume   outlives the task and is shared by every member of the org. Hydrated  credentials belong on container storage that dies with the task.
- **The store path must have one resolver.** `config.constants` exported  `INTEGRATIONS_STORE_PATH`, computed at import time from the host home. It had  **zero consumers**, but it was the obvious-looking name and it ignored both org
  scoping and the new override — anyone reaching for it would write to the shared  mount. Removed rather than left as a trap.

### DIFF
--- a/config/constants/__init__.py
+++ b/config/constants/__init__.py
@@ -189,7 +189,6 @@ from config.constants.opensre_infra_aws import (
 from config.constants.pagerduty import PAGERDUTY_API_KEY_ENV, PAGERDUTY_BASE_URL_ENV
 from config.constants.paths import (
     CONTEXT_ROOT_ENV,
-    INTEGRATIONS_STORE_PATH,
     OPENSRE_HOME_DIR,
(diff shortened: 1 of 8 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4622, opened 2026-07-31
Mirror silo-side integration changes back to the webapp vault

YauhenBichel:
A silo already reads its integrations from its Secrets Manager blob at boot. A change made inside the agent had nowhere to go — it stayed in the local store and was overwritten on the next restart. This closes the loop in the other direction.

integrations/webapp_vault.py gains push_webapp_org_integration() and delete_webapp_org_integration() alongside the existing fetch, sharing a _write_target() helper. Both are best-effort by design: they return False and log on any failure, and skip entirely when no webapp target is configured. The local store is already the working copy, so a network blip must not fail the connect flow an operator is running, and a laptop must not call out at all.

Hooked at the four connect/disconnect funnels — setup_flow.py and github/login.py after upsert_integration, cli.py and the interactive shell's _handle_remove after a successful remove_integration.

Deliberately out of scope:

the wizard configurators (laptop onboarding, four more call sites);
any hook inside store.upsert_*, which would sweep in the Slack/Telegram inbound allowlist — gateway-internal policy that doesn't belong in the org vault;
suppressing the control plane's refresh, so an agent-side connect or remove restarts that silo. Correct, just heavy; the flag can come later if it proves annoying.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4629, opened 2026-07-31
feat: interactive setup for opensre remote-sync (#4554)

### LINKED ISSUE #4554: [Intermediate] opensre remote-sync setup — interactive configuration
### Complexity
Intermediate — multi-file design. Testable with fakes; live cloud optional.

### Context
Users should not have to hand-export env vars to enable remote sync. Settings live under `remote_sync` in `~/.opensre/config.yml` (env still wins for a one-off run). Ambient cloud credentials are never written to that file.

### Done when
- [x] `opensre remote-sync setup` prompts (or accepts flags) for provider, bucket, prefix, region, profile.
- [x] Writes the `remote_sync` section via `local_settings`.
- [x] `/remote-sync setup` works in the shell (flags; interactive prompts on CLI).
- [x] `RemoteSyncSubcommand.SETUP` keeps CLI and slash aligned.
- [ ] Optional follow-up: verify-by-listing the store after setup.

### Onboarding
This is the remote-sync setup path (not `opensre onboard`). New providers should work with:
`opensre remote-sync setup --provider <name> --bucket <store>`

---

arnavpopley:
Fixes #4554

Adds an `opensre remote-sync setup` / `/remote-sync setup` flow that configures provider, bucket, prefix, region, and profile, and writes them via a new shared `write_remote_sync_config` in `platform/filestorage/operations.py`. Refuses an org-scoped turn, same as `get_sync_status`/`run_remote_sync`.

**Update 1:** closed after Greptile flagged headless `console.input()` calls and a silently-accepted empty bucket. Fixed: `setup` gated on `repl_tty_interactive()`, and the shared write function rejected an empty bucket.

**Update 2:** closed after Greptile flagged `save_local_settings()` leaking a bare `OSError` on an unwritable settings path, and Ctrl-D during setup surfacing as a generic turn error. Fixed at the source: `save_local_settings` now wraps `OSError` into `LocalSettingsError` like its sibling `load_local_settings` already does, and a `_prompt()` helper catches `EOFError`/`KeyboardInterrupt` for a clean cancellation.

**Update 3:** closed after Greptile and Copilot found four more issues, all fixed:
1. Rerunning setup while sync was already enabled kept the stored `enabled: true` while replacing the destination, so a new, unverified bucket went live immediately. `write_remote_sync_config` now always writes `enabled: false`, so redirecting a destination never skips the explicit re-enable step.
2. An unsupported or misspelled provider saved cleanly and only failed later, at sync time. `write_remote_sync_config` now validates against `registered_providers()` and rejects an unknown one immediately, mirroring `build_object_store`'s own error message.
3. The issue's acceptance criteria calls for `/remote-sync setup` to work in the shell via flags, with interactive prompts reserved for the CLI. The slash handler previously refused every headless call outright. It now parses `--provider/--bucket/--prefix/--region/--profile`, using them directly when present (gateway-safe, no prompting) and falling back to interactive prompts only in a real terminal.
4. Ctrl-D on the very first setup prompt still asked the remaining four before reporting cancellation. The prompts are now asked one at a time in a loop that returns immediately on the first `None`.

```
$ make lint && make typecheck && make check-imports
All checks passed! / Success: no issues found in 1545 source files / All 3 import checks passed.

$ uv run pytest tests/filestorage/ tests/cli/test_remote_sync_command.py tests/cli/test_gateway_remote_sync.py tests/interactive_shell/test_remote_sync_cmds.py tests/surfaces/test_remote_sync_surface_contract.py tests/config/ -q
356 passed
```

---

For the enabled-state finding, the fix had to close the gap for every future call, not just the one Greptile spotted, so `write_remote_sync_config` always sets `enabled: false` unconditionally rather than only when it was previously absent. For the provider finding, reused the same registry (`registered_providers()`) and message shape that `build_object_store` already uses at sync time, so the two failure points read as one consistent contract instead of inventing new wording. For the flags finding, re-read the issue text directly rather than trusting my own earlier interpretation: it explicitly says "`/remote-sync setup` works in the shell (flags; interactive prompts on CLI)" and gives `opensre remote-sync setup --provider <name> --bucket <store>` as the example, which the original implementation didn't satisfy for the shell surface at all. Added a small `--key value` parser scoped to the five known flags, used only when at least one is present, keeping the interactive path as an unconditional fallback for a real terminal. For the cancellation finding, replaced the five separate prompt calls with a loop over `(key, label)` pairs that returns `None` the moment a prompt is cancelled, so no later prompt fires.

---

### DIFF
--- a/config/local_settings.py
+++ b/config/local_settings.py
@@ -53,11 +53,14 @@ def read_section(name: str) -> dict[str, Any]:
 
 
 def save_local_settings(data: dict[str, Any]) -> None:
-    import yaml
-
     path = local_settings_path()
-    path.parent.mkdir(parents=True, exist_ok=True)
-    path.write_text(yaml.safe_dump(data, sort_keys=False), encoding="utf-8")
+    try:
+        import yaml
+
+        path.parent.mkdir(parents=True, exist_ok=True)
+        path.write_text(yaml.safe_dump(data, sort_keys=False), encoding="utf-8")
+    except OSError as exc:
+        raise LocalSettingsError(f"could not write {path}: {exc}") from exc
 
 
 def update_section(name: str, values: dict[str, Any]) -> None:
--- a/platform/filestorage/enums.py
+++ b/platform/filestorage/enums.py
@@ -40,6 +40,7 @@ class RemoteSyncSubcommand(StrEnum):
 
     STATUS = "status"
     SYNC = "sync"
+    SETUP = "setup"
 
 
 __all__ = [
--- a/platform/filestorage/operations.py
+++ b/platform/filestorage/operations.py
@@ -21,6 +21,10 @@ from __future__ import annotations
 from dataclasses import dataclass
 from pathlib import Path
 
+from config.constants.filestorage import (
+    DEFAULT_REMOTE_SYNC_PREFIX,
+    DEFAULT_REMOTE_SYNC_PROVIDER,
+)
 from config.scope_context import current_scope
 from platform.filestorage.config import RemoteSyncConfig, load_remote_sync_config
 from platform.filestorage.engine import SyncReport, resolve_direction, run_sync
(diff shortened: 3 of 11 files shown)


arnavpopley:
Closing to reopen after fixing two more findings from Greptile's round-4 review: writing enabled: false to the stored config did nothing when OPENSRE_REMOTE_SYNC was already set in the environment (env takes precedence by design), so a redirected destination could still go live immediately; both surfaces now check remote_sync_enabled() after a successful write and warn instead of printing the enable hint. Also hardened the slash setup flag parser so a flag missing its value, or followed by another flag instead of a value (--bucket --provider aws), reports a usage error instead of silently saving a malformed value.

PR #4630, opened 2026-07-31
feat: interactive setup for opensre remote-sync (#4554)

### LINKED ISSUE #4554: [Intermediate] opensre remote-sync setup — interactive configuration

arnavpopley:
Fixes #4554

Adds an `opensre remote-sync setup` / `/remote-sync setup` flow that configures provider, bucket, prefix, region, and profile, and writes them via a new shared `write_remote_sync_config` in `platform/filestorage/operations.py`. Refuses an org-scoped turn, same as `get_sync_status`/`run_remote_sync`.

**Update 1:** closed after Greptile flagged headless `console.input()` calls and a silently-accepted empty bucket. Fixed: `setup` gated on `repl_tty_interactive()`, and the shared write function rejected an empty bucket.

**Update 2:** closed after Greptile flagged `save_local_settings()` leaking a bare `OSError` on an unwritable settings path, and Ctrl-D during setup surfacing as a generic turn error. Fixed at the source: `save_local_settings` wraps `OSError` into `LocalSettingsError` like its sibling `load_local_settings` already does, and a `_prompt()` helper catches `EOFError`/`KeyboardInterrupt` for a clean cancellation.

**Update 3:** closed after Greptile and Copilot found four issues: rerunning setup while sync was already enabled kept the stored `enabled: true` while replacing the destination, so a new, unverified bucket went live immediately; an unsupported provider saved cleanly and only failed later, at sync time; the issue's own acceptance criteria calls for `/remote-sync setup` to work in the shell via flags, which the slash handler didn't support at all; and Ctrl-D on the first prompt still asked the remaining four before reporting cancellation. All four fixed with regression tests.

**Update 4:** closed after Greptile found that writing `enabled: false` to the stored config does nothing when `OPENSRE_REMOTE_SYNC` is already set in the environment, since env takes precedence by design, so a redirected destination could still go live immediately. `write_remote_sync_config` can only control the stored section, so the fix lives in both surfaces: CLI and slash now call `remote_sync_enabled()` after a successful write and print a warning instead of the "set the env var" hint whenever the environment is what is actually keeping sync on. Copilot also flagged that the setup flag parser would treat `--provider` in `--bucket --provider aws` as the bucket's value; the parser now returns a clear usage error whenever a recognized flag is missing its value or is followed by what looks like another flag.

```
$ make lint && make typecheck && make check-imports
All checks passed! / Success: no issues found in 1545 source files / All 3 import checks passed.

$ uv run pytest tests/filestorage/ tests/cli/test_remote_sync_command.py tests/cli/test_gateway_remote_sync.py tests/interactive_shell/test_remote_sync_cmds.py tests/surfaces/test_remote_sync_surface_contract.py tests/config/ -q
361 passed

$ uv run pytest tests/interactive_shell/ -q
1394 passed, 1 skipped
```

---

`write_remote_sync_config` only ever touches the stored `remote_sync` section of `config.yml` — it has no business mutating the caller's shell environment, and doing so would be surprising and dangerous. So the env-override finding is fixed one layer up: both `opensre remote-sync setup` and `/remote-sync setup` call the existing `remote_sync_enabled()` helper (which already implements the env-over-stored precedence documented in `platform/filestorage/config.py`) right after a successful write, and print a distinct warning instead of the usual "enable syncing" hint whenever the environment variable is what is actually keeping sync live. This reuses an existing helper rather than inventing new precedence logic. For the flag-parser finding, the fix distinguishes three parser outcomes instead of two: an empty dict (no recognized flags, fall back to prompts), a populated dict (proceed), and `None` (a recognized flag was malformed) so a dangling `--provider` at the end or a `--bucket` immediately followed by another `--flag` reports a usage error rather than silently saving the wrong value. Also tightened `write_remote_sync_config`'s docstring, which previously implied the function alone guarantees sync stays off; it now says explicitly that it only controls the stored switch and points to the surface-level check for the environment case.

---

### DIFF
--- a/config/local_settings.py
+++ b/config/local_settings.py
@@ -53,11 +53,14 @@ def read_section(name: str) -> dict[str, Any]:
 
 
 def save_local_settings(data: dict[str, Any]) -> None:
-    import yaml
-
     path = local_settings_path()
-    path.parent.mkdir(parents=True, exist_ok=True)
-    path.write_text(yaml.safe_dump(data, sort_keys=False), encoding="utf-8")
+    try:
+        import yaml
+
+        path.parent.mkdir(parents=True, exist_ok=True)
+        path.write_text(yaml.safe_dump(data, sort_keys=False), encoding="utf-8")
+    except OSError as exc:
+        raise LocalSettingsError(f"could not write {path}: {exc}") from exc
 
 
 def update_section(name: str, values: dict[str, Any]) -> None:
--- a/platform/filestorage/enums.py
+++ b/platform/filestorage/enums.py
@@ -40,6 +40,7 @@ class RemoteSyncSubcommand(StrEnum):
 
     STATUS = "status"
     SYNC = "sync"
+    SETUP = "setup"
 
 
 __all__ = [
--- a/platform/filestorage/operations.py
+++ b/platform/filestorage/operations.py
@@ -21,6 +21,10 @@ from __future__ import annotations
 from dataclasses import dataclass
 from pathlib import Path
 
+from config.constants.filestorage import (
+    DEFAULT_REMOTE_SYNC_PREFIX,
+    DEFAULT_REMOTE_SYNC_PROVIDER,
+)
 from config.scope_context import current_scope
 from platform.filestorage.config import RemoteSyncConfig, load_remote_sync_config
 from platform.filestorage.engine import SyncReport, resolve_direction, run_sync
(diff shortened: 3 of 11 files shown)


arnavpopley:
TBC

muddlebee:
hey @arnavpopley thank you for efforts, but already merged as part of #4625. closing this, I will take the issue up

PR #4631, opened 2026-07-31
fix(filestorage): reject incomplete Vercel blob metadata

muddlebee:
Fixes: Greptile P1 finding on re-trigger of #4625 (targets that feature branch — `platform/filestorage/providers/vercel.py` doesn't exist on `main` yet). Merge into `feat/vercel-blob-remote-sync-provider`, not `main`.

A Vercel Blob list-response entry missing `pathname` or `uploadedAt` was silently defaulted (empty-string key, `datetime.now()` timestamp) instead of rejected. That let incomplete metadata look authoritative: an empty key masks the real remote key, and a "now" timestamp can make stale/incomplete remote data outrank a genuinely newer local file during sync — same class of bug the existing malformed-shape handling in this file already guards against, just one level deeper (per-item fields, not just the collection shape).

- `_blobs_from_list_payload` now requires a non-empty string `pathname` per blob item, alongside its existing checks for missing/non-list `blobs` and non-object entries.
- `_parse_uploaded_at` now raises `ValueError` instead of returning `datetime.now(tz=UTC)` for a missing, null, or unparseable `uploadedAt` — `list_objects`'s existing `except (httpx.HTTPError, TypeError, ValueError)` already converts that to `RemoteSyncUnavailableError`, so no new error path was needed.
- `list_objects` reads `blob["pathname"]` directly now that presence is guaranteed upstream, rather than re-defaulting with `.get("pathname", "")`.

```
$ uv run python -m pytest tests/filestorage/test_vercel_provider.py -q
..........                                                              [100%]
10 passed in 1.00s
```

New test `test_incomplete_blob_metadata_is_remote_sync_unavailable` covers: missing `pathname`, blank `pathname`, missing `uploadedAt`, null `uploadedAt`, unparseable `uploadedAt` string, and wrong-typed `uploadedAt` — each now raises `RemoteSyncUnavailableError` instead of silently producing a phantom/mistimed entry.

`make lint`, `make format-check`, `make typecheck`, and the full `tests/filestorage/` suite (76 passed) are green.

---

Greptile flagged that incomplete per-blob metadata (missing `pathname`/`uploadedAt`) was defaulted rather than rejected during Vercel Blob listing, which can silently corrupt sync decisions. The fix extends the same fail-closed validation the file already applies at the collection level (`_blobs_from_list_payload` rejecting missing/non-list `blobs`, non-object entries) down to the two per-item fields the sync engine actually depends on for correctness: the object's key and its comparison timestamp. `_parse_uploaded_at` raising instead of defaulting was the minimal change — the calling code (`list_objects`) already had a `ValueError`-catching except clause built for exactly this shape, from the earlier malformed-payload fix in this same PR chain, so no new error-handling path was required.

---

---

### DIFF
--- a/platform/filestorage/providers/vercel.py
+++ b/platform/filestorage/providers/vercel.py
@@ -71,7 +71,8 @@ class VercelBlobObjectStore:
         out: list[RemoteObject] = []
         try:
             for blob in self._iter_blobs(full_prefix):
-                pathname = str(blob.get("pathname", ""))
+                # _blobs_from_list_payload already guaranteed a non-empty str.
+                pathname = blob["pathname"]
                 out.append(
                     RemoteObject(
                         key=self._strip_prefix(pathname),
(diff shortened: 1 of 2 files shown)


muddlebee:
@greptile review

muddlebee:
@greptile review

muddlebee:
@greptile review

PR #4632, opened 2026-07-31
feat(filestorage): add Google Cloud Storage provider and remote-sync setup

farizanjum:
Fixes #4563

#### Describe the changes you have made in this PR -

Adds Google Cloud Storage as a second built-in object-store provider for remote sync (`OPENSRE_REMOTE_SYNC_PROVIDER=gcs`), plus the `opensre remote-sync setup` command from the issue's Onboarding section. Engine, CLI sync paths, REPL, and gateway are untouched; everything goes through the existing open/closed registry.

**Provider (`platform/filestorage/providers/gcs.py`)**
- Four-method `ObjectStore` implementation over `google-cloud-storage`, registered via `_BUILTIN_MODULES` (lazy import).
- Change detection: converts GCS's base64 `md5Hash` to the hex tag the engine compares. Strict decode (`validate=True`, exactly 16 bytes); composite, foreign, or malformed objects fall back to timestamp comparison, same as S3 multipart. The version-style `blob.etag` is never used - it is not a content hash.
- Credentials are ambient (Application Default Credentials). No new env vars; nothing credential-shaped is stored anywhere.
- All vendor and transport failures wrap to `RemoteSyncUnavailableError`. The catch is broad on purpose: `google-resumable-media` / `requests` exceptions (DataCorruption, InvalidResponse, connection errors) do not inherit `GoogleAPIError`, and the engine and chat surfaces may only ever see `RemoteSyncError` (CWE-209 boundary unchanged).
- `list_objects` force-materializes the lazy listing inside the try and maps to `RemoteObject` outside it, so provider-side mapping bugs can never be masked as availability errors.

**Setup (`opensre remote-sync setup --provider gcs`)**
- New `platform/filestorage/setup.py`: validates the provider against the registry first, then persists exactly six non-secret values (enabled/provider/bucket/prefix/region/profile) into the `remote_sync` section of `~/.opensre/config.yml`. `update_section` merges, so pre-existing keys survive.
- CLI flags for everything; TTY prompts for provider/bucket/prefix when omitted (region/profile are flags-only); clear error when run in a pipe without `--bucket`.
- Provider-aware credential hints (`gcloud auth application-default login` for GCS, profile/SSO for AWS) via a small `credential_hint_for_provider` registry helper and a `CREDENTIAL_HINT` constant per provider module.
- `--disabled` writes `enabled: false` and the confirmation correctly reports "Remote sync is off."
[...]

PR #4633, opened 2026-07-31
feat(filestorage): add Google Cloud Storage provider for remote sync

### LINKED ISSUE #4563: [Advanced] Google Cloud Storage provider for remote sync
### Complexity
Advanced — needs a cloud account. Requires a GCP account.

### Context
Implement the four-method `ObjectStore` protocol in `platform/filestorage/providers/gcs.py` and register it. Engine, CLI, and REPL need no changes — already proven by `tests/filestorage/test_multi_provider_support.py`.

Including the `_BUILTIN_MODULES` entry. A third party registering from their own package needs no core edit at all.

### Done when
- A full push-and-restore cycle works against a real GCS bucket.
- The contributor confirms it themselves.
- Built-in registration includes `_BUILTIN_MODULES` in `providers/registry.py` (lazy import avoids a package self-import cycle).

### Onboarding
Support setup via `opensre remote-sync setup --provider gcs` (and env). Ambient GCP credentials; never store secrets in config.yml. Not part of `opensre onboard`.

---

farizanjum:
Fixes #4563. Supersedes #4632 (closed as outdated after #4625 merged; this branch is rebased on current main and reconciled with the merged Vercel Blob provider).

Adds Google Cloud Storage as a built-in object-store provider for remote sync (`OPENSRE_REMOTE_SYNC_PROVIDER=gcs`), plus hardening of the shared `remote-sync setup` flow that the issue's Onboarding section requires.

**GCS provider (`platform/filestorage/providers/gcs.py`, new)**
- Four-method `ObjectStore` implementation against the GCS JSON API over `AuthorizedSession` + `requests` (google-auth is already a hard dependency), registered via `_BUILTIN_MODULES` (lazy import). No GCS SDK: `pyproject.toml`, `uv.lock`, and `mypy.ini` are untouched - zero new packages for any install. Engine, CLI sync paths, REPL, and gateway are unchanged.
- Change detection: converts the listing's base64 `md5Hash` to the hex tag the engine compares (strict decode, exactly 16 bytes). Composite, foreign, or malformed objects get no tag and fall back to timestamp comparison, same as S3 multipart. The version-style resource `etag` is never used - it is not a content hash.
- Credentials are ambient (Application Default Credentials): `google.auth.default` resolves them and the session refreshes the bearer token per call. No new env vars; nothing credential-shaped is stored anywhere.
- Error boundary: transport errors, credential-refresh errors (`RefreshError`), HTTP errors via `raise_for_status`, and malformed success payloads all become `RemoteSyncUnavailableError`; chat surfaces only ever see `RemoteSyncError` (CWE-209 boundary unchanged). Untrustworthy or offset-less timestamps fail closed - the sync stops at the listing rather than uploading over a remote object whose age is unknown.
- Hand-rolled pagination (`nextPageToken`), percent-encoded object names in the download path, `uploadType=media` single-request uploads - the same plain-HTTP shape as `vercel.py`.

**Setup hardening (on top of the merged #4625 flow)**
- `save_remote_sync_settings` now validates the provider against the registry, so a typo (`--provider gogle`) fails at setup with the known-provider list instead of at first sync.
- New `disable_remote_sync()`: `opensre remote-sync setup --disabled` (no bucket) now just writes `enabled: false` and keeps the stored provider/bucket for later - previously it demanded a bucket (Greptile P2 on #4632). `--disabled` with explicit setup flags and no bucket is rejected instead of silently dropping them.
- `--disabled` output is state-accurate on both CLI and REPL: it says sync is off and never suggests an immediate sync.
- GCS credential hint ("run `gcloud auth application-default login`") registered alongside the provider factory, matching the AWS/Vercel pattern.
- One source of truth for the built-in provider list (`builtin_providers()`): CLI help, the setup wizard prompt, REPL notes, and `DISABLED_HELP` all compute from it, so help text stays correct as providers land - and community registrations are never mislabeled as built-in.

**Docs**
- GCS provider section: ADC setup, IAM note - `storage.objects.list/get/create/delete`; GCS requires delete to *replace* an existing object (changed sessions re-upload), even though opensre itself never deletes. Default encryption at rest.
- Providers table gains a GCS row; setup/env examples cover all three built-ins; `.env.example`, package docstrings, and constants comments updated.

https://github.com/user-attachments/assets/f51ebfbb-6da1-45e5-a1f6-b629f45667a9

The recording shows the full zero-to-working run on the plain-JSON-API build: the SDK being absent, the provider tests, bucket creation, first sync, the second-sync etag skip, the overwrite path, the pull-restore with byte-identical hashes, and cleanup.

Terminal evidence (live, personal GCP account via ADC; every bucket deleted after its run):

```
$ opensre remote-sync setup --provider gcs --bucket opensre-setup-verify-fariz --prefix opensre
Remote sync settings saved -> gcs / opensre-setup-verify-fariz/opensre
GCP credentials stay ambient: run `gcloud auth application-default login` once per machine with the account that owns the bucket; opensre stores nothing.

$ opensre remote-sync status          # no env vars set; reads stored config
Remote sync is on (gcs) -> opensre-setup-verify-fariz/opensre

$ opensre remote-sync sync
Sync complete - 0 downloaded, 3 uploaded (486 B), 0 already current (486 B total).

$ opensre remote-sync sync            # second run, no changes: etag skip path
Sync complete - 0 downloaded, 0 uploaded, 6 already current.

# after editing one local session file (overwrite path):
Sync complete - 0 downloaded, 1 uploaded (267 B), 5 already current (267 B total).

# restore: sessions dir moved away, then `sync --pull-only`
Sync complete - 2 downloaded (450 B), 0 uploaded, 1 already current (450 B total).
# all restored files verified byte-identical (matching Get-FileHash MD5 pairs)
```

The disabled flow (the Greptile finding on #4632), live:

```
$ opensre remote-sync setup --disabled
Remote sync is off. Saved enabled: false to ~/.opensre/config.yml (stored provider/bucket kept for when you turn it back on).

remote_sync:
  enabled: false
  provider: gcs
  bucket: opensre-setup-verify-fariz
  prefix: opensre
```

Additional live checks from independent QA runs: a planted `credentials.json` in the sessions dir is refused by the deny-list (`Sync failed: refusing to upload ...`), an unknown provider fails with `known: aws, gcs`, and the stored `remote_sync` section contains no credential material anywhere.

The issue required a GCS ObjectStore registered as a built-in, proven by a real push-and-restore cycle, plus a setup command that never stores secrets. I mirrored the existing `aws.py` provider because the registry was designed for exactly that shape (open/closed: new module plus one register call), and - after the SDK-versus-API discussion on this PR - implemented it the `vercel.py` way: plain JSON API over `AuthorizedSession`, since a bearer-token vendor needs no SDK of its own and `pyproject.toml`/`uv.lock` stay untouched. The one GCS-specific problem is change detection: the resource `etag` is a generation tag, not a content hash, so the provider converts the listing's base64 `md5Hash` into the hex MD5 the engine compares; composite or malformed objects degrade to timestamp comparison, the same as S3 multipart. Credentials are ambient ADC, so setup persists only non-secret values and points the user at `gcloud auth application-default login`. Alternatives considered and rejected: using the resource `etag` directly (breaks skip-on-unchanged), CRC32C (would require engine changes, which the issue forbids), storing a service-account path in config.yml (violates "never store secrets"), and the `google-cloud-storage` SDK itself (four extra packages for every install; its real wins - resumable uploads and managed retries - do not apply to KB-scale files, and retry policy is #4555's scope). Fail-closed choices where metadata cannot be trusted: malformed listings, credential-refresh errors, and unknown or offset-less timestamps all stop the sync with `RemoteSyncUnavailableError` rather than guessing an ordering.

---

---

## Notes for reviewers

- Tests: 24 GCS provider tests, plus setup/CLI/registry/enum/message coverage. Full filestorage/CLI/REPL/gateway/surface suites pass (204); ruff, ruff format, and mypy (1550 files) are clean.
- Five pre-existing test failures reproduce on *unmodified* `main` on Windows only (text-mode `write_text("\n")` writes CRLF; byte assertions then mismatch): four in `test_multi_provider_support.py` / `test_remote_sync.py` and `test_vercel_provider.py::test_engine_push_restore_via_registry` from #4625. Linux CI is unaffected.
- Resolved review rounds on this PR and #4632: the overwrite-requires-delete IAM documentation, the setup command itself (issue Onboarding section), strict base64/16-byte md5Hash parsing, the Greptile 4/5 disabled-without-bucket issue, the SDK-to-REST rework, credential-refresh (`RefreshError`) wrapping, malformed payload/timestamp fail-closed behavior, enabled-aware setup output on CLI and REPL, and one built-in provider source for all help text.


farizanjum:
@greptile review

farizanjum:
@greptile review

Devesh36:
Nice one 👍  can you fix the conflicts please

farizanjum:
@greptile review

muddlebee:
Nice work on this — the etag/md5 handling, the `--disabled` fix, and the setup validation are all solid. One design question before merge: do we need the `google-cloud-storage` SDK at all here?

This PR adds `google-cloud-storage` as a hard dependency, which pulls in three more packages (`google-cloud-core`, `google-resumable-media`, `google-crc32c`) for *every* opensre install, not just GCS users — `pyproject.toml` and `uv.lock` both grow regardless of whether anyone touches GCS remote sync.

Compare to how the two existing providers handle their SDK dependency:
- `aws.py` uses `boto3` — justified, since S3 needs SigV4 request signing, which you don't want to hand-roll.
- `vercel.py` uses plain `httpx` against `vercel.com/api/blob` — no SDK — because Vercel Blob auth is just a bearer token.

GCS auth is in the `vercel.py` bucket, not the `aws.py` bucket: the JSON API just needs an OAuth2 bearer token, and `google-auth` (already a hard dependency via `google-api-python-client`) ships `google.auth.transport.requests.AuthorizedSession` for exactly that — ADC resolution + token refresh, for free, with zero new packages.

Concretely, the four `ObjectStore` methods map onto plain REST calls:
- list: `GET storage.googleapis.com/storage/v1/b/{bucket}/o?prefix=...` (loop on `nextPageToken`)
- get: `GET storage.googleapis.com/storage/v1/b/{bucket}/o/{object}?alt=media`
- put: `POST storage.googleapis.com/upload/storage/v1/b/{bucket}/o?uploadType=media&name=...`

`md5Hash` comes back on the JSON object resource either way, so the base64→hex conversion logic in this PR is unaffected — this is a swap of the transport layer only, not the change-detection design.

Would you be up for reworking `gcs.py` to use `AuthorizedSession` + `requests` instead of the SDK, and dropping the `pyproject.toml`/`uv.lock` additions? Happy to help sketch the request/response shapes if useful. Everything else here looks ready to go.

muddlebee:
@farizanjum IMO its ideal to use the API endpoints rather than the google-cloud-storage SDK. thoughts?

farizanjum:
> @farizanjum IMO its ideal to use the API endpoints rather than the google-cloud-storage SDK. thoughts?

@muddlebee Yes, I think the SDK is overkill for three REST calls, and the vercel.py shape is the right precedent. I'll rework gcs.py onto AuthorizedSession + requests (google-auth is already a hard dep), so pyproject.toml and uv.lock drop out of the diff entirely, along with the mypy.ini google.cloud overrides.

 The transport is all that changes: md5Hash rides the JSON resource the same way, so the base64 to hex conversion, prefix handling, and the RemoteSyncUnavailableError boundary all stay as-is. I'll re-run the full live push-and-restore cycle on the new transport plus the test suites, then ping you. I've got the three endpoint shapes mapped (list with nextPageToken, get with alt=media, media upload POST) 

will tag you if need any help.

muddlebee:
thank you @farizanjum you can still explore if google sdk is worth it or for our use case API endpoints will work just fine, either way if you plan the API route please add a nice demo :)

farizanjum:
> thank you @farizanjum you can still explore if google sdk is worth it or for our use case API endpoints will work just fine, either way if you plan the API route please add a nice demo :)

So I dug into both options properly before picking one.

- **SDK side (google-cloud-storage):** the good stuff is managed retries, resumable uploads with checksums, and Google handling API changes for us through SDK updates. But none of that matters much here. We upload tiny session and memory files, syncs run when the user asks, and retries already have their own issue (#4555) to get solved once for all providers. The cost is real though: 4 extra packages in every opensre install, even for people who never touch GCS, and it breaks the pattern vercel.py already set (plain HTTP, no SDK).

- **Plain API side (AuthorizedSession + requests):** zero new dependencies, since google-auth and requests are already in the project pyproject and uv.lock stay untouched. The whole thing is about 30 lines of readable HTTP, same shape as the Vercel provider. The tricky bits we have to own are pagination, URL-encoding object names in the download path, and parsing the timestamp/hash fields all small, and I added a test for each (pagination across pages, %20 in the URL, malformed md5, missing fields, connection errors).

So for three simple calls with a bearer token, the plain API route is the better fit. The SDK would make sense if we were moving big files around or needed its retry machinery, which we don't. So, I am working on it and I will be pushing the new commits with a demo video late night.

muddlebee:
Ok great

farizanjum:
@greptile review

muddlebee:
@farizanjum good work, thank you for the PR :partying_face:

PR #4634, opened 2026-08-01
Make the CLI entry readable: extract the REPL launch, inject the shell console, move the definition out of __main__.py

YauhenBichel:
Three small, behavior-preserving steps along the path a user actually takes — `main.py` → `surfaces/cli` → the interactive shell — plus one type fix that fell out of them.

Tracing that path end to end surfaced three rough spots, in order of what they cost a developer.

**`cli()` decided whether to open the shell inline.** A 29-line, four-level-deep block inside a Click callback: parameter-source introspection, a three-branch ternary chain, a config load, and `raise SystemExit(run_repl(...))`. It was the hardest part of the file and the part a newcomer most needs to follow.

**The shell's console was bound at import.** One module-level `Console(force_terminal=True, ...)` served the splash, the ready box, the login gate, the resume output, and the controller. Nothing could redirect any of it, and it emitted ANSI even when piped — so the shell was not embeddable.

**`__main__.py` held the whole CLI.** 304 lines: the launcher *and* the group, every option, and the exception ladder. `gateway/main.py` already shows the intended shape — 17 lines, with its definition in `gateway/runtime/manager.py`. The CLI was the outlier, and 24 test files were importing from a dunder module whose name means "run me", not "import me".


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4637, opened 2026-08-01
feat(filestorage): add fail-closed client-side encryption to remote sync

### LINKED ISSUE #4562: [Advanced] Client-side encryption of synced content
### Complexity
Advanced — needs a cloud account or careful design. Security-sensitive design. Read the Warning in `docs/configuration/remote-sync.mdx` before starting.

### Context
Server-side encryption means the cloud provider can still read incident history.

### Done when
- Content is encrypted before upload under a user-held key.
- Key handling and rotation are documented.
- Losing the passphrase fails loudly (never silently syncs plaintext).

---

parthchilwerwar:
Fixes #4562

I added opt-in, fail-closed client-side encryption for remote-sync sessions and memory.

- Added a content-codec boundary without changing the existing four-method `ObjectStore` protocol.
- Added deterministic authenticated encryption using AES-SIV, with a per-store key derived from an ambient passphrase using Scrypt.
- Kept `OPENSRE_REMOTE_SYNC_PASSPHRASE` out of config models, `config.yml`, logs, and user-facing output.
- Added an encrypted verification object for each initialized prefix. It validates the passphrase before any upload can modify remote content.
- Prevented mode crossover: encryption refuses nonempty unverified prefixes, while plaintext mode refuses prefixes already initialized for encryption.
- Made encryption flags strict so a typo cannot silently disable encryption and upload plaintext.
- Preserved ETag-based unchanged-file detection by encrypting identical content deterministically for the same object path.
- Kept `--pull-only` read-only: an empty encrypted prefix is not initialized until an upload-capable sync.
- Added `--encrypt` support to CLI and `/remote-sync setup`, plus status/setup guidance.
- Documented setup, plaintext migration, passphrase rotation, recovery limits, downgrade protection, and provider-visible metadata.

```text
$ uv run python -m pytest -q tests/filestorage \
    --deselect <five existing Windows CRLF-only assertions>
91 passed, 5 deselected

$ uv run python -m pytest -q \
    tests/interactive_shell/test_remote_sync_cmds.py \
    tests/surfaces/test_remote_sync_surface_contract.py \
    tests/integrations/test_verification_registry.py \
    tests/integrations/test_registry.py
54 passed

$ uv run python -m ruff check config core gateway integrations platform surfaces tools tests/
All checks passed!

$ uv run python -m ruff format --check config core gateway integrations platform surfaces tools tests/
3075 files already formatted

Success: no issues found in 1550 source files
```

The five deselected filestorage assertions are the repository's existing Windows CRLF fixture failures; the same tests run normally on Linux CI. The final corrective commit passed CI, CodeQL, synthetic deterministic tests, and the interactive-shell workflow.

---

The provider abstraction already has the right responsibility: moving opaque bytes. I kept it unchanged and inserted a codec at the local/object-store boundary. Plaintext configurations use an identity codec, while encrypted configurations derive one in-memory AES-SIV key from the ambient passphrase and store identity.

I considered randomized authenticated encryption, but it would produce different bytes on every run and defeat the engine's ETag comparison, causing unchanged files to be uploaded repeatedly. AES-SIV provides authenticated deterministic encryption when used without a nonce. The logical object path is included inside the authenticated plaintext envelope, so moving ciphertext to another path fails closed. The versioned outer header leaves room for a future format migration.

Ciphertext alone cannot prove that the current passphrase is correct without reading it. Conflict resolution can legitimately avoid downloading an older remote file, so relying only on pull-time decryption would allow a newer local file to overwrite remote data under a wrong key. To close that gap, the first encrypted upload writes a small authenticated verification object. Every later sync verifies it before uploads. The same marker prevents plaintext mode from writing into an encrypted prefix. A nonempty prefix without a marker is rejected in encryption mode, which forces migration to a new prefix rather than mixing formats.

The passphrase is ambient rather than a config field; only the opt-in boolean is persisted. Boolean parsing is strict because treating an unknown value as false would be an unsafe downgrade. Existing plaintext prefixes migrate by pulling first and then pushing to a new prefix. Rotation uses another new prefix so the old data remains recoverable until the replacement is verified.

---


docs/configuration/remote-sync.mdx
```diff
@@ -113,6 +120,40 @@ opensre remote-sync sync --pull-only    # new machine, fetch only
...
+
+Add `--encrypt` during setup, or set `OPENSRE_REMOTE_SYNC_ENCRYPTION=1`, to
+encrypt sessions and memory before upload. Keep
+`OPENSRE_REMOTE_SYNC_PASSPHRASE` in your shell secret manager and provide the
+same value on every machine. The passphrase is never written to
+`~/.opensre/config.yml`.
+
+Missing or incorrect passphrases, changed object paths, tampered ciphertext,
+and plaintext objects all stop the sync with an error. There is no recovery
+without the original passphrase.
```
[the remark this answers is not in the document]
parthchilwerwar: Addressed. The docs now state the exact pull behavior instead of claiming every object is always decoded. I also tightened the implementation: the first encrypted upload creates an authenticated verification object, later syncs validate it before any upload, plaintext mode refuses an initialized encrypted prefix, and an empty encrypted `--pull-only` run remains read-only. This makes the passphrase/downgrade guarantee explicit without overstating per-object validation.

parthchilwerwar:
Thanks for reviewing this. I noticed that I had made a mistake in the earlier update, including the formatting failure and a couple of encryption edge cases. I’ve corrected the implementation so setup preserves the existing encryption mode when the flag is omitted, unknown or misspelled flags fail safely, and local files are validated before the encrypted prefix is initialized.

All CI, CodeQL, synthetic, and interactive-shell checks are now passing.

(I'm human sorry for the mistake. It has been fixed in the committed code.)

platform/filestorage/encryption.py
```diff
@@ -0,0 +1,171 @@
...
+            raise RemoteSyncEncryptionError(
+                "Remote encrypted content belongs to a different object path."
+            )
+        return payload[key_end:]
+
+
+def _store_salt(config: RemoteSyncConfig) -> bytes:
+    """Stable, store-specific Scrypt salt shared by matching configurations."""
+    identity = "\x00".join((config.provider, config.bucket, config.prefix)).encode("utf-8")
+    return hashlib.sha256(_KDF_CONTEXT + identity).digest()[:_KDF_SALT_BYTES]
```
Davidson3556, line 90: Blocking. The salt is a pure function of provider/bucket/prefix — I recomputed it outside the codebase and got the same 16 bytes — so there's no per-store randomness, and `_VERIFICATION_PAYLOAD` gives an attacker a known plaintext to test guesses against offline. Measured: 136 ms/guess at N=2**14 vs 1127 ms at 2**17.

Two fixes, I'd take both: raise `_KDF_N` to `2**17` (derivation is once per `run_remote_sync`, which is CLI/slash only, so ~1s is fine), and store a random 16-byte salt in the marker. The second needs reordering, since `content_codec_for` runs before `prepare_remote_content` reads the marker.

platform/filestorage/encryption.py
```diff
@@ -0,0 +1,171 @@
...
+    RemoteSyncEncryptionError,
+)
+from platform.filestorage.ports import ObjectStore, RemoteObject
+
+_ENVELOPE_MAGIC = b"opensre-remote-sync\x00"
+_ENVELOPE_VERSION = b"\x01"
+_ENVELOPE_HEADER = _ENVELOPE_MAGIC + _ENVELOPE_VERSION
+_KDF_CONTEXT = b"opensre.remote-sync.encryption.v1\x00"
+_KDF_KEY_BYTES = 64
+_KDF_N = 2**14
```
Davidson3556, line 29: `2**17` per the OWASP scrypt minimum. See the note on `_store_salt` for the measured costs and why this one matters more than usual here.

platform/filestorage/encryption.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def __init__(self, key: bytes) -> None:
+        self._cipher = AESSIV(key)
+
+    def encode(self, key: str, data: bytes) -> bytes:
+        """Encrypt and authenticate ``data`` for exactly ``key``."""
+        key_bytes = key.encode("utf-8")
+        payload = len(key_bytes).to_bytes(_KEY_LENGTH_BYTES, "big") + key_bytes + data
+        ciphertext = self._cipher.encrypt(payload, [])
+        return _ENVELOPE_HEADER + ciphertext
```
Davidson3556, line 56: AES-SIV takes associated data natively — `self._cipher.encrypt(data, [key.encode("utf-8")])` gives the same binding, fails with `InvalidTag` on a wrong path, and lets `decode` drop lines 71-83, `_KEY_LENGTH_BYTES`, and both bounds checks. Also 16 bytes overhead instead of 64. The current parsing isn't exploitable (it's inside the AEAD), just ~15 lines you don't need to own.

platform/filestorage/encryption.py
```diff
@@ -0,0 +1,171 @@
...
+def content_codec_for(config: RemoteSyncConfig) -> ContentCodec:
+    """Build the configured codec without retaining the passphrase."""
+    if not config.encryption:
+        return PLAINTEXT_CONTENT_CODEC
+    passphrase = os.getenv(REMOTE_SYNC_PASSPHRASE_ENV, "")
+    if not passphrase.strip():
+        raise RemoteSyncConfigError(
+            f"Client-side encryption is enabled but {REMOTE_SYNC_PASSPHRASE_ENV} is not set."
+        )
+    return EncryptedContentCodec(_derive_key(config, passphrase))
```
Davidson3556, line 113: Line 109 strips, line 113 doesn't, so `"hunter2"` and `"hunter2\n"` both pass and derive different keys — a prefix that won't decrypt with nothing pointing at whitespace. Strip once and use the stripped value for both, or compare against `""` and stay byte-exact.

platform/filestorage/content_codec.py
```diff
@@ -0,0 +1,36 @@
...
+class PlaintextContentCodec:
+    """Identity codec used by existing, unencrypted configurations."""
+
+    def encode(self, _key: str, data: bytes) -> bytes:
+        """Return ``data`` unchanged."""
+        return data
+
+    def decode(self, _key: str, data: bytes) -> bytes:
+        """Return ``data`` unchanged."""
+        return data
```
Davidson3556, line 27: Plaintext `decode` accepts anything, so a deleted marker (lifecycle rule, manual cleanup) means ciphertext gets written over local session files — I reproduced it, `a.jsonl` ends up starting with `opensre-remote-sync\x00\x01`. `prepare_remote_content` can't guard what the marker no longer says. Two lines here close it:

```python
def decode(self, _key: str, data: bytes) -> bytes:
    if data.startswith(_ENVELOPE_MAGIC):
        raise RemoteSyncEncryptionError(
            "Remote content is encrypted but this run is configured for plaintext."
        )
    return data
```

platform/filestorage/config.py
```diff
@@ -87,6 +91,50 @@ def _env_or_stored(
...
+    unrelated settings file is damaged. A valid stored encryption opt-in is
+    still honored when the other fields happen to come from the environment.
+    """
+    env = os.getenv(env_name)
+    if env is not None and env.strip() != "":
+        return _strict_flag(env, source=env_name)
+    try:
+        value = stored().get(stored_key)
+    except RemoteSyncConfigError:
+        return False
```
Davidson3556, line 111: The one place the PR fails open: an unreadable `config.yml` returns `False`, meaning plaintext. That contradicts `_strict_flag` right below, whose whole rationale is that an unknown value must not become false. If the file can't be read you don't know the setting — let the error propagate like `_env_or_stored` does; the surfaces already catch `RemoteSyncError`.

platform/filestorage/engine.py
```diff
@@ -195,11 +204,15 @@ def _local_path_for(obj: RemoteObject, by_name: dict[str, SyncRoot]) -> Path | N
...
+def _should_download(
+    obj: RemoteObject,
+    target: Path,
+    codec: ContentCodec,
+) -> bool:
     if not target.exists():
         return True
     tag = comparable_etag(obj)
-    if tag and tag == content_tag(target.read_bytes()):
+    if tag and tag == content_tag(codec.encode(obj.key, target.read_bytes())):
```
Davidson3556, line 215: Every file is now encrypted twice per `run_sync` — once here for the comparison tag, once in `push` for the body. Instrumented a steady-state sync with 2 unchanged files: 4 encodes, i.e. four AES-SIV passes over the whole corpus to transfer nothing. Cache by `(key, sha256(data))` for the duration of one sync, or compute the tags once and share them.

Davidson3556 (commented):
Ran it: 99 filestorage tests pass here, including the five you flagged as Windows CRLF. Design is right — codec at the boundary with `ObjectStore` untouched, AES-SIV for determinism, path bound into the envelope, marker closing the gap ciphertext can't.

I probed the crypto instead of reading it. One blocker: the salt is derivable from public config and the marker is a known plaintext, so at N=2**14 (136 ms/guess) an offline dictionary attack is cheap for exactly the adversary #4562 names. Details inline.

One surface gap: `remote_sync_cmds.py:167` swallows all detail, correct for provider messages under CWE-209, but the encryption errors are your own static wording — same as `OrgScopeNotSupportedError` two lines up, which that handler deliberately shows. So a missing passphrase fails closed everywhere but only fails legibly on the CLI. Suggest raising it as `RemoteSyncEncryptionError` and adding that class to the visible branch; keep `RemoteSyncConfigError` quiet since `setup.py` builds it from `LocalSettingsError`.

Four smaller ones inline. Branch also conflicts with main. Docs page is the strongest part.

also missing demo video

muddlebee:
@parthchilwerwar thank you for the efforts, but needs more discussions and approval before opening PRs. feel free to discuss at github or discord.

PR #4661, opened 2026-08-01
perf: stop re-scanning hot-path data structures on every turn

YauhenBichel:
Six places on per-turn paths did the same work more than once, or scaled with data that only grows. Each is now resolved once and reused. No behaviour changes; the measurements below are from benchmarks run against the real registry and realistic file sizes.

| Path | Before | After |
| --- | --- | --- |
| `flush()` closing a 4,000-record session | 33.9 ms | 11.2 ms |
| `session.history` after 1,000 turns | 2.5 MB | 0.46 MB |
| Gateway tool-label cache miss (267 tools) | 7.0 us | 0.1 us |
| Memory index per turn (300 memories) | 19.96 ms | 2.74 ms |
| Tool-payload redaction per tool call | 260 us | 126 us |

- **Session flush** parsed the whole JSONL three times. The two re-reads were  dead: every counter matches only `custom_message`/`turn_stub` records, and  neither intervening append writes one, so the re-parses returned identical
  answers.
- **`session.history`** grew without bound because each row can carry a full  agent reply. Rows are still never removed — see the note below — but a row  outside a 20-turn window now sheds its response body, which nothing reads.
- **Gateway tool labels** resolved a name with a linear scan of the registry on  every LRU miss. The registry already caches a `{name: tool}` map; this adds  the missing single-tool accessor rather than using `get_registered_tool_map`,
  which copies the whole map per call.
- **Long-term memory** was read and parsed in full on every action-agent turn.  Records are now reused until a per-file `(name, mtime_ns, size)` signature  changes.
- **Tool-payload redaction** deep-copied a tool's input when the call was  announced, then deep-copied the same input again to attach the output. The  second walk now reuses the first.
- **Tool planning** hoisted `secondary_tool_sources()` out of the per-tool  scoring loop.

Alongside the above, some typing and style hardening:

- `AgentMode` becomes a `StrEnum` instead of a `Literal`, with state dumps  switched to `model_dump(mode="json")` so the runtime dict keeps a  JSON-stable shape.
- Protocol methods in `turn_snapshot.py` and `tool_planning.py` use  docstring-only bodies instead of `raise NotImplementedError`, per AGENTS.md.
- Anthropic's non-standard 529 gets a named constant.
- CONTRIBUTING documents rebasing onto `main` rather than merging `main` in.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -154,6 +154,31 @@ pytest tests/synthetic/ -k "test_scenario"                      # no live infra 
 
 Follow the PR template (see below). Link the relevant issue and describe what changed and why.
 
+### 6. Keep the Branch Up to Date with `main` (Rebase)
+
+PRs must be current with `main` before merge (branch protection). **Update by rebasing onto `main` — do not merge `main` into your feature branch.**
+
+Merging `main` into the PR branch creates noisy merge commits and trains the GitHub “Update branch” button toward that path. Rebase keeps history linear and surfaces conflicts on your commits, where they belong.
+
+**Preferred (local):**
+
+```bash
+git fetch origin main
+git rebase origin/main
+# resolve any conflicts, then:
+git push --force-with-lease
+```
+
+**On GitHub:** if the PR shows “This branch is out-of-date with the base
+branch”, prefer **Update with rebase** when the UI offers it. Avoid **Update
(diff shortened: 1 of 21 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4670, opened 2026-08-02
fix(ci): refresh stale auto-merge PRs and keep runtime paths human-merged

YauhenBichel:
Two problems on the auto-merge path, plus a policy change.

**PRs stall when `main` moves.** With "require branches to be up to date" enabled, every merge marks all other open PRs `BEHIND`. Nothing automated reacted to that: `automerge.yml` subscribes to `pull_request_target` and `workflow_run`, but not to `push` on `main` — so at the exact moment PRs go stale, no workflow runs. Someone had to press "Update branch" and wait for CI.

That does not scale at our merge rate. `main` took 156 commits in the last 7 days (22/day, 47 on the busiest). Inside an 8-hour window at that peak a merge lands roughly every 10 minutes, while the gating CI jobs are budgeted at 15–30. A PR goes stale during its own CI run, so the manual loop cannot terminate.

**Anything can auto-merge.** A change to the agent runtime, the multi-tenant platform, the chat gateway, or to the merge machinery itself could land with no human on the button.

What changed

- `automerge.yml` now also triggers on `push` to `main`. That pass has no single  PR, so the script sweeps instead.
- The sweep **advances one PR per pass and stops**. Merging pushes `main`, which  triggers the next pass, so the queue drains serially. Refreshing every PR on  every merge would cost roughly 110 extra CI runs a day at five labelled PRs;
  front-of-line costs about 22.
- PRs are served **oldest first**, so the one that has waited longest goes next.
- A PR that cannot be refreshed no longer strands the queue. `gh pr  update-branch` answers **403** on forks that disallow maintainer edits; that  used to abort the whole sweep and fail the workflow.
- Only **green** PRs are refreshed. A failing PR would otherwise re-run the full  suite on every merge to `main`, forever.
- Push sweeps share one concurrency group so they queue rather than overlap, and  do not cancel — a cancelled sweep can interrupt an in-flight merge.

Paths a machine never merges

```
core/  platform/  gateway/  .github/workflows/  .github/scripts/
```

The first three carry the agent runtime, the multi-tenant platform and the chat gateway. The last two are the merge machinery itself, which must not be able to widen its own permissions unattended. One protected file taints the whole PR; prefixes match on the directory boundary, so `coreutils/` is not `core/`.

### DIFF
--- a/.github/scripts/automerge_pr.py
+++ b/.github/scripts/automerge_pr.py
@@ -10,12 +10,19 @@ import sys
 from typing import Any
 
 AUTOMERGE_LABEL = "automerge"
+#: Process exit code. Every path through ``main`` reports success — a PR that
+#: is skipped is not an error. Real failures raise ``CalledProcessError`` and
+#: exit with the code ``gh`` returned.
+EXIT_SUCCESS = 0
+#: How many protected paths the refusal names before summarising the rest.
+#: Enough to show the reader why without pasting a whole diff into the log.
+PROTECTED_PATHS_LOGGED = 3
+#: Page size ``gh pr view --json files`` returns; a full page may be cut short.
+FILE_PAGE_SIZE = 100
 #: Paths a machine must never merge on its own. The first three carry the
 #: agent runtime, the multi-tenant platform and the chat gateway — a bad
 #: change there reaches every user. The last two are the merge machinery
 #: itself, which must not be able to widen its own permissions.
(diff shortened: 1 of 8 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4677, opened 2026-08-02
Fixes: investigation pipeline resilience (alert classification, tool failure handling, stage rollback)

Eros483:

Three fixes in the investigation pipeline, all in the same failure-resilience domain:
1. **fallback_details defaults to noise** (`core/domain/alerts/extraction.py:151`)
   When LLM-based alert classification fails (network error, auth error, timeout), the
   deterministic fallback now returns `is_noise=True` instead of `is_noise=False`.
   Previously a casual chat message or greeting hitting an LLM error would run through
   all five pipeline stages as a spurious investigation.

2. **execute_tools failure preserves evidence** (`agent.py:324-393`, `loop.py:151-197`)
   The agent loop already handled LLM invoke failures gracefully by returning degraded
   state with all collected evidence intact. Tool execution failures had no such handling
   — an unhandled exception from `execute_tools` or any individual tool result iterator
   would propagate out of `run()` and discard every prior iteration's data. Both points
   are now wrapped with try/except guards that call a new `degraded_investigation_from_tool_failure`
   helper, returning the accumulated evidence, messages, and tool context.

3. **Per-stage state rollback on failure** (`lifecycle.py:19-32`)
   `_run_stage` mutates `AgentState` in-place via `apply_state_updates`. If a later stage
   raised after earlier stages wrote to state, the except handler in `run_connected_investigation`
   saw a partially-corrupt state dict. Each stage now snapshots state with `copy.deepcopy`
   before execution and restores on failure, so only the failing stage's own mutations are
   undone while earlier successful stages' data is untouched.

- A.1 — before: LLM failure on a hello message → full investigation
$ opensre investigate --text "thanks!"  (LLM auth broken)
extract_alert LLM alert extraction failed, using fallback: <error>
Running investigation with 5 stages...
now: falls back to noise and skips investigation
$ opensre investigate --text "thanks!"  (LLM auth broken)
extract_alert LLM alert extraction failed, using fallback: <error>
Message classified as noise - skipping investigation

- A.2 — before: tool raises → all seed evidence lost
now: tool raises → seed evidence preserved in degraded state
root_cause_category: "tool_execution_error"
evidence: {"seed_tool": {"ok": true}}

- A.3 — before: stage 4 raises → state has half-baked fields from stages 1-3
now: stage 4 raises → state rolled back to exactly what stage 3 left it

### Test results

79 passed in 2.31s (tests/tools/investigation/ + tests/core/domain/alerts/)
8 new tests: 2 noise default, 3 tool failure handling, 3 stage rollback
make lint: clean on all changed files

---

- [x] Yes, I used AI assistance via OpenCode

The investigation pipeline had three failure-mode gaps that shared a common root cause: no
defensive handling when a component fails partway through. Each fix follows an existing
pattern already established elsewhere in the codebase:

- **A.1 (`is_noise=True` fallback):** 

`fallback_details` is the deterministic path when the LLM is unavailable. It cannot reliably distinguish noise from an alert — no regex or heuristic can match the LLM's classification accuracy. 
  
Defaulting to noise is the conservative choice: it prevents wasted compute and avoids confusing delivery channels with half-baked investigations. The function already extracts `alert_name` and `severity` from the raw payload best-effort, so the metadata is still available if needed.

- **A.2 (tool failure evidence preservation):** 

The agent loop at line 264 already has a mature failure pattern for LLM invoke errors: catch, classify, call `degraded_investigation_from_llm_failure`, return. Tool execution had no equivalent. 

I added a parallel `degraded_investigation_from_tool_failure` helper in `loop.py` (same
signature shape, different error messaging) and wrapped both the `execute_tools` call
and the per-result `next(fresh_results)` iteration. Two guards because `execute_tools`
could raise before yielding or an individual tool could raise mid-iteration — both must
be caught to protect evidence. I considered catching only at the outer level but that
misses per-tool failures after `execute_tools` has already yielded some results.

- **A.3 (stage rollback):** 

`_run_stage` mutates state in-place. The fix is a snapshot
before each stage, restored on exception. 

I chose `copy.deepcopy` over a key-level delta approach because (a) the investigation state dict is not on a hot path — this runs once per investigation, not per tool call — and (b) deeply nested values
(evidence, agent_messages, masking_map) are passed by reference into downstream code, so
a shallow copy wouldn't protect against internal mutation. 

Each stage only undoes its own changes; earlier successful stages' data persists — this is the correct semantic because a stage 4 failure that deleted stage 1's integration resolution would prevent the caller
from diagnosing the failure.

---

### DIFF
--- a/core/domain/alerts/extraction.py
+++ b/core/domain/alerts/extraction.py
@@ -124,7 +124,12 @@ def needs_full_json_prompt(raw_alert: dict[str, Any]) -> bool:
 
 
 def fallback_details(state: Mapping[str, Any], raw_alert: Any) -> AlertDetails:
-    """Best-effort field extraction when the LLM path is unavailable."""
+    """Best-effort field extraction when the LLM path is unavailable.
+
+    Defaults to noise when LLM extraction fails — the fallback cannot reliably
+    distinguish a casual message from a real alert, and running a full
+    investigation on a failed extraction wastes resources.
+    """
     alert_name = state.get("alert_name", "unknown")
     severity = state.get("severity", "unknown")
 
@@ -148,7 +153,7 @@ def fallback_details(state: Mapping[str, Any], raw_alert: Any) -> AlertDetails:
         )
 
     return AlertDetails(
-        is_noise=False,
+        is_noise=True,
         alert_name=alert_name or "unknown",
         severity=severity or "unknown",
     )
--- a/core/domain/alerts/extraction.py
+++ b/core/domain/alerts/extraction.py
@@ -126,9 +126,9 @@ def needs_full_json_prompt(raw_alert: dict[str, Any]) -> bool:
 def fallback_details(state: Mapping[str, Any], raw_alert: Any) -> AlertDetails:
     """Best-effort field extraction when the LLM path is unavailable.
 
-    Defaults to noise when LLM extraction fails — the fallback cannot reliably
-    distinguish a casual message from a real alert, and running a full
-    investigation on a failed extraction wastes resources.
+    Returns ``is_noise=False`` — when the LLM can't classify, the safe
+    default is to investigate. Callers should flag the degraded state so
+    downstream stages can adjust confidence accordingly.
     """
     alert_name = state.get("alert_name", "unknown")
     severity = state.get("severity", "unknown")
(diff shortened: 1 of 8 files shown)


arnabwithab:
@greptile review

arnabwithab:
@cerencamkiran what's up? any issues?

cerencamkiran:
I closed it because I couldn't find a linked issue.

arnabwithab:
> I closed it because I couldn't find a linked issue.

seemed pretty straightforward to me. will open an issue and link it.

arnabwithab:
fixes #4685

please review @cerencamkiran

PR #4680, opened 2026-08-02
refactor(harness): make the turn path readable from the entrypoint down

YauhenBichel:
Tracing the call path from `main.py` to the ReAct loop, the middle layers had
grown signatures and functions that could only be read by reading their callers.
This makes that path legible without changing behaviour.

**Long parameter lists became objects.** The worst offenders took 13, 12, 12 and
10 parameters, mixing collaborators that live for the whole surface with data
that changes every turn. They now separate:

| new type | holds |
| --- | --- |
| `ActionTurnRunner` | output, tools, deps, error reporter, tool hooks — bound once |
| `ShellActionRunner` | the same, for one shell session |
| `ActionRequest` | one turn's wiring |
| `AnswerRequest` | one answer's wiring |
| `_ShellTurnBindings` | replaces three ad-hoc closures |

`headless_dispatch` already held every one of those fields and restated them on
every call; it now builds the runner once.

**Seams became Protocols.** `StreamAnswerFn`, `EvidenceGatherer` and
`ExecuteActions` were `Callable[..., Any]`, so nothing checked the keywords
callers actually get wrong. They now name their parameters.

**The longest function lost its unrelated jobs.** `_run_action_turn` went from
168 lines to 120, with counting, response composition and rendering extracted
into `_count_turn`, `_compose_response` and `_render_turn_response`.

**Route dispatch is now exhaustive.** `run_turn` chose between three routes with
an `if`/`elif` chain that rebuilt `TurnResult` in each arm and ended in
`raise AssertionError("Unknown route intent")`. It is a `match` with
`assert_never`, one `TurnResult`, and the arms hold only what differs. Adding a
fourth intent now fails type-check — verified by adding one:

```
Argument 1 to "assert_never" has incompatible type "Literal['new_route']"
```

## Two bugs this surfaced

**An architecture guard that could not fail.** `test_agent_shapes` asserted the
action driver was assignable to `ExecuteActions` — but that alias was
`Callable[..., ToolCallingTurnResult]`, so a function with a completely
different signature passed. Demonstrated, then replaced with a test that pins
the keyword contract; removing `turn_plan` from the runner now fails it.

**An import cycle.** Making `StreamAnswerFn` a Protocol pulled `TurnPlan` in
under `TYPE_CHECKING`, and `turn_plan` imports `SessionStore` back from `ports`.
`check-imports` counts typing-only imports, so this was a real 2-cycle.

### DIFF
--- a/core/agent_harness/AGENTS.md
+++ b/core/agent_harness/AGENTS.md
@@ -44,7 +44,7 @@ subpackage. Default port implementations live with the concern they serve, not i
     what the turn knows. Downstream components (e.g.
     `action_driver._resolved_integrations_for_turn`) read it from there rather
     than re-resolving. Do NOT reintroduce per-component integration resolution.
-  - `action_driver.py` — `run_action_agent_turn`: one action tool-calling turn
+  - `action_driver.py` — `ActionTurnRunner`: one action tool-calling turn
     over the ports, via a `_build_action_agent` factory that returns an
     `ActionTurnPlan`.
   - `evidence_driver.py` — bounded evidence-gather loop, via a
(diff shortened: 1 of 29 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4687, opened 2026-08-03
refactor(infra): drop the opensre-infra-aws submodule and settle on one organization-id env var

YauhenBichel:
One name for the organization

The product read a prefixed `OPENSRE_ORGANIZATION_ID` alongside the bare `ORGANIZATION_ID`. Two names for one fact is what caused the original outage: a Fargate silo injected only the bare name, so credits metering, usage attribution
and Slack principal resolution all resolved to an empty string and Slack turns were refused. A dual-name resolver papered over it. Now there is one name.

The infra submodule is gone

`platform/deployment_multi_tenant` pinned `Tracer-Cloud/opensre-infra-aws`, which is archived; its content now lives with the web application. The pin is removed, along with everything that existed only to serve it.

One production path depended on it: the gateway imported `ControlPlaneDbClient` to build its remote-run repository. The architecture was already right — `RemoteRunWorker` depends on the `AgentRunRepository` **Protocol**, which lives
here — so only the implementation had to move. `gateway/storage/agent_runs/` now provides it, mirroring the pooling used by `gateway/storage/investigations/`. The queue semantics stay in SQL: `FOR UPDATE SKIP LOCKED` to claim, and every write re-checks `claimed_by` and the lease so a worker whose lease lapsed cannot finalize a run another worker has taken.

The Protocol also narrowed from five methods to three. `enqueue_agent_run` and `fetch_agent_run_by_id` served the control plane; nothing here calls them.

## Removed

| | why |
| --- | --- |
| submodule, `.gitmodules`, `.git/modules` entry | the pin |
| `tests/deployment/control_plane/`, `tests/platform/deployment_multi_tenant/` | 14 files testing archived code; they never ran in CI, which checks out without submodules |
| `config/constants/opensre_infra_aws.py` | all 14 constants dead — their only consumer lived in the submodule |
| `conftest` skip logic, two `mypy.ini` sections, `cdk-verify`, two workflow references | |

`DEPLOYMENT.md`'s Fargate section now points at `opensre-webapp/opensre-infra-aws/`
rather than being deleted — that deployment still exists, just elsewhere.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
I can't find OPENSRE_ORGANIZATION_ID what greptile mentioned

PR #4598, opened 2026-07-31
fix(remote-sync): add dry-run preview mode

santhiprakash:
Problem
- Closes #4549. `opensre remote-sync sync` could only perform the transfer; there was no preview mode to list planned uploads/downloads without mutating state.

Triage / Root cause
- `platform.filestorage.engine.run_sync()` always called `pull()` / `push()` in write mode, and the CLI / slash adapters had no `--dry-run` flag to request a no-write pass.

Fix
- Added a shared `dry_run` flag through `run_sync()`, `pull()`, `push()`, and `run_remote_sync()`.
- Wired `--dry-run` through `opensre remote-sync sync` and `/remote-sync sync`.
- Updated `format_report_lines()` so preview output reads as a dry run while still listing the planned transfers.
- Added in-memory regression coverage that proves dry-run leaves the fake store untouched and surfaces the new flag on both adapters.

Verification
- `uv run pytest tests/filestorage/test_remote_sync.py -q`
- `uv run pytest tests/cli/test_remote_sync_command.py -q`
- `uv run pytest tests/interactive_shell/test_remote_sync_cmds.py -q`
- `python3 -m py_compile platform/filestorage/engine.py platform/filestorage/messages.py platform/filestorage/operations.py surfaces/cli/commands/remote_sync.py surfaces/interactive_shell/command_registry/remote_sync_cmds.py tests/filestorage/test_remote_sync.py tests/cli/test_remote_sync_command.py tests/interactive_shell/test_remote_sync_cmds.py`

Notes / Risks
- Preview mode still reads local files and remote listings to determine what would happen; it just avoids `put_object()` and atomic local writes.
- The repo's `uv run` emits a harmless `/home/ubuntu/.hermes/.env: line 30: dhbt: command not found` warning before pytest starts.

Model Used
- GPT-5.4-mini

Checklist
- [x] Linked the issue
- [x] Added focused regression tests
- [x] Kept the change scoped to the shared remote-sync path
- [x] Ran targeted verification

PR #4688, opened 2026-08-03
feat(remote-sync): add --dry-run preview to sync

### LINKED ISSUE #4549: [Good first issue] Add --dry-run to opensre remote-sync sync
### Complexity

### Context
There is no way to see what a sync would move before it moves it. `status` only shows whether sync is on and which roots exist — it does not preview transfers. `push()` and `pull()` in `engine.py` already decide upload / download / skip / keep-remote; the work is reporting that without `put_object` or local writes.

### Done when
- `--dry-run` lists what would upload, download, skip, and be kept.
- Nothing changes locally or remotely.
- CLI (`opensre remote-sync sync --dry-run`) and `/remote-sync sync --dry-run` both expose it.
- Implementation lives in operations / engine (surfaces stay thin).
- Covered by an in-memory fake store test (no cloud).

---

muddlebee:
Fixes #4549

`opensre remote-sync sync --dry-run` and `/remote-sync sync --dry-run` now preview what would upload, download, skip, and be kept back — without calling `store.put_object`, `store.get_object`, or writing anything to disk.

**Why this needed care:** `run_sync` does pull-then-push sharing one `SyncReport`. In a real full sync, if the remote has a newer copy of a file that also exists locally, `pull` overwrites the local file, so `push` re-reads it afterward and finds it already matches the remote (`skipped`). A dry run that just skips the writes but lets `push` read the file naively re-derives the wrong answer from the still-stale local bytes, reporting the same key as both `downloaded` *and* `kept_remote` — a self-contradictory preview. This is exactly the P1 Greptile caught on the first attempt at this issue (#4598, closed after CI also failed).

**Fix:** `pull` and `push` already share one `SyncReport` instance when `run_sync` runs both directions. Under `dry_run`, `push` builds `previewed_pulls = set(result.downloaded)` once and, for any key pull already claimed, trusts that decision (`skipped`) instead of re-comparing against on-disk bytes it never wrote. `pull` itself just appends to `result.downloaded`/`downloaded_bytes` (using the listing's `size`, no `get_object` call) and returns — no network read, no disk write, beyond the one `list_objects` call a normal sync already makes.

This is a smaller footprint than the closed attempt, which fetched the actual bytes during a dry-run pull (an extra `store.get_object()` per previewed file) purely to feed push's content-hash comparison — unnecessary in a mode whose whole point is "touch nothing," and it also had a latent gap on multipart ETags (`comparable_etag` returns `""`, so push falls back to an mtime comparison that could disagree with a real sync's outcome). Trusting pull's decision outright sidesteps that instead of half-fixing it.

`--push-only --dry-run` (no pull ever runs) is unaffected — `previewed_pulls` is empty, push just compares against real on-disk content as it does today, with `put_object` suppressed.

```
$ uv run pytest tests/filestorage/test_remote_sync.py tests/filestorage/test_service_and_registry.py \
    tests/cli/test_remote_sync_command.py tests/cli/test_gateway_remote_sync.py \
    tests/interactive_shell/test_remote_sync_cmds.py -q
219 passed
```

Full `make test-scope` (escalated to cover `tests/cli/` + `tests/interactive_shell/` since this touches the shared engine plus both surfaces): `2436 passed, 4 skipped`. Also ran `make lint`, `make format-check`, `make typecheck`, `make verify-integrations-smoke` — all clean.

New/changed coverage:
- `test_dry_run_preview_matches_what_a_real_full_sync_would_settle_on` — the exact contradictory-preview scenario, asserting the key lands only in `downloaded`, never `uploaded`/`kept_remote`, and nothing on disk or in the store changes.
- `test_dry_run_writes_nothing_locally_or_remotely` — full two-way sync (one remote-only, one local-only file), store and disk untouched.
- `test_push_only_dry_run_does_not_upload`
- `format_report_lines(..., dry_run=True)` wording test.
- CLI (`--dry-run` flag forwarding, `--help` documents it) and slash-command (`/remote-sync sync --dry-run` forwarding + gateway dispatch) coverage.

---

I read the issue and the previously closed attempt (#4598) to understand what tripped it up: CI failed outright, and separately Greptile flagged a real bug — the dry-run preview could report a file as both downloaded and kept-back for the same key, because push re-derived its decision from a local file that a dry run correctly declines to write. Rather than reuse #4598's fix (fetch the real bytes during dry-run pull and feed them into push's normal content-hash comparison), I used the fact that `pull` and `push` already share one `SyncReport` object when `run_sync` calls both directions — so push can just check "did pull already resolve this key" and, if so, defer to that instead of re-evaluating. This needed no new fields on `SyncReport`, no extra parameter threaded through `push`, and no extra `store.get_object()` calls during a dry-run pull (a real sync's pull already does that GET; the dry-run path deliberately skips it and uses the object size from the listing instead). I verified the fix against the specific "newer remote, existing local file" scenario that broke the previous attempt, then threaded `dry_run` through `operations.run_remote_sync`, `messages.format_report_lines` (wording only), the CLI command, and the slash command — plus found and fixed a test at `tests/cli/test_gateway_remote_sync.py` whose fake `run_remote_sync` needed the new keyword, which `make test-scope`'s escalated run caught.

---

---

### DIFF
--- a/platform/filestorage/engine.py
+++ b/platform/filestorage/engine.py
@@ -128,8 +128,16 @@ def push(
     report: SyncReport | None = None,
     remote: list[RemoteObject] | None = None,
     exclusions: ExclusionRules = NO_EXCLUSIONS,
+    dry_run: bool = False,
 ) -> SyncReport:
-    """Upload local files whose contents differ from the bucket."""
+    """Upload local files whose contents differ from the bucket.
+
+    Under ``dry_run`` nothing is sent, and a key ``pull`` already previewed as
+    a download (``result.downloaded``, when both share one report) is trusted
+    rather than re-compared against the still-unwritten local file — a real
+    full sync would have overwritten it first, so re-reading stale bytes here
+    would report it as both downloaded and kept back.
+    """
     roots = roots if roots is not None else syncable_roots()
     result = report if report is not None else SyncReport()
     listing = remote if remote is not None else store.list_objects("")
@@ -137,6 +145,8 @@ def push(
     # Resolve each root once rather than per file: this loop touches every
     # session and memory file on the machine.
     allowed = resolved_roots(roots)
+    # Built once, not per file: membership in a loop needs a set, not a list.
+    previewed_pulls = set(result.downloaded) if dry_run else frozenset[str]()
 
     # Check every candidate before uploading any of them. A denied file found
     # halfway through must not leave earlier files already in the store.
@@ -156,6 +166,9 @@ def push(
             planned.append((key, path))
 
     for key, path in planned:
+        if key in previewed_pulls:
+            result.skipped += 1
+            continue
         data = path.read_bytes()
         existing = by_key.get(key)
         if existing is not None:
@@ -167,7 +180,8 @@ def push(
                 # destroy it. Same rule pull applies in the other direction.
                 result.kept_remote.append(key)
                 continue
-        store.put_object(key, data)
+        if not dry_run:
+            store.put_object(key, data)
         result.uploaded.append(key)
         result.uploaded_bytes += len(data)
     return result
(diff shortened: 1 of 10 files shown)


muddlebee:
@greptile review

PR #4689, opened 2026-08-03

### LINKED ISSUE #4482: [BUG] Support using separate AWS sessions for infrastructure investigation and Bedrock model invocation
### Summary

 **Current behaviour**

**Expected behaviour**

### Steps to reproduce

**Question**

Yes

### How often does it occur?

Every time

### Operating system

macOS

### Scenario

I have two AWS accounts:

```
```

Example:

```
```

```
```

Example:

```
```

## Use case

arnavpopley:
Fixes #4482

Bedrock and infrastructure investigation (EC2, CloudWatch, ...) both resolved AWS identity from the same ambient default credential chain, with no way to give them different identities. Investigation tools already work correctly cross-account because they read the process's single ambient `AWS_PROFILE` directly. Bedrock had no equivalent override, so a laptop configured for a Bedrock account reachable only via an AssumeRole profile in a different account than the infra being investigated had no way to give Bedrock a different identity than whatever profile was active for everything else, exactly the symptom in the issue's repro.

Adds a Bedrock-only identity override, isolated from investigation's ambient chain: `BEDROCK_AWS_PROFILE` (named profile), `BEDROCK_AWS_ROLE_ARN` + `BEDROCK_AWS_EXTERNAL_ID` (explicit assume-role, for deployments with no `~/.aws/config` file), and `BEDROCK_AWS_REGION`. Naming mirrors the existing `AWS_ROLE_ARN`/`AWS_EXTERNAL_ID` convention in `config/constants/aws.py`.

A new `resolve_bedrock_aws_session()` in `core/llm/transports/sdk/bedrock_converse.py` builds a `boto3.Session` from the override (role-arn takes precedence) or returns `None` when neither is set, so callers fall through to today's exact ambient-chain behavior. `frozen_bedrock_credentials()` extracts concrete key material for `AnthropicBedrock`'s constructor kwargs (confirmed it accepts `aws_access_key`/`aws_secret_key`/`aws_session_token` directly), raising a clear error instead of an `AttributeError` if a session can't resolve credentials.

Wired into all three Bedrock client construction sites: `BedrockAgentClient` and `BedrockConverseAgentClient` in `agent_clients.py`, and `BedrockLLMClient` in `llm_clients.py`. `BedrockLLMClient`'s existing permissive region default (falls back to `us-east-1` rather than raising) was deliberately preserved.

Out of scope: the `OPENSRE_LLM_TRANSPORT=litellm` Bedrock path, and investigation-tool credential resolution (already correctly isolated via ambient `AWS_PROFILE`).

```
355 passed

$ uv run python -m ruff check config/ core/ docs/ tests/core/runtime/llm/ && uv run python -m ruff format --check config/ core/ tests/core/runtime/llm/
All checks passed! / 313 files already formatted

1588 passed, 29 skipped
```

14 new tests: 9 unit tests on the new session/credential-resolution helpers (profile path, assume-role path, role-arn-over-profile precedence, missing-ExternalId, unresolved-credentials error), and 5 wiring tests proving each of the three Bedrock client constructors actually uses the resolved session over the ambient default.

---

Traced every AWS/boto3 client construction site in the repo before writing anything, to confirm the root cause: both Bedrock and investigation tools call plain `boto3.client(...)`/`AnthropicBedrock(...)` against the single ambient default credential chain, with investigation's cross-account support coming entirely from the OS-level `AWS_PROFILE` env var rather than any application code. That meant the fix only needed to touch Bedrock's side. Checked `AnthropicBedrock`'s actual constructor signature before assuming it could accept explicit credentials. Found `integrations/aws/verifier.py::_build_sts_client` already implements the identical assume-role pattern for the AWS integration's verify probe, so the new assume-role branch mirrors its exact shape instead of inventing new semantics. Unified `AnthropicBedrock` (discrete key kwargs) and the raw `boto3.client("bedrock-runtime", ...)` Converse path (wants a Session) behind one `boto3.Session`-returning function. While touching `BedrockLLMClient`, noticed it silently defaults to `us-east-1` unlike the other two Bedrock clients, which raise; deliberately preserved that existing asymmetry instead of unifying it as a side effect of an unrelated credential fix.

---

### DIFF
--- a/config/constants/aws.py
+++ b/config/constants/aws.py
@@ -9,6 +9,19 @@ AWS_ACCESS_KEY_ID_ENV = "AWS_ACCESS_KEY_ID"
 AWS_SECRET_ACCESS_KEY_ENV = "AWS_SECRET_ACCESS_KEY"
 AWS_SESSION_TOKEN_ENV = "AWS_SESSION_TOKEN"
 
+# Bedrock-only identity override, isolated from the vars above. Investigation
+# tools (EC2, CloudWatch, ...) already read the ambient AWS_PROFILE directly,
+# so a laptop configured for one AWS account already investigates that
+# account correctly. Bedrock had no equivalent override, so a cross-account
+# setup (Bedrock reachable only via an AssumeRole profile in a different
+# account than the infra being investigated) could not give Bedrock a
+# different identity than whatever profile was active for everything else.
+# See #4482.
+BEDROCK_AWS_PROFILE_ENV = "BEDROCK_AWS_PROFILE"
+BEDROCK_AWS_ROLE_ARN_ENV = "BEDROCK_AWS_ROLE_ARN"
+BEDROCK_AWS_EXTERNAL_ID_ENV = "BEDROCK_AWS_EXTERNAL_ID"
+BEDROCK_AWS_REGION_ENV = "BEDROCK_AWS_REGION"
+
 __all__ = [
     "AWS_ACCESS_KEY_ID_ENV",
     "AWS_EXTERNAL_ID_ENV",
@@ -16,4 +29,8 @@ __all__ = [
     "AWS_ROLE_ARN_ENV",
     "AWS_SECRET_ACCESS_KEY_ENV",
     "AWS_SESSION_TOKEN_ENV",
+    "BEDROCK_AWS_EXTERNAL_ID_ENV",
+    "BEDROCK_AWS_PROFILE_ENV",
+    "BEDROCK_AWS_REGION_ENV",
+    "BEDROCK_AWS_ROLE_ARN_ENV",
 ]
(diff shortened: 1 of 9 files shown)


arnavpopley:
Closing to reopen after fixing three findings from Greptile's review: frozen_bedrock_credentials() was permanently freezing a BEDROCK_AWS_PROFILE session's credentials even when the underlying profile was a refreshable role_arn + source_profile pair, so a long-lived cached client would start failing once the snapshot expired; the fix passes the profile straight through as aws_profile= instead, preserving AnthropicBedrock's own refresh (BEDROCK_AWS_ROLE_ARN remains a documented one-time snapshot since AnthropicBedrock has no hook for a refreshable provider at all). resolve_bedrock_aws_session() also called the strict require_aws_region() internally, so BEDROCK_AWS_ROLE_ARN with no region env var would reject a config that BedrockLLMClient's own permissive us-east-1 default should have accepted; region is now a parameter each caller resolves itself. And BedrockLLMClient's region resolution used nested os.getenv(name, default) calls, so BEDROCK_AWS_REGION="" (present but empty) would short-circuit instead of falling through to AWS_REGION/AWS_DEFAULT_REGION/us-east-1.

PR #4690, opened 2026-08-03

### LINKED ISSUE #4482: [BUG] Support using separate AWS sessions for infrastructure investigation and Bedrock model invocation

arnavpopley:
Fixes #4482

Bedrock and infrastructure investigation (EC2, CloudWatch, ...) both resolved AWS identity from the same ambient default credential chain, with no way to give them different identities. Investigation tools already work correctly cross-account because they read the process's single ambient `AWS_PROFILE` directly. Bedrock had no equivalent override, so a laptop configured for a Bedrock account reachable only via an AssumeRole profile in a different account than the infra being investigated had no way to give Bedrock a different identity than whatever profile was active for everything else, exactly the symptom in the issue's repro.

Adds a Bedrock-only identity override, isolated from investigation's ambient chain: `BEDROCK_AWS_PROFILE` (named profile), `BEDROCK_AWS_ROLE_ARN` + `BEDROCK_AWS_EXTERNAL_ID` (explicit assume-role, for deployments with no `~/.aws/config` file), and `BEDROCK_AWS_REGION`. Naming mirrors the existing `AWS_ROLE_ARN`/`AWS_EXTERNAL_ID` convention in `config/constants/aws.py`. Wired into all three Bedrock client construction sites in `core/llm/transports/sdk/`.

**Update 1:** closed after Greptile found three issues: (1) freezing a `BEDROCK_AWS_PROFILE` session's credentials at construction time discarded whatever refresh the underlying profile supported, so a long-lived cached `AnthropicBedrock` client would fail once the snapshot expired, fixed by passing the profile straight through as `aws_profile=` instead (`AnthropicBedrock` resolves and refreshes it internally); `BEDROCK_AWS_ROLE_ARN` stays a documented one-time snapshot since `AnthropicBedrock` has no hook for a refreshable provider at all; (2) the session resolver called a strict region-requiring helper internally, so `BEDROCK_AWS_ROLE_ARN` with no region env var would reject a config `BedrockLLMClient`'s own permissive `us-east-1` default should have accepted, fixed by making region a parameter each caller resolves itself; (3) `BedrockLLMClient`'s region resolution used nested `os.getenv(name, default)`, so `BEDROCK_AWS_REGION=""` would short-circuit instead of falling through, fixed with the same `or`-chain `require_aws_region()` already used.

I also did an independent deeper self-review afterward, including verifying against the actual installed `anthropic` SDK source that `aws_profile=` genuinely re-resolves credentials on every request, not just at construction. That pass found one minor issue: the STS credentials helper was typed `dict[str, str]`, but the real response also includes a non-string `Expiration` field; harmless today since only the three string fields are read, but corrected to `dict[str, Any]`.

```
359 passed

$ uv run python -m ruff check core/llm/ docs/ tests/core/runtime/llm/ && uv run python -m ruff format --check core/llm/ tests/core/runtime/llm/
All checks passed! / 63 files already formatted

1592 passed, 29 skipped
```

Confirmed the two new regression tests (region-conflict, empty-string region) fail against the pre-fix code and pass after.

---

Checked `AnthropicBedrock`'s actual constructor signature: it accepts `aws_profile` directly, so passing a named profile straight through uses the SDK's own supported refresh mechanism instead of resolving a `boto3.Session` and freezing static credentials out of it. Split credential resolution into two functions: `resolve_bedrock_aws_session()` for `boto3.client()` consumers, `resolve_bedrock_anthropic_kwargs()` for `AnthropicBedrock` consumers, since only the former can accept a refreshable session. For `BEDROCK_AWS_ROLE_ARN`, considered building a self-refreshing session via botocore's `AssumeRoleCredentialFetcher`, but this codebase has no precedent for that pattern and `AnthropicBedrock` can't use a refreshable provider regardless, so documented the one-time-snapshot limitation instead of adding new refresh infrastructure. For the region-conflict finding, moved region resolution to a parameter each caller passes in, removing the second, conflicting policy that lived inside the shared helper. For the empty-string finding, reused the `or`-chain pattern `require_aws_region()` already used correctly, instead of writing new logic.

---

### DIFF
--- a/config/constants/aws.py
+++ b/config/constants/aws.py
@@ -9,6 +9,19 @@ AWS_ACCESS_KEY_ID_ENV = "AWS_ACCESS_KEY_ID"
 AWS_SECRET_ACCESS_KEY_ENV = "AWS_SECRET_ACCESS_KEY"
 AWS_SESSION_TOKEN_ENV = "AWS_SESSION_TOKEN"
 
+# Bedrock-only identity override, isolated from the vars above. Investigation
+# tools (EC2, CloudWatch, ...) already read the ambient AWS_PROFILE directly,
+# so a laptop configured for one AWS account already investigates that
+# account correctly. Bedrock had no equivalent override, so a cross-account
+# setup (Bedrock reachable only via an AssumeRole profile in a different
+# account than the infra being investigated) could not give Bedrock a
+# different identity than whatever profile was active for everything else.
+# See #4482.
+BEDROCK_AWS_PROFILE_ENV = "BEDROCK_AWS_PROFILE"
+BEDROCK_AWS_ROLE_ARN_ENV = "BEDROCK_AWS_ROLE_ARN"
+BEDROCK_AWS_EXTERNAL_ID_ENV = "BEDROCK_AWS_EXTERNAL_ID"
+BEDROCK_AWS_REGION_ENV = "BEDROCK_AWS_REGION"
+
 __all__ = [
     "AWS_ACCESS_KEY_ID_ENV",
     "AWS_EXTERNAL_ID_ENV",
@@ -16,4 +29,8 @@ __all__ = [
     "AWS_ROLE_ARN_ENV",
     "AWS_SECRET_ACCESS_KEY_ENV",
     "AWS_SESSION_TOKEN_ENV",
+    "BEDROCK_AWS_EXTERNAL_ID_ENV",
+    "BEDROCK_AWS_PROFILE_ENV",
+    "BEDROCK_AWS_REGION_ENV",
+    "BEDROCK_AWS_ROLE_ARN_ENV",
 ]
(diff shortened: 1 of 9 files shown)


arnavpopley:
Closing to reopen after fixing two more findings from Greptile: BEDROCK_AWS_ROLE_ARN's one-time sts:AssumeRole snapshot was frozen into every boto3-backed Bedrock client (Converse/non-Anthropic models), so a long-lived cached client failed once the assumed role's session duration elapsed. resolve_bedrock_aws_session() now builds the role-arn session via AssumeRoleCredentialFetcher + DeferredRefreshableCredentials (the same mechanism boto3 uses internally for a role_arn + source_profile profile), so it re-assumes the role on demand instead of going stale. This only covers the boto3-backed clients: AnthropicBedrock (Claude models) has no hook for a refreshable credentials provider at all, so BEDROCK_AWS_ROLE_ARN there remains a documented one-time snapshot with BEDROCK_AWS_PROFILE recommended instead. Also fixed BedrockLLMClient's region resolution not stripping whitespace, unlike require_aws_region(), so a shell-provided BEDROCK_AWS_REGION with surrounding spaces would reach the SDK verbatim and fail client initialization.

PR #4692, opened 2026-08-03

### LINKED ISSUE #4482: [BUG] Support using separate AWS sessions for infrastructure investigation and Bedrock model invocation

arnavpopley:
Fixes #4482

Bedrock and infrastructure investigation (EC2, CloudWatch, ...) both resolved AWS identity from the same ambient default credential chain, with no way to give them different identities. Investigation tools already follow the process's ambient `AWS_PROFILE` correctly cross-account; Bedrock had no equivalent, so a Bedrock account reachable only via a different AssumeRole profile than the infra account had no way to be isolated.

Adds a Bedrock-only identity override: `BEDROCK_AWS_PROFILE`, `BEDROCK_AWS_ROLE_ARN` + `BEDROCK_AWS_EXTERNAL_ID` (assume-role, for deployments with no `~/.aws/config`), and `BEDROCK_AWS_REGION`. Naming mirrors the existing `AWS_ROLE_ARN`/`AWS_EXTERNAL_ID` convention in `config/constants/aws.py`. Wired into all three Bedrock client construction sites in `core/llm/transports/sdk/`.

**Two rounds of Greptile fixes since opening:** (1) a `BEDROCK_AWS_PROFILE` session's credentials were frozen at construction instead of passed through as `aws_profile=`, discarding the profile's own refresh; (2) the session resolver required a region internally, rejecting configs `BedrockLLMClient`'s permissive `us-east-1` default should accept, fixed by making region a caller-supplied parameter; (3) `BedrockLLMClient`'s `os.getenv(name, default)` didn't treat `BEDROCK_AWS_REGION=""` as unset; (4) `BEDROCK_AWS_ROLE_ARN` credentials for the `boto3`-backed clients (Converse/non-Anthropic models) were a one-time snapshot that went stale on a long-lived cached client, fixed using botocore's public `AssumeRoleCredentialFetcher`/`DeferredRefreshableCredentials` (the same mechanism boto3 uses for a `role_arn` + `source_profile` profile) so that path now self-refreshes; `AnthropicBedrock` (Claude models) still has no hook for a refreshable provider at all, so it remains a documented one-time snapshot with `BEDROCK_AWS_PROFILE` recommended instead; (5) region resolution wasn't stripping whitespace.

```
361 passed

1594 passed, 29 skipped
```

Every regression test confirmed failing against the pre-fix code and passing after. The self-refresh fix is proven by asserting the role-arn session's credentials are a `DeferredRefreshableCredentials` instance rather than a static snapshot.

---

Checked `AnthropicBedrock`'s actual constructor signature before assuming it could refresh a profile: it accepts `aws_profile` directly, using the SDK's own mechanism rather than freezing static credentials out of a resolved session. Split credential resolution into two functions since only `boto3.client()` consumers can accept a refreshable session. For the `BEDROCK_AWS_ROLE_ARN` refresh finding, verified via `inspect.signature`/botocore's source that `AssumeRoleCredentialFetcher`/`DeferredRefreshableCredentials` are real, public classes before wiring them up, and tested by asserting the resulting object's type rather than reproducing botocore's own expiry-timing logic.

---

### DIFF
--- a/config/constants/aws.py
+++ b/config/constants/aws.py
@@ -9,6 +9,19 @@ AWS_ACCESS_KEY_ID_ENV = "AWS_ACCESS_KEY_ID"
 AWS_SECRET_ACCESS_KEY_ENV = "AWS_SECRET_ACCESS_KEY"
 AWS_SESSION_TOKEN_ENV = "AWS_SESSION_TOKEN"
 
+# Bedrock-only identity override, isolated from the vars above. Investigation
+# tools (EC2, CloudWatch, ...) already read the ambient AWS_PROFILE directly,
+# so a laptop configured for one AWS account already investigates that
+# account correctly. Bedrock had no equivalent override, so a cross-account
+# setup (Bedrock reachable only via an AssumeRole profile in a different
+# account than the infra being investigated) could not give Bedrock a
+# different identity than whatever profile was active for everything else.
+# See #4482.
+BEDROCK_AWS_PROFILE_ENV = "BEDROCK_AWS_PROFILE"
+BEDROCK_AWS_ROLE_ARN_ENV = "BEDROCK_AWS_ROLE_ARN"
+BEDROCK_AWS_EXTERNAL_ID_ENV = "BEDROCK_AWS_EXTERNAL_ID"
+BEDROCK_AWS_REGION_ENV = "BEDROCK_AWS_REGION"
+
 __all__ = [
     "AWS_ACCESS_KEY_ID_ENV",
     "AWS_EXTERNAL_ID_ENV",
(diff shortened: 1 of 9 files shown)


arnavpopley:
Closing to open a fresh PR after fully fixing the repeated Greptile P1 finding (static role credentials expire for the AnthropicBedrock/BEDROCK_AWS_ROLE_ARN path). Previous rounds documented this as an accepted limitation; this round adds a _RefreshingBedrockAnthropicClient that re-assumes the role in place before each request when credentials are near expiry, so Claude models on Bedrock now self-refresh the same way non-Claude models and BEDROCK_AWS_PROFILE already did. New PR to follow.

PR #4696, opened 2026-08-04
fix(reporting): render WhatsApp and SMS links as plain text

### LINKED ISSUE #4695: [BUG] WhatsApp/SMS report renders Slack link syntax instead of plain text
### Summary

The WhatsApp RCA report renders evidence citations and S3 trace links as raw Slack link markup (`<url|label>`) instead of plain text. SMS is affected too, because `ReportMessages.sms_text` returns `whatsapp_text`.

**Expected:** WhatsApp and SMS are plain-text channels, so a cited evidence link should read as text with a tappable URL — both platforms auto-linkify bare URLs.

```
• Pool saturation began at 14:02 UTC. [E1 (https://grafana.example.com/d/abc/db?viewPanel=7)]
```

**Actual:** the reader sees Slack's link syntax verbatim.

```
• Pool saturation began at 14:02 UTC. [<https://grafana.example.com/d/abc/db?viewPanel=7|E1>]
```

**Cause.** `format_whatsapp_message` (`tools/investigation/reporting/formatters/report.py`) builds its body from Slack-flavoured helpers, two of which emit `<url|label>` via `format_slack_link`:

| Helper | Line | What leaks |
|---|---|---|
| `_render_claim_lines` | `report.py:114` | evidence citations `[E1]` |
| `build_investigation_trace` | `infrastructure.py:288,294` | "Input data inspected" / "Audit trail found" S3 console links |

Telegram avoids this with a dedicated `_render_claim_lines_telegram` that calls `format_html_link` (`report.py:191`). WhatsApp is the only channel with no plain-text path — the same shape as the Telegram sanitizer gap fixed in #3249.

### Steps to reproduce

```python
from config.platform_bootstrap import ensure_project_platform_package

ensure_project_platform_package()

from tools.investigation.reporting.formatters.messages import build_report_messages

ctx = {
    "severity": "critical",
    "root_cause": "Connection pool exhausted on orders-db.",
    "validated_claims": [
        {"claim": "Pool saturation began at 14:02 UTC.", "evidence_ids": ["e1"]},
    ],
    "evidence_catalog": {
        "e1": {"display_id": "E1", "url": "https://grafana.example.com/d/abc/db?viewPanel=7"},
    },
}

print(build_report_messages(ctx).whatsapp_text)
```

Output on `main` @ `23d5deac`:

```
[CRITICAL] OpenSRE Investigation

Connection pool exhausted on orders-db

*Findings*
• Pool saturation began at 14:02 UTC. [<https://grafana.example.com/d/abc/db?viewPanel=7|E1>]
```

The Telegram body from the same context renders `<a href="…">E1</a>` correctly, and Slack renders `<url|E1>` correctly — WhatsApp/SMS is the odd one out.

Yes

### How often does it occur?

Every time — any investigation whose cited evidence carries a URL, or whose trace includes an S3 object or audit trail.

### Operating system

macOS

### Logs and error output

_No response_

### Additional context

Suggested fix: convert at the WhatsApp boundary rather than per-helper, so citations, trace links, and any Slack link added upstream later are all covered. `renderers/terminal.py:254` already does exactly this for plain-text terminal output, and its `_strip_slack_links` (`terminal.py:52`) already implements the `label (url)` conversion — so the inverse of `format_slack_link` can move next to it in `formatters/base.py` and be shared instead of duplicated.

I'd like to take this one. Could a maintainer assign it to me? I'll open a linked PR with a regression test once assigned.

cloudenochcsis:
Fixes #4695

WhatsApp and SMS recipients were shown Slack's link markup verbatim — `[<https://grafana…|E1>]` instead of a readable citation. `format_whatsapp_message` builds its body from Slack-flavoured helpers, and two of them emit `<url|label>` through `format_slack_link`:

| Helper | Line | What leaked |
|---|---|---|
| `_render_claim_lines` | `report.py:114` | evidence citations `[E1]` |
| `build_investigation_trace` | `infrastructure.py:288,294` | "Input data inspected" / "Audit trail found" S3 console links |

Telegram already dodges this with its own `_render_claim_lines_telegram` calling `format_html_link`. WhatsApp was the only channel with no plain-text path, and `ReportMessages.sms_text` returns `whatsapp_text`, so SMS inherited the bug.

The conversion happens at the WhatsApp boundary rather than per-helper. That covers citations, trace links, and any Slack link a future change adds upstream, in one place — the same approach `_render_plain_report` already takes for terminal output (`renderers/terminal.py:242`).

`label (url)` keeps the URL rather than dropping it, because WhatsApp and SMS auto-linkify bare URLs — a bare `E1` would lose the link entirely. That form was already the repo's convention, pinned by the existing terminal-renderer tests.

**On the shared helper.** The `label (url)` conversion already existed as `_strip_slack_links` in `renderers/terminal.py`, byte-identical to what WhatsApp needs. Rather than add a second copy, it moves next to its inverse `format_slack_link` in `formatters/base.py` and both callers use it. `_rich_line_with_links` keeps using the same regex, now imported. The existing parametrized cases in `test_terminal_renderer.py` follow the rename and become coverage for the shared helper.

Not touched: `integrations/telegram/markdown.py:16` holds a third copy of that regex, but it drives an HTML conversion rather than plain text, so folding it in would change Telegram behavior. Left for a separate change.

Same `ReportContext` in both runs (the reproduction from #4695, plus a remediation step).

**Before**

```
[CRITICAL] OpenSRE Investigation

Connection pool exhausted on orders-db

*Findings*
• Pool saturation began at 14:02 UTC. [<https://grafana.example.com/d/abc/db?viewPanel=7|E1>]

*Recommended Actions*
• Raise max_connections to 200.
```

**After**

```
[CRITICAL] OpenSRE Investigation

Connection pool exhausted on orders-db

*Findings*
• Pool saturation began at 14:02 UTC. [E1 (https://grafana.example.com/d/abc/db?viewPanel=7)]

*Recommended Actions*
• Raise max_connections to 200.
```

Slack and Telegram are byte-identical before and after — `test_slack_and_telegram_keep_their_own_link_syntax` guards that, and it passed against unpatched `main` as well as after the fix.

Local checks, per `CI.md`. `tools/investigation/` is on the §3 escalation list, so this ran the full suite rather than scoped tests:

```
$ make lint
All checks passed!

$ make format-check
3105 files already formatted

$ make typecheck
Success: no issues found in 1577 source files

$ make test-cov
========== 13431 passed, 59 skipped, 31 warnings in 145.12s (0:02:25) ==========

$ make verify-integrations-smoke
============================== 31 passed in 1.08s ==============================
```

---

Tooling: Claude Code (`claude-opus-5`). I found and reproduced the bug, chose the boundary-conversion approach over the alternatives below, and ran the full harness locally.

The problem is that `format_whatsapp_message` reuses helpers written for Slack. Slack's `<url|label>` is meaningless on WhatsApp, so the reader gets punctuation soup where a citation should be. It is the same failure mode as #3249 — one of several sibling callers missing the channel-specific path.

I considered three fixes. The first was threading a `link_fn` parameter through `_render_claim_lines`, defaulting to `format_slack_link`, with WhatsApp passing a plain variant. That mirrors `evidence.py:22`, which already takes exactly such a parameter, so it fits the file's grain — but it only fixes evidence citations and leaves the S3 trace links broken, since `build_investigation_trace` is shared with Slack and Telegram and would need the same treatment. The second was a WhatsApp-specific claim-lines function, which is the smallest diff but adds a third near-duplicate of the same loop to a file that already has two, and still misses the trace.

I chose the third: convert once on the assembled body. One call covers both leak sites and anything similar added later, and there is precedent for it in this exact codebase — `_render_plain_report` already runs the Slack body through the same conversion for plain terminal output. The trade-off is that it is a text pass over the finished string rather than structured formatting, so it depends on the link syntax being unambiguous. It is: the regex only matches `<` followed immediately by `http://` or `https://`, so ordinary prose containing angle brackets is untouched.

Key components. `SLACK_LINK_RE` is the pattern for Slack's two link forms, `<url|label>` and bare `<url>`. `slack_links_to_plain_text` substitutes each match with `label (url)`, or just `url` when the link carries no label — the capture group is optional, so `m.group(2)` is `None` in that case and the branch is real, not defensive. `format_whatsapp_message` calls it on its return value.

Edge cases I tested: a link with no label (returns the bare URL), several links in one string (each rewritten independently), a bare URL with no angle brackets (passed through untouched — this helper is not a linkifier), a string with no links, and the empty string. Those cases exist in `test_terminal_renderer.py` and now exercise the shared helper. I also added a guard test asserting Slack and Telegram bodies still carry their own syntax, so a future change cannot quietly apply the plain-text pass to the wrong channel.

One thing I checked deliberately: WhatsApp does support `*bold*`, so the conversion touches links only and leaves the `*Findings*` / `*Provenance*` section markers alone. Stripping mrkdwn here would have been a regression.

---

---

## Post-Deploy Monitoring & Validation

No new runtime surface, external call, or configuration. The observable change is the text of outbound WhatsApp/SMS report bodies. Validation signal: a delivered WhatsApp report whose citation reads `E1 (https://…)` with a tappable URL and no `<`/`|` markup. Failure signal would be link text disappearing from a channel body — Slack `<url|label>` or Telegram `<a href>` going missing — which `test_slack_and_telegram_keep_their_own_link_syntax` would have caught first. Rollback is reverting one commit; no data migration or state involved.


tools/investigation/reporting/formatters/report.py
```diff
@@ -723,7 +724,8 @@ def format_whatsapp_message(ctx: ReportContext) -> str:
     if meta_bits:
         parts.append(" | ".join(meta_bits))
 
-    return "\n\n".join(p for p in parts if p)
+    # Shared helpers emit Slack <url|label> links; WhatsApp and SMS render no markup.
+    return slack_links_to_plain_text("\n\n".join(p for p in parts if p))
```
Davidson3556, line 721: Links are fixed, but `*Findings*` still ships as-is. WhatsApp renders that bold; SMS shows literal asterisks, since `sms_text` returns this string unchanged. Keeping mrkdwn is right for WhatsApp, so the real fix is SMS getting its own body rather than aliasing — out of scope for #4695, worth its own issue.

Worth softening the new test file's docstring too: "WhatsApp and SMS are plain-text channels: no Slack link markup may reach them" reads as a broader claim than the code makes.

Davidson3556 (commented):
Great job — bug is real, and the boundary conversion was the right pick of the three; it's the only one that catches the trace links too.

One follow-up inline: `*Findings*` still reaches SMS as literal asterisks. Separate issue. Branch is behind main.

Davidson3556:
please add a screenshot and a demo video

muddlebee:
thx but not a priority atm.

PR #4715, opened 2026-08-04
refactor(prompts): assemble every prompt in tiered layers and make scheduled deliveries work

YauhenBichel:
Gives prompt assembly the three-layer structure from [Hermes](https://hermes-agent.nousresearch.com/docs/developer-guide/prompt-assembly), plus a fourth tier deliberately outside the cached prefix.

| Tier | Changes when | Cached |
| --- | --- | --- |
| `STABLE` | release or config | yes |
| `CONTEXT` | project or session | yes |
| `VOLATILE` | explicit rebuild | yes |
| `EPHEMERAL` | every turn | **no** |

`PromptTier` and `PromptBlockKind` are `StrEnum`s. `kind` says what a block *is*; `tier` says *when it changes*, which is what decides cacheability. All three builders — action, gather, assistant — return a `PromptEnvelope` and
declare a tier per block. `_render()` sorts by tier, so layer order is a property of the envelope, not of the order a builder appends in.

Measured before the change, two consecutive turns:

| | |
| --- | --- |
| Shared byte prefix | 50,422 chars (~12,605 tokens) |
| Prefix diverges at | **98.4%** |

`cached_system()` puts one `cache_control` marker at the end of the whole system string, so a tail that changes every turn invalidated everything before it. 98.4% identical, 0% cached. After: the cached half is byte-identical across turns on both wired paths — action ~31.9k chars, assistant 7,308.

Scheduled deliveries now work end to end

`platform/scheduler/` existed and no prompt mentioned it — zero hits for `schedul`, `cron`, `recurring` across the prompt package. Verified live: playbook loads → briefing composed from real data → canonical offer → `yes`
→ `/cron add --kind daily_summary --cron '0 8 * * 1-5' …` → task created.

Confirmation goes through a structured `PendingScheduleOffer`, not prose scraping. The offer is consumed only after the command succeeds, so a rejected `cron add` no longer burns it.

Also fixed, all found by hand-testing

- **`propose_scheduled_delivery` was unreachable.** Added to `TOOL_MODULES` but  never pinned in the descriptor index, so its module never loaded for the  `action` surface. `tests/tools/test_registry_index.py` was already red and  named it exactly.
- **A skill body crashed the turn.** `sed 's/\]\]>//'` parsed as Rich markup —  `MarkupError` killed the whole turn. Tool output and model replies are now  escaped before printing.
- **Grafana discovery ran once per concurrent caller.** Check-then-act race; ten  threads, ten discoveries, each burning a 10s timeout against a dead host.
- **An unreachable service dumped ~60 lines of urllib3 stack** into the shell.  Classification is by exception chain, not method name, so a genuine bug keeps  its traceback.
- **The offer tool advertised itself as the way to run a report**, so the model  called it instead of doing the work. Metadata rewritten and an execution  guard added — the guard is the part that works.
- **The fetch guard was session-wide**, so a stale `wttr.in` call from any  earlier turn unlocked a fresh offer. Now turn-scoped via `history_start`.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4669, opened 2026-08-02
fix(github): log soft-fail exception in identity lookup (#4648)

Shivang9983:
Fixes #4648

Added exception logging (logger.warning with exc_info=True) to the soft-fail catch block in integrations/github/identity.py. When get_integration("github") fails or encounters an error, it now logs the exception details for observability while keeping the original fallback behavior (return "") completely intact.

### Demo/Screenshot for feature changes and bug fixes - 

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ x ] No, I wrote all the code myself
- [ ] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [ ] I have reviewed every single line of the AI-generated code
- [ ] I can explain the purpose and logic of each function/component I added
- [ ] I have tested edge cases and understand how the code handles them
- [ ] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**

---

## Checklist before requesting a review
- [ x ] I have added proper PR title and linked to the issue
- [ x ] I have performed a self-review of my code
- [ x ] **I can explain the purpose of every function, class, and logic block I added**
- [ x ] I understand why my changes work and have tested them thoroughly
- [ x ] I have considered potential edge cases and how my code handles them
- [ x ] If it is a core feature, I have added thorough tests
- [ x ] My code follows the project's style guidelines and conventions

---

Note: Please check **Allow edits from maintainers** if you would like us to assist in the PR.

PR #4730, opened 2026-08-04
fix(pagerduty): log swallowed exception in client factory soft-fail path

### LINKED ISSUE #4649: SM-115: Log soft-fail in PagerDuty client factory
## Summary
Same pattern as SM-114: the PagerDuty client factory in `integrations/pagerduty/client.py` catches `Exception` and returns `None` silently. Keep the soft-fail return, add a log line (debug/warning with exception info).

## Do not
Edit a second file — one issue, one file, one PR.

## Done when
- [ ] Soft-fail path logs the exception
- [ ] Return behavior unchanged
- [ ] lint/typecheck pass

## Files
`integrations/pagerduty/client.py`

ousamabenyounes:
Fixes #4649

`make_pagerduty_client` in `integrations/pagerduty/client.py` wrapped client
construction in `try/except Exception` and returned `None` on any failure — but
swallowed the exception silently. A misconfigured PagerDuty integration would
then fail to build with **no diagnostic signal** anywhere.

This adds a `logger.warning(..., exc_info=True)` in the soft-fail path while
keeping the `None` return unchanged, so the failure is diagnosable without
changing behavior. It follows the same pattern requested by the SM-11x series
and already merged for the GitHub identity lookup (SM-114, #4669) and the
Discord verifier (#4718).

Scope: one issue, one production file (plus its companion test).

#### Demo/Screenshot for feature changes and bug fixes -

Characterization test proves the behavior (RED → GREEN):

```text
# BEFORE the fix (silent swallow) — new test fails:
tests/integrations/pagerduty/test_client.py::test_make_client_logs_soft_fail_on_construction_error FAILED
E   assert False   # no WARNING record captured

# AFTER the fix — full file green:
tests/integrations/pagerduty/test_client.py .................................  [100%]
============================== 33 passed in 0.90s ==============================
```

`make lint`, `make format-check`, `make typecheck`, and
`make verify-integrations-smoke` all pass.

---

The factory's `except Exception` branch returned `None` with no logging, so an
operator got zero signal when PagerDuty client construction failed. I bind the
exception (`as exc`, matching the identifier used everywhere else in this
module) and log it at WARNING with `exc_info=True` for the traceback, then keep
the existing `return None` so the soft-fail contract is untouched. I chose
`warning` over `debug` because the issue anchors on SM-114 (#4669), which logs
at `warning`. The test monkeypatches `PagerDutyConfig` to raise, then asserts
both that the factory still returns `None` and that a WARNING record carrying
`exc_info` was captured — which fails against the old silent code and passes
with the fix.

---

### DIFF
--- a/integrations/pagerduty/client.py
+++ b/integrations/pagerduty/client.py
@@ -464,5 +464,6 @@ def make_pagerduty_client(
         if base_url:
             config_kwargs["base_url"] = base_url
         return PagerDutyClient(PagerDutyConfig(**config_kwargs))
-    except Exception:
+    except Exception as exc:
+        logger.warning("Failed to create PagerDuty client: %s", exc, exc_info=True)
         return None
--- a/tests/integrations/pagerduty/test_client.py
+++ b/tests/integrations/pagerduty/test_client.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import logging
 from typing import Any
 
 import pytest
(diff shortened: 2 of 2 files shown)


ousamabenyounes:
@greptile review

PR #4731, opened 2026-08-04
fix(alertmanager): log swallowed exception in client factory soft-fail path

### LINKED ISSUE #4651: SM-117: Log soft-fail in Alertmanager client
## Summary
Same pattern as SM-114/SM-115/SM-116: the client path in `integrations/alertmanager/client.py` swallows exceptions silently. Keep the soft-fail behavior, add a log line (debug/warning with exception info).

## Do not
Edit a second file — one issue, one file, one PR.

## Done when
- [ ] Soft-fail path logs the exception
- [ ] Return behavior unchanged
- [ ] lint/typecheck pass

## Files
`integrations/alertmanager/client.py`

ousamabenyounes:
Fixes #4651

`make_alertmanager_client` (`integrations/alertmanager/client.py`) wrapped client
construction in a bare `except Exception: return None`, so a misconfigured
Alertmanager integration failed to build **silently** — no log line, no signal.

This keeps the soft-fail contract (still returns `None`) but logs the swallowed
exception at `WARNING` with `exc_info`, matching the SM-114/SM-115 precedent
already merged for the GitHub identity lookup (#4669) and the PagerDuty client
factory (#4649). One issue, one source file (+ its test).

```python
    except Exception as exc:
        logger.warning("Failed to create Alertmanager client: %s", exc, exc_info=True)
        return None
```

The log goes only to the module logger (server-side); callers still receive
`None`, so no exception detail reaches any external surface (CWE-209 safe).

Added a characterization test (`test_make_client_logs_soft_fail_on_construction_error`)
that forces config construction to raise and asserts the factory still returns
`None` **and** that a `WARNING` with `exc_info` is emitted.

**RED — new test on the unmodified factory (silent swallow):**

```
tests/integrations/alertmanager/test_client.py::test_make_client_logs_soft_fail_on_construction_error F
    assert any(record.levelno == logging.WARNING and record.exc_info is not None ...)
E   assert False
1 failed in 2.04s
```

**GREEN — after adding the log line (whole client test module):**

```
tests/integrations/alertmanager/test_client.py ......................................
38 passed in 0.98s
```

Local CI (`make lint`, `make format-check`, `make typecheck`,
`make verify-integrations-smoke`) is green on both the upstream `main` baseline
and this branch.

---

The problem is an observability gap: the client factory swallowed *every*
construction error and returned `None`, so an operator with a broken
Alertmanager config got no diagnostic. I deliberately kept the soft-fail return
(callers depend on `None` meaning "not configured / unavailable") and only added
a `logger.warning(..., exc_info=True)` so the stack trace lands in server logs.
I considered re-raising or returning an error object, but that would change the
factory contract and break every call site — out of scope for this cleanup
ticket, and inconsistent with the two already-merged sibling fixes (#4669,
#4649), which I matched exactly for consistency. The test monkeypatches
`AlertmanagerConfig` to raise, exercising the real `except` branch with a fake
(no live API), and asserts both the `None` return and the logged WARNING.

---

### DIFF
--- a/integrations/alertmanager/client.py
+++ b/integrations/alertmanager/client.py
@@ -16,6 +16,7 @@ from typing import Any
 
 import httpx
 
+from integrations._validation_helpers import report_validation_failure
 from integrations.config_models import AlertmanagerIntegrationConfig
 from integrations.probes import ProbeResult
 from platform.observability.errors.service import capture_service_error
@@ -235,5 +236,15 @@ def make_alertmanager_client(
             )
         )
     except Exception as exc:
-        logger.warning("Failed to create Alertmanager client: %s", exc, exc_info=True)
+        # The caught exception can be a pydantic ValidationError whose text embeds
+        # the rejected bearer_token/password via ``input_value=``. Route through the
+        # repo's sanitized reporter (fixed one-line local message + scrubbed Sentry
+        # capture) instead of logging the raw exception, so a misconfigured
+        # integration stays diagnosable without leaking credentials into local logs.
+        report_validation_failure(
+            exc,
+            logger=logger,
+            integration="alertmanager",
+            method="make_alertmanager_client",
+        )
         return None
--- a/integrations/alertmanager/client.py
+++ b/integrations/alertmanager/client.py
@@ -234,5 +234,6 @@ def make_alertmanager_client(
                 password=password or "",
             )
         )
-    except Exception:
+    except Exception as exc:
+        logger.warning("Failed to create Alertmanager client: %s", exc, exc_info=True)
         return None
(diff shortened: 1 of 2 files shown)


ousamabenyounes:
@greptile review

ousamabenyounes:
Good catch on the credential exposure — fixed in the follow-up commit. The soft-fail path now logs only `type(exc).__name__` (no `exc` value, no `exc_info`), since a rejected credential-bearing config can echo the bearer token/password through a Pydantic `ValidationError` or the traceback frames (CWE-532). The characterization test now plants a secret-shaped token in the raised error and asserts it never appears in the log record. This is stricter than the SM-114/SM-115 precedent, which logs `exc` directly — those factories carry the same risk and may be worth the same treatment in a follow-up.

ousamabenyounes:
Correction on the mechanism: the landed fix routes through the repo's existing `report_validation_failure` helper (fixed one-line local message + scrubbed Sentry capture) rather than logging `type(exc).__name__` directly — the more idiomatic version of the same sanitization. Net effect is unchanged: the bearer token / password can no longer reach local logs on a soft-fail.

PR #4742, opened 2026-08-05
Fix development command examples

TheophilusChinomona:
## Summary

- Update the onboarding and investigation examples in `docs/DEVELOPMENT.md` to use `uv run opensre`.
- Keep the examples aligned with the setup guidance and ensure they run against the checkout's virtual environment.

## Root cause

The guide recommended `uv run opensre` but the examples used plain `opensre`, which can fail when the editable CLI is not on `PATH`.

## Validation

- `git diff --check`
- Documentation-only change; per `CI.md`, the docs shortcut applies.


Davidson3556:
Which issue are you working on? Also, the PR description doesn't follow the required format. Please update it accordingly.

PR #4755, opened 2026-08-05
fix: soft-fail corrupt config when remote sync is fully env-driven

### LINKED ISSUE #4552: [Good first issue] Soft-fail optional stored settings when env already configures sync
### Complexity

### Context
Docs teach enabling with only `OPENSRE_REMOTE_SYNC` + `OPENSRE_REMOTE_SYNC_BUCKET`. Optional fields (prefix, provider, …) still fall through to `config.yml`. A corrupt YAML can fail that documented happy path even though required values are fully in env. There is already a test for "every field in env ignores corrupt settings"; extend coverage to the two-variable docs path.

### Done when
- With only the two documented env vars set, a corrupt `~/.opensre/config.yml` does not block sync (defaults apply for unset optionals).
- A unit test pins that path.
- Still raise `RemoteSyncConfigError` when the stored section is required (env switch unset / bucket only in file).

---

AshSgDe29071999:
## Summary
Docs teach enabling remote sync with only `OPENSRE_REMOTE_SYNC` + `OPENSRE_REMOTE_SYNC_BUCKET`. Optional fields still fell through to `config.yml`, so a corrupt YAML blocked that documented happy path.

## Changes
- When switch + bucket are already set in the environment, soft-fail a corrupt stored `remote_sync` section to `{}` so optionals use defaults.
- Keep hard `RemoteSyncConfigError` when the stored section is actually required (env switch unset / bucket only in file).
- Unit test for the two-variable docs path with a damaged `config.yml`.

Closes #4552

### DIFF
--- a/platform/filestorage/config.py
+++ b/platform/filestorage/config.py
@@ -135,17 +135,27 @@ def _exclusions(stored: Callable[[], dict[str, Any]]) -> ExclusionRules:
     return parse_exclusions(stored().get("exclude"))
 
 
-def _lazy_stored() -> Callable[[], dict[str, Any]]:
+def _lazy_stored(*, soft: bool = False) -> Callable[[], dict[str, Any]]:
     """Read the settings file at most once, and only if something needs it.
 
     A run configured entirely through the environment must not fail because an
     unrelated section of ``config.yml`` is damaged.
+
+    When ``soft`` is true (docs happy path: switch + bucket already in env), a
+    corrupt stored section falls back to ``{}`` so optional fields use defaults
+    instead of blocking sync.
     """
     cache: dict[str, dict[str, Any]] = {}
 
     def read() -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


muddlebee:
Thanks for the fix! Greptile flagged a gap worth closing before merge: the soft-fail fallback only covers the case where the whole stored `config.yml` section is unreadable. If the file parses fine but an *individual optional field* (e.g. prefix/region/profile) is malformed, that still raises today even when `OPENSRE_REMOTE_SYNC` + `OPENSRE_REMOTE_SYNC_BUCKET` are fully set per the docs. Could you extend the soft-fail path (or the optional-field validation) to cover that case too, so the documented two-env-var setup is fully resilient to a damaged `config.yml`? Happy to take another look once that's in.

Davidson3556:
closing this needs to be assign to you first before picking it up.

PR #4565, opened 2026-07-30
refactor(probes): promote verification ProbeStatus to StrEnum

santhiprakash:
Fixes #4516

#### Describe the changes you have made in this PR -

Promote integration verification probe outcomes from duplicated string literals to a closed `ProbeStatus` `StrEnum`, wire `ProbeResult` factories/`ok` through the enum, and key `integrations/verify.py` status styling on enum members. String values stay identical (`passed` / `failed` / `missing`) so verifier dicts and existing comparisons keep working.

### Demo/Screenshot for feature changes and bug fixes -

N/A (refactor-only; behavior unchanged).

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [x] I have reviewed every single line of the AI-generated code
- [x] I can explain the purpose and logic of each function/component I added
- [x] I have tested edge cases and understand how the code handles the code handles them
- [x] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**

Probe outcomes were hard-coded in `ProbeResult` factories and again in `_STATUS_STYLE`, so typos only fail at runtime. I added `ProbeStatus(StrEnum)` beside `ProbeResult`, typed the dataclass field, coerced plain strings in `__post_init__` (frozen dataclass requirement from the issue), and switched factories/`ok` to enum members. `verify.py` now styles rows by resolving the string status to `ProbeStatus` while still rendering the original string for output. Added `tests/integrations/test_probe_status_enums.py` following the existing fleet/filestorage enum test pattern.

---

## Problem

Integration verifiers return one of three probe statuses as bare strings. Those literals are repeated in `ProbeResult` factories and `integrations/verify.py` styling, so a typo (e.g. `"passsed"`) only surfaces at runtime and UI code can drift from factories.

## Triage / Root cause

`integrations/probes.py` typed `ProbeResult.status` as `str` and factories passed raw strings; `integrations/verify.py` duplicated the same three strings in `_STATUS_STYLE`.

## Fix

- Add `ProbeStatus` `StrEnum` with unchanged values.
- Type/coerce `ProbeResult.status`; factories return enum members; `ok` checks `ProbeStatus.PASSED`.
- Key `_STATUS_STYLE` on `ProbeStatus` and resolve row status through the enum in `format_verification_results`.

## Verification

[...]

PR #4759, opened 2026-08-05
refactor: promote ToolSurface / EvidenceType / SideEffectLevel to StrEnums

### LINKED ISSUE #4530: refactor: ToolSurface / EvidenceType / SideEffectLevel StrEnums
## Why we need this
Every registered tool declares metadata using string Literals for surface, evidence type, and side-effect level. StrEnums give exhaustiveness at definition time. **`EvidenceSource` must stay `str`** — it is an open vendor-key registry.

## Current code
```python
# core/domain/types/tools.py
ToolSurface = Literal["investigation", "chat", "action"]

# core/tool_framework/metadata.py
EvidenceType = Literal["logs", "metrics", "traces", ...]
SideEffectLevel = Literal["none", "read_only", "mutating", "external"]
```

## Target shape
StrEnums with identical string values; leave `EvidenceSource = str` untouched.

## Done when
- [ ] Three StrEnums mirroring current Literals.
- [ ] Do not change `EvidenceSource`.
- [ ] Registry/decorator paths still accept the same strings.
- [ ] Contract tests updated if needed.

**Files:** `core/domain/types/tools.py`, `core/tool_framework/metadata.py`

<details>
<summary><b>StrEnum conventions &amp; gotchas (shared across these refactors)</b></summary>

- Use `enum.StrEnum` (not plain `Enum`) so values stay usable as strings in env, JSON, Click, and `==` checks.
- Put the enum in a **leaf** module (no import cycles). Env-var **names** stay in `config/constants/`; only **values** become enums.
- Keep string values **identical** to today (analytics, frontmatter, persisted JSON).
- Prefer `Enum("value")` / member compares; use `.value` only when an external API needs a plain `str`.
- Add a small unit test that pins members and round-trips from the string (see `tests/filestorage/test_enums.py`).
- Precedent: `platform/filestorage/enums.py`, `platform/common/task_types.py`.

**Gotchas**
- **Frozen dataclass:** coerce in `__post_init__` via `object.__setattr__(self, "field", Enum(self.field))` — annotating the field converts nothing, so a plain string passed by a caller stays a `str` while mypy/tests stay green.
- **Pydantic:** `model_dump()` returns the member, not the string; use `model_dump(mode="json")` / `model_dump_json()` where persisted shape matters, and pin it with a round-trip test on a real saved file.
- **YAML:** call `.value` before `yaml.safe_dump` (it raises `RepresenterError` on a StrEnum member).
- **Hot paths:** `Enum(value)` is a lookup, not a compare — resolve once and reuse.

**Do not enum** env-var names, free-form user text, or open registries (`EvidenceSource`, community object-store provider names, the full `LLMProvider` catalog).
</details>

themid6t:
Fixes #4530

Promotes the three tool-metadata `Literal` aliases to closed `enum.StrEnum`s with **identical string values**, so the wire values (`"investigation"`, `"read_only"`, `"metrics"`, …) are unchanged for analytics, persisted JSON, and existing comparisons. `EvidenceSource` is left as `str` — it is an open vendor-key registry.

The change is split into two commits so the issue's actual ask can be reviewed in isolation from the typecheck fallout.

**Commit 1 — the enums (3 files)**

- `core/domain/types/tools.py` — `ToolSurface` → `StrEnum`.
- `core/tool_framework/metadata.py` — `EvidenceType`, `SideEffectLevel` → `StrEnum`.
- `core/tool_framework/registry_metadata.py` — **not optional.** `normalize_surfaces` validated with `set(get_args(ToolSurface))`, and `get_args()` returns the members for a `Literal` but `()` for a `StrEnum`. Left as-is, `_VALID_SURFACES` is empty and *every* tool registration raises `Unsupported tool surface` at import. Now validates against the enum directly.

No coercion code was needed: `RegisteredTool.__post_init__` already round-trips every field through `ToolMetadata.model_validate`, so tools declaring plain strings keep working and resolve to enum members for free.

**Commit 2 — the annotation boundary (7 files + tests)**

A `StrEnum` is a *subclass* of `str`, so while every member **is a** `str` (equality, `isinstance`, and `json.dumps` all keep working), a plain `str` is **not** assignable to the enum. Every tool still declaring a bare wire value therefore failed typecheck — **84 errors across 57 files**, all one of two shapes:

```
Incompatible types in assignment (expression has type "str",
  base class "BaseTool" defined the type as "SideEffectLevel | None")

Argument 1 to "get_registered_tools" has incompatible type "str";
  expected "ToolSurface | None"
```

Every one of those 84 errors traces back to just **seven annotation sites**. Widening those to `Enum | str` clears all of them without touching a single call site:

| File | Change |
| --- | --- |
| `core/tool_framework/base.py` | 3 `ClassVar`s (`evidence_type`, `side_effect_level`, `surfaces`) |
| `core/tool_framework/tool_decorator.py` | `evidence_type` / `side_effect_level` across the 4 `@tool` overloads |
| `core/tool_framework/registered_tool.py` | 3 dataclass fields + the builder kwargs |
| `tools/registry.py` | the 3 `surface=` lookup params |

Plus three tools that *restated* `surfaces: ClassVar[tuple[ToolSurface, ...]]` rather than inheriting it (`railway/inspect_tool.py`, `railway/redeploy_tool.py`, `slack_thread_replay_tool/tool.py`) — the redundant annotation is dropped so they inherit the widened one.

**On scope:** the alternative was migrating ~50 tool modules to enum members (`side_effect_level=SideEffectLevel.READ_ONLY`), which would put the whole tool tree in the diff for a refactor that changes no behavior. These 7 annotation files are in the diff *precisely so those 50 are not.* As a side effect `tools/registry_index.py` needs no change either — its static AST index only understands string literals in `surfaces=`, and since call sites still hold strings, its whole suite passes untouched.

Widening `evidence_type` to accept `str` does not lose typo protection, it just moves the stage: previously `Literal` rejected `"reed_only"` at mypy; now `ToolMetadata.model_validate` raises `ValidationError` at import.

**Tests** — `tests/core/tool_framework/test_enums.py` pins the member values against the previous `Literal`s, string round-trips, closed membership, and that `normalize_surfaces` still coerces plain strings. Follows the `tests/integrations/test_probe_status_enums.py` precedent from #4565.

Refactor only — no behavior or wire-value change. Every tool in the repo still declares bare strings and is coerced transparently:

```
$ uv run python -c "from tools.registry import get_registered_tools; ..."
registry loaded: 247 investigation tools
sample: describe_rds_instance
  side_effect_level = <SideEffectLevel.READ_ONLY: 'read_only'>  (type SideEffectLevel)
  == plain string   : True
  surfaces          = (<ToolSurface.INVESTIGATION: 'investigation'>,)
```

Values verified identical to the previous `Literal`s, member for member:

```
OK  ToolSurface:     order+values identical = True
OK  EvidenceType:    order+values identical = True
OK  SideEffectLevel: order+values identical = True
```

Full gate:

```
$ make typecheck
Success: no issues found in 1581 source files

$ make lint
All checks passed!

$ make format-check
3045 files already formatted

$ make test-cov
13574 passed, 62 skipped
```

`make test-cov` also reports 2 failures in `tests/cli/wizard/test_env_sync.py` and 1 in `tests/cli/test_smoke.py` — these reproduce identically on an untouched upstream `main` (they attempt real outbound network calls, blocked in my sandbox) and are unrelated to this change.

---

The three vocabularies were bare string `Literal`s, so there was no single closed source of truth and a typo only surfaced wherever an annotation happened to catch it. Promoting each to a `StrEnum` with identical values makes the set exhaustive at definition time while members still behave as strings everywhere they are already used.

The interesting problem was scope. Because a `StrEnum` is a `str` subclass, the subtyping runs the wrong way for assignment — which is why ~50 tool modules that declare `side_effect_level = "read_only"` suddenly failed typecheck even though nothing about their behavior changed. The obvious fix is to rewrite all of them to use enum members, but that produces a very large diff for a refactor that changes no behavior, and it drags in `tools/registry_index.py`, whose static AST index only resolves string literals in `surfaces=`.

I chose the opposite direction: keep the call sites untouched and widen the handful of framework annotations they bind to. I traced all 84 errors back to their declaration sites and found they collapse to seven, so widening those to `Enum | str` clears the entire set. I used `Enum | str` rather than plain `str` deliberately — plain `str` orphans the `EvidenceType`/`ToolSurface` imports in those modules and leaves the signatures with no pointer to the valid values, whereas the union keeps the canonical type visible in the API that tool authors actually read.

Runtime correctness comes free from machinery that already existed: `RegisteredTool.__post_init__` validates through `ToolMetadata`, so a plain string is coerced to a member at registration. I verified that end to end rather than assuming it — all 247 investigation tools load with their declarations unchanged and come back holding enum members. I also checked the two serialization traps the issue calls out (`yaml.safe_dump` raising on a member, and `model_dump()` returning members rather than strings) and confirmed neither path touches these three fields; the only `yaml.safe_dump` call sites serialize fleet-agent and CLI config, and `json.dumps` is safe regardless since a `StrEnum` is a `str`.

---

---


Davidson3556 (commented):
Great job — clean refactor, and catching the `get_args()` breakage in `registry_metadata.py` was the sharp bit.

One thing before merge: `| str` gives back the static check the `Literal` was doing — 3 typos caught before, 0 now. Runtime still catches all three at import, so nothing ships broken, but #4530 asked for closed *and* typed and only half lands. Worth a follow-up to migrate the ~50 call sites and drop the `| str`.

core/tool_framework/base.py
```diff
@@ -45,8 +45,11 @@ class BaseTool(ABC):
     input_model: ClassVar[type[BaseModel] | None] = None
     source: ClassVar[EvidenceSource]
     source_id: ClassVar[str | None] = None
-    evidence_type: ClassVar[EvidenceType | None] = None
-    side_effect_level: ClassVar[SideEffectLevel | None] = None
+    # Declared as ``... | str`` so subclasses may keep using the plain wire
+    # values (``"logs"``, ``"read_only"``); ``ToolMetadata`` validation in
+    # ``__init_subclass__`` coerces them to the enum member.
+    evidence_type: ClassVar[EvidenceType | str | None] = None
+    side_effect_level: ClassVar[SideEffectLevel | str | None] = None
```
Davidson3556, line 52: This is the part I'd want stated in the issue rather than only in the PR description. `| str` gives back the static check the `Literal` was doing. Same class with three typos:

```
pre-refactor (Literal): 3 errors — reed_only, metricz, investigaton
this branch  (| str)  : Success: no issues found
```

Your defence holds up — I confirmed all three still raise `ValidationError` at import, `surfaces` included via `normalize_surfaces`. So nothing ships broken, the check just moves from edit time to import time.

Still, #4530 asked for these vocabularies to be closed *and* typed, and right now only half of that lands. Worth a follow-up issue to migrate the ~50 call sites to members and drop the `| str`, so the widening reads as a deliberate staging step instead of the end state. Fine to merge as-is with that filed.

themid6t: Filed as #4767 — and thanks, agreed on the direction. Scoping it against the tree turned up something I didn't expect though, so I ended up splitting it in two.

Your ~50 lines up with the 57 files that actually failed typecheck. But there are 138 non-test files declaring bare wire values, and the extra ones aren't files I missed — the `@tool` decorator's `surfaces` param is `tuple[str, ...]` on `main` as well, untouched by this PR. There's no `| str` there to drop, because it was never typed to the vocabulary in the first place, not even in the `Literal` era. So the ~134 files declaring `surfaces=` never type-errored and aren't being masked by anything I did here.

Your three-typo class is a nice illustration of that, actually: it must have been a `BaseTool` subclass, since that's the only shape where `surfaces` was typed on `main`. Put `investigaton` inside a `@tool(surfaces=(...))` call instead and it typechecks clean on `main` too — same gap, just smaller.

Which means dropping the `| str` is a smaller job than either of us assumed: 57 files, `side_effect_level` and `evidence_type` plus a handful of `BaseTool` subclass ClassVars. That's the part that closes the regression this PR introduced, and it's Phase A in the issue.

Typing `surfaces` properly is the separate, older problem. It needs `registry_index.py` taught to resolve `ast.Attribute` first — the AST scan only reads string literals, so migrated chat/action tools would quietly fall back to investigation-only. The index contract tests do catch it, but it means the scanner has to land before any declaration moves. It also drags in the `ToolRegistry` port at `core/agent_harness/ports.py:113,116`, which types `surface: str` too. That's Phase B.

Both phases are checklisted in the issue with counts and blockers. Your typo evidence is in the Context section as the rationale for the whole thing, and Phase A finishes by re-adding that check as a regression test.

Davidson3556: You're right, and the correction is on my side: my probe was a `BaseTool` subclass, which is the one shape where `surfaces` was typed on `main`. A surfaces-only typo inside `@tool(surfaces=(...))` typechecks clean on both branches — so that gap predates this PR and isn't something the `| str` is masking. The regression this PR actually introduces is `evidence_type` / `side_effect_level` plus the `BaseTool` ClassVars, which is your Phase A.

Your Phase B blocker is real too. I migrated one tool to enum members and `test_descriptor_surfaces_match_registry` and `test_surface_scoped_load_equals_full_filtered` both fail — the scanner does have to land first, and the contract tests do catch it.

#4767 splits it better than my comment did. Approving.

Davidson3556 (approved):
Great job — the scope analysis in #4767 is sharper than my original comment. Phase A is the real regression here, Phase B predates this PR.

Davidson3556:
@greptile review

Davidson3556:
LGTM

PR #4760, opened 2026-08-05
feat(prompts): ground the assistant in the operator's real setup

YauhenBichel:
The assistant prompt had no awareness of goals, scheduling, or the user's progress. Every mention of scheduling lived in `action_agent_system_prompt.py` — the agent that decides actions — while the assistant that writes what the user
reads had none. The CONTEXT tier carried only machine state (environment, repo map, runtime facts) and never user progress. An agent cannot guide someone toward an outcome whose state it cannot see.

Two changes, both in this diff.

1. Setup state as CONTEXT-tier facts

`platform/setup_state.py` renders what the install actually has:

    --- Setup state ---
    Integrations connected: posthog_mcp, slack
    Scheduled tasks configured: 23
    Last scheduled delivery: failed

The CLI preamble is reframed around outcomes rather than command help, and forbids claiming something is configured when the block says otherwise. Every technical fact from the previous preamble is preserved verbatim — slash-command
argv semantics, the `!` prefix rule, and the "does not run investigations itself" grounding.

`SurfaceProfile.setup_state` gates the block to the CLI. One installation's connected integrations are not a shared Slack channel's business, and a test asserts the marker never appears in a gateway prompt.

**Facts go in the CONTEXT block; instructions stay in the STABLE persona.** `test_states_facts_without_instructing_the_model` fails if wording like "you should" or "offer to" leaks into the fact block, so a per-turn state change cannot silently rewrite the rules.

2. Prompt package restructure

`assistant_agent_prompt.py` (529 lines) is split into an `assistant/` package — `text.py` holds the literals, `blocks.py` the per-block contributors, `assemble.py` the ordering, `parts.py` the inputs — and `prompt_context.py` becomes
`context/provider.py`. Block assembly is now a list of contributors rather than a long conditional body, which is what made adding a gated block a few lines instead of another branch.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4764, opened 2026-08-06
feat: added azure blob storage exposure check with warning if public

### LINKED ISSUE #4726: [Intermediate] Warn when the target Azure Blob Storage container is publicly readable
### Context

Follow-up to #4556 (AWS) / #4719 and #4720 (Vercel) / #4722 — `status` warns
when the configured bucket/store is publicly readable via an opt-in
`public_access_checker` registered per provider (see
`platform/filestorage/exposure.py`, `platform/filestorage/providers/registry.py`).
AWS and GCS have it (#4719); Vercel Blob has it (#4722, as a follow-up —
the original design assumed a per-object access model and had to be
corrected once the real one turned out to be store-level).

Azure Blob Storage support itself is tracked separately in #4564 (PR #4671,
in progress). Filing this as its own issue rather than adding scope to
#4564 mid-flight — pick this up once the base provider lands.

### Scope

Azure Storage containers have a **container-level** public-access setting
(`None` / `Blob` / `Container`), readable in one call via the container's
properties (`ContainerClient.get_container_properties().public_access` in
the Python SDK, or the equivalent REST call) — the same shape as S3's
`GetBucketPolicyStatus` and GCS's `getIamPolicy`
(`providers/aws.py::check_public_access`, `providers/gcs.py::check_public_access`),
not a per-object scan.

### Done when

- `status` warns when the target container's public-access level is not
  `None`.
- Registered via the optional `public_access_checker` on
  `register_object_store` — never raises; a missing permission or an
  unreachable container degrades to `BucketExposure.UNKNOWN` with a short
  note, same contract as AWS/GCS/Vercel.
- `detail` on that note never carries raw exception/API response text — it
  can be echoed straight into a gateway chat surface (Slack/Telegram
  `/remote-sync status`), which isn't vetted for that (CWE-209). See the
  docstring on `providers/aws.py::check_public_access` for the reasoning.
- Behavior is mockable without a real public container (fake client/session
  injected, same pattern `S3ObjectStore`/`GCSObjectStore`/`VercelBlobObjectStore`
  already use).
- Adds an Azure section to `tests/filestorage/test_bucket_exposure.py` (the
  precedent file — covers AWS/GCS/Vercel today) rather than a new test file.
- Documents the exposure-check permission explicitly in
  `docs/configuration/remote-sync.mdx`'s Azure section, the way
  `s3:GetBucketPolicyStatus` / `storage.buckets.getIamPolicy` are called out
  for AWS/GCS.

---

1. Read `platform/filestorage/ports.py` and `platform/filestorage/exposure.py`
   first — exposure checks are opt-in per provider via `public_access_checker`
   on `register_object_store`, not a method on the core `ObjectStore` protocol.
2. `check_bucket_exposure()`'s contract is that a checker never raises.
3. Prefer `StrEnum` for new closed vocab. Env var names stay in
   `config/constants/filestorage.py`, or a leaf `config/constants/<vendor>.py`
   for a vendor-wide constant shared across packages (see
   `config/constants/vercel.py`'s `VERCEL_API_BASE_URL`, added in #4722 to
   avoid the same host string being duplicated).

_Source: follow-up from #4556 / #4719 and #4720 / #4722. Depends on #4564 / #4671 landing first._

SaptarshiSarkar12:
Fixes #4726 

I have made the following changes in the PR:
- Added code for checking blob access
- Added tests and docs

https://github.com/user-attachments/assets/f9b63cfd-435d-49b4-860f-c15deaa4d36b

---

- [X] Yes, I used AI assistance (continue below)

- [X] I have reviewed every single line of the AI-generated code
- [X] I can explain the purpose and logic of each function/component I added
- [X] I have tested edge cases and understand how the code handles them
- [X] I have modified the AI output to follow this project's coding standards and conventions

The code addressed the missing Azure blob storage access check. It first gets the Azure storage container/blob data, checks the header value of `x-ms-blob-public-access` and gives a warning if that header field exists and has a value either `blob` or `container`. The complete procedure happens via a single Azure's REST API call.

---

- [X] I have added proper PR title and linked to the issue
- [X] **I can explain the purpose of every function, class, and logic block I added**
- [X] I have considered potential edge cases and how my code handles them
- [X] If it is a core feature, I have added thorough tests
- [X] My code follows the project's style guidelines and conventions

---

### DIFF
--- a/platform/filestorage/providers/azure.py
+++ b/platform/filestorage/providers/azure.py
@@ -9,10 +9,10 @@ from urllib.parse import quote
 
 import httpx
 
-from platform.filestorage import BucketExposure, PublicAccessStatus
 from platform.filestorage.config import RemoteSyncConfig
-from platform.filestorage.enums import BuiltInProvider, RemoteSyncField
+from platform.filestorage.enums import BucketExposure, BuiltInProvider, RemoteSyncField
 from platform.filestorage.errors import RemoteSyncUnavailableError
+from platform.filestorage.exposure import PublicAccessStatus
 from platform.filestorage.ports import RemoteObject
 from platform.filestorage.providers.registry import SetupExtraField, register_object_store
 
@@ -23,6 +23,11 @@ CREDENTIAL_HINT = (
 EXTRA_FIELDS = (
     SetupExtraField(RemoteSyncField.PROFILE, "Azure Storage Account Name (e.g., opensre)"),
 )
+_API_VERSION = "2026-04-06"
+
+
+def _get_account_name(config: RemoteSyncConfig) -> str:
+    return config.profile or os.getenv("AZURE_STORAGE_ACCOUNT_NAME", "")
 
 
 class AzureBlobObjectStore:
@@ -36,13 +41,11 @@ class AzureBlobObjectStore:
         client: httpx.Client | None = None,
     ) -> None:
         self._config = config
-        self._account_name = config.profile or os.getenv("AZURE_STORAGE_ACCOUNT_NAME", "")
-
+        self._account_name = _get_account_name(config)
         if not self._account_name:
             raise RemoteSyncUnavailableError(
                 "Azure storage account name is missing. Set it during setup or via AZURE_STORAGE_ACCOUNT_NAME."
             )
-
         self._token = token if token is not None else _get_azure_access_token()
         self._client = client if client is not None else httpx.Client(timeout=30.0)
 
@@ -129,7 +132,7 @@ class AzureBlobObjectStore:
         return quote(self._config.key_for(key), safe="/")
 
     def _auth_headers(self) -> dict[str, str]:
-        return {"Authorization": f"Bearer {self._token}", "x-ms-version": "2026-04-06"}
+        return {"Authorization": f"Bearer {self._token}", "x-ms-version": _API_VERSION}
 
     def _strip_prefix(self, full_key: str) -> str:
         prefix = f"{self._config.prefix.rstrip('/')}/"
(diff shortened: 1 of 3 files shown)


platform/filestorage/providers/azure.py
```diff
@@ -9,6 +9,7 @@
 
 import httpx
 
+from platform.filestorage import BucketExposure, PublicAccessStatus
```
Davidson3556: `aws.py` and `gcs.py` both take these from the leaf modules — `BucketExposure` from `platform.filestorage.enums`, `PublicAccessStatus` from `platform.filestorage.exposure`. This is the only provider reaching for the package root.

I tested it and it doesn't break: the parent package is always initialised before a submodule, so the names are bound by the time this runs. But `registry.py:62` says built-ins are imported on first use "so this package never imports itself," and this is the edge AGENTS.md's `py/cyclic-import` note says not to add. Two-line change to match the siblings.

While here: `gcs.py` lists `check_public_access` in `__all__`; this file doesn't.

platform/filestorage/providers/azure.py
```diff
@@ -214,8 +215,54 @@ def _factory(config: RemoteSyncConfig) -> AzureBlobObjectStore:
...
+
+    url = f"https://{account_name}.blob.core.windows.net/{config.bucket}"
+    owns_client = client is None
+    cl = client if client is not None else httpx.Client(timeout=30.0)
+
+    try:
+        response = cl.get(
+            url,
+            params={"restype": "container"},
+            headers={"Authorization": f"Bearer {token}", "x-ms-version": "2026-04-06"},
```
Davidson3556: Both lines are copies of `AzureBlobObjectStore.__init__`/`_auth_headers` (39 and 132). The API version in particular is now in two places, so a bump has to find both. Worth a module-level `_API_VERSION = "2026-04-06"` and a small `_account_name(config)` helper that both the store and the checker call.

Davidson3556 (commented):
Great job. Two nits inline, neither blocking.

muddlebee:
@SaptarshiSarkar12 good work, might have a look at review points above

SaptarshiSarkar12:
> @SaptarshiSarkar12 good work, might have a look at review points above

@muddlebee @Davidson3556 Thanks for the reviews. I have made the requested changes.
@greptile review

PR #4775, opened 2026-08-06
fixed CI

YauhenBichel:
On main, coverage-report used always(), so it still ran after test was cancelled. No shard artifacts → coverage combine . → No data to combine.

only run that job when needs.test.result == 'success'. CI Gate already treats skipped coverage as OK.


YauhenBichel:
@greptile review

PR #4784, opened 2026-08-06
fix(scheduler): stop accepting scheduled deliveries that can never arrive

YauhenBichel:
## The problem, from a real install

A developer machine had **37 scheduled tasks**. Two distinct schedules; 35 byte-identical duplicates. One run recorded across all 37, and it failed. Every Slack task had an empty `chat_id` against a Slack whose auth was `invalid_auth`.

The user was told "task created" 37 times. Nothing ever delivered, and nothing ever said so. The agent then offered to *run* one of the dead crons, because the only thing it could see was a count.

Three defects, each fixed at the point that makes bypass impossible.

## 1. Setup state reports health, not count

The CONTEXT block rendered `Scheduled tasks configured: 29`, which reads as
progress. It now reads:

Scheduled tasks: 37 configured, 1 able to deliver
Last scheduled delivery: failed

Deliverability is not a new rule. `task_can_deliver` in `platform/scheduler/delivery.py` composes what already existed —
`slack_can_deliver` for Slack (a webhook carries its own destination, so an empty `chat_id` is legitimate there), credentials plus an explicit `chat_id` elsewhere, and never for providers with no delivery path. Each provider resolves credentials from the task's own `params` exactly as the executor does at delivery time, so a task carrying its own webhook is judged correctly.

A task whose deliverability cannot be evaluated counts as undeliverable rather than raising: the count runs across every task, so one malformed row must not collapse the snapshot and report a configured install as empty.

## 2. Identical schedules are stored once

`add_task` returns the matching stored schedule instead of appending. Identity is full configuration — `kind`, `cron`, `timezone`, `provider`, `chat_id`, `window_hours`, `params` — excluding `id`, `name` and run bookkeeping, which
differ between two confirmations of the same thing.

Deliberately conservative: three tests assert that a different destination, cron, or params stays *separate*. Under-merging is the bug we had; over-merging would silently discard a report the user asked for.

## 3. The schedule offer stops exempting Slack

`propose_scheduled_delivery` hard-coded Slack as never needing a `chat_id`:

```python
if provider not in {Provider.SLACK.value, Provider.INTERACTIVE_SHELL.value} and not chat_id:

```

Slack can skip an explicit channel only while a webhook supplies one. `/cron add` already asks requires_explicit_chat_id; this path did not, so it was the way undeliverable Slack tasks got in. It now asks the same question.

Also in this branch
PendingInvestigationOffer and a DispatchablePendingOffer Protocol make "yes" to an investigation offer structured state rather than scraped prose — the shape PendingScheduleOffer already used.

A handed-off turn now speaks once. assistant_handoff is a tool, so handled = planned_count > 0 marked the turn handled and printed the action's closing; the router then saw action_handled with handoff_contents and routed on to the LLM, which answered again. Saying "good morning" produced two replies. _self_recording_tools_only exists for this and did not fire — it excludes assistant_handoff then returns bool(names), so a handoff-only turn yields an empty list.

Merged origin/main:
Upstream landed work-item management, Buzz messaging, scheduler loops and goal-reviewed turns while this branch was open. Five conflicts, all from the same collision — both sides added a pending-offer field to `SessionCore`:

- `pending_investigation_offer` (ours) and `pending_recovery_note` (theirs) are independent; both kept. `_CORE_FIELD_COUNT` went 21 → 22, since each side had updated it for one field.
- `tests/scheduler/test_delivery.py` was deleted upstream. All seven delivery functions it covers still exist, so the file is kept rather than dropping  coverage for live code.
- The characterization snapshot was regenerated from the upstream base rather than hand-merged.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

core/agent_harness/prompts/assistant/text.py
```diff
@@ -191,8 +191,11 @@
     "those as available thread context for follow-up questions; do not ask the "
     "user to paste values that are already present there.\n\n"
     "When the user asked a cause/why question and this turn gathered quick "
-    "evidence, lead with what the evidence shows and close with **Want me to:** "
-    "run a full investigation. Make the offer once — skip it when the evidence "
+    "evidence, lead with what the evidence shows and close with exactly "
+    "**Want me to:** run a full investigation — that exact phrase, once, as "
+    "the final line. Do NOT offer paste-an-alert, /integrations setup, or a "
+    "dual menu instead of that closer (the runtime arms a structured accept on "
+    "it; dual closers break bare yes). Skip the offer only when the evidence "
```
YauhenBichel:
```suggestion
    "it, dual closers break bare yes). Skip the offer only when the evidence "
```

PR #4798, opened 2026-08-06
feat(bootstrap): process-level bootstrap layer; fix doubled investigation progress rendering

YauhenBichel:
Two changes, one structural and one behavioral:

**1. Process bootstrap layer (`bootstrap/`)**
Every entrypoint (CLI, interactive shell, gateway, report/digest runners, loop scheduler) previously wired harness ports and adapters through its own scattered startup path (`surfaces/shared/runtime_bootstrap.py`, per-runner install calls, gateway `runtime/bootstrap.py`). This PR introduces a single top-layer `bootstrap/` package — `process.py` owns process startup ordering, `adapters.py` owns adapter installation — and every entrypoint now goes through it. The old per-surface bootstrap modules are removed in the same change (no forwarding shims). `docs/ARCHITECTURE.md` documents the new layer.

**2. Single-painter fix for streamed investigation progress**
Every stage of a foreground investigation rendered twice (doubled READ/PLAN/INVEST lines with independent timings). Root cause: the streamed pipeline silences the process-wide progress tracker so `StreamRenderer` is the only painter, but `repl_foreground_renderer` re-armed a real global tracker for the run — giving pipeline stages a second painter on the same
console. The foreground path no longer re-arms the tracker before the run; `StreamRenderer` (which receives the caller's console) is the single painter, and the tracker is restored after the stream so later REPL turns keep their progress display.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4816, opened 2026-08-07
refactor(gateway): extract channels from the manager and cut the gateway→surfaces cycle

YauhenBichel:
Implements the "Improvements To Gateway Structure" note. Three things, plus a naming pass.

**Channels are no longer wired inside the manager.** `GatewayManager` had a worker field, a `_start_*` method, and a stop branch per transport. It now calls `start_channels()` and holds one `ChannelsHandle`. The new `gateway/channels/`
package is the sole composer of web + Telegram/Slack/Discord; `chat.py` drives every transport through one table (`TRANSPORTS`) keyed by a `TransportName` StrEnum. Adding a transport is now one row plus a `startup.py`, and the manager imports no transport class at all.

Startup outcomes are uniform: missing credentials report `not configured`, a transport that starts but never becomes usable reports `failed` (a new `GatewayTransportFailedError`), and the rest still start. Discord's readiness wait moved out of the manager and into `transports/discord/startup.py`, where the other transport-specific detail already lives.

**The daemon no longer names a surface.** `gateway/core/runtime/daemon.py` hardcoded `python -m surfaces.cli.gateway_entry` in its subprocess argv, so a surface started the daemon and the daemon started a surface back. Import-linter cannot see a dependency spelled as a string literal, so this passed every boundary check in CI. `start_gateway_daemon(argv=...)` is now required and each surface supplies its own entrypoint via `surfaces/shared/gateway_entrypoint.py`. `gateway_entry_argv()` resolves at spawn time and falls back to `gateway start --foreground` for frozen builds, where `python -m` would hand `-m` to Click.

**`wiring.py` → `startup.py`.** Each of those modules contained exactly one function, `start_<name>_worker`; the name described nothing. `startup.py` matches the existing `surfaces/cli/startup.py` precedent, and the surrounding vocabulary
moved with it.

Also on this branch: Telegram inline-keyboard approvals (`transports/telegram/approvals.py`, with the transport-neutral broker staying in `core/runtime/approvals.py`), a Telegram `principal.py` matching the Slack/Discord org+actor border, and `concurrency.py` reduced to the process gate with the test-only callback wrapper moved under `gateway/tests/`.

### DIFF
--- a/.importlinter
+++ b/.importlinter
@@ -24,9 +24,10 @@ forbidden_modules =
     gateway.transports
     gateway.web
 ignore_imports =
-    # Composition root: wires transports and the web surface together.
-    gateway.core.runtime.manager -> gateway.transports.**
-    gateway.core.runtime.manager -> gateway.web.web_server
+    # Composition root: the one edge into the channels composer, which owns all
+    # transport and web wiring. Pinned by gateway/tests/test_package_borders.py
+    # (test_only_manager_imports_channels_module).
+    gateway.core.runtime.manager -> gateway.channels
 
 [importlinter:contract:gateway-web-no-transports]
 name = Gateway web surface never imports chat transports
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -266,6 +266,11 @@ Steps:
 
 ## 3. Footguns (common mistakes to avoid)
 
+- Constant-condition toggles: never disable product logic with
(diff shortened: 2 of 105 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile-apps

PR #4826, opened 2026-08-09
fix(filestorage): soft-fail optional stored settings when env already configures sync

### LINKED ISSUE #4552: [Good first issue] Soft-fail optional stored settings when env already configures sync

activateLabhesh:
Fixes #4552

The docs teach enabling remote sync with only two env vars (OPENSRE_REMOTE_SYNC + OPENSRE_REMOTE_SYNC_BUCKET). Before this fix, that documented happy path broke whenever ~/.opensre/config.yml was corrupt: the switch and bucket were satisfied from env, but optional fields (prefix, provider, region, profile, exclude) still fell through to the stored remote_sync section, whose unreadable state raised RemoteSyncConfigError.

platform/filestorage/config.py now treats an unreadable stored section as "no stored settings" (the same as a missing file), so optionals fall back to their defaults (opensre prefix, aws provider, no exclusions). The stored section is still required — and still fails loudly — whenever a value has no env source: the enabled switch with OPENSRE_REMOTE_SYNC unset, and bucket with OPENSRE_REMOTE_SYNC_BUCKET unset. Malformed values in a readable file are still rejected at read time, naming the offending key.

Screenshot from 2026-08-09 20-37-18

Screenshot from 2026-08-09 20-39-28

Screenshot from 2026-08-09 20-40-00

---

- [ ✓ ] Yes, I used AI assistance (continue below)

- [ ✓ ] I have reviewed every single line of the AI-generated code
- [ ✓ ] I can explain the purpose and logic of each function/component I added
- [ ✓ ] I have tested edge cases and understand how the code handles them
- [ ✓ ] I have modified the AI output to follow this project's coding standards and conventions

-- What problem does your code solve?

Docs teach how to enable remote sync with only two env vars (OPENSRE_REMOTE_SYNC + OPENSRE_REMOTE_SYNC_BUCKET). Everything else defaults (opensre prefix, aws provider, no exclusions). Before this change, that happy path would break whenever ~/.opensre/config.yml had been broken: the switch and bucket were satisfied from env, but the optional fields still fell thru to the stored remote_sync section — and an unreadable file raised RemoteSyncConfigError there. A user followed the docs and got sync disabled (or crashed) due to an unrelated damaged file, despite having provided everything require.

-- What alternative approaches did you consider?

If any env var is set, ignore the stored file completely. Rejected - that would silently drop stored settings (e.g. a stored exclude list) when a user exported a single var for one run. Env precedence should be field-based, not file-based.
2. Soft-fail only for fields that have env values Rejected – prefix provider region profile exclude – each falls thru to file independently. If only two env vars set and all would read corrupt file and raise. Special case 5 all.
3. Treat an unreadable stored section as "no stored settings" (as I did). This matches the existing behavior for a missing file (read_section returns {} -> defaults apply) so corrupt files behave like missing files, the least

--Why did you choose this specific implementation?

A corrupt file is indistinguishable from “no stored settings” at the config layer, and the missing-file path was already documented behavior. The risk of “exclude nothing” is bounded: the credential deny-list in syncable.py is enforced separately at sync time, so no pattern (or lack of one) can ever upload integrations.json/llm-auth.json. And the fail-closed guaranty is preserved exactly where it should be – the two values that can only come from the file still raise.

Key functions/components:
- _stored_section() → dict | None: returns None on LocalSettingsError instead of raising, so callers can distinguish "no stored settings" from "stored settings exist".
- _required_stored() (new): raises RemoteSyncConfigError when the section is None. Used only where a value has no env source: enabled when OPENSRE_REMOTE_SYNC is unset (in _enabled), and bucket when OPENSRE_REMOTE_SYNC_BUCKET is unset (in load_remote_sync_config). This is what keeps the two fail-closed paths from the issue working: switch unset / bucket only in file.
- _env_or_stored() and _exclusions(): return the empty/default when the section is None, so optionals fall back to opensre, aws, empty region/profile, NO_EXCLUSIONS.
- _lazy_stored(): still caches the result (including None), so the file is parsed at most once per run.
- Unchanged: _validated_scalar() still rejects malformed values in a readable file (bucket: [list] etc.), so typos in a valid YAML file still fail loudly and name the key.

---

- [ ✓ ] I have added proper PR title and linked to the issue
- [ ✓ ] I have performed a self-review of my code
- [ ✓ ] **I can explain the purpose of every function, class, and logic block I added**
- [ ✓ ] I understand why my changes work and have tested them thoroughly
- [ ✓ ] I have considered potential edge cases and how my code handles them
- [ ✓ ] My code follows the project's style guidelines and conventions

---


activateLabhesh:
@greptile review

activateLabhesh:
@Davidson355, I think this pr is ready to be merged! just look into it once when you are free.

Issue #4426, opened 2026-07-28
[FEATURE] Background RCA notifications and retrieval: non-email channels, inbound commands, preference model


larsspinetta12:
## Background

#2655 was the post-v1 follow-up to #2654 (background investigation mode with email-first notifications). The messaging gateway (#1483) has shipped and provides the infrastructure foundation. This issue tracks the remaining acceptance criteria from #2655 that aren't covered by the gateway alone.

### 1. Background RCA notifications on non-email channels

Wire at least one non-email channel (Telegram is the primary candidate given existing gateway support) end-to-end for background RCA completion delivery. When a background investigation finishes, the user should receive the summary notification on their configured non-email channel without any manual retrieval step.

Decisions needed:
- Which channel gets priority when multiple destinations are configured
- Whether background notifications fan out to multiple channels or use one preferred default
- How delivery failures surface when one channel fails and another succeeds

### 2. External RCA retrieval from communication channels

Allow users to retrieve completed background RCA results from a messaging channel rather than only from the local REPL. Examples:
- List recent background jobs from Telegram
- Request a completed RCA summary by task ID via message command
- Resend findings/next-steps into a preferred destination on demand

### 3. Inbound command surfaces for background RCA

Evaluate and implement inbound command/reply surfaces for background RCA workflows:
- Trigger a fresh investigation from an external channel
- Query a completed RCA by task ID
- Ask follow-up questions against a completed RCA from an external channel

Build on `gateway.dispatch_inbound()` (#1483) rather than introducing a parallel architecture.

### 4. Notification preference model

Define and implement the policy model for multi-channel notification preferences:
- Preferred channel ordering
- Multi-channel fanout vs first-success delivery
- Per-channel recipient overrides
- User-specific vs machine-local defaults
- Per-job notification override behaviour

## Acceptance criteria

- At least one non-email channel is wired end-to-end for background RCA completion delivery
- Users can retrieve a completed background RCA summary from an external communication channel
- At least one inbound command surface is implemented for background RCA lookup
- Notification preference handling is documented and implemented for multi-channel setups
- Implementation uses `MessagingGateway` (#1483) rather than direct platform calls

## Related

- #2654 — Core background investigation mode (completed)
- #1483 — MessagingGateway / unified messaging provider (completed)
- #1482 — Conversational loop / inbound messaging (completed)
- #2655 — Predecessor issue (closed; superseded by this issue)

SawhneySatvik:
[...]

PR #4840, opened 2026-08-10
fix(interactive-shell): stop /background raising on headless sessions

SawhneySatvik:
Part of #4426

A `/`-prefixed message from Telegram, Slack or Discord is dispatched straight to
`slash_invoke`, and `/background` is in the command enum. But chat turns run against a
`SessionCore`, which has no terminal facet, and `/background` dereferenced
`session.terminal` at eleven sites, so every form raised `AttributeError` out of
`dispatch_slash`.

Nobody saw it. The react loop logs the exception server-side and the slash outcome is
recorded before the handler runs, so the turn came back reported as succeeded with no
content, which is harder to diagnose than a failure.

`core/agent_harness/session/terminal_access.py` already exists for this split, reading the
shell terminal when present and falling back otherwise. `/background` was the one command
bypassing it. Read forms now go through two new accessors beside the existing
`background_mode_enabled`. Write forms have no headless equivalent, so `on`, `off` and
`notify set` name the REPL command to run, following `/onboard` and `/integrations setup`.

Read forms report empty state on a gateway session, because records are process-local to
the REPL. Sharing them across processes is part of the remaining #4426 scope.

Recorded against the real Telegram Bot API through a running `opensre gateway`, not mocked.
Same chat, same session, two minutes apart, with only the checked-out branch changed
between them.

| Sent from Telegram | `main` | this branch |
| --- | --- | --- |
| `/background list` | `(succeeded)` and nothing else | `(succeeded)` + `no background investigations tracked in this session.` |
| `/background on` | same silent failure | `(failed)` + `Run it in the REPL: uv run opensre` |
| `/background notify list` | same silent failure | `(succeeded)` + `background notify channels: none` |

From one continuous gateway log:

```
07:14:23 WARNING core.execution | [tool:slash_invoke] failed: 'SessionCore' object has no attribute 'terminal'
07:14:23 INFO  gateway | tool result name=slash_invoke ok=False size=61
07:14:24 INFO  gateway | outbound text='slash /background list (succeeded)'

07:16:19 INFO  gateway | tool result name=slash_invoke ok=True size=115
07:16:20 INFO  gateway | outbound text='slash /background list (succeeded) no background investigations tracked in this session.'
```

On `main` the tool returns `ok=False` and the user is told `(succeeded)` with nothing after
it. That is the defect: not a crash anyone can report, but a turn that looks like it worked.

One condition worth stating plainly: #4816 added a principal gate to the Telegram inbound
handler, so this capture was taken with `ORGANIZATION_ID` set. The defect and the fix are
unchanged, but post-#4816 the path is only reachable on an org-bound deployment, and I
would rather flag that than present an org-bound run as a stock local one.

---

I reproduced it before changing anything: `/background list` against a `SessionCore` raises
`AttributeError` straight out of `dispatch_slash`.

Two alternatives were rejected. Giving `SessionCore` a terminal facet pushes REPL
presentation state into the type every surface uses, to serve one command. Guarding each of
the eleven sites with `getattr` duplicates what `terminal_access.py` already centralises,
which is how `/status` already works headlessly.

The read/write split is the substantive decision: reads have a sensible headless answer,
writes do not, and failing loudly beats a silent no-op that leaves the user believing
background mode is on. Accessors are typed `Any` to match the module, because `core/` cannot
import the record type from `surfaces/` under the layering contract. Tests cover the command
layer against `SessionCore` and a round trip through `GatewayTurnHandler`.

---


muddlebee:
@greptile review

muddlebee:
@SawhneySatvik good work :+1:

SawhneySatvik:
Thanks @muddlebee !

PR #4850, opened 2026-08-10
fix(agent-harness): typed session-goal fields, and stop empty goal state reaching the transcript

YauhenBichel:
Follow-up hardening on the outer-goal work. Six defects, all found by running
the suite rather than reading it.

**A packed string became typed fields.** `session_goal` accepted three
spellings of one concept: `continue`, `max_turns=<n>`, and
`max_turns=<n>;steps=<n>`. A live planner picked whichever it liked, so a
fixture pinning one spelling failed against a schema-legal answer. Attach is
now a boolean and the cap is an integer `session_goal_max_turns`, so there is
exactly one way to say each thing. The packed grammar is gone from the tool
schema, the parser, and the planner prompt.

A checklist now implies attach. Observed live: the planner emitted
`session_goal_items` and omitted `session_goal`, a half-state the host had to
guess at. Items only mean something under a goal, so attach is derived.

**Empty goal state was written to every transcript.** The session flush
guarded on `hasattr(session, "session_goal")` — always true for a dataclass
field — so every flush appended
`{'session_goal': None, 'offered_upgrade_ctas': [], ...}`. That record became
the transcript's last line and re-parented leaf entries, which is what
`test_flush_leaf_parents_to_last_real_entry_not_trace_span` caught. Now guarded
on content, matching the sibling `if session.accumulated_context:`.

**A gateway streaming regression.** An unconditional
`_finish_streamed_response` on `gather_and_answer` fired whenever the
investigation-offer branch was skipped — exactly the gateway, which disables
that capability. An already-streamed answer was pushed through `finalize`, so
the user saw it twice. The flush is now conditional on the text actually having
been rewritten.

**An import cycle** between `assistant_handoff` and `evidence_need`, broken
structurally per AGENTS.md by extracting `EvidenceKind` to a leaf both import.
That also deleted `_EVIDENCE_KIND_VALUES`, a frozenset duplicating the enum's
four members.

**Payload keys are enums.** `HandoffField` / `HandoffTag` in `handoff_keys.py`.
These names appear in three places that must agree — the tool schema, the
decoder, and the prefix-keyed guidance table — and there were 26 literal
occurrences. Adding a handoff field is now an enum member plus one decode line.

**Two stale imports** after the orchestrator split into `answer_finalize` /
`handoff_policy` / `turn_route`, and a type leak where `policy_for` was
imported from the wrong module, making its `bool` return `Any`.

    pytest (CI scope)   14,138 passed, 32 skipped, 1 failed (see below)
    mypy                Success: no issues found in 1707 source files
    ruff check/format   All checks passed, 3332 files
    check_imports       All 3 import checks passed

Before, every flush wrote this even with no goal:

    {"session_goal": null, "offered_upgrade_ctas": [], "pending_integration_setup_offer": null}

Before, the planner could emit any of:

    session_goal: "continue"
    session_goal: "max_turns=5"
    session_goal: "max_turns=5;steps=5"

Now:

    session_goal: true
    session_goal_max_turns: 5


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4852, opened 2026-08-10
fix(session-goal): judge goal completion on host evidence, not the model's claim

YauhenBichel:
An outer `SessionGoal` used to finish when the model emitted
`session_goal:achieved`. That tag is the agent grading its own work, so a turn
could end with the goal marked done and nothing to show for it.

**The host now judges independently.** `session_goal_evaluate.py` holds the
structured verdict:

* checklist complete (via host-tracked `done=` indices) → achieved
* `achieved` claimed but the checklist is incomplete → **tag ignored**, stay
  active, and report which item is next
* condition-only goal → require tool evidence on the turn, else stay active

`session_goal_review.py` adds an optional second opinion from the model, but
only for the condition-only path — the case with no structured signal to check.
Checklist completion stays structured-only. The reviewer **fails open to
ACTIVE** on any LLM error, so a broken reviewer cannot false-complete a goal.

**Evidence means work that worked.** The evidence predicate was
`success > 0 or executed > 0`. The first clause is dead (a success is an
execution), so the rule collapsed to "any tool ran" — meaning a tool that ran
and *errored* satisfied it, letting an `achieved` claim through on a failed
call. It now requires `executed_success_count > 0`.

Note the behaviour change for review: a goal that previously completed off a
failed tool call now stays active until its budget runs out. That is the
intended semantics, but it does change when goals finish.

**One paint per outer turn.** The loop gained `_finish_outer_turn`, a single
"refresh → evaluate → paint" path, so progress is reported with the evaluated
reason instead of a pre-evaluation guess. The gateway paints a compact status
line to chat sinks and the full checklist to debug logs, matching the shell.

Also: named `MAX_REVIEWED_REPLY_CHARS` for the reviewer's context budget, and
named character budgets in `session_goal.py` (`MAX_GOAL_REASON_CHARS`,
`MAX_GOAL_CONDITION_CHARS`, `MAX_GOAL_STATUS_LINE_CONDITION_CHARS`) replacing
four `limit` / `limit - 3` pairs where the `-3` silently encoded `len("...")`.
Those now delegate to the existing `truncate_message` helper.

    pytest (CI scope)   14,174 passed, 32 skipped, 0 failed
    mypy                Success: no issues found in 1709 source files
    ruff check/format   All checks passed, 3336 files
    check_imports       All 3 import checks passed

Before — a claim with no successful work completed the goal:

    executed_count=1, executed_success_count=0, reply contains "session_goal:achieved"
    → ACHIEVED

After:

    → ACTIVE, reason "achieved tag ignored; no tool evidence yet"

---

The split between a structured judge and an optional LLM confirm is deliberate.
Checklist progress is already host-tracked, so asking a model to re-judge it
adds cost and a way to be wrong; the model is only consulted where there is no
structured signal. The reviewer fails open to ACTIVE for the same reason — the
failure mode of a second judge should be "keep working", never "declare
victory".

The evidence fix changes the definition rather than the threshold. `executed`
answers "did the agent try", which is not what completion needs; the existing
test asserted the old behaviour, so it was rewritten to state all three cases
explicitly rather than deleted.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4874, opened 2026-08-11
refactor(gateway/slack): group transport by concern and split the streamed-reply state machine

YauhenBichel:
`gateway/transports/slack/` was 16 flat modules. Nothing indicated which ones
receive a Slack event, which deliver a reply, and which own the Socket Mode
connection — so the grouping now says it:

slack/
client.py, settings.py, startup.py   # Slack API surface + entrypoint
connection/   worker.py, heartbeat.py
inbound/      events, security, principal, dispatcher, attachments, thread_history
outbound/     output_sink, turn_stream, approvals, feedback, channel_intro

`startup.py` stays at the package root — it is the only module imported from
outside the package (`gateway/channels/chat.py`), so the public entrypoint is
unchanged.

## Separation of concerns

**`output_sink.py` held two classes.** `SlackOutputSink` (the sink adapter) and
`_TurnStream`, a ~190-line `chat.startStream` state machine owning stream
lifecycle, chunk throttling, and open-task bookkeeping. Every other transport's
`output_sink.py` — Discord, Telegram, Buzz — contains exactly one class; Slack
was the outlier, and `_TurnStream` already had its own test file.


YauhenBichel:
@greptile review

PR #4876, opened 2026-08-11
feat(gateway/slack): Events API HTTP inbound so production is not on Socket Mode

YauhenBichel:
Socket Mode was the only inbound transport that existed. That is single-consumer, which caps production: no horizontal scale, no rolling deploy without dropping events, and a second replica splits the event stream rather than sharing it. The constraint was already costing us — `prep_ec2_deployment.py` and the Telegram gateway lifecycle deliberately withhold `SLACK_*` tokens so a deployed gateway will not compete for the socket.

Slack now chooses its inbound transport. Socket Mode stays the default for local development — no public URL, works behind NAT. Production can run Events API over HTTP, where any replica can serve a signed POST.

The config layer already anticipated this: `SlackBotConfig.signing_secret` existed with "MUST be set for Events API HTTP" and was never read. Only the transport was missing.

**Slack serves its own HTTP listener** (`connection/http_server.py`) on its own port — not a route bolted onto `gateway.web`. That app is the health and alert surface, and the package DAG forbids it importing a chat transport. This also
keeps the two inbound transports symmetrical: Socket Mode owns its connection in `connection/worker.py`, Events API owns its server next to it.

Two routes, because Slack posts button clicks to a **separate** request URL. Without `/slack/interactivity`, approvals and feedback stop working while turns still run — a silent failure.

**`connection/signature.py`** — the security boundary Socket Mode provided implicitly. HMAC over the raw request bytes, `hmac.compare_digest` so rejection is constant-time, and a ±5-minute replay window. A non-numeric timestamp fails closed rather than parsing to `0`, which would read as "very old" and give the right answer for the wrong reason.

**`connection/http_receiver.py`** — framework-agnostic admission returning a typed `SlackHttpStatus` (`CHALLENGE` / `REJECTED` / `IGNORED` / `ACCEPTED`), so the `url_verification` handshake, signature failures, duplicates, and real events are distinguishable without a server in the test.

**`connection/turn_stack.py`** — the collaborators both transports need (`WebClient`, executor, dispatcher, greeter, approval broker, binding store, bot user id) built once. Previously assembled inline in the socket worker; duplicating it for HTTP would have been ~35 lines of copy.

**Transport selection** in `startup.py` is a table keyed on the enum, one row per transport rather than a branch chain. Added `inbound_transport`, `signing_secret`, and `http_port` to settings, with per-transport credential validation — Socket Mode requires `app_token`, HTTP requires `signing_secret`. Previously `choose_app_token` raised unconditionally, so a correct HTTP config was rejected at boot.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4916, opened 2026-08-12
chore(graft): add repo context graph config for agent tooling

joseedson18jc:
No linked issue — this is tooling-only config, no product change.

Adds the output of `graft init` (Graft 0.9.1), which indexes this repo into a
context graph so coding agents can locate code without grepping or reading whole
files.

What is included:

- **MCP server registration** for the agents this repo already carries config
  for: `.mcp.json` (Claude Code), `.cursor/mcp.json`, `.gemini/settings.json`,
  `.kiro/settings/mcp.json`, `opencode.json`. Each one just runs `graft mcp`
  over stdio; no keys, no network calls, no credentials.
- **Usage instructions** appended to `AGENTS.md` between `<!-- graft:start -->`
  / `<!-- graft:end -->` markers, and mirrored into `GEMINI.md`,
  `.github/copilot-instructions.md`, `.cursor/rules/graft.mdc`,
  `.windsurf/rules/graft.md`, and `.kiro/steering/graft.md`.
- **`.claude/helpers/graft-hooks.cjs` and `graft-statusline.cjs`** plus the
  matching `hooks`/`statusLine` entries and `Bash(graft:*)` permissions in
  `.claude/settings.json`.
- **`.gitignore`**: the generated graph (`graft/`) is not committed — it is
  regenerable with `graft build` (deterministic, no API key). **`.ignore`**
  re-admits that directory to ripgrep, which reads `.ignore` before
  `.gitignore`, so the cards stay greppable locally.

No application code, dependencies, or CI configuration is touched. Contributors
who do not use these agents are unaffected; nothing here runs during build,
test, or deploy.

Querying the graph for gateway auth, from this branch:

```
$ graft ask "gateway request authentication" --in gateway/
graft ask — "gateway request authentication"  (lexical)

1. request · method  [symbol]
   gateway/slack/approvals.py:L106-L144
...
5. _gateway_auth_error · function  [symbol]
   gateway/http/webapp.py:L105-L123
   def _gateway_auth_error(request: Request) -> JSONResponse | None
7. receive_alert · function  [symbol]
   gateway/http/webapp.py:L127-L171
```

That pointed straight at the real answer (bearer-token check with a loopback
fallback, shared by `/alerts` and `/investigate`) from one query plus a 30-line
read, instead of opening `gateway/http/webapp.py` in full.

MCP server handshake, verified over stdio:

```
$ printf '...initialize...tools/list...' | graft mcp
initialize OK: {'name': 'graft', 'version': '0.9.1'}
tools: graft_find_code, graft_file_api, graft_check_freshness,
       graft_trace_calls, graft_find_all, graft_repo_map
```

---

To be precise about provenance: the committed files were generated by the
`graft init` CLI, not written by a model. Claude Code was used to review the
generated output, decide what to stage, and write this PR. No hand-written or
model-written application logic is included.

The problem: agents working in this repo burn context re-reading large files to
answer "where does X happen". Graft builds a local graph of small markdown nodes
carrying exact `file:line` spans, so a question resolves to a ranked set of spans
instead of a file dump.

Alternatives considered: (a) do nothing and keep relying on grep — works, but
costs context on every task; (b) commit the generated `graft/` graph so
contributors skip the build — rejected, since it is derived data that would churn
on every code change and pollute diffs; (c) keep all of this untracked and local
per-developer — rejected because the instruction blocks and MCP registrations are
exactly the parts worth sharing, and they are inert for anyone who does not
install Graft.

Key pieces: `.mcp.json` and its per-agent equivalents register `graft mcp` as a
stdio MCP server exposing six read-only query tools. The `AGENTS.md` block tells
agents to query the graph before grepping. `graft-hooks.cjs` reacts to
Write/Edit and Bash tool events to keep the local graph fresh and report token
savings; `graft-statusline.cjs` renders that in the status line. Both are
local-only developer conveniences.

Edge cases: the graph directory is ignored by git but explicitly re-admitted to
ripgrep via `.ignore`, so a stale or absent `graft/` degrades to "run
`graft build`" rather than breaking anything. Contributors without the `graft`
binary see the MCP server simply fail to start in their agent, with no effect on
the repo itself.

---

### DIFF
--- a/.claude/helpers/graft-hooks.cjs
+++ b/.claude/helpers/graft-hooks.cjs
@@ -37,7 +37,16 @@ function entry(name) {
     const f = path.join(d, name);
     if (fs.existsSync(f)) return f;
   }
-  return path.join(dir, 'dist', 'claude', name); // last-ditch; import will no-op if absent
+  // No trusted candidate holds the entrypoint, so there is nothing to run.
+  // This used to fall back to `<project>/dist/claude/<name>` -- which, on a
+  // machine without graft installed, imports and executes that file straight
+  // out of whatever repository happens to be open. A clone carrying its own
+  // dist/claude/hooks.js would run on session-start, post-edit and stop with
+  // no install step and no prompt. Absent graft, the hook must do nothing.
+  return null;
 }
 
-import(pathToFileURL(entry("hooks.js")).href).then((m) => m.main(process.argv[2])).catch(() => { /* graft unavailable — no-op */ });
+const hooksEntry = entry("hooks.js");
+if (hooksEntry) {
+  import(pathToFileURL(hooksEntry).href).then((m) => m.main(process.argv[2])).catch(() => { /* graft unavailable — no-op */ });
+}
--- a/.claude/helpers/graft-hooks.cjs
+++ b/.claude/helpers/graft-hooks.cjs
@@ -1,0 +1,43 @@
+#!/usr/bin/env node
+const path = require('path');
+const fs = require('fs');
+const { pathToFileURL } = require('url');
+const { execFileSync } = require('child_process');
+const dir = process.env.CLAUDE_PROJECT_DIR || process.cwd();
+// Optional per-developer override: absolute path to @nanonets/graft's dist/claude.
+const BAKED = process.env.GRAFT_CLAUDE_DIR || "";
+
+// The dist/claude dir of @nanonets/graft resolved from a base whose node_modules is searched.
+function fromPkg(base) {
+  try {
+    const pkg = require.resolve('@nanonets/graft/package.json', { paths: [base] });
+    return path.join(path.dirname(pkg), 'dist', 'claude');
+  } catch { return null; }
+}
+
+// The global node_modules dir per npm (handles Homebrew/Windows/volta). Queried on demand.
+function globalRoot() {
+  try {
+    const root = execFileSync('npm', ['root', '-g'], { encoding: 'utf8', stdio: ['ignore', 'pipe', 'ignore'], shell: process.platform === 'win32' }).trim();
+    return root || null;
+  } catch { return null; /* npm unavailable */ }
+}
+
+function candidates() {
+  const out = [];
+  if (BAKED) out.push(BAKED);
+  const local = fromPkg(dir); if (local) out.push(local);
+  const legacy = fromPkg(path.join(path.dirname(process.execPath), '..', 'lib')); if (legacy) out.push(legacy);
+  const gr = globalRoot(); if (gr) out.push(path.join(gr, '@nanonets', 'graft', 'dist', 'claude'));
+  return out;
+}
+
+function entry(name) {
+  for (const d of candidates()) {
+    const f = path.join(d, name);
+    if (fs.existsSync(f)) return f;
+  }
+  return path.join(dir, 'dist', 'claude', name); // last-ditch; import will no-op if absent
+}
+
+import(pathToFileURL(entry("hooks.js")).href).then((m) => m.main(process.argv[2])).catch(() => { /* graft unavailable — no-op */ });
--- a/.claude/helpers/graft-hooks.cjs
+++ b/.claude/helpers/graft-hooks.cjs
@@ -32,21 +32,37 @@ function candidates() {
   return out;
 }
 
-function entry(name) {
+// Every trusted candidate that actually holds the entrypoint, best first.
+//
+// There is deliberately no fallback to `<project>/dist/claude/<name>`: on a
+// machine without graft installed that imported and executed the file straight
+// out of whatever repository happened to be open, so a clone carrying its own
+// dist/claude/hooks.js would run on session-start, post-edit and stop with no
+// install step and no prompt. Absent graft, these helpers do nothing.
+function entries(name) {
+  const out = [];
   for (const d of candidates()) {
     const f = path.join(d, name);
-    if (fs.existsSync(f)) return f;
+    if (fs.existsSync(f)) out.push(f);
   }
-  // No trusted candidate holds the entrypoint, so there is nothing to run.
-  // This used to fall back to `<project>/dist/claude/<name>` -- which, on a
-  // machine without graft installed, imports and executes that file straight
-  // out of whatever repository happens to be open. A clone carrying its own
-  // dist/claude/hooks.js would run on session-start, post-edit and stop with
-  // no install step and no prompt. Absent graft, the hook must do nothing.
-  return null;
+  return out;
 }
 
-const hooksEntry = entry("hooks.js");
-if (hooksEntry) {
-  import(pathToFileURL(hooksEntry).href).then((m) => m.main(process.argv[2])).catch(() => { /* graft unavailable — no-op */ });
-}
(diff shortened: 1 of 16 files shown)


joseedson18jc:
Addressed the Greptile finding on `.claude/helpers/graft-hooks.cjs` and `.claude/helpers/graft-statusline.cjs`.

Both helpers baked an absolute, author-specific install path as the first module-resolution candidate. That path is now an optional environment override, so resolution starts from portable lookups:

```js
// Optional per-developer override: absolute path to @nanonets/graft's dist/claude.
const BAKED = process.env.GRAFT_CLAUDE_DIR || "";
```

Order is now: `GRAFT_CLAUDE_DIR` (unset by default) → project `node_modules` → node prefix `lib` → `npm root -g`. Verified locally with no override set — the status line still resolves and renders (`graft · 28070 nodes / 60669 edges · ✓ synced`), now via the `npm root -g` candidate.

Also rebased onto current `main` to clear a `.gitignore` conflict; both sides' entries were kept.

joseedson18jc:
@nowhere-in-space @YauhenBichel — tooling-only PR, no application code, no dependency or CI changes.

It adds the output of `graft init`: MCP server registration for the agents this repo already carries config for (Claude Code, Cursor, Gemini, Kiro, OpenCode) plus the matching usage instructions in `AGENTS.md` and its mirrors. The generated graph itself stays gitignored — it is regenerable with `graft build`, and `.ignore` re-admits it to ripgrep so the cards stay locally greppable. Contributors who do not use these agents are unaffected.

Greptile scored it 4/5 with a single finding: both Claude helpers baked an author-specific absolute path as the first module-resolution candidate. That is fixed in `e21595cb` — the path is now an optional `GRAFT_CLAUDE_DIR` override, with portable lookups first. The branch is also rebased onto current `main` (one `.gitignore` conflict, both sides' entries kept).

Would appreciate a look when you have a moment.

joseedson18jc:
@greptileai review

Re-review requested: the P2 finding on `.claude/helpers/graft-hooks.cjs` and `.claude/helpers/graft-statusline.cjs` (author-specific absolute path taking module-resolution precedence) was fixed in `e21595cb`. The baked path is now an optional `GRAFT_CLAUDE_DIR` environment override, unset by default, so resolution starts from portable lookups.

The last review covered `f58802c9`, which predates that fix.

joseedson18jc:
@nowhere-in-space @YauhenBichel — following up on this one.

The Greptile score still shows 4/5, but that review covers `f58802c9` and predates the fix. Its single P2 — the two Claude helpers baking an author-specific absolute path ahead of portable module resolution — is addressed in `e21595cb`: the path is now an optional `GRAFT_CLAUDE_DIR` override, unset by default, so resolution starts from `node_modules`, then the node prefix, then `npm root -g`. The bot has not re-run against the new head, so the open thread is stale rather than outstanding.

The branch is rebased onto current `main` and CI is green.

Scope is tooling-only — `graft init` output plus matching `AGENTS.md` guidance. No application code, dependencies, or CI changes. It merges into `.claude/settings.json` rather than replacing it, the generated graph stays gitignored, and contributors who do not use these agents are unaffected.

Happy to split or drop any part that is unwelcome.

joseedson18jc:
@greptile review

joseedson18jc:
Pushed `dcdbfc06` with two fixes found by automated review on a sibling repository carrying the same `graft init` output.

**1. The helpers could import code out of the open repository.** `entry()` ended with a fallback to `<project>/dist/claude/<name>`, and that path went straight into a dynamic import:

```js
return path.join(dir, "dist", "claude", name); // last-ditch
...
import(pathToFileURL(entry("hooks.js")).href).then((m) => m.main(process.argv[2]))
```

With `dir = CLAUDE_PROJECT_DIR || cwd()`, a contributor without graft installed would have that file imported and executed straight out of whatever repository was open — on session-start, post-edit and stop, with no install step and no prompt. A clone shipping its own `dist/claude/hooks.js` would simply run. `entry()` now returns `null` when no trusted candidate holds the entrypoint, and the import is skipped.

Verified both directions: with the old resolution a planted `dist/claude/hooks.js` executed; with the new one nothing runs, and the status line still resolves normally where graft is installed.

**2. The instructions contradicted `.gitignore`.** Six agent instruction files described `graft/` as "kept in sync with the code through git" while `.gitignore` excludes it, so a fresh clone was told to query a graph it does not have. They now describe it as a local regenerable cache and say to run `graft build` when it is absent or stale.

The previous 5/5 covered `e21595cb`, so it no longer applies to this head.

@greptile review

joseedson18jc:
Pushed `e0cd5ecb`, closing the remaining P1.

Every MCP registration runs a bare `graft`, so a contributor following the setup without the CLI on `PATH` got command-not-found at server start and no context service, with nothing in the instructions explaining why. The prerequisite is now stated up front in all six instruction files, along with `npm i -g @nanonets/graft`, the fact that nothing here vendors or installs it, and that contributors who skip it are otherwise unaffected.

That is all three findings addressed across `dcdbfc06` and `e0cd5ecb`: the helper import fallback, the `graft/` lifecycle wording, and now the CLI prerequisite.

@greptile review

joseedson18jc:
Pushed `d3d2285c` with two more fixes, both surfaced by review on a sibling repo carrying the same `graft init` output.

**Candidate resolution stopped at the first hit.** The launchers took the first candidate path that existed and imported only that one, so a stale or half-installed graft masked a working installation further down the list. The failure mode is a hook that silently does nothing — indistinguishable from graft not being installed. They now walk the candidates in order, and only a *load* failure falls through: once `main()` has been reached it owns the outcome, since retrying the next candidate after it ran would repeat its side effects. A module without a callable `main()` is skipped as unusable rather than trusted.

Verified with a broken first candidate and a valid second: the second runs. The status line still resolves normally where graft is installed, and there is still no fallback into the open repository.

**`.windsurf/rules/graft.md` had no frontmatter.** A Windsurf workspace rule without a `trigger` field is never activated, so that file has been inert since it was added. It now declares `trigger: always_on`.

@greptile review

joseedson18jc:
Pushed `c8a26954`. While auditing my own earlier fix I found it had narrowed the problem rather than closed it, so this corrects the record.

Removing the `<project>/dist/claude/<name>` fallback shut the obvious path, but `candidates()` still calls `fromPkg(dir)`, which resolves `@nanonets/graft` through the **open repository's own node_modules** — and that lookup was ranked *second*, ahead of the installation the machine's owner actually chose. A clone shipping `node_modules/@nanonets/graft/dist/claude/hooks.js` therefore still executed on session-start, post-edit and stop, beating a real global install, with nobody running an install step.

Candidates are now ordered by how much the machine's owner controls them rather than by proximity, so the project-local lookup comes last. A project-local graft stays supported; it is simply reached only when nothing trusted holds the entrypoint.

Verified both directions by printing the resolution order:
- hostile `node_modules/@nanonets/graft` present **and** a real global install → global wins, repo copy passed over
- no global install → project-local still resolves

Worth flagging that all five defects on this PR are in `graft init`'s generated output, so they exist in every repository it has wired, not just this one.

@greptile review

muddlebee:
such big change need issue approval

Issue #4903, opened 2026-08-11
[GOOD FIRST ISSUE] SM-SR-LOG-3: Log soft-fail in Jira client


muddlebee:
Supports Python API / repeat usage — teammate.
Audience: community
Kind (maintainer): n/a
Task id: SM-SR-LOG-3
Code PRs: TDD; AAA (Arrange→Act→Assert); clear names; in-tree patterns; SOLID.
Not: `core/` product edits.

---

**Selected task:** `SM-SR-LOG-3`

**Selected file:** `integrations/jira/client.py`

| Field | Value |
|-------|-------|
| **Audience** | Community |
| **Level** | Junior |
| **Complexity** | S |
| **Good first issue** | Yes |
| **North star** | Quality |

**Summary:** Same pattern as SM-114: keep soft-fail return, add
`debug`/`warning` log with `exc_info=True` (or exception type). Soft-fail
without log still present in Vercel, Argo CD, and Jira clients (verified).

**GitHub identity soft-fail:** already filed as
[#4648](https://github.com/Tracer-Cloud/opensre/issues/4648) **SM-114** —
**claim that issue**; do **not** invent SM-SR-LOG for GitHub identity.

| ID | File |
|----|------|
| SM-SR-LOG-1 | `integrations/vercel/client.py` (`except Exception: return None` on factory path) |
| SM-SR-LOG-2 | `integrations/argocd/client.py` |
| SM-SR-LOG-3 | `integrations/jira/client.py` |

**Do not:** Change return types. One ID = one file = one PR.

**Done when:** [ ] Soft-fail still returns the previous sentinel
[ ] Failure path logs at debug/warning with exception context

**Files:** exactly one client module per ID

---

deepspace28:
Hi maintainers — I'd like to pick this one up (SM-SR-LOG-3, `integrations/jira/client.py`). Could I be assigned?

I'm already assigned to the sibling **SM-SR-LOG-1** (#4901) and its PR is open at #4919, which came back **Greptile 5/5, no actionable defects**. This would be the same change applied to the Jira factory, as a separate PR per the "one ID = one file = one PR" rule.

Approach, unchanged from the reviewed sibling:

- Keep the soft-fail contract exactly — `make_jira_client()` still returns `None`, no return-type or signature change.
- Bind the exception in the existing `except` at the end of the factory and log at `WARNING` with `exc_info=True`, mirroring `integrations/pagerduty/client.py`.
[...]

PR #4919, opened 2026-08-12
fix(integrations/vercel): log soft-fail when client construction fails

### LINKED ISSUE #4901: [GOOD FIRST ISSUE] SM-SR-LOG-1: Log soft-fail in Vercel client
Audience: community
Kind (maintainer): n/a
Task id: SM-SR-LOG-1
Not: `core/` product edits.

---

**Selected task:** `SM-SR-LOG-1`

**Selected file:** `integrations/vercel/client.py`

| Field | Value |
|-------|-------|
| **Audience** | Community |
| **Level** | Junior |
| **Complexity** | S |
| **North star** | Quality |

| ID | File |
|----|------|

---

deepspace28:
Fixes #4901

`make_vercel_client()` ended in a bare `except Exception: return None`. Returning
construction failure (bad token shape, malformed team id, a `VercelConfig`
validation error) was indistinguishable from "Vercel is simply not configured" —
the exception was discarded entirely.

- `integrations/vercel/client.py` — `except Exception as exc:` now logs at
- `tests/integrations/vercel/test_client.py` — new AAA unit test that forces
  `VercelConfig` construction to raise and asserts **both** that the factory
  still returns `None` **and** that a `WARNING` record carrying exception
  context was emitted.

Return type, signature, and every success path are untouched.

The wording and log level deliberately mirror the already-merged soft-fail log in
`integrations/pagerduty/client.py` (SM-115), and the test mirrors
`tests/integrations/pagerduty/test_client.py::test_make_client_logs_soft_fail_on_construction_error`,
so the three sibling clients stay consistent.

Per the issue's "one ID = one file = one PR" rule, this covers **SM-SR-LOG-1**
only. `integrations/argocd/client.py` (#4902) and `integrations/jira/client.py`
(#4903) are left alone.

Same forced failure (`VercelConfig` raising `ValueError`), run on `main` and on
this branch:

**Before — silent**

```
$ python demo.py

```

```
$ python demo.py
WARNING integrations.vercel.client: Failed to create Vercel client: api_token must not contain whitespace
  File "D:\oss\opensre\integrations\vercel\client.py", line 587, in make_vercel_client
    return VercelClient(VercelConfig(api_token=token, team_id=team_id or ""))
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    raise ValueError("api_token must not contain whitespace")
ValueError: api_token must not contain whitespace

```

```
$ pytest tests/integrations/vercel/test_client.py -k soft_fail -q
FAILED tests/integrations/vercel/test_client.py::test_make_vercel_client_logs_soft_fail_on_construction_error
1 failed, 79 deselected

# after the one-line fix (GREEN)
$ pytest tests/integrations/vercel/ -q
80 passed in 22.28s
```

**Required checks**

```
All checks passed!

3369 files already formatted

$ pytest tests/integrations -n auto -q
2 failed, 3242 passed, 7 skipped
```

The two failures are `tests/integrations/git/test_local.py::test_changed_paths_handles_quoted_filenames`
and `tests/integrations/llm_cli/test_codex_oauth.py::test_codex_auth_payload_and_write_auth_shape`.
Both reproduce identically on a clean checkout of this branch's base commit with
my changes stashed — they are pre-existing Windows-environment failures,
unrelated to this PR, and neither touches the Vercel integration.

---

"no Vercel client available" and carry on. That part is correct and I did not
want to change it. What is wrong is that the bare `except Exception` bound no
name and logged nothing, so the exception object was destroyed before anyone
could see it. Debugging a misconfigured Vercel integration meant reading the
source to discover that a silent swallow was even possible.

   against `Client | None`. It would also make an optional integration able to
   break startup.
2. *Route it through `capture_service_error`*, which the request methods in this
   same module use. Rejected — that helper is for failures of a live service
   call (it takes a `method=` and reports to the error pipeline). This is a
   local construction failure with no request behind it, and the sibling
   PagerDuty factory does not use it either.
   because it matches PagerDuty exactly, and because reaching this branch means
   the user supplied a non-empty token that then failed to build — that is a
   genuine misconfiguration worth surfacing, not routine noise. The empty-token
   case returns earlier and stays silent, so a user with no Vercel setup at all
   sees nothing new.

**Why this implementation.** `logger.warning("Failed to create Vercel client: %s", exc, exc_info=True)`
scan, and attaches the full traceback via `exc_info=True` for anyone who needs
the stack. It sits inside the existing `except`, so no control flow moved.

**The test.** `test_make_vercel_client_logs_soft_fail_on_construction_error`
`integrations.vercel.client.VercelConfig` for `_raise_runtime_error`, a helper
taking `**_kwargs` because the factory calls it with keyword arguments. Patching
the config rather than the client is what makes the failure land inside the
`try` block. *Act* — call the factory under
`caplog.at_level(logging.WARNING, logger="integrations.vercel.client")`, scoped
to this module's logger so an unrelated warning elsewhere cannot make it pass.
*Assert* — two things, because both halves of the contract matter: the return is
still `None`, and some record is `WARNING` **with `exc_info is not None`.
Asserting on `exc_info` rather than on message text is deliberate: it pins the
behaviour the issue asks for (exception context preserved) without freezing the
wording, so a future copy edit to the message will not break the test.
`monkeypatch` undoes the patch automatically, leaving no global state behind.

**Edge cases I checked.** Empty / `None` / whitespace-only token still return
`None` before the `try` and log nothing (covered by the three pre-existing
factory tests, still green). The success path never enters the `except`. A
token containing a secret is never passed to the logger — only `str(exc)` and
the traceback are logged, and the factory does not interpolate the token into
any message.

---

### DIFF
--- a/integrations/vercel/client.py
+++ b/integrations/vercel/client.py
@@ -585,5 +585,6 @@ def make_vercel_client(api_token: str | None, team_id: str | None = None) -> Ver
         return None
     try:
         return VercelClient(VercelConfig(api_token=token, team_id=team_id or ""))
-    except Exception:
+    except Exception as exc:
+        logger.warning("Failed to create Vercel client: %s", exc, exc_info=True)
         return None
--- a/tests/integrations/vercel/test_client.py
+++ b/tests/integrations/vercel/test_client.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import json
+import logging
 from typing import Any
 
 import httpx
@@ -588,6 +589,30 @@ def test_make_vercel_client_forwards_team_id() -> None:
     client = make_vercel_client("tok_test", "team_xyz")
     assert client is not None
     assert client.config.team_id == "team_xyz"
+
+
+def _raise_runtime_error(**_kwargs: Any) -> None:
+    raise RuntimeError("construction failure")
+
+
+def test_make_vercel_client_logs_soft_fail_on_construction_error(
+    monkeypatch: pytest.MonkeyPatch, caplog: pytest.LogCaptureFixture
+) -> None:
+    # Arrange: force config construction inside the factory to raise, exercising
+    # the soft-fail `except Exception` path. The factory must still return None,
+    # but must no longer swallow the exception silently.
+    monkeypatch.setattr("integrations.vercel.client.VercelConfig", _raise_runtime_error)
+
+    # Act
+    with caplog.at_level(logging.WARNING, logger="integrations.vercel.client"):
+        result = make_vercel_client("tok_test")
+
+    # Assert
+    assert result is None
+    assert any(
+        record.levelno == logging.WARNING and record.exc_info is not None
+        for record in caplog.records
+    ), caplog.text
 
 
 def _raise_value_error() -> None:


deepspace28:
Cross-reference for anyone reviewing this alongside its siblings: the Argo CD PR in this set (#4920) drew a P1 from Greptile — logging a `pydantic.ValidationError` can echo the rejected input via `input_value=`, and for a **model-level** failure that mapping is the whole config, credentials included. Local `exc_info` logging bypasses Sentry's scrubber, which is the hazard `integrations/_validation_helpers.report_classify_failure` documents. I've guarded it there.

I checked whether the same applies here before leaving this PR alone:

- `VercelIntegrationConfig` does carry one model-level validator, `_reject_unknown_fields`, inherited from `StrictConfigModel`. So the mechanism exists.
- It is not reachable from `make_vercel_client`, which passes a fixed pair of known keyword arguments (`api_token`, `team_id`). No unknown field can be introduced through the factory, so that validator cannot fire on this path.
- The remaining validators on the model are field-level (`normalize_str` on `api_token` and `team_id`). A field-level error renders only the offending field's own input, and both are passed as already-stripped `str`, so the reachable failure surface here does not include a whole-mapping echo.

So I do not believe this PR needs the same change, and I would rather not churn an approved diff on a hypothetical. That said, if you would prefer all three SM-SR-LOG PRs to carry the identical guard for uniformity — so nobody has to re-derive this reasoning the next time a validator is added to `VercelIntegrationConfig` — say the word and I will add it here and in #4921. It is six lines and the same technique already used in `_validation_helpers`.

PR #4920, opened 2026-08-12
fix(integrations/argocd): log soft-fail when client construction fails

### LINKED ISSUE #4902: [GOOD FIRST ISSUE] SM-SR-LOG-2: Log soft-fail in Argo CD client
Audience: community
Kind (maintainer): n/a
Task id: SM-SR-LOG-2
Not: `core/` product edits.

---

**Selected task:** `SM-SR-LOG-2`

**Selected file:** `integrations/argocd/client.py`

| Field | Value |
|-------|-------|
| **Audience** | Community |
| **Level** | Junior |
| **Complexity** | S |
| **North star** | Quality |

| ID | File |
|----|------|

---

deepspace28:
Fixes #4902

Sibling of #4919 (SM-SR-LOG-1, Vercel), which came back **Greptile 5/5 with no
actionable defects**. Same change, same reasoning, applied to the Argo CD factory
as a separate PR per the issue's "one ID = one file = one PR" rule.

`make_argocd_client()` ended in a bare `except Exception: return None`. Returning
construction failure (an `ArgoCDConfig` validation error — plain HTTP that isn't
loopback, ambiguous auth methods, a malformed `verify_ssl`) was indistinguishable
from "Argo CD is simply not configured." The exception was discarded entirely.

- `integrations/argocd/client.py` — `except Exception as exc:` now logs at
- `tests/integrations/argocd/test_client.py` — new AAA unit test that forces
  `ArgoCDConfig` construction to raise and asserts **both** that the factory
  still returns `None` **and** that a `WARNING` record carrying exception
  context was emitted.

Return type, signature, and every success path are untouched. The two earlier
guard returns — missing base URL, or neither a bearer token nor a
username/password pair — stay silent, so a user with no Argo CD setup at all
sees nothing new. Only a genuine construction failure logs.

Wording and log level deliberately mirror the already-merged soft-fail log in
`integrations/pagerduty/client.py` (SM-115), so the sibling clients stay
consistent.

Same forced failure (`ArgoCDConfig` raising `ValueError`), run on `main` and on
this branch:

**Before — silent**

```
$ python demo.py

```

```
$ python demo.py
WARNING integrations.argocd.client: Failed to create Argo CD client: bearer_token and username/password are mutually exclusive
  File "D:\oss\opensre\integrations\argocd\client.py", line 536, in make_argocd_client
    ArgoCDConfig(
    raise ValueError("bearer_token and username/password are mutually exclusive")
ValueError: bearer_token and username/password are mutually exclusive

```

```
$ pytest tests/integrations/argocd/test_client.py -k soft_fail -q
FAILED tests/integrations/argocd/test_client.py::test_make_argocd_client_logs_soft_fail_on_construction_error
1 failed, 23 deselected

# after the one-line fix (GREEN)
$ pytest tests/integrations/argocd/ tests/tools/test_argocd_tools.py tests/integrations/test_argocd_catalog_verify.py -q
37 passed in 19.91s
```

**Required checks**

```
All checks passed!

3369 files already formatted

```

---

"no Argo CD client available" and carry on. That part is correct and I did not
want to change it. What is wrong is that the bare `except Exception` bound no
name and logged nothing, so the exception object was destroyed before anyone
could see it. `ArgoCDConfig` is a pydantic model with real validators in this
repo — it rejects plain HTTP for non-loopback hosts and rejects ambiguous auth
— so this branch is genuinely reachable from user misconfiguration, not just
theoretically.

   against `Client | None`. It would also let an optional integration break
   startup.
2. *Let the `pydantic.ValidationError` propagate and catch it more narrowly.*
   Rejected — it would change behaviour for the exact case the soft-fail exists
   to absorb, and the issue scope is a log, not a control-flow change.
   to match PagerDuty exactly, and because reaching this branch means the user
   supplied a URL **and** a working auth method that then failed to build —
   that is a genuine misconfiguration worth surfacing, not routine noise.

**Why this implementation.** `logger.warning("Failed to create Argo CD client: %s", exc, exc_info=True)`
scan, and attaches the full traceback via `exc_info=True`. It sits inside the
existing `except`, so no control flow moved.

**The test.** `test_make_argocd_client_logs_soft_fail_on_construction_error`
`integrations.argocd.client.ArgoCDConfig` for `_raise_runtime_error`, a helper
taking `**_kwargs` because the factory calls it with keyword arguments. Patching
the config rather than the client is what makes the failure land inside the
`try`. I pass a valid URL and bearer token so the call gets past the guard
returns and actually reaches the `try`. *Act* — call the factory under
`caplog.at_level(logging.WARNING, logger="integrations.argocd.client")`, scoped
to this module's logger so an unrelated warning elsewhere cannot make it pass.
*Assert* — both halves of the contract: the return is still `None`, and some
record is `WARNING` **with `exc_info is not None`**. Asserting on `exc_info`
rather than on message text pins the behaviour the issue asks for without
freezing the wording. `monkeypatch` undoes the patch automatically.

**Edge cases I checked.** Empty URL, and a URL with no auth at all, still return
`None` before the `try` and log nothing — covered by the pre-existing
`test_make_argocd_client_requires_base_url_and_auth`, still green. The success
path never enters the `except`. No credential is passed to the logger: only
`str(exc)` and the traceback are logged, and the factory does not interpolate
the token or password into any message. I also ran the Argo CD tool and catalog
tests, not just the client tests, to confirm nothing downstream depended on the
silence.

---

### DIFF
--- a/integrations/argocd/client.py
+++ b/integrations/argocd/client.py
@@ -16,6 +16,7 @@ from typing import Any
 from urllib.parse import quote
 
 import httpx
+from pydantic import ValidationError
 
 from integrations.config_models import ArgoCDIntegrationConfig
 from integrations.probes import ProbeResult
@@ -544,5 +545,15 @@ def make_argocd_client(
             )
         )
     except Exception as exc:
-        logger.warning("Failed to create Argo CD client: %s", exc, exc_info=True)
+        # `ValidationError` renders the rejected input inline via `input_value=`,
+        # and for a model-level failure that is the whole config mapping — bearer
+        # token and password included. Local `exc_info` logging bypasses Sentry's
+        # scrubber, so swap in a message-only error that keeps the traceback (but
+        # no `__cause__`, which would print the raw text again). Same technique as
+        # `integrations._validation_helpers.report_classify_failure`.
+        safe_exc: BaseException = exc
+        if isinstance(exc, ValidationError):
+            safe_exc = ValueError("Argo CD config validation failed")
+            safe_exc.__traceback__ = exc.__traceback__
+        logger.warning("Failed to create Argo CD client: %s", safe_exc, exc_info=safe_exc)
         return None
--- a/integrations/argocd/client.py
+++ b/integrations/argocd/client.py
@@ -543,5 +543,6 @@ def make_argocd_client(
                 verify_ssl=_normalize_verify_ssl(verify_ssl),
             )
         )
-    except Exception:
+    except Exception as exc:
+        logger.warning("Failed to create Argo CD client: %s", exc, exc_info=True)
         return None
--- a/tests/integrations/argocd/test_client.py
+++ b/tests/integrations/argocd/test_client.py
@@ -80,6 +80,40 @@ def test_make_argocd_client_logs_soft_fail_on_construction_error(
     ), caplog.text
 
 
+def test_make_argocd_client_soft_fail_log_does_not_echo_config_input(
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    # Arrange: supplying both auth methods is a real misconfiguration that gets
+    # past the factory's guards and reaches ArgoCDConfig's model validator.
+    # A model-level pydantic failure renders the whole input mapping via
+    # `input_value=...`, which is where a bearer token or password would escape.
+    secret_token = "s3cr3t-argocd-bearer-token"
+    secret_password = "s3cr3t-argocd-password"
+
+    # Act
+    with caplog.at_level(logging.WARNING, logger="integrations.argocd.client"):
+        result = make_argocd_client(
+            "https://argocd.example.com",
+            bearer_token=secret_token,
+            username="admin",
+            password=secret_password,
+        )
+
+    # Assert: the soft-fail contract and the warning both survive, but no part
+    # of the submitted config is echoed. Asserting on `input_value` rather than
+    # only on the literal secrets is deliberate: pydantic elides the middle of a
+    # long mapping, so a secret-substring check passes by accident today and
+    # would stop protecting us if field order or pydantic's truncation changed.
+    assert result is None
+    assert any(
+        record.levelno == logging.WARNING and record.exc_info is not None
+        for record in caplog.records
+    ), caplog.text
+    assert "input_value" not in caplog.text
+    assert secret_token not in caplog.text
+    assert secret_password not in caplog.text
+
+
 def test_probe_access_success(monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.setattr(
         ArgoCDClient,
--- a/tests/integrations/argocd/test_client.py
+++ b/tests/integrations/argocd/test_client.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import json
+import logging
 from typing import Any
 
 import httpx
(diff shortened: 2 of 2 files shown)


deepspace28:
Thanks — fixed in `0166486`. The underlying concern is right and I've guarded it. But I want to be straight about one thing rather than just claiming the finding reproduced, because the difference matters for how you read the test.

## The fix

`ValidationError` is swapped for a message-only `ValueError` that reuses the original `__traceback__` but sets no `__cause__`:

```python
except Exception as exc:
    safe_exc: BaseException = exc
    if isinstance(exc, ValidationError):
        safe_exc = ValueError("Argo CD config validation failed")
        safe_exc.__traceback__ = exc.__traceback__
    logger.warning("Failed to create Argo CD client: %s", safe_exc, exc_info=safe_exc)
    return None
```

This is the same technique `integrations/_validation_helpers.report_classify_failure` already uses, and its docstring describes this exact hazard — that local `exc_info` logging bypasses Sentry's `before_send` scrubber. Note `exc_info=safe_exc` rather than `exc_info=True`: inside an `except` block `True` resolves via `sys.exc_info()` to the *original* `ValidationError`, which would have reprinted everything the swap just removed.

Log output after the change — traceback still points at the failing model, nothing echoed:

```
WARNING integrations.argocd.client: Failed to create Argo CD client: Argo CD config validation failed
  File "integrations/argocd/client.py", line 537, in make_argocd_client
    ArgoCDConfig(
  File ".../pydantic/main.py", line 263, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
ValueError: Argo CD config validation failed
```

## What I could not reproduce

I tried to write a test that fails on the old code by asserting the credentials were absent from the log. **It passed against the unguarded code** — so I went looking for why before assuming the test was fine.

Pydantic truncates the middle of a long `input_value` mapping, keeping only the head and tail. Going through `make_argocd_client`, the mapping always ends with the three fields the factory appends after the credentials:

```
input_value={'base_url': 'https://arg... '', 'verify_ssl': True}, input_type=dict
```

`bearer_token`, `username` and `password` are structurally in the elided middle. I tested the shortest legal URL (`http://[::1]`), loopback, normal https, and credentials embedded in the URL itself, with credential lengths from 1 to 60 characters — no credential reached the log in any of them.

My earlier direct-construction probe *did* leak, but only because calling `ArgoCDConfig(...)` with four kwargs puts `password` last, inside the retained tail. The factory never produces that shape.

So the P1 as described — dual-auth through the factory disclosing the token and password — does not reproduce on this branch. I'd rather flag that than let a "verified" label stand on something I couldn't confirm.

## Why I fixed it regardless

The safety above is an accident, not a property. It depends on pydantic's truncation threshold (an implementation detail, and this repo pins 2.13), on `ArgoCDIntegrationConfig` field ordering, and on the factory continuing to pass `project` / `app_namespace` / `verify_ssl` after the credentials. Any of those can change without anyone connecting it to a credential disclosure. Six lines to remove the dependency is a good trade, and it makes the local log consistent with what the Sentry path already guarantees.

## The test

For that reason the regression test asserts `"input_value" not in caplog.text` rather than only checking the literal secrets:

```python
assert "input_value" not in caplog.text
assert secret_token not in caplog.text
assert secret_password not in caplog.text
```

The secret-substring assertions alone pass on the unguarded code, so they would have been a false green. Asserting that no rejected input is echoed at all is order-independent and survives a pydantic upgrade. This test does fail on the pre-fix code.

`ruff check`, `ruff format --check`, and `mypy` are clean across all source paths; Argo CD client, tool, and catalog tests are 38 passed.

@greptile review

Issue #4905, opened 2026-08-11
[GOOD FIRST ISSUE] SM-SR-TO-2: Spell out timeout constant units in Jira client


muddlebee:
Supports Python API / repeat usage — teammate.
Audience: community
Kind (maintainer): n/a
Task id: SM-SR-TO-2
Code PRs: TDD; AAA (Arrange→Act→Assert); clear names; in-tree patterns; SOLID.
Not: `core/` product edits.

---

**Selected task:** `SM-SR-TO-2`

**Selected file:** `integrations/jira/client.py`

| Field | Value |
|-------|-------|
| **Audience** | Community |
| **Level** | Junior |
| **Complexity** | S |
| **Good first issue** | Yes |
| **North star** | Quality |

**Summary:** Rename `_DEFAULT_TIMEOUT` → `_DEFAULT_TIMEOUT_SECONDS` (value
unchanged; verified `_DEFAULT_TIMEOUT = 30` in PagerDuty and Jira clients) and
update references in **one** client package. Matches SM-62…SM-71 style.

| ID | File |
|----|------|
| SM-SR-TO-1 | `integrations/pagerduty/client.py` |
| SM-SR-TO-2 | `integrations/jira/client.py` |

**Do not:** Change the numeric value or public API timeouts.

**Done when:** [ ] Constant renamed with `_SECONDS` suffix
[ ] All in-file references updated; value still `30`

**Files:** exactly one client module per ID

---

activateLabhesh:
@muddlebee, can i be assigned this issue?

Davidson3556:
go for it @activateLabhesh

PR #4921, opened 2026-08-12
fix(integrations/jira): log soft-fail when client construction fails

### LINKED ISSUE #4903: [GOOD FIRST ISSUE] SM-SR-LOG-3: Log soft-fail in Jira client

deepspace28:
Fixes #4903

Sibling of #4919 (SM-SR-LOG-1, Vercel), which came back **Greptile 5/5 with no
actionable defects**, and #4920 (SM-SR-LOG-2, Argo CD). Same change, same
reasoning, applied to the Jira factory as a separate PR per the issue's
"one ID = one file = one PR" rule. This completes the SM-SR-LOG set.

`make_jira_client()` ended in a bare `except Exception: return None`. Returning
construction failure (a `JiraIntegrationConfig` validation error) was
indistinguishable from "Jira is simply not configured." The exception was
discarded entirely.

- `integrations/jira/client.py` — `except Exception as exc:` now logs at
- `tests/integrations/test_jira_search_and_factory.py` — new AAA unit test that
  forces `JiraIntegrationConfig` construction to raise and asserts **both** that
  the factory still returns `None` **and** that a `WARNING` record carrying
  exception context was emitted. Placed alongside the existing
  `make_jira_client` factory tests in that file.

Return type, signature, and every success path are untouched. The earlier guard
returns — missing base URL, email, or API token — stay silent, so a user with no
Jira setup at all sees nothing new. Only a genuine construction failure logs.

Wording and log level deliberately mirror the already-merged soft-fail log in
`integrations/pagerduty/client.py` (SM-115), so the sibling clients stay
consistent.

**Scope note:** this is the soft-fail log only. The `_DEFAULT_TIMEOUT` rename in
the same file is tracked separately as SM-SR-TO-2 (#4905) and is deliberately
not touched here.

Same forced failure (`JiraIntegrationConfig` raising `ValueError`), run on `main`
and on this branch:

**Before — silent**

```
$ python demo.py

```

```
$ python demo.py
WARNING integrations.jira.client: Failed to create Jira client: base_url must be an https Atlassian site URL
  File "D:\oss\opensre\integrations\jira\client.py", line 293, in make_jira_client
    config = JiraIntegrationConfig(
             ^^^^^^^^^^^^^^^^^^^^^^
    raise ValueError("base_url must be an https Atlassian site URL")
ValueError: base_url must be an https Atlassian site URL

```

```
$ pytest tests/integrations/test_jira_search_and_factory.py -k soft_fail -q
FAILED tests/integrations/test_jira_search_and_factory.py::test_make_jira_client_logs_soft_fail_on_construction_error
1 failed, 8 deselected

# after the one-line fix (GREEN) — client, factory, and all four Jira tools
$ pytest tests/integrations/test_jira_search_and_factory.py tests/integrations/test_jira_client.py tests/tools/test_jira_*.py -q
45 passed in 8.98s
```

**Required checks**

```
All checks passed!

3369 files already formatted

```

---

"no Jira client available" and carry on. That part is correct and I did not want
to change it. What is wrong is that the bare `except Exception` bound no name and
logged nothing, so the exception object was destroyed before anyone could see it.
Of the three clients in this SM-SR-LOG set, Jira is the one where the silence
hurts most in practice: it takes three separate credentials (site URL, account
email, API token), so there are more ways for a user to get the shape wrong and
end up staring at a Jira integration that simply does nothing.

   against `Client | None`. It would also let an optional integration break
   startup.
2. *Widen the scope and also do the `_DEFAULT_TIMEOUT` rename* while I was in
   this file. Rejected — that is SM-SR-TO-2 (#4905), a separate issue, and the
   task says one ID per PR. Mixing them would make both harder to review.
   to match PagerDuty exactly, and because reaching this branch means the user
   supplied all three credentials and they still failed to build — a genuine
   misconfiguration worth surfacing, not routine noise.

**Why this implementation.** `logger.warning("Failed to create Jira client: %s", exc, exc_info=True)`
scan, and attaches the full traceback via `exc_info=True`. It sits inside the
existing `except`, so no control flow moved. Note the `try` block here wraps both
the config construction and the `JiraClient(config)` call, so a failure in either
is now reported.

**The test.** `test_make_jira_client_logs_soft_fail_on_construction_error`
`integrations.jira.client.JiraIntegrationConfig` for `_raise_runtime_error`, a
helper taking `**_kwargs` because the factory calls it with keyword arguments. I
patch the name as bound in the client module, not in `integrations.config_models`,
because that is the reference the factory actually resolves at call time. I pass
all three valid credentials so the call gets past the guard returns and actually
reaches the `try`. *Act* — call the factory under
`caplog.at_level(logging.WARNING, logger="integrations.jira.client")`, scoped to
this module's logger so an unrelated warning elsewhere cannot make it pass.
*Assert* — both halves of the contract: the return is still `None`, and some
record is `WARNING` **with `exc_info is not None`**. Asserting on `exc_info`
rather than on message text pins the behaviour the issue asks for without
freezing the wording. `monkeypatch` undoes the patch automatically.

**Edge cases I checked.** Missing URL, missing email, missing token, and all-`None`
arguments still return `None` before the `try` and log nothing — covered by the
four pre-existing factory tests, still green. The success path never enters the
`except`. No credential is passed to the logger: only `str(exc)` and the
traceback are logged, and the factory does not interpolate the email or API token
into any message. I ran the four Jira tool test files as well as the client tests
to confirm nothing downstream depended on the silence.

---

### DIFF
--- a/integrations/jira/client.py
+++ b/integrations/jira/client.py
@@ -297,5 +297,6 @@ def make_jira_client(
             project_key=(project_key or "").strip(),
         )
         return JiraClient(config)
-    except Exception:
+    except Exception as exc:
+        logger.warning("Failed to create Jira client: %s", exc, exc_info=True)
         return None
--- a/tests/integrations/test_jira_search_and_factory.py
+++ b/tests/integrations/test_jira_search_and_factory.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import logging
 from unittest.mock import MagicMock, patch
 
 import httpx
@@ -123,3 +124,27 @@ def test_make_jira_client_returns_none_missing_token() -> None:
 
 def test_make_jira_client_returns_none_all_none() -> None:
     assert make_jira_client(None, None, None) is None
+
+
+def _raise_runtime_error(**_kwargs: object) -> None:
+    raise RuntimeError("construction failure")
+
+
+def test_make_jira_client_logs_soft_fail_on_construction_error(
+    monkeypatch: pytest.MonkeyPatch, caplog: pytest.LogCaptureFixture
+) -> None:
+    # Arrange: force config construction inside the factory to raise, exercising
+    # the soft-fail `except Exception` path. The factory must still return None,
+    # but must no longer swallow the exception silently.
+    monkeypatch.setattr("integrations.jira.client.JiraIntegrationConfig", _raise_runtime_error)
+
+    # Act
+    with caplog.at_level(logging.WARNING, logger="integrations.jira.client"):
+        result = make_jira_client("https://myteam.atlassian.net", "user@example.com", "token")
+
+    # Assert
+    assert result is None
+    assert any(
+        record.levelno == logging.WARNING and record.exc_info is not None
+        for record in caplog.records
+    ), caplog.text


deepspace28:
Cross-reference for anyone reviewing this alongside its siblings: the Argo CD PR in this set (#4920) drew a P1 from Greptile — logging a `pydantic.ValidationError` can echo the rejected input via `input_value=`, and for a **model-level** failure that mapping is the whole config, credentials included. Local `exc_info` logging bypasses Sentry's scrubber, which is the hazard `integrations/_validation_helpers.report_classify_failure` documents. I've guarded it there.

I checked whether the same applies here before leaving this PR alone:

- `JiraIntegrationConfig` does carry one model-level validator, `_reject_unknown_fields`, inherited from `StrictConfigModel`. So the mechanism exists.
- It is not reachable from `make_jira_client`, which passes a fixed set of four known keyword arguments (`base_url`, `email`, `api_token`, `project_key`). No unknown field can be introduced through the factory, so that validator cannot fire on this path.
- The remaining validators are field-level: `normalize_url` on `base_url` and `normalize_str` on `email` / `api_token` / `project_key`. A field-level error renders only the offending field's own input. The reachable one is `base_url`, which is not a credential — and all four are passed as already-stripped `str`.

Unlike Argo CD, Jira has no equivalent of `_no_dual_auth` — a user-triggerable model-level validator — so there is no path here that produces a whole-mapping echo.

So I do not believe this PR needs the same change, and I would rather not churn an approved diff on a hypothetical. That said, if you would prefer all three SM-SR-LOG PRs to carry the identical guard for uniformity — so nobody has to re-derive this reasoning the next time a validator is added to `JiraIntegrationConfig` — say the word and I will add it here and in #4919. It is six lines and the same technique already used in `_validation_helpers`.

PR #4928, opened 2026-08-12
fix(scheduler): write the task store atomically and never clobber an unreadable one (#4927)

### LINKED ISSUE #4927: [BUG] Scheduler task store writes non-atomically; a torn write silently deletes every scheduled task
### Summary

`platform/scheduler/store.py` writes the task store with a plain
`Path.write_text`, which truncates the file before rewriting it. If the process dies
during that window (SIGKILL, container eviction, power loss, ENOSPC), the JSON is left
truncated. `_load_raw` then swallows the `JSONDecodeError` and returns `[]`, so the
scheduler reports **no tasks** — and the next `add_task` writes a fresh list over the
damaged file, making the loss permanent.

Every other local store in the repo already writes by atomic rename
(`integrations/store.py::_atomic_write`, `platform/analytics/provider.py::_write_text_atomic`,
`platform/filestorage/engine.py::_write_atomically`, `integrations/sentry/uptime.py`).
`platform/filestorage/__init__.py` states the convention outright: "stores here write by
atomic rename." The scheduler task store is the one that does not.

**Expected:** a crash during a store write leaves the previous task list intact, and a
store that cannot be parsed is never overwritten by a subsequent write.

**Actual:** the write is non-atomic, and an unparseable store is silently treated as
"zero tasks" by both readers and writers. Every scheduled digest, report, and uptime
watch disappears, and the only signal is a single `logger.warning`.

The two halves compound:

- `_save_raw` (L44–47) — `store_path.write_text(...)`, no temp file, no `os.replace`.
- `_load_raw` (L30–41) — returns `[]` on `JSONDecodeError`/`OSError`, so `add_task`,
  `remove_task`, and `update_task` all proceed as though the store were empty.

### Steps to reproduce

```python
from pathlib import Path
import tempfile
from platform.scheduler.store import add_task, list_tasks
from platform.scheduler.types import Provider, ScheduledTask, TaskKind

store = Path(tempfile.mkdtemp()) / "scheduler_tasks.json"
for hour in (7, 8, 9):
    add_task(
        ScheduledTask(
            name=f"digest-{hour}", kind=TaskKind.DAILY_SUMMARY, cron=f"0 {hour} * * *",
            timezone="UTC", provider=Provider.TELEGRAM, chat_id="-100", window_hours=24,
        ),
        store_path=store,
    )
print([t.name for t in list_tasks(store)])

# Simulate a crash midway through the non-atomic write in _save_raw:
full = store.read_text(encoding="utf-8")
store.write_text(full[: len(full) // 2], encoding="utf-8")
print([t.name for t in list_tasks(store)])

add_task(
    ScheduledTask(
        name="digest-10", kind=TaskKind.DAILY_SUMMARY, cron="0 10 * * *",
        timezone="UTC", provider=Provider.TELEGRAM, chat_id="-100", window_hours=24,
    ),
    store_path=store,
)
print([t.name for t in list_tasks(store)])
```

Output:

```
stored tasks: ['digest-7', 'digest-8', 'digest-9']
after partial write, list_tasks -> []
after next add_task -> ['digest-10']
```

The three schedules are gone from disk, not merely unreadable. The only log line is
`Failed to read scheduler store: Expecting value: line 24 column 15 (char 522)`.

Yes

### How often does it occur?

Under specific conditions (process killed or disk full during a store write)

### Operating system

Other (any — the write path is platform-independent)

### Additional context

Distinct from #4685, which covers state handling inside the investigation pipeline;
this is the scheduler's on-disk task store.

Suggested fix, mirroring `integrations/store.py::_atomic_write`:

1. Write via `tempfile.mkstemp` in the destination directory, `flush` + `os.fsync`, then
   `os.replace` — atomic on POSIX and Windows.
2. Do not let a mutating operation overwrite a store it could not parse. Rename the
   unreadable file aside (`scheduler_tasks.json.corrupt-<timestamp>`) and log at error
   level, so the schedules stay recoverable instead of vanishing.

**Done when:** a torn write cannot destroy the task list, and an unparseable store is
preserved rather than silently replaced.

```bash
uv run opensre cron add …          # a few tasks
uv run opensre cron list
uv run opensre cron start &        # then SIGKILL it mid-write
uv run opensre cron list
```

**Pass:** the previous task list survives the kill. **Fail (today):** an
unparseable store reads as zero tasks, and the next `cron add` overwrites it
permanently.

The store is `infrastructure/scheduling/scheduler/store.py` now, not
`platform/scheduler/store.py`.

Chirag6722:
Fixes #4927

The scheduler task store wrote with a plain `Path.write_text`, which truncates the file
before rewriting it. A process killed inside that window left truncated JSON, `_load_raw`
swallowed the `JSONDecodeError` and returned `[]`, and the next `add_task` wrote a fresh
list straight over the damage. Every digest, report, and uptime watch disappeared, with
one `logger.warning` as the only signal.

Every other local store in the repo already writes by atomic rename
(`integrations/store.py::_atomic_write`, `platform/analytics/provider.py`,
`platform/filestorage/engine.py`, `integrations/sentry/uptime.py`) —
`platform/filestorage/__init__.py` states the convention outright. This store was the
outlier. Two changes, addressing the cause and the amplifier:

**1. Atomic write.** `_save_raw` writes via `tempfile.mkstemp` in the destination
directory, `flush` + `os.fsync`, then `os.replace` (atomic on POSIX and Windows), and
unlinks the temp file if anything fails. Shaped after `integrations/store.py::_atomic_write`
rather than inventing a new pattern.

**2. Never overwrite what you could not read.** `_read_rows` now distinguishes "no file
yet" (safe to write over) from "file exists but will not parse" (not safe). Read-only
`list_tasks` still degrades to an empty list so listing never raises. The three mutating
operations use `_load_raw_for_write`, which renames the damaged file to
`scheduler_tasks.json.corrupt-<timestamp>` and logs at **error** level before writing —
the schedules stay on disk and recoverable instead of vanishing.

A JSON object where a list belongs is now treated as damage too, for the same reason it
previously read as "zero tasks": it isn't.

`list_tasks`, `get_task`, and the dedup behavior in `add_task` are untouched.

The repro from the issue, **before** this change — three schedules silently reduced to one:

```
stored tasks: ['digest-7', 'digest-8', 'digest-9']
after partial write, list_tasks -> []
after next add_task -> ['digest-10']
```

**After** — the damaged file is preserved and the operator gets a loud error instead of
silence:

```
Scheduler store at ...\scheduler_tasks.json was unreadable and has been preserved at
...\scheduler_tasks.json.corrupt-1786515332. Scheduled tasks it held are not loaded;
recover them from that file.
```

Four regression tests in `TestStoreDurability`. They **fail on the pre-fix code and pass
after**, verified by stashing only `platform/scheduler/store.py`:

```
# tests against the OLD store.py:
FAILED tests/scheduler/test_store.py::TestStoreDurability::test_failed_write_leaves_the_previous_tasks_intact
FAILED tests/scheduler/test_store.py::TestStoreDurability::test_failed_write_does_not_leak_a_temp_file
FAILED tests/scheduler/test_store.py::TestStoreDurability::test_unreadable_store_is_preserved_rather_than_overwritten
FAILED tests/scheduler/test_store.py::TestStoreDurability::test_non_list_json_is_treated_as_unreadable
4 failed in 3.52s

# with the fix, whole scheduler suite:
$ uv run python -m pytest tests/scheduler -q
141 passed in 5.13s

$ python -m ruff check config core gateway integrations platform surfaces tools tests
All checks passed!

$ python -m ruff format --check config core gateway integrations platform surfaces tools tests
3366 files already formatted

Success: no issues found in 1728 source files

$ uv run python .github/ci/check_imports.py
All 3 import checks passed.
```

---

Atomic write alone would have been an incomplete fix. It prevents *new* torn writes, but
any store already damaged — or damaged by something outside this code path, like a disk
error — still gets silently erased by the next `add_task`. The read side is what turns a
recoverable file into permanent loss, so both halves had to change.

The read-side split is deliberate rather than uniform: making all reads raise would mean
`opensre cron list` blows up with a traceback on a damaged store, which is worse for an
operator trying to diagnose. Listing degrading to "no tasks" is acceptable because it
destroys nothing; a *write* proceeding on that same assumption is not. Hence one
`_read_rows` returning `None` for unreadable, and two thin callers that make opposite
choices for opposite reasons.

I chose to quarantine-and-continue over raising from `add_task` because the scheduler
daemon also mutates the store; raising there would turn a recoverable file into a crash
loop. Renaming aside keeps the data, keeps the command working, and leaves an error-level
log naming the exact recovery path.

Edge cases covered by the tests: the rename failing partway (original must be byte-identical
afterwards), temp-file debris on a failed write (a retry loop must not fill the directory),
truncated JSON, and structurally-wrong-but-valid JSON.

---

### DIFF
--- a/platform/scheduler/store.py
+++ b/platform/scheduler/store.py
@@ -76,17 +76,23 @@ def _load_raw_for_write(store_path: Path) -> list[dict[str, object]]:
 
 
 def _quarantine_unreadable(store_path: Path) -> None:
-    """Move an unreadable store aside so the next write cannot erase it."""
+    """Move an unreadable store aside so the next write cannot erase it.
+
+    Raises if the rename fails. Continuing would hand the caller an empty list
+    and let it write over the only copy of the schedules — the exact loss this
+    preservation step exists to prevent, so it fails closed instead.
+    """
     backup = store_path.with_name(f"{store_path.name}.corrupt-{int(time.time())}")
     try:
         os.replace(store_path, backup)
     except OSError:
-        logger.warning(
-            "Could not preserve unreadable scheduler store at %s; leaving it in place",
+        logger.error(
+            "Could not preserve unreadable scheduler store at %s; refusing to "
+            "overwrite it. Move or repair the file by hand to continue.",
             store_path,
             exc_info=True,
         )
-        return
+        raise
     logger.error(
         "Scheduler store at %s was unreadable and has been preserved at %s. "
         "Scheduled tasks it held are not loaded; recover them from that file.",
--- a/platform/scheduler/store.py
+++ b/platform/scheduler/store.py
@@ -2,8 +2,12 @@
 
 from __future__ import annotations
 
+import contextlib
 import json
 import logging
+import os
+import tempfile
+import time
 from collections.abc import Mapping
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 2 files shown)


Chirag6722:
Good catch — that P1 was real and it defeated the point of the change. If the quarantine rename failed I logged and returned, and the caller then wrote its new task straight over the file I meant to preserve.

Fixed in 375471f0: the rename failure now propagates, so the mutation aborts instead of overwriting. Continuing is only safe once the data is somewhere else; when preservation fails, refusing the write is what keeps it. The error log now names the manual recovery step.

Also added the regression you flagged as missing — quarantine rename fails → `add_task` raises and the damaged file is left byte-identical. Scheduler suite is 142 passed.

@greptile review

Chirag6722:
Also right, thanks. `.corrupt-<int(time.time())>` gave two same-second quarantines the same path, so the second `os.replace` destroyed the first backup — losing exactly what the first quarantine had preserved.

Fixed by reserving the name with `tempfile.mkstemp` in the store directory, which closes the race atomically instead of narrowing it the way a finer-grained timestamp would.

Regression added: damage and quarantine twice in immediate succession, then assert both backups exist and the earlier one still holds its schedule. Scheduler suite: 143 passed.

@greptile review

PR #4932, opened 2026-08-12
fix(sandbox): gate chat code execution on approval and close the write-guard gaps (#4931)

### LINKED ISSUE #4931: [SECURITY] Sandbox restrictions are advisory, not a boundary (details redacted — reported via wrong channel)
> **Redacted by the reporter.** This was filed as a public issue in error —
> [SECURITY.md](https://github.com/Tracer-Cloud/opensre/blob/main/SECURITY.md) asks for
> vulnerabilities to go privately to support@opensre.com, not to a public issue. The
> reproduction steps and proof-of-concept that were originally here have been removed.
> Maintainers: I'm happy to send the full write-up, probe scripts, and verified output to
> support@opensre.com on request, or to have this issue deleted entirely — your call.

## Summary (non-exploitable detail only)

The Python sandbox in `platform/sandbox/runner.py` enforces its restrictions by
monkeypatching builtins and module attributes **inside the same interpreter that runs the
supplied code**. Those patches are guard rails against accidental damage, not a containment
boundary: code running under them can restore the originals, and several write paths were
never covered by the guard at all.

This matters because `execute_python_code` is served on the **chat** surface with
`requires_approval` unset, so the code author on that surface is a remote chat user rather
than the operator, and the tool injects a live credential into the process.

## Impact

Untrusted chat input, or injected content the agent reads, could reach code execution on the
host without a human in the loop.

## Proposed direction

1. Gate the tool behind the existing human-in-the-loop approval mechanism on messaging
   surfaces (`gateway/core/runtime/approvals.py`).
2. Cover the write paths the guard missed, as accident-prevention rather than containment.
3. Stop documenting containment guarantees the runner does not provide.
4. Structural follow-up, needing a maintainer decision: real containment requires OS-level
   isolation (container/jail per execution, rlimits, dedicated low-privilege user, no ambient
   credentials in the child environment). Worth its own design issue.

Apologies for the process error.

Chirag6722:
Closes #4931

`platform/sandbox/runner.py` enforces its restrictions by monkeypatching `builtins.open`,
`subprocess.*`, and `socket.*` **inside the same interpreter that runs the supplied code**.
Code that is trying to get out can undo them. `execute_python_code` is served on the
**chat** surface with `requires_approval` unset, so a Slack/Telegram/Discord user — or
prompt injection in any content the agent reads — reached arbitrary host code execution
unattended, in a process holding an injected `GITHUB_TOKEN`.

Verified against `main` before changing anything:

| Probe | Result on `main` |
|---|---|
| `builtins.open` write outside temp | BLOCKED (the one guarded path) |
| `pathlib.Path.write_text` outside temp | **ALLOWED** |
| `os.open` + `os.write` outside temp | **ALLOWED** |
| `importlib.reload(subprocess)` then spawn a child | **ALLOWED — child ran** |
| child sees the injected credential | **`child sees GITHUB_TOKEN: ghp_FAKE_TOKEN_FOR_AUDIT_ONLY`** |

The last row is the important one, and it is not a hole to be patched: a child interpreter
never receives the preamble, so it has no write guard and no network block by construction.
Blocking `importlib.reload` would just move the bypass to `ctypes`, `os.posix_spawn`, or a
re-import of `_posixsubprocess`.

So this PR does not pretend to make the sandbox a boundary. It does three things:

1. **Puts a human in front of the risk.** `PythonExecutionTool.requires_approval = True`
   routes chat-surface executions through the approve/deny gate the repo already has
   (`gateway/core/runtime/approvals.py`). That gate is gateway-only, so local CLI and
   investigation turns are unchanged.
2. **Closes the two write-guard gaps** — `io.open` (what `pathlib` resolves) and `os.open`
   — as accident-prevention on its own merits. Reads stay unrestricted; writes inside
   `OPENSRE_TMP_DIR` still work.
3. **Stops claiming containment that does not hold.** The tool description told the model
   "Subprocess spawning is always blocked"; the runner docstring now states these are guard
   rails, not a security boundary, and points callers at the approval gate.

**Two encoding bugs surfaced while testing and are fixed here** (both pre-existing, both in
the lines this PR already touches):

- The script was written with `NamedTemporaryFile(mode="w")` in the locale encoding while
  Python reads `.py` as UTF-8 — any non-ASCII character in generated code died with
  `SyntaxError: Non-UTF-8 code ...` under cp1252.
- The child's stdout was decoded as UTF-8 while the child encoded in its locale codec, so
  non-ASCII output came back as mojibake (`caf� � na�ve`). Fixed with `PYTHONIOENCODING`.

Same probe re-run after the change — the two write bypasses are closed, and the subprocess
escape is unchanged and openly acknowledged rather than papered over:

```
- builtins.open write outside temp     -> BLOCKED: Write access denied outside the OpenSRE temp directory
- pathlib.Path.write_text outside temp -> BLOCKED: Write access denied outside the OpenSRE temp directory
- os.open + os.write outside temp      -> BLOCKED: Write access denied outside the OpenSRE temp directory
- subprocess after importlib.reload    -> ALLOWED after reload: child ran   <-- by design; see #4931
```

Tests added: `TestSandboxWriteGuardCoversEveryOpenPath` (pathlib and `os.open` writes
refused; writes inside `OPENSRE_TMP_DIR` and read-only `os.open` still work),
`TestSandboxSourceEncoding`, and a `requires_approval` assertion pinning the gate.

```
$ uv run python -m pytest tests/sandbox tests/platform/sandbox tests/tools/test_python_execution_tool.py tests/tools/test_python_execution_runtime_inputs.py -q
67 passed, 2 failed

$ uv run python -m pytest tests/tools tests/quality -q
2917 passed, 4 failed

$ python -m ruff check config core gateway integrations platform surfaces tools tests
All checks passed!

$ python -m ruff format --check config core gateway integrations platform surfaces tools tests
3366 files already formatted

Success: no issues found in 1728 source files
```

Those 4 failures are pre-existing and Windows-only — I confirmed each reproduces on clean
`main` with this branch stashed. Two are `test_subprocess_primitives` (process signalling);
two are `allow_network` socket tests failing with `WinError 10106` because `_sandbox_env`
drops `SystemRoot`, which Winsock needs. That last one is a genuine Windows bug in the same
function, but it is unrelated to this change and I did not want to bundle it into a security
fix — happy to file it separately.

---

The tempting fix here is to keep patching: block `importlib.reload`, block `os.posix_spawn`,
block `ctypes`. I deliberately did not, because that produces a sandbox that *looks* stronger
while remaining escapable — the worst of both, since the false assurance is what lets an
untrusted-input path stay wired to it. An in-process guard cannot contain the interpreter it
lives in, so I fixed the reachability instead of the symptom.

`requires_approval` was the right lever because it already exists and is transport-neutral:
Slack, Discord, and Telegram all render it, and it is read only by the gateway, so the local
operator (who *is* the code author) sees no new friction. The alternative — dropping the
`chat` surface — would have removed a feature people use, for a risk a human check addresses.

The write-guard extension is scoped to writes: `_write_allowed` is shared by all three
patched entry points, and `os.open` is only refused when the flags request write access, so
read-only `os.open` keeps working (covered by a test, since silently breaking reads would be
a regression in a tool whose main job is filesystem introspection).

The encoding fixes are included because I hit them while testing this change and both live in
lines this PR already modifies; separating them would have meant shipping a sandbox whose own
test for non-ASCII output fails.

---

### DIFF
--- a/platform/sandbox/runner.py
+++ b/platform/sandbox/runner.py
@@ -93,8 +93,15 @@ _SANDBOX_PREAMBLE = textwrap.dedent(f"""\
     _original_os_open = _os_module.open
 
     def _restricted_os_open(path, flags, *args, **kwargs):
+        # O_TRUNC belongs here even though it is not an access mode: on POSIX
+        # ``os.open(p, O_RDONLY | O_TRUNC)`` empties the file, so a mask of
+        # access modes alone let generated code destroy any file it could name.
         _write_flags = (
-            _os_module.O_WRONLY | _os_module.O_RDWR | _os_module.O_APPEND | _os_module.O_CREAT
+            _os_module.O_WRONLY
+            | _os_module.O_RDWR
+            | _os_module.O_APPEND
+            | _os_module.O_CREAT
+            | _os_module.O_TRUNC
         )
         if flags & _write_flags and not _write_allowed(path):
             raise PermissionError(
--- a/platform/sandbox/runner.py
+++ b/platform/sandbox/runner.py
@@ -1,4 +1,13 @@
-"""Python sandbox runner with timeout and restricted access."""
+"""Python sandbox runner with timeout and restricted access.
+
+**These guards are not a security boundary.** They are injected into the same
+interpreter that runs the supplied code, so code that is *trying* to get out
+can: ``importlib.reload`` restores any patched module, ``ctypes`` and
+``os.posix_spawn`` were never patched, and a child interpreter started by any
+route receives no preamble at all. Treat the restrictions as protection against
+an agent damaging the host by accident, and gate the calling tool on approval
+wherever the code author is not the operator.
+"""
 
 from __future__ import annotations
 
@@ -51,9 +60,17 @@ _NETWORK_BLOCK_PREAMBLE = textwrap.dedent("""\
 # Preamble always injected before user code: restricts filesystem writes and subprocesses.
 _SANDBOX_PREAMBLE = textwrap.dedent(f"""\
     import builtins as _builtins_module
+    import io as _io_module
     import os as _os_module
 
     _ALLOWED_WRITE_ROOTS = ({_SANDBOX_TMP_ROOT!r},)
+
+    def _write_allowed(path):
+        abs_path = _os_module.path.realpath(_os_module.fspath(path))
+        return any(
+            abs_path == root or abs_path.startswith(root + _os_module.sep)
+            for root in _ALLOWED_WRITE_ROOTS
+        )
 
     _original_open = _builtins_module.open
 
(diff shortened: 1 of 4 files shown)


Chirag6722:
Right catch — `O_TRUNC` is not an access mode, so my mask missed it, and on POSIX `os.open(path, O_RDONLY | O_TRUNC)` empties the file without ever requesting write access. That is a destructive bypass of exactly the guard this PR adds.

Fixed: `O_TRUNC` is now in the mask, with a comment explaining why a non-access-mode flag belongs there.

The regression test is skipped on Windows — it rejects that flag combination outright with Errno 22, so the truncation risk is POSIX-only, which is where the gateway runs. Sandbox suite: 28 passed, 1 skipped.

@greptile review

PR #4941, opened 2026-08-12
fix(agent): correct the runtime facts the agent reports — cloud, OS, cross-turn evidence, connected integrations

YauhenBichel:
Four defects, one cause: the agent was reporting facts it did not have, or facts
that were wrong at the source. None of them were reasoning failures — each was a
probe publishing something untrue, or a fact never reaching the path that needed
it.

## 1. AWS_REGION alone claimed the process ran in AWS

"What environment are you running in?" answered `Cloud: aws / us-east-1` on a
MacBook. Reported as a hallucination; it wasn't. The probe manufactured it:

```python
if not region and aws_region:
    region = aws_region
    if not provider:
        provider = "aws"        # invents the provider from AWS_REGION alone
```

.env.example ships AWS_REGION=us-east-1 for the AWS integration, so any
laptop with it set claimed AWS and the model quoted it faithfully.

CLOUD_PROVIDER is now required, and the AWS region fallback is gated on the
provider actually being AWS — CLOUD_PROVIDER=gcp with an ambient AWS_REGION
no longer reports gcp / us-east-1.

2. The host OS was absent, then wrong
There was no OS fact at all, so the environment question had only a hostname to
infer from. Added os_family via sys.platform.

Deliberately no version: platform.release() is the kernel release — on
macOS the Darwin number (25.5.0), not the OS version (26.5.2) — so publishing it
as the OS release states a false fact in the block that exists to prevent them.
sys.platform rather than platform.system() also avoids importing this repo's
own shadowing platform package from config/ (precedent:
config/secrets/os_keyring.py).

3. A continuation turn did not know what earlier turns found
Live: turn 1 fetched weather and news and printed the figures; a later turn ran
one slash command, saw no weather tool among its results, and answered
"current weather and news retrieval is unavailable in this environment".

Each continuation is a fresh chat call carrying only the nudge, and
record_conversation_turn stores assistant prose — never tool payloads. So
nothing handed turn N+1 what turn N established.

SessionGoal gains findings; the loop records each evidence-bearing turn's
answer and the nudge surfaces them ("Already established in earlier turns …
treat these as done and do not report them as unavailable"). Bounded to the last
four, truncated to the existing reason limit.

4. The answer path was never told which integrations are connected
_build_integration_guard fired only when the set was empty. With
integrations configured, the answer path was told nothing about them, so
"MySQL is not connected" could not be completed with what is.

The connected set now reaches the answer path.

Also in this branch
gateway/tests/slack/test_routes.py captured its signing timestamp at module
import. The signature check enforces a five-minute replay window, so in a full
suite (~7 min) those signatures are stale and all six tests fail — they pass
alone, which is how it merged in #4876. Timestamps are now per request.

Unrelated to the rest; included because main is currently red.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4959, opened 2026-08-12
feat(background): notification adapter

SawhneySatvik:
Part of #4426

The foundation you scoped: an outbound dispatch seam, adapter registration with capability
metadata, and a durable background-investigation store.

Background RCA notices ran through a hard-coded `if channel == ...` chain in the CLI package
with each vendor imported inline, so no channel could declare background support. That chain is
now an adapter registry in `platform/notifications/` behind an eight-line dispatch loop, and the
four channel bodies live in `integrations/<vendor>/background_adapter.py`.

Records persist on completion under `opensre_home()`, so they outlive the originating REPL
session and another process can read them. `/new` resets the session without deleting them or
the notification preferences.

Placement follows your note: `platform/notifications/`, nothing under `gateway/` or named
`MessagingGateway`. The store follows `task_registry.py` with the `file_bindings.py` write
recipe. Capability is a `frozenset` per adapter, and channel validation now reads the registry,
so the hardcoded channel list is gone and a new adapter becomes selectable without editing the
command module. Registration is triggered when the command runs rather than at REPL boot, which
costs nine modules and pulls no vendor transport.

Left for the second PR, as you scoped it: preferences that survive the process, `/background
list` reading the store instead of session memory, and suppressing the completion notice when
the full report already reached that destination, which means knowing what `publish_findings`
delivered.

Two deliberate non-changes: adapter exceptions still propagate to the runner, and Rocket.Chat's
no-webhook-fallback branch moved byte-identical though the adapter beside it behaves oppositely.

The commit that swaps the chain for registry dispatch touches `bootstrap/adapters.py` and the
entry point only: two files, zero test files, zero `runner.py` lines. The existing suite is the
characterization evidence rather than something rewritten alongside. Git scores the four moved
modules as renames between 70 and 91 percent similarity, so no delivery logic was reworded into
place.

Cross-process retrieval is exercised with two real processes, not a second in-process instance.
Store and runner tests were written failing first: rotation, timestamp coercion, and a raising
store that must neither escape the worker thread nor go unreported. Zero new
`.importlinter.strict` entries.

---

I moved existing channel bodies behind a seam rather than rewriting them, and mirrored
`platform/reporting/delivery_registry.py` rather than inventing a shape, so the diff is mostly
relocation.

Correcting an earlier comment of mine on #4426: I said the bootstrap would live in `tools/`.
#4798 landed `bootstrap/` after that, and it carries no `ignore_imports` entries, so it costs no
layering debt where `tools/` would have cost four lines.

Deliberately no `iter_` on the registry: the report path fans one report out to every vendor,
while this delivers to an ordered list the user chose, rendered in that order.

The SMTP config lookup sat above the chain and reaches the OS keyring, so it ran on every
completion even with no channel configured. It now lives in the email adapter. That saves the
keyring lookup; the `keyring` module itself still loads through the vendor package chain.

Persistence sits in a `finally`, guarded by `except Exception` so a locked store cannot dump a
traceback into the terminal from the daemon thread, and not `BaseException` so Ctrl-C is still
honoured. The worker copies its context so a bound org scope survives the thread, matching
`memory_extraction._schedule_coalesced`.

Self-review caught three defects. `list_recent` sorted with `reverse=True`, which a stable sort
hands back oldest-first when timestamps tie. The sort key called `float()` raw where `from_dict`
coerces, so one hand-edited timestamp would have ended persistence silently. And the
registration step relied on an import side effect, so it could not refill a cleared registry
while still reporting success, which mattered once channel validation started reading that
registry. All three have regression tests that fail against the old code.

Unrelated: `tests/synthetic/**_axis2.py` are marked `axis2` but not `synthetic`, so
`-m "not synthetic"` still runs live-LLM scenarios.

---


SawhneySatvik:
@greptile review

muddlebee:
@SawhneySatvik looks good, but for change this big we would need a proper e2e demo covering each functionality and showing the opensre integration and CLI etc.

muddlebee:
Also I notice docs aren't update for this feature PR, please do that as well

SawhneySatvik:
hey @muddlebee cool, will update the docs as well as will add screenshots like below for verification.

image

SawhneySatvik:
This is the first; the second is now up as #4974, and the two together close the issue.

#4974 is where a completed RCA becomes retrievable from Telegram, Slack or Discord in plain text, and where notify preferences survive a restart. It is also where docs/background-investigations.mdx changes, for the reason above.

SawhneySatvik:
Terminal Screenshots for the above PR

99-full-session
07-list-after
08-show
4549-b93f-44fabe101687" />
10-tests
06-investigation

SawhneySatvik:
Telegram screenshot
Screenshot 2026-08-13 at 12 46 03 PM

SawhneySatvik:
@muddlebee I am still working on the second half of the issue #4974, I have attached all the screenshots of tui and telegram above, can you review these once. Also completed the e2e.

muddlebee:
@SawhneySatvik thank you I will take a look today

muddlebee:
@SawhneySatvik nice work :+1:

PR #4968, opened 2026-08-13
fix(sandbox): forward SystemRoot so allow_network works on Windows

### LINKED ISSUE #4937: [BUG] Sandbox allow_network is broken on Windows: SystemRoot missing from the subprocess env breaks Winsock (WinError 10106)
## Summary

## Root cause analysis

`socket.socket()` raises

```
```

```python
import os, subprocess, sys
for label, env in (
):
```

```
```

## Steps to reproduce

```
```

Yes

## Operating system

Windows

## Proposed solution

pass there.

rawsun007:
Fixes #4937

`_sandbox_env` in `platform/sandbox/runner.py` builds the subprocess environment from the `_BASE_ENV_KEYS` allowlist. `SystemRoot` was not on it. Windows resolves the Winsock service provider catalogue relative to `%SystemRoot%`, so every socket call in the sandboxed child raised `WinError 10106` before any sandbox rule was consulted, and `allow_network=True` could not work there at all.

This adds `SystemRoot` to the allowlist, plus a regression test.

Two details worth calling out:

- The bug was invisible by default. With `allow_network=False` the preamble replaces `socket.socket` with a stub that raises `PermissionError`, so real Winsock initialization never happens and the network-block tests pass for an unrelated reason. Only the `allow_network=True` paths reach real sockets.
- No platform branch is needed. `SystemRoot` is absent on POSIX, and the existing `if value:` guard in `_sandbox_env` skips it there.

The variable in isolation, independent of OpenSRE (Windows):

```
```

Focused suite run locally (macOS), per CI.md section 2 — `platform/sandbox/` maps to `tests/sandbox/`:

```
$ uv run python -m pytest tests/sandbox/ -q
24 passed in 5.12s

$ uv run python -m pytest tests/tools/test_python_execution_tool.py tests/tools/test_python_execution_runtime_inputs.py tests/platform/sandbox/ -q
40 passed in 8.69s
```

The new test fails on `main` and passes with the fix:

```
$ git stash push platform/sandbox/runner.py && uv run python -m pytest tests/sandbox/test_runner.py -k system_root -q
FAILED tests/sandbox/test_runner.py::TestSandboxEnvironment::test_system_root_forwarded_and_allowlist_stays_narrow
1 failed, 23 deselected
```

Baseline checks: `make lint`, `make format-check`, `make typecheck` (plus mypy directly on the changed test file, since CI typecheck does not cover `tests/`), and `make verify-integrations-smoke` — all clean.

I could not run the two Windows tests named in the issue on Windows hardware, so that half rests on the analysis in the issue. The new test is platform-agnostic on purpose: it monkeypatches `SystemRoot` into `os.environ` and asserts `_sandbox_env` forwards it, so it guards the allowlist entry on every CI platform rather than only on the one where the symptom appears.

---

The problem is a missing entry in an environment allowlist. The sandbox deliberately hands the child process a narrow environment, and that narrowness accidentally broke a platform primitive: on Windows the loader needs `%SystemRoot%` to bring up Winsock, so the child could not create a socket even when the caller had explicitly opted into network access.

I considered two alternatives:

1. A Windows-only branch in `_sandbox_env` (`if sys.platform == "win32": ...`). Rejected — it adds a platform conditional to a function that currently has none, for no behavioral gain. The existing `if value:` guard already makes the entry a no-op on POSIX, where the variable does not exist.
2. Copying a broader set of Windows variables (`SystemDrive`, `windir`, `COMSPEC`). Rejected as scope creep. `SystemRoot` is the one the reported failure needs; widening the allowlist speculatively weakens the isolation the allowlist exists to provide.

So the fix is the single allowlist entry, with a comment recording *why* it is there — otherwise it looks like an arbitrary Windows path and is a prime candidate for a future cleanup to delete.

The test asserts both halves of the contract in one case, because they are the same decision: `SystemRoot` is forwarded, and an unrelated variable (`AWS_SECRET_ACCESS_KEY`) is still dropped. A test that only checked the first would pass just as well if someone "fixed" this by forwarding all of `os.environ`. It calls `_sandbox_env` directly rather than spawning a subprocess, so it is fast and does not depend on the host OS.

Edge cases considered: the variable being absent (POSIX — skipped by the existing falsy check, which is why `tests/sandbox/` still passes on macOS); an empty value (also skipped by the same guard); and `extra_env` overriding it, which still works since `extra_env` is applied after the base loop.

---

### DIFF
--- a/platform/sandbox/runner.py
+++ b/platform/sandbox/runner.py
@@ -25,6 +25,10 @@ _BASE_ENV_KEYS = (
     "PYTHONPATH",
     "REQUESTS_CA_BUNDLE",
     "SSL_CERT_FILE",
+    # Windows resolves the Winsock service provider catalogue relative to
+    # %SystemRoot%; without it every socket call in the child raises
+    # WinError 10106. Absent on POSIX, where the lookup below skips it.
+    "SystemRoot",
     "TMPDIR",
 )
 
--- a/tests/sandbox/test_runner.py
+++ b/tests/sandbox/test_runner.py
@@ -5,10 +5,13 @@ from __future__ import annotations
 import os
 import tempfile
 
+import pytest
+
 from config.constants import OPENSRE_TMP_DIR, ensure_opensre_tmp_dir
 from platform.sandbox.runner import (
     MAX_TIMEOUT,
     SandboxResult,
+    _sandbox_env,
     run_python_sandbox,
 )
 
@@ -72,6 +75,22 @@ class TestSandboxRunnerInputInjection:
         result = run_python_sandbox("x = 1", inputs=None)
         assert result.success
         assert result.inputs == {}
+
+
+class TestSandboxEnvironment:
+    def test_system_root_forwarded_and_allowlist_stays_narrow(
+        self, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        # Windows needs %SystemRoot% to initialize Winsock in the child, so a
+        # sandbox without it fails every socket call with WinError 10106 before
+        # allow_network is ever consulted.
+        monkeypatch.setenv("SystemRoot", r"C:\Windows")
+        monkeypatch.setenv("AWS_SECRET_ACCESS_KEY", "must-not-leak")
+
+        env = _sandbox_env(None)
+
+        assert env["SystemRoot"] == r"C:\Windows"
+        assert "AWS_SECRET_ACCESS_KEY" not in env
 
 
 class TestSandboxNetworkRestrictions:


rawsun007:
Greptile came back 5/5 with no actionable findings, and all 13 required checks are green, so no code changes on my side.

One thing a reviewer should know before reading that green tick as platform proof: `windows quality` and `windows test` both show `skipping`. Per [`ci-labels-windows.yml`](https://github.com/Tracer-Cloud/opensre/blob/main/.github/workflows/ci-labels-windows.yml) those jobs are gated on `if: contains(github.event.pull_request.labels.*.name, 'ci:windows')`, and I cannot add labels from a fork.

That is the one CI path that would actually exercise this fix on the platform it targets — the two tests named in the issue (`test_network_access_can_be_enabled`, `test_socket_reachability_check_works_with_allow_network`) are in the `windows test` job's pytest selection. Could a maintainer add the `ci:windows` label?

Everything currently green ran on Ubuntu, where `SystemRoot` does not exist and the new allowlist entry is a no-op by design — that confirms no POSIX regression, which is the other half worth checking, but it cannot confirm the fix.

PR #4976, opened 2026-08-13
fix(agent): finish the live query in evidence gather instead of stopping at discovery

YauhenBichel:
Problem. Asked a metric question against a connected MCP source, the agent spent its gather budget listing tools and probing schemas, then replied with the HogQL it would have run. The turn looked successful — tools executed, no errors — so nothing rejected it. Three separate mechanisms let that through: the gather reviewer counted discovery calls as evidence and was told "when in doubt, GOAL_REACHED"; the circuit breaker was tool-scoped, so every sibling tool on a dead host re-paid its own connect timeout; and breaker state was per-gather, so each session-goal continuation started blind and re-paid them again.

Change.

- gather_discovery_budget.py (new) classifies a call as discovery by exact tool-target name. Prefix matching was tried first and rejected — it read search_tweets as search.
- The gather reviewer now taps (tool_name, args) rather than names alone, so call_posthog_tool(tool_name=…) is distinguishable. A turn where every call was discovery is rejected deterministically, before the LLM review. The review prompt's tiebreak flipped to NOT_REACHED if no metric query ran.
- SourceCircuitBreaker marks the source when a host has never answered this run, not just the one tool. A later success clears the source mark, so a vendor that just returned evidence is never blocked.
- gather_unreachable.py (new) carries those marks on SessionCore across run_until_session_goal iterations; cleared by /new and /resume.
- The answer path is given the valid /integrations setup ids and preferred-source coverage, so it stops inventing vendors.
- PostHog contributes its own gather-prompt fragment via register_gather_prompt_fragment instead of vendor rules living in the shared prompt.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4982, opened 2026-08-13
fix(agent): finish the metric query and keep vendor dialect out of core

YauhenBichel:
Problem. Two failure modes with the same shape: the agent stops short of the data and states something it did not check.

A metric ask against a connected MCP source could spend the whole gather listing tools and probing schemas, then reply with the query it would have run. The turn looked successful — tools executed, no errors — so nothing rejected it. Separately, a signup/retention ask could substitute user_signed_in for a signup event and publish a retention percentage derived from logins. And on a laptop with no cloud configured, replies asserted cloud provider aws / region us-east-1 — a training prior no prompt wording displaces.

Change.

- Gather now distinguishes discovery from a real fetch by exact tool-target name, caps discovery spend, and rejects a discovery-only conclusion deterministically before the LLM review.
- SourceCircuitBreaker marks the source when a host has never answered, so sibling tools don't each re-pay a connect timeout; gather_unreachable carries those marks across session-goal turns and clears on /new and /resume.
- An unformed metric answer gets a labeled draft-query fence plus one setup line, then stops — never a fabricated number.
- Vendor dialect moved out of core/. HogQL fences, PromQL fences, and signup-event parsing now register through platform.harness_ports (register_metric_query_draft, register_metric_cohort_resolver, register_metric_cohort_goal_matcher, register_metric_cohort_unverified_detector) from integrations/posthog_mcp/ and integrations/grafana/. Core keeps the host protocol in cohort_identity.py and stays source-agnostic, so a new integration opts in where it lives instead of editing core/.
runtime_fact_check.py reports replies that contradict the authoritative runtime-facts block, so the turn composes a correction rather than restating the prior.
- The answer path is told which integrations are connected, and the valid setup ids only when something is connected — on an empty session that roster is several hundred characters of vendor names that crowd out the question.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4986, opened 2026-08-13
fix(secrets): stop reading the OS keychain, so onboarding never asks permission

YauhenBichel:
Problem. Launching opensre raised a macOS dialog — "opensre wants to use your confidential information stored in opensre.llm" — and it came back on every run. Every secret lookup went through the keychain, and macOS gates each read. Someone had already noticed and added OPENSRE_USE_KEYRING to make keychain writes opt-in, with a comment saying it was so "Mac 'allow' prompts do not block onboarding" — but only writes were covered, so the reads kept prompting.

Change. The keychain is no longer a storage tier. Reads, writes and deletes all resolve environment → owner-only local file (mode 0600), the file tier that already existed as a fallback.

Secrets written by older versions are moved across once, by config/secrets/keychain_import.py. It checks whether an entry exists before reading it — existence costs no approval dialog, only the read does — so a user approves once per key they actually have, not once per provider in the catalog. A marker file then keeps the keychain closed for good.

Retired completely rather than left inert: OPENSRE_USE_KEYRING, keyring_writes_enabled(), SecretLookup.keyring_error and keyring_unreachable (both permanently empty once nothing reads a keychain), and the credential-resolution branch that existed only to tolerate a keychain hiccup.

Verified on a real machine. A full uv run opensre session completed with no dialog while resolving a live OpenAI key, and a direct security find-generic-password still prompted — confirming the entry is present and gated, and that opensre simply stops asking. The migration ran for real: OPENAI_API_KEY and ANTHROPIC_API_KEY moved into ~/.opensre/credentials.json.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4991, opened 2026-08-13
fix(packaging): add the missing package markers that drop code from every wheel

YauhenBichel:
Problem. 
A directory without __init__.py still imports from a source checkout — Python treats it as a namespace package — but [tool.setuptools.packages.find] only collects directories that have one. So eight directories were invisible to every installed build while passing lint, typecheck and the entire test suite.

The visible symptom was ~65 failed to import lines on startup and Skipping tools.system.*, with integrations silently non-functional after the wizard reported them connected.

### DIFF
--- a/gateway/tests/main/__init__.py
+++ b/gateway/tests/main/__init__.py
@@ -1,0 +1,1 @@
+"""Gateway agent life-cycle tests."""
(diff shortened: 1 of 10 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #4992, opened 2026-08-13
Fix: gather_unreachable persists dead-host marks past recovery (#4988)

### LINKED ISSUE #4988: [BUG] gather_unreachable persists dead-host marks forever, even after a tool/source recovers (regression in #4952)
### Summary

gather_unreachable persists dead-host marks forever, even after the tool/source recovers (setdefault-only merge in store_gather_unreachable, regression in #4952)

Expected: Once a previously-unreachable tool or source succeeds again within a session, later SessionGoal turns should retry it. The per-run SourceCircuitBreaker already clears the mark for that run — per its own docstring in source_circuit_breaker.py: "A success for a tool is sticky for the run and clears the source mark... otherwise the next gather iteration blocks a vendor that just returned evidence."

Actual: Once a tool/source is marked unreachable in any turn, it stays blocked for the rest of the session — even after it recovers — until an explicit /new or /resume. store_gather_unreachable (core/agent_harness/turns/gather_unreachable.py:29-40) only calls setdefault when merging the breaker's snapshot into the session-level map, so a key no longer present in the new snapshot is never removed from the persisted store. Every later turn seeds a fresh breaker from this stale map (evidence_driver.py:332-333), so it re-blocks a source that just returned evidence in the current turn.

No exception or error is raised — this is a silent behavioral bug, not a crash. The visible symptom is that the agent stops attempting a data source it previously failed to reach, even after that source is confirmed working again, with no indication why.

### Steps to reproduce

Start a SessionGoal run that queries a tool/source which is temporarily unreachable (e.g. a connect timeout to a Grafana/Kubernetes host). It gets recorded via store_gather_unreachable in the session's gather_unreachable_tools / gather_unreachable_sources map.
Let the tool/source recover (e.g. the transient network issue resolves).
Continue the same session into a later SessionGoal turn that queries the same tool/source again.
Observe that the now-healthy tool/source is still treated as unreachable and skipped, because evidence_driver.py:332-333 seeds a fresh SourceCircuitBreaker from the stale session-level map, and that map was never reconciled to remove the recovered entry.

Also confirmable by code inspection alone, without a live environment:

source_circuit_breaker.py:178-181 — a successful call pops the tool/source out of _tool_down / _source_down before snapshot() is taken.
gather_unreachable.py:29-40 — store_gather_unreachable merges via tool_map.setdefault(...) / source_map.setdefault(...) only, so a key no longer present in the new snapshot is never removed from the persisted session map.

Yes

### How often does it occur?

Every time

### Operating system

Windows

### Logs and error output

No error is thrown — this is a silent logic bug, not a crash. There is no stack trace or exit code to attach. The bug is confirmed via static code inspection of the files below rather than a runtime log:

core/agent_harness/turns/gather_unreachable.py:29-40
core/agent_harness/turns/source_circuit_breaker.py:178-181
core/agent_harness/turns/evidence_driver.py:332-333, 367-368

### Additional context

This is a regression introduced in #4952 (fix(gather): stop re-paying dead-host timeouts and MCP discovery loops across SessionGoal turns). Before #4952, each turn started a fresh SourceCircuitBreaker and would retry a previously-dead host on the next turn. #4952's stated goal was to stop re-paying connect timeouts on hosts that are still dead across SessionGoal turns — but the setdefault-only merge in store_gather_unreachable also permanently blocks hosts that have since recovered, which is a regression relative to pre-#4952 behavior.

Suggested fix: store_gather_unreachable should reconcile the session map against the new snapshot — removing any key no longer present in the snapshot — rather than only ever adding via setdefault.

Bhumika-1432006:
Fixes #4988

`store_gather_unreachable` (`core/agent_harness/turns/gather_unreachable.py`) merged the
`SourceCircuitBreaker`'s per-run snapshot into the session-level unreachable-tool/source map
using `setdefault` only. `setdefault` can add a key but never remove one, so once a tool or
source was marked unreachable in any turn, it stayed marked unreachable for the rest of the
session — even after it recovered — because a recovered name is *absent* from later snapshots
(the breaker pops a name out of `_tool_down`/`_source_down` on success, before `snapshot()` is
taken), and nothing ever reconciled that absence back into the persisted session map. Every
later `SessionGoal` turn then seeded a fresh breaker from the stale map and silently skipped a
source that had just returned evidence, with no error or indication why.

The fix changes `store_gather_unreachable` to rebuild the session map from the new snapshot on
every call: a name present in the snapshot is kept (and keeps its original down-reason — first
reason wins, unchanged from before), and a name **absent** from the snapshot is dropped, since
its absence means the breaker itself already cleared it. This makes the session-level map track
reality instead of growing monotonically.

This is a session-state bookkeeping fix with no UI surface, so the regression test is the
demonstration:

```
tests/core/agent_harness/test_source_circuit_breaker.py::test_store_gather_unreachable_drops_recovered_tool_from_session PASSED
tests/core/agent_harness/test_source_circuit_breaker.py::test_store_gather_unreachable_keeps_first_reason_for_still_down_tool PASSED
```

The first test reproduces the bug end-to-end at the `store_gather_unreachable` level: mark a
tool/source unreachable, then call it again with an empty snapshot (simulating recovery), and
assert the tool/source is gone from the session map rather than still present.

---

The problem is that `store_gather_unreachable` only ever added keys to the session-persisted
unreachable map (via `dict.setdefault`), with nothing to remove a key once the corresponding
tool/source recovered. I confirmed the mechanism by reading `source_circuit_breaker.py:168-199`:
a successful tool call pops that tool (and, unless another tool on the source already succeeded,
the source) out of `_tool_down`/`_source_down` under lock, *before* `snapshot()` returns a copy
of those two dicts. So `breaker.snapshot()` is not a delta of new failures — it is the breaker's
full, current down-set. That distinction is what made the fix straightforward: since the
snapshot is authoritative for "what's down right now," the session map for each call can simply
be *rebuilt* from the snapshot rather than merged into indefinitely.

I considered two approaches:
1. Explicitly diff the old and new maps and delete only the keys no longer present (an
   in-place `del` for the difference), preserving the rest as-is.
2. Rebuild the new map from the snapshot's keys, looking up each key's existing reason (falling
   back to the snapshot's own reason for a name that's newly down).

I chose (2) because it's shorter and has no separate "compute the removed set" step — the
snapshot itself already defines exactly which keys should exist afterward, so building the new
map directly from it can't drift from that invariant. It also keeps the "first reason wins"
semantics for currently-down names for free: `tool_map.get(name, snapshot_reason)` returns the
existing reason when there is one, otherwise falls back to the reason on the new snapshot,
without needing a separate `setdefault` pass.

Key components:
- `load_gather_unreachable(session)` — unchanged; returns a plain-dict snapshot of whatever is
  currently stored on `session`.
- `store_gather_unreachable(session, *, tools, sources)` — now builds `new_tool_map` /
  `new_source_map` as dict comprehensions over the incoming `tools`/`sources` snapshot, each
  entry keeping the pre-existing reason when the name was already tracked. It then mutates the
  existing session-level dict in place (`clear()` + `update()`) when one is present, or sets a
  fresh dict via `setattr` otherwise — this part is unchanged, since `SessionCore`'s `/new` /
  `clear()` path and any other live references to that dict need to keep seeing the same object.
- `clear_gather_unreachable` — unchanged; not part of this bug.

Edge case I specifically tested: a name that is still down across two consecutive
`store_gather_unreachable` calls must keep its *original* reason, not the newer call's reason —
covered by `test_store_gather_unreachable_keeps_first_reason_for_still_down_tool`.

---

---

**Testing:**

Two new regression tests were added to
`tests/core/agent_harness/test_source_circuit_breaker.py` (the existing home of
`store_gather_unreachable`/`load_gather_unreachable` coverage, alongside the `SourceCircuitBreaker`
tests they integrate with):
- `test_store_gather_unreachable_drops_recovered_tool_from_session`
- `test_store_gather_unreachable_keeps_first_reason_for_still_down_tool`

Ran per this repo's `core/` → `tests/core/` scope mapping in
`.github/ci/test_scope_rules.py`:

- `uv run python -m pytest tests/core/agent_harness/test_source_circuit_breaker.py -q` →
  **18 passed** (16 existing + 2 new).
- `uv run python -m pytest tests/core/ -q` → **2091 passed, 33 skipped, 1 failed**. The one
  failure, `tests/core/agent_harness/test_chat_api.py::test_only_chat_api_module_defines_dispatch_chat_turn`,
  is pre-existing and unrelated to this change: it compares a `pathlib.Path.relative_to(...)`
  result against a hardcoded forward-slash literal (`"core/agent_harness/turns/chat_api.py"`),
  which fails on Windows because `Path` renders with backslashes there. Nothing in this PR
  touches `chat_api.py` or any path-formatting code, and the failure reproduces identically on
  `main` before this change.
- `uv run ruff check` and `uv run ruff format --check` on both touched files → clean.
- `uv run mypy` over the full `make typecheck` scope (`bootstrap config core gateway
  integrations platform surfaces tools`) → 69 pre-existing errors from missing third-party type
  stubs in this dev environment (`psycopg2`, `slack_sdk`, `sentry_sdk`, etc.); none reference
  `gather_unreachable.py` or the changed test file.

### DIFF
--- a/core/agent_harness/turns/gather_unreachable.py
+++ b/core/agent_harness/turns/gather_unreachable.py
@@ -32,26 +32,36 @@ def store_gather_unreachable(
     tools: dict[str, str],
     sources: dict[str, str],
 ) -> None:
-    """Merge unreachable marks onto ``session`` (setdefault — first reason wins)."""
+    """Reconcile ``session`` with the latest breaker snapshot.
+
+    ``tools``/``sources`` is the breaker's full current down-set, not a
+    delta: a name absent from it recovered (the breaker pops a name on
+    success before snapshotting), so it is dropped from the session-level
+    map rather than kept forever. A name still present keeps its original
+    reason — first reason wins — rather than being overwritten by the new
+    snapshot's summary for the same name.
+    """
     tool_map, source_map = load_gather_unreachable(session)
-    for name, summary in tools.items():
-        tool_map.setdefault(str(name), str(summary))
-    for name, summary in sources.items():
-        source_map.setdefault(str(name), str(summary))
+    new_tool_map = {
+        str(name): tool_map.get(str(name), str(summary)) for name, summary in tools.items()
+    }
+    new_source_map = {
+        str(name): source_map.get(str(name), str(summary)) for name, summary in sources.items()
+    }
     # Prefer mutating existing maps when SessionCore fields are present so
     # ``/new`` ``clear()`` and live references stay consistent.
     existing_tools = getattr(session, _TOOLS_ATTR, None)
     existing_sources = getattr(session, _SOURCES_ATTR, None)
     if isinstance(existing_tools, dict):
         existing_tools.clear()
-        existing_tools.update(tool_map)
+        existing_tools.update(new_tool_map)
     else:
-        setattr(session, _TOOLS_ATTR, tool_map)
+        setattr(session, _TOOLS_ATTR, new_tool_map)
     if isinstance(existing_sources, dict):
         existing_sources.clear()
-        existing_sources.update(source_map)
+        existing_sources.update(new_source_map)
     else:
-        setattr(session, _SOURCES_ATTR, source_map)
+        setattr(session, _SOURCES_ATTR, new_source_map)
 
 
 def clear_gather_unreachable(session: Any) -> None:
--- a/tests/core/agent_harness/test_source_circuit_breaker.py
+++ b/tests/core/agent_harness/test_source_circuit_breaker.py
@@ -388,6 +388,46 @@ def test_session_core_clear_drops_gather_unreachable_carry() -> None:
     assert sources == {}
 
 
+def test_store_gather_unreachable_drops_recovered_tool_from_session() -> None:
+    """A tool absent from a later snapshot recovered and must not stay blocked.
+
+    Regression for #4988: ``store_gather_unreachable`` used to merge via
+    ``setdefault`` only, so a name once marked down was never removed even
+    after the breaker itself stopped reporting it as down.
+    """
+    session = SimpleNamespace()
+
+    # Arrange: first gather turn marks the tool unreachable.
+    store_gather_unreachable(
+        session,
+        tools={"query_grafana_metrics": "timeout"},
+        sources={"grafana": "timeout"},
+    )
+    assert load_gather_unreachable(session)[0] == {"query_grafana_metrics": "timeout"}
+
+    # Act: the tool succeeds on a later turn, so the fresh breaker snapshot no
+    # longer reports it (or its source) as down.
+    store_gather_unreachable(session, tools={}, sources={})
+
+    # Assert: the recovered tool and source are gone, not merely unchanged.
+    tools, sources = load_gather_unreachable(session)
+    assert tools == {}
+    assert sources == {}
+
+
+def test_store_gather_unreachable_keeps_first_reason_for_still_down_tool() -> None:
+    """A name still present in the snapshot keeps its original reason."""
+    session = SimpleNamespace()
+
+    store_gather_unreachable(session, tools={"query_grafana_metrics": "timeout"}, sources={})
+    store_gather_unreachable(
+        session, tools={"query_grafana_metrics": "different reason"}, sources={}
+    )
+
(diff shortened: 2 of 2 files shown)


Bhumika-1432006:
@greptile review

PR #4993, opened 2026-08-13
fix(prompts): give the action agent the same runtime facts as the assistant

YauhenBichel:
Problem. Two speakers can end a turn in user-visible prose — the action agent and the conversational assistant — and only one of them was told where it was running. runtime_facts appeared exclusively under prompts/assistant/; the action prompt carried none, and never mentioned the word "cloud" at all.

A speaker asked to answer without facts fills them in. That's the shape behind a laptop turn reporting it ran in AWS, and behind the Slack turn that denied being the bot it was addressed as.

Change. The action envelope now carries render_static_runtime_facts(capture_runtime_facts()) — the same producer and the same block, verbatim, as the assistant. Placed at PromptTier.STABLE, so it rides in the cached half rather than moving the breakpoint each turn; verified it lands in render_cached() and does not leak into render_ephemeral().

Also in here, because the change forced them:

PromptBlockId(StrEnum) — 14 block ids across the action and gather envelopes. require_block on a mistyped literal returns a missing block rather than raising, so producer and consumer now agree by construction. Zero id literals remain in prompt code.
A complexity guardrail failed (build_action_system_prompt_envelope 7 → 8). Fixed by collapsing five if content: blocks.append(...) guards into _optional_block(...) returning a 0-or-1 list, rather than raising the cap.
Frozen static facts in the characterization test. The real block carries hostname, kubeconfig path and scratchpad dir; regenerating unfrozen would have baked one developer's laptop into a snapshot every other machine then fails. Verified no machine-specific values leaked.


YauhenBichel:
@greptile review

PR #4994, opened 2026-08-13
fix(onboarding): let a blank LLM key defer setup instead of blocking the wizard

YauhenBichel:
Problem. The LLM credential prompt was the one onboarding step with no "later". Its outcomes were repick, cancel, save_anyway and continue_unsaved — so a user without a key to hand could only go back to the provider menu or abandon setup.

It was worse than a missing option. _prompt_value is while True and returns only on non-empty input, so a blank answer loops forever. The regression test written for this didn't fail — it hung for 600 seconds and had to be killed. In a terminal you experience it as a wizard that won't move on until you Ctrl-C.

Change. A blank answer now returns DEFERRED: onboarding completes, nothing is validated or persisted, and the summary tells you how to finish:

not set yet — run `opensre auth login openai` when you have a key
Everything downstream already supported this — resolve_for_request raises an actionable MissingLLMCredentialError, and the REPL never gated startup on an LLM key. So a deferred key lands the user in a working shell that names what's missing, instead of a dead end.

The prompt label now says — leave blank to set up later, and allow_empty is off for host credentials, where a bare Enter would accept the pre-filled default.

Also: credential_kind string literals → WizardCredentialKind. 17 comparisons across 6 files. The enum already existed and the field was already typed as it; the literals bypassed it, which is how a typo becomes a silent no-match rather than an error.

### DIFF
--- a/surfaces/cli/llm_auth/providers.py
+++ b/surfaces/cli/llm_auth/providers.py
@@ -5,7 +5,12 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import Literal
 
-from surfaces.cli.wizard.config import PROVIDER_BY_VALUE, SUPPORTED_PROVIDERS, ProviderOption
+from surfaces.cli.wizard.config import (
+    PROVIDER_BY_VALUE,
+    SUPPORTED_PROVIDERS,
+    ProviderOption,
+    WizardCredentialKind,
+)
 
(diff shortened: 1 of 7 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5000, opened 2026-08-14
fix(shell): answer key-less LLM turns with auth-login guidance instead of raw provider errors

YauhenBichel:
Follow-up to #4994. Deferring the LLM key during onboarding is now reachable, but the first shell turn on a key-less install answered with the provider SDK's raw exception text ("Missing credentials. Please pass an `api_key`, `workload_identity`, `admin_api_key`, or set the `OPENAI_API_KEY` …") — no hint of what to actually do.

- `core/llm_invoke_errors.py` — new `remediate_missing_llm_credentials()`:  recognizes the "no key at all" provider phrasings and returns actionable  text naming the exact command (`opensre auth login <provider>`, or  `opensre onboard` to rerun setup), with the provider detail kept in  parentheses. Every other failure kind (invalid key, quota, timeout)  returns None so existing rendering is unchanged.
- `core/agent_harness/turns/action_driver.py` — the action-agent failure path renders and returns the remediation as the turn's response; telemetry  still stages the raw error so analytics classification is unaffected.
- `core/agent_harness/turns/orchestrator.py` — same treatment on the  answer-stream path; the `assistant failed: …` fallback stays for other  errors.
- Tests: unit coverage of the helper against the exact SDK message (plus  invalid-key / quota / timeout negatives), and an action-path regression  asserting the reply carries the login command while staged telemetry keeps  the raw text.

Demo

image

### DIFF
--- a/core/agent_harness/turns/action_driver.py
+++ b/core/agent_harness/turns/action_driver.py
@@ -58,6 +58,7 @@ from core.execution import (
 )
 from core.llm.failure_classification import is_context_length_overflow
 from core.llm.types import AgentLLMResponse, ToolCall
+from core.llm_invoke_errors import remediate_missing_llm_credentials
 from core.tool_framework.tags import SUMMARIZE_OBSERVATION_TAG
 from platform.analytics.react_turn import run_react_agent_with_telemetry
 from platform.observability.trace.prompts import persist_turn_system_prompt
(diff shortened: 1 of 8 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5002, opened 2026-08-14
refactor(llm-errors): one ordered classifier for provider failures

YauhenBichel:
Follow-up to #5000. The provider-failure string matching in `core/llm_invoke_errors.py` had grown into five overlapping pattern tuples (`_NOT_CONFIGURED_PATTERNS`, `_QUOTA_PATTERNS`, `_MISSING_KEY_PATTERNS`, `_REJECTED_KEY_VETO_PATTERNS`, `_AUTH_PATTERNS`) feeding two independent classifiers that could disagree: Anthropic's missing-key message ("Could not resolve authentication method…") was bucketed as `auth` for analytics while the shell correctly showed missing-key login guidance for the same text.

- One `ProviderFailureKind` StrEnum (`missing_key | rejected_key | quota |  not_configured | provider_error`) and one ordered first-match-wins rule  table (`_FAILURE_RULES`), most specific first. The five tuples are removed  completely — no forwarding names.
- `classify_llm_provider_failure()` is the single classification; both  consumers derive from it:
  - `remediate_missing_llm_credentials()` fires iff the kind is    `missing_key` (the #5000 rejected-key protection becomes rule *order*    instead of a separate veto tuple).
  - `classify_provider_error_kind()` maps the enum onto the existing    four-value analytics vocabulary (`missing_key`→`not_configured`,    `rejected_key`→`auth`), so PostHog dashboards keep filtering on the same  strings.
- Behavior-preservation is characterization-tested: the pre-existing 15-row  analytics table passes unchanged, as do all rejected-key-with-env-name  negatives from #5000. A new property test over seven real provider  messages pins the invariant that remediation, kind, and analytics can  never disagree again.

### DIFF
--- a/core/llm_invoke_errors.py
+++ b/core/llm_invoke_errors.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
+from enum import StrEnum
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


YauhenBichel:
@greptile review

PR #5003, opened 2026-08-14
fix(shell): explain first-run choices — greeting, loops picker, alert listener, verify next steps

YauhenBichel:
Onboarding dogfood polish: four places where the shell showed a state without
saying what to do about it, found by walking the first-run experience by hand.

- **First-run greeting** — the ready panel greeted every user with  "Welcome back", including installs where the wizard has never run. The  identity block now uses the same `_is_first_run()` signal the tips column  already uses: fresh installs get "Welcome <name>!", returning installs keep  "Welcome back <name>!".
- **Loops picker explains itself** — the startup picker asked users to "pick  one to set up" without saying what a loop is, what selecting does, or  whether it can be undone. A dim explainer line now precedes the menu:  loops are scheduled weekday check-ins, picking one runs a first pass and  then offers the schedule, nothing is saved until confirmed, manage with  `/loops`. The wording matches the actual mechanics (canned prompt →  `propose_scheduled_delivery` → `/cron add` confirmation). The console is  passed in from `main.py` — optional parameter, no globals.
- **Alert listener says how to enable it** — the banner line becomes  "○ not configured — /alerts for setup", and `/alerts` on an inactive  listener now prints the enable paths (`alert_listener_enabled: true` in  `~/.opensre/config.yml`, or `OPENSRE_ALERT_LISTENER_ENABLED=1`) instead of  only a warning.
- **`/integrations verify` names the fix** — after the table, the summary  adds "Reconfigure with `/integrations setup <service>` — the detail column  above names what is missing", in both the all-integrations and  single-service forms.

image

### DIFF
--- a/surfaces/interactive_shell/command_registry/alerts.py
+++ b/surfaces/interactive_shell/command_registry/alerts.py
@@ -15,6 +15,11 @@ def _cmd_alerts(_session: Session, console: Console, _args: list[str]) -> bool:
     inbox = get_current_inbox()
     if inbox is None:
         console.print(f"[{WARNING}]alert listener is not active.[/]")
+        console.print(
+            f"[{DIM}]Enable it with alert_listener_enabled: true in "
+            "~/.opensre/config.yml (interactive section) or "
+            "OPENSRE_ALERT_LISTENER_ENABLED=1, then restart the shell.[/]"
+        )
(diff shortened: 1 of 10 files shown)


YauhenBichel:
@greptile review

PR #5004, opened 2026-08-14
refactor(gateway): one transport API — contract, middleware, and a conformance ledger

YauhenBichel:
Problem. The four chat transports shared a real bridge to the harness at only two points — a start function and the turn callback. Everything between was reinvented per transport, by copying the nearest neighbour: Discord's security module diffs empty against Slack's after substituting the platform name, and Buzz's own docstrings say it mirrors Telegram. The cost was concrete — the newest transport shipped with the safety half of the copy dropped: no principal binding (sessions silently degraded to legacy empty-key rows), no turn timeout, no /stop, no credit metering — and the entire border-test suite stayed green, because every cross-transport test pins what transports may not import, and none pinned what each must contain.

Change, in four steps.

Identity is required, not defaulted. Buzz gets a principal resolver (silo org + sender pubkey, mirroring Telegram), binds the storage scope around session resolution and inside the executor-thread turn body, and SessionResolver.resolve/rotate no longer default principal/actor to None — forgetting them is now a TypeError; legacy empty-key rows require an explicit None. The missing twin of Telegram's scope-propagation test now exists for Buzz.

Four-copy concerns hoisted to gateway/core. identity_policy (the load/save/persist trio, one platform-parameterized module), inbound_decision (the persist → reply → deny → rotate/resolve choreography, with per-transport residue reduced to a send closure, deny style, and an optional resolver lock), and one ROTATE_SESSION constant replacing two definitions and six bare literals. Net −144 lines; the dead per-transport wrappers are removed with no forwarding.

A conformance contract with a shrinking ledger. test_transport_contract.py auto-discovers transport packages (never hand-listed), asserts the registry lists them all, checks seven concerns per transport by source marker, and forbids reintroducing hoisted copies. Known gaps are asserted with exact equality in both directions — losing a concern fails, and closing a gap fails until its ledger entry is deleted. The ledger is the remaining work list: Buzz lacks a turn timeout, /stop and credit metering; Telegram lacks credit metering. New transports get no entries.

The tree now names the roles, API-framework style:

channels/              facade      — registry + start_transports
core/transport_api/    API         — TransportSpec, TransportWorker, GatewaySink, GatewayAgentCallback
core/middleware/       middleware  — inbound_decision, identity_policy, approvals,
                                     active_turns, attention, conversation_locks, terminal_outcome
core/runtime/          service     — turn_handler (harness bridge), manager, daemon
transports/            controllers — slack/ discord/ telegram/ buzz/
sink_protocol.py is deleted into transport_api; the contract types leave chat.py, which is now purely the facade. 61 files re-imported, git mv so history follows, AGENTS.md updated — including finally listing Buzz, absent since it shipped.

One deliberate behavior change. Slack/Discord posted NEW_SESSION_MESSAGE while holding the process-wide resolver lock — an HTTP call serializing every conversation. The shared step releases the lock before sending.

Tests. 14,854 passing; lint, format, typecheck, strict layer contract, and all 3 import checks green. New: 4 identity-policy tests, 6 inbound-decision tests (including "the rotate sentinel is never echoed to the user"), 9 contract tests, 4 Buzz scope tests — ledger mechanics and the sentinel guard mutation-checked.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5005, opened 2026-08-14
feat(gateway/buzz): soft turn timeout and /stop — close two transport-ledger gaps

YauhenBichel:
Buzz was born as a copy of Telegram minus two behaviors the transport conformance ledger (#5004) records as known gaps: a hung turn hung its conversation forever, and nothing could stop a running turn. This PR closes both; only `credit metering` remains in the buzz ledger entry.

- Soft turn timeout: `turn_timeout_seconds` (default 240s, same as the other  chat transports, env `GATEWAY_TURN_TIMEOUT_SECONDS`). On expiry the  placeholder is finalized with the shared timeout message and  `sink.turn_cancel` is set for cooperative cancellation.
- `/stop` handling: the poll loop consumes `/stop` before dispatch — it cancels  the active turn for that conversation via the runtime's new  `ActiveTurnRegistry`, answers "nothing to stop" when idle, and never starts  an agent turn for a cancellation.
- One terminal outcome per turn: timeout, user stop, error, and normal  completion arbitrate through `TerminalOutcomeArbiter.claim()`.

### DIFF
--- a/gateway/core/middleware/active_turns.py
+++ b/gateway/core/middleware/active_turns.py
@@ -1,10 +1,17 @@
 """In-flight turn cancel registry for gateway chat hosts.
 
 Transports serialize turns per conversation, so a ``/stop`` message cannot
-wait behind the same lock as the running turn. Instead each turn registers its
-``turn_cancel`` Event here under a conversation key; ``/stop`` looks up the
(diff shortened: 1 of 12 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5007, opened 2026-08-14
refactor(harness): gateway reaches the agent harness only through its curated public surface

YauhenBichel:
Problem. 

The gateway imported eleven internal harness submodules — turns.host_cancel, turns.headless_dispatch, turns.gather_ports, turns.default_headless_agent, session_goal.{goal,progress,run_until}, accounting.turn_accounting, session.integration_resolution, harness, ports. The harness already has a public host API — the curated, lazily-resolved export table in core/agent_harness/__init__.py, which its own AGENTS.md calls "the package's public surface" — but nothing stopped a host from reaching past it, so the gateway grew around it and any harness-internal rename broke the gateway at a distance.

Change.

The curated table now covers what a host genuinely needs — 11 names added (33 total): SessionCore, SessionManager, SessionGoal, run_until_session_goal, format_session_goal_progress, format_session_goal_status_line, DefaultTurnAccounting, GatherPorts, ensure_turn_cancel, OutputSink, has_resolved_integrations. All lazy (PEP 562), so importing the root stays cheap and interactive-shell boot doesn't drag the turn-driver stack — the property that table exists to protect.
Thirteen gateway files migrated to the single import shape from core.agent_harness import … — the four deep-coupled runtime files (turn_handler, session_agents, live_sink's neighbours, cancel_console), transport_api, the resolver, middleware, and all four transports. Zero core.agent_harness.<submodule> references remain in gateway source.
The border is enforced, not documented: gateway/tests/test_harness_api_border.py scans gateway sources and names any submodule import. Written red-first — it failed with the full eleven-module inventory before the migration. Widening the coupling now requires editing the harness's own export table, a visible reviewable act, instead of a quiet deeper import.

This is the mirror image of #5004: that PR gave transports a contract on the gateway side of the bridge (transport_api); this one gives the gateway a contract on the harness side. Both directions of gateway↔harness are now tested boundaries.

### DIFF
--- a/core/agent_harness/__init__.py
+++ b/core/agent_harness/__init__.py
@@ -16,6 +16,7 @@ import importlib
 from typing import TYPE_CHECKING, Any
 
 if TYPE_CHECKING:
+    from core.agent_harness.accounting.turn_accounting import DefaultTurnAccounting
     from core.agent_harness.harness import (
         AgentSession,
         ChatDispatcher,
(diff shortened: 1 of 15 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5008, opened 2026-08-14
refactor: surfaces harness-import ledger + gateway manager.py renamed to controller.py

YauhenBichel:
Two boundary cleanups, same theme: names and imports that say what they are.

1. Surfaces reach the harness through the curated surface — or the ledger
The twin of #5007. Measuring first changed the shape: surfaces/ imports 29 harness submodules across 35 files, versus the gateway's 11. The interactive shell legitimately assembles the agent stack itself (session persistence, prompt internals, tool providers), so hoisting everything into the curated core.agent_harness.__init__ table at once would un-curate it — the opposite of the point.

So this follows the transport-conformance pattern instead:

10 files migrated now — every import whose names the curated table already exports (SessionManager, AgentSession/SessionConfig, run_until_session_goal, dispatch_chat_turn, OutputSink, TurnResult, …).
tests/interactive_shell/test_harness_api_border.py — the same AST scanner as the gateway twin (catches aliased, line-continued, and importlib.import_module(...) forms), plus a 29-entry _KNOWN_DEEP_IMPORTS ledger asserted with exact equality in both directions: a new deep import fails immediately, and a module that stops being imported must be deleted from the ledger. Mutation-checked both ways.
The ledger doubles as the harness owners' curation worklist — each entry resolves by deliberately exporting the name or deciding it's shell-private, and either way the set only shrinks

2. gateway/core/runtime/manager.py → controller.py
GatewayManager → GatewayController — the class is the gateway's process controller (boot, credential hydrate, compose surfaces, scheduler, signals, stop), and "manager" said none of that. Renamed via git mv, ~27 files updated across bootstrap/, platform/, the CLI entry, tests and docs.

Two things surfaced by the rename, both fixed:

A vocabulary collision, resolved deliberately. The transport docs called transports "the controllers" (framing introduced in #5004). Two meanings of controller would be worse than the old vague manager, so transports are now consistently "ingress adapters" — the phrase turn_handler.py's docstring already used — and controller means exactly one thing.
The minimal .importlinter allowlist was doubly stale: it still permitted gateway.core.runtime.manager -> gateway.channels, naming both the old module and a package deleted two PRs ago. Now controller -> gateway.startup. The stale-named border tests went with it — test_only_manager_imports_channels_module was checking gateway.startup under its old name; it and two siblings are now named for what they actually assert.

### DIFF
--- a/.importlinter
+++ b/.importlinter
@@ -26,8 +26,8 @@ forbidden_modules =
 ignore_imports =
     # Composition root: the one edge into the channels composer, which owns all
     # transport and web wiring. Pinned by gateway/tests/test_package_borders.py
-    # (test_only_manager_imports_channels_module).
-    gateway.core.runtime.manager -> gateway.channels
+    # (test_only_controller_imports_the_startup_module).
+    gateway.core.runtime.controller -> gateway.startup
 
 [importlinter:contract:gateway-web-no-transports]
 name = Gateway web surface never imports chat transports
--- a/.importlinter.strict
+++ b/.importlinter.strict
@@ -41,7 +41,7 @@ ignore_imports =
     platform.scheduler.executor -> integrations.telegram.formatting
     # surfaces -> gateway (cli + repl gateway commands)
     surfaces.cli.commands.gateway -> gateway.core.runtime.daemon
-    surfaces.cli.gateway_entry -> gateway.core.runtime.manager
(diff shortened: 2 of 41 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5009, opened 2026-08-14
refactor(gateway): worker initialization moves out of the gateway root into transports/startup.py

YauhenBichel:
controller.start_surfaces()
  → gateway/startup.py: start_gateway()          # facade: web + chat, one handle
      - → gateway/transports/startup.py: start_transports()   # registry + worker loop
         - - → transports/{telegram,slack,discord,buzz}/startup.py

gateway/startup.py dropped from ~170 lines to ~80 and knows nothing about workers — no registry, no per-transport start functions, no skip/fail loop. All of that lives in gateway/transports/startup.py, next to the platforms it initializes. Its docstring records the peer-isolation reasoning: importing one platform never loads the composer, so one platform ≠ four SDK stacks still holds.


YauhenBichel:
@greptile review

PR #5014, opened 2026-08-15
fix(shell,aws): straight /integrations table under recording + role-ARN setup that explains itself

YauhenBichel:
## Summary

Two user-reported problems, unrelated, fixed on one branch.

### 1. `/integrations` table rendered as a diagonal staircase

Every slash command runs inside `capture_console_segment`, which turns on
`console.record` so the analytics transcript gets a plain-text copy of the
output. `print_repl_table` already had a TTY-safe path (render to a buffer,
one write, `\r\n` per row) but its guard treated a recording console as
"keeps its own copy — a direct stdout write would be invisible", skipped the
safe path, and fell back to row-by-row `console.print`. Under
`patch_stdout(raw=True)` each of those rows ends in a bare `\n`, which moves
down without returning to column 0, so every row inherits the previous row's
ending column. Recording and the safe path were mutually exclusive, and
recording always won; it only looked intermittent because the drift needs
the cursor displaced first (a long spinner run with wrapped probe warnings).

- `rendering.py`: the guard now only bails inside an open `capture()`. Under
  plain recording it takes the CRLF write and feeds the rendered plain text
  into the owning console's record buffer, so the transcript still sees the
  table. Applied to `print_repl_table`, `print_repl_json`, and the launch
  poster. A leading `\r` makes the write immune to wherever a spinner
  teardown left the cursor.
- The REPL had no root log handler, so library WARNINGs fell through to
  Python's last-resort stderr writer, racing the status spinner on the tty.
  `platform/logging/shell_handler.py` prints WARNING+ records through the
  shell's own Rich console (atomic with a live spinner, one row per record)
  and is installed once at REPL start; a non-REPL host keeps its own
  handlers.
- Noise that this made visible, cleaned at the source: `urllib3
  .connectionpool` retry chatter floored to ERROR in `quiet_third_party.py`
  (the outcome is reported by the caller); the shell handler's floor is
  ERROR because a WARNING such as `[kubernetes] probe_access failed`
  duplicates the table's own `failed` row; and `capture_service_error` now
  logs an unreachable service at WARNING, matching the rule the same
  function already used to drop its traceback ("an operational fact, not a
  fault in our code").

Verified by driving the real REPL in a pty at the reporter's 210×59 geometry
and rendering the raw bytes through a VT emulator: baseline STAIRCASE (row
indents 27, 24, 21, …), fixed STRAIGHT on five consecutive runs, and the
`Retrying (Retry(total=2, …))` / `[kubernetes] probe_access failed` lines no
longer appear above the table while the `kubernetes │ failed │ …` row still
does.

Before:
image

After:
image

### 2. AWS "IAM Role ARN" setup was a dead end without base credentials

Assuming a role needs base credentials boto3 can find; the wizard's role mode
collected only the ARN and external ID, then failed STS with the raw
`Unable to locate credentials`. On a fresh laptop that is every user.

- `integrations/aws/credential_chain.py`: local, network-free probe of the
  ambient boto3 chain.
- `surfaces/cli/wizard/configurators/aws.py`: when "IAM Role ARN" is picked
  and the chain is empty, explain why in one sentence and offer three paths —
  switch to access key + secret (default, in place), configure base
  credentials first (prints the `aws configure` / env-var instruction and
  leaves setup cleanly via `WizardBack`), or continue with the role for
  EC2/ECS/Lambda deployments with an attached role.
- `spec_configurator.py`: one generic `on_mode_chosen` hook so any vendor can
  gate a mode on a prerequisite; no AWS knowledge in the shared loop.
- `integrations/aws/verifier.py`: the non-wizard path (`AWS_ROLE_ARN` set by
  hand) returns an actionable message instead of the raw boto3 error; mode
  label and `docs/aws.mdx` no longer imply a role ARN alone is sufficient.

Before:

image

After:

image

image

### DIFF
--- a/config/constants/aws.py
+++ b/config/constants/aws.py
@@ -9,6 +9,13 @@ AWS_ACCESS_KEY_ID_ENV = "AWS_ACCESS_KEY_ID"
 AWS_SECRET_ACCESS_KEY_ENV = "AWS_SECRET_ACCESS_KEY"
 AWS_SESSION_TOKEN_ENV = "AWS_SESSION_TOKEN"
 
+#: boto3 reads these two as one unit: an access key id in the environment
+#: without its secret does not fall through to the next credential source, it
+#: raises ``PartialCredentialsError``. The secret is keyring-only, so a plain
+#: ``.env`` can legitimately hold the id alone — it must then stay out of the
+#: process environment.
+AWS_STATIC_KEY_PAIR_ENV: tuple[str, str] = (AWS_ACCESS_KEY_ID_ENV, AWS_SECRET_ACCESS_KEY_ENV)
+
 __all__ = [
     "AWS_ACCESS_KEY_ID_ENV",
     "AWS_EXTERNAL_ID_ENV",
@@ -16,4 +23,5 @@ __all__ = [
     "AWS_ROLE_ARN_ENV",
     "AWS_SECRET_ACCESS_KEY_ENV",
     "AWS_SESSION_TOKEN_ENV",
+    "AWS_STATIC_KEY_PAIR_ENV",
 ]
--- a/docs/aws.mdx
+++ b/docs/aws.mdx
@@ -8,7 +8,7 @@ OpenSRE uses AWS to map your environment: Lambda functions, EKS clusters, S3 buc
 ## Prerequisites
 
 - AWS account with IAM permissions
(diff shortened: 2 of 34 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5029, opened 2026-08-17
refactor(harness): four import doors by role — ports is the Implement door, spi split into role modules, llm_factory replaces ToolCallingDeps

YauhenBichel:
Follow-up to #5020/#5021/#5023. Goal: an engineer opening `core/agent_harness` should see, in one place each, what to **embed**, what to **implement**, what to **call**, and what **builds and runs** the agent. Reviewers: read `tests/core/agent_harness/test_public_api.py` first — it is the whole contract on one screen.

**1. `core.agent_harness.ports` is the fourth door (Implement).**
All 12 port protocols plus the aliases live there and it is now public: border tests treat it (and `spi.<role>`) as doors, the shell ledger entry for it is retired. `OutputBindable` and the new `LlmFactory` alias are exported.

**2. `.spi` is a package of role modules, not a 48-name facade.**
`spi.py` → `spi/{session_goal,session_flags,cancel,accounting,prompt_chrome,integrations,grounding,defaults}.py` (42 names). The package itself exports nothing (pinned), so a host imports a role on purpose. `stream_answer`, `dispatch_chat_turn`, `ChatTurnBindings` are no longer exported (engine internals; only a test helper used them). `TurnResult`/`ToolCallingTurnResult` move to the api door (`core.agent_harness` is now 7 names — embedders read results). 37 importers repointed.

**3. `ToolCallingDeps` deleted; the seam is `llm_factory: LlmFactory` end-to-end**
(runner, `HeadlessAgent`, `DefaultPorts.agent`, shell `run_action_tool_turn`). It was a one-field wrapper whose field was the name that belonged on the seam.

**4. Naming / embedder API.**
`DefaultPorts.reporter` → `error_reporter` (resolved value private). `AgentSession.start`, `run_headless_turn` and `bootstrap.embedded.start_embedded_session` take explicit kwargs — no `**agent_kwargs`, so an unknown kwarg fails loudly. `resolve_env_variables` / `load_or_create_session` / `load_context` are private steps of `startup()` (only tests called them): `AgentSession` 12 → 9 public methods.

**5. Wide protocol tax removed where it bit.**
`ShellPromptContextProvider` now subclasses `DefaultPromptContextProvider` and overrides three methods (`cli_reference`, `log_diagnostics`, `bind_session`) instead of wrapping it and delegating nine verbatim — and it is now `SessionBindable`, which it silently was not. `SessionState` audited: all 12 attributes have 3–32 harness readers, so it stays as is.

[...]

PR #5031, opened 2026-08-17
refactor(harness): HeadlessAgent.handle is the one host loop for shell and gateway; HeadlessPorts + DefaultPorts are the only construction path

YauhenBichel:
Follow-up to #5029. Two things an engineer reading the harness should now see without doubt: **one per-message API that both hosts call**, and **one way to construct an agent**.

**1. One host loop, provided by the harness — `HeadlessAgent.handle`**
Both hosts hand-rolled the same loop in different words: `bind_turn(TurnBinding(...))`, a `_chat` closure, `run_until_session_goal(...)`, and an `AgentSession` used only as a wrapper around `agent.dispatch` (the shell built a throwaway `AgentSession(SessionConfig(load_env=False))` per turn; the gateway kept `self._session_api`). Now:

```python
handle(text, binding, *, accounting_factory=None, cancel_requested=None, on_progress=None) -> TurnResult
```

binds the turn, dispatches, and continues while a session goal is attached. Gateway and shell each call it once per message and nothing else; dispatch is the single-turn verb underneath. accounting_factory (message → TurnAccounting) is applied per outer turn because the goal loop may dispatch several with different text; None uses the agent's default per-message accounting — so the gateway no longer passes DefaultTurnAccounting by hand.

2. One construction verb — <family>.agent(...)
HeadlessAgent.__init__ is now ports-required (no None → in-memory adapter chain). The in-memory family has a name: HeadlessPorts(session?, output?, reasoning?).agent(tools=…), beside DefaultPorts in core/agent_harness/turns/port_families.py (default_ports.py retired). Hosts, scripts and the ~25 test sites that constructed HeadlessAgent(...) directly all say <family>.agent(...). Runtime door: +HeadlessPorts (7 names).

Hosts after this PR

gateway/core/runtime/turn_handler.py: agent.handle(text, TurnBinding(session, tool_hooks, console, is_tty=False), cancel_requested=…, on_progress=…); _session_api removed.
surfaces/interactive_shell/runtime/shell_turn_execution.py: agent.handle(text, binding, accounting_factory=ShellTurnAccounting per message, cancel_requested=host_cancel_requested(sink), on_progress=print checklist); the throwaway AgentSession is gone.

Docs / internal notes: docs/hosting-the-agent.mdx host loop is build → handle, both shipped hosts shown as the same call, HeadlessPorts documented; core/agent_harness/AGENTS.md table rows updated.

No user-visible behaviour change.

Tests

test_handle_is_the_one_host_loop_binding_each_outer_turn — bind with the factory's accounting, then dispatch; hooks/tty carried in the binding.
test_both_hosts_reach_the_agent_through_handle_only — neither host source contains run_until_session_goal( or AgentSession(; both contain agent.handle(.
tests/shared/fake_agent.py — a mock agent whose handle is the real loop over the mock's bind_turn/dispatch, so gateway tests' bind_turn.call_args / dispatch.call_count assertions keep their meaning.
test_default_ports.py → test_port_families.py; direct constructions converted to HeadlessPorts(...).agent(...); runtime-door pin updated.

### DIFF
--- a/core/agent_harness/AGENTS.md
+++ b/core/agent_harness/AGENTS.md
@@ -22,8 +22,8 @@ Process boot (`configure_process`) and headless construction
 | Happy path (already booted) | `AgentSession.start()` → repeated `.chat` / `.investigate` |
 | Embedded / script | `start_embedded_session()` → repeated `.chat` / `.investigate` |
 | Multi-step / keep-going | `start` / `start_embedded_session` → `.chat_until_goal(...)` (`SessionGoal` loop) |
-| Custom host | **`DefaultPorts(...).agent(...)`** (only construction seam) → `attach_agent` once → many `.chat` |
-| Gateway | `SessionAgentPool` → factory once / session → `bind_turn` → `run_until_session_goal` (same SessionGoal policy as shell) |
+| Custom host | **`DefaultPorts(...).agent(...)`** (or `HeadlessPorts` in-memory; the only construction seam) → `agent.handle(text, TurnBinding(...))` per message |
+| Gateway | `SessionAgentPool` → `DefaultPorts.agent` once / session → `agent.handle(text, TurnBinding(...))` per message (the goal loop lives inside `handle`; same policy as shell) |
 | Scheduled one-shot | `AgentSession.run_headless_turn(...)` (not the multi-turn pattern) |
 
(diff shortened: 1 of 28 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5034, opened 2026-08-17
refactor(wizard): simplify integration configuration and update wizard mode handling

Devesh36:
- Removed the 'mode' parameter from `_configure_selected_integrations` and adjusted the prompt message.
- Updated the wizard mode handling to default to 'quickstart' instead of 'focused' for legacy modes.
- Consolidated next steps rendering logic in `_render_next_steps` to remove conditional checks.
- Deleted the now-redundant focused integrations test file


YauhenBichel (approved):
Thanks Devesh

PR #5036, opened 2026-08-17
refactor(harness): close the shell deep-import ledger to zero; spi.session_state; schema helpers and provider-model resolution leave the harness

YauhenBichel:
- Shell deep-import ledger 8 → 0 — every harness name surfaces/ uses now comes through a door, as the gateway already did. _KNOWN_DEEP_IMPORTS is frozenset().
- Doors: TurnPlan via .runtime (it's the stage protocols' parameter type; .ports would cycle); SELF_RECORDING_ACTION_TOOL_NAMES in a leaf (accounting/self_recording_tools.py) exported by spi.accounting, so spi stays import-cheap; spi.session_flags → spi.session_state (it was already state, not only flags) gaining compact_session_branch, format_recovery_note; JsonlSessionStore in spi.defaults; load_skill_body in spi.grounding.
- Out of the harness: JSON-schema builders (object_schema, string_property, string_array_property, ToolSchema) → core/tool_framework/utils/schema.py (15 importers, one ToolSchema definition); resolve_provider_models → core/llm/provider_models.py, rewritten to read each CLI provider's model env from the provider catalog instead of a hand-copied set + env-name table (repo rule: env names live under config/).
- Pins: test_public_api role lists, tests/shared/harness_doors.py role set (rename), shell ledger. Docs/AGENTS.md role name updated. No behaviour change.


YauhenBichel:
@greptile review

PR #5038, opened 2026-08-17
refactor(harness): HeadlessAgent enforces one turn at a time (AgentBusyError); headless_agent.py, public-first method order, shell seams own their adapters

YauhenBichel:
Follow-up to #5036. Two things: the agent now owns its last invariant that was held by discipline elsewhere, and a readability pass over the harness core files so a cold reader meets the public API first and every docstring is true.

**1. One turn at a time, enforced by the agent**
`HeadlessAgent.bind_turn` mutates per-turn state, so two overlapping turns on one agent would read each other's binding. Until now that was prevented only by the gateway pool's per-session lock and the shell being single-threaded. The agent now holds a reentrant lock acquired **non-blocking** in `handle` and `dispatch` (RLock so `handle → dispatch` nests on the same thread); a second thread raises **`AgentBusyError`** (a `RuntimeError` subclass, exported via `core.agent_harness.runtime`) instead of interleaving. One private context manager guards both entry points; zero cost on the happy path.

**2. Readability review (SoC / clean code)**
- `core/agent_harness/turns/headless_dispatch.py` → **`headless_agent.py`** — the file is named after what it holds. Its module docstring example constructed `HeadlessAgent(...)` directly, which has been impossible since #5031; replaced by the real program (`HeadlessPorts().agent(...).handle(...)`). `__all__` is now only what the module defines (`AgentBusyError`, `HeadlessAgent`); the in-memory adapters are imported from `headless_adapters` by everyone (19 importers repointed). Methods are **public-first in call order**: `handle`, `dispatch`, `bind_turn`, `bind_session`, `bind_stages`, then private helpers.
- `core/agent_harness/harness.py` — docstring rewritten (embed → attach → run; host pointer to `DefaultPorts` + `handle`); dropped drift (`bind_turn each message`, a "there is no `dispatch_message_to_headless_agent`" line, "Path-2"). `AgentSession` public methods first, the three private startup steps last.
- `surfaces/interactive_shell/runtime/shell_turn_execution.py` — the three test-injection stage adapters and their binder moved into **`turn_seams.py`** (the seams and their adapters belong together; `bind_injected_stages` is the public name). The turn module is one screen: build (or reuse) → `TurnBinding` → seams (no-op in production) → `handle`. 210 → 105 lines.

`gateway/core/runtime/turn_handler.py` and `turns/port_families.py` were reviewed and left as they read cleanly.

Not changed, deliberately: `AgentSession.chat_until_goal`. It and `HeadlessAgent.handle` are one-line callers of the single loop function `run_until_session_goal` — the embedder verb and the host verb; delegating one to the other would add a conditional path because an attached agent may be any `ChatDispatcher`.

No user-visible behaviour change.

- [x] Yes, I used AI assistance (Claude Code)

Problem: after seven passes the harness API was right but two things still hurt a cold reader — the agent's "one turn at a time" rule lived in the pool, not the agent, so misuse was silent; and the core files carried a stale docstring, a module named after an old verb, a second door for the in-memory adapters, private helpers ahead of the public API, and a turn module dominated by test-only adapters.

Alternatives considered: (a) a boolean in-turn flag — rejected, an RLock gives the same non-blocking check plus correct same-thread nesting for `handle → dispatch` without a second flag; (b) guarding only `handle` — rejected, `dispatch` is a public entry point too and the invariant is the agent's; (c) delegating `chat_until_goal` to `handle` — rejected after reading it: both call the one loop function; a delegation would add a conditional path.

Why this shape: the invariant is stated once (one context manager, one named error a host can catch); each core file now reads top-down as "what a host uses, then how it works"; seams and their adapters share a module so the turn module shows only the turn.

Key pieces: `HeadlessAgent._one_turn_at_a_time` / `AgentBusyError`, `turns/headless_agent.py`, `turn_seams.bind_injected_stages`.

---

### DIFF
--- a/bootstrap/process.py
+++ b/bootstrap/process.py
@@ -7,7 +7,7 @@ and a profile cannot invent a different sequence.
 Does **not** configure gateway or CLI logging — that stays with the surface
 composition root (``GatewayController.configure_logging``, CLI stderr).
 
-Does **not** construct :class:`~core.agent_harness.turns.headless_dispatch.HeadlessAgent`
+Does **not** construct :class:`~core.agent_harness.turns.headless_agent.HeadlessAgent`
 or run turns — that is ``DefaultPorts.agent`` /
 :class:`~core.agent_harness.harness.AgentSession` after boot. Bootstrap and
 headless construction are separate layers, not duplicated stacks.
--- a/core/agent_harness/AGENTS.md
+++ b/core/agent_harness/AGENTS.md
@@ -5,7 +5,7 @@ tool-calling loop (`core.agent.Agent` via `build_agent`) and the direct-answer
 path (`stream_answer` via the `StreamAnswerFn` seam in `ports.py`, no tools).
 It was extracted out of `interactive_shell` so the same harness can run the
 interactive terminal and be invoked headlessly via
-`agent_harness.turns.headless_dispatch`.
+`agent_harness.turns.headless_agent`.
 
 ## Host API (teach this)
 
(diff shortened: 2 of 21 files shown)


YauhenBichel:
@greptile review

PR #5039, opened 2026-08-17
refactor(gateway): PostgresDatabase owns connection bounds and takes max_connections; investigations store is its second adapter; slack dedup gains connect timeout

YauhenBichel:
Follow-up to #5020, which introduced `gateway/core/storage/postgres.py` (`PostgresDatabase`) with one adapter (Slack event de-dup). This PR gives it its second adapter and moves the connection bounds into it, so the gateway has one pool implementation and one place to tune pool size and timeouts.

**`PostgresDatabase` owns the connection bounds — as its default for every connection**
- The connect-timeout clamp (libpq waits forever when `connect_timeout` is unset or 0; default 10 s, ceiling 30 s, operator-tunable through the DSN) and TCP keepalives move from `investigations/postgres.py` into the leaf. `bounded_connect_timeout` is public and tested.
- `PostgresDatabase(dsn, *, max_connections=5)`: the default fits a transport that issues one statement per request; the investigations store passes 10 (worker + a burst of API threads).

**`PostgresInvestigationStore` is the second adapter**
- Its own `ThreadedConnectionPool` / `_get_pool` / `_connection` (~40 lines) are deleted; it uses `self._db.transaction()` and `run_schema(_SCHEMA)`, with a `database=` injection seam like de-dup.

**Behaviour change (deliberate):** Slack event de-dup previously opened connections with **no** connect timeout — the exact wait-forever hazard the investigations store guarded against on its own. It now gets the same bounds for free. Pinned by `test_dedup_connections_open_under_the_same_bounds_as_every_store`.

**Also:** `docs/hosting-the-agent.mdx` "rules a host must keep" now names `AgentBusyError` (from #5038): an overlapping `handle`/`dispatch` raises rather than interleaves; queue before `handle`, do not catch to retry.

- [x] Yes, I used AI assistance (Claude Code)

Problem: the gateway had two copies of the same Postgres pool — the shared leaf (used by de-dup, without a connect timeout) and the investigations store (with a timeout clamp and keepalives). One adapter on the leaf meant the seam was hypothetical, and the safer bounds lived in only one of the two places.

Alternatives considered: (a) pass the bounds as an option and leave de-dup unbounded — rejected, the wait-forever hazard is not store-specific; (b) leave both pools and share only the clamp helper — rejected, still two pools to keep in step; (c) a per-store `connect_kwargs` override — not needed today; `max_connections` is the only knob the two stores actually differ on, so it is the only parameter.

Why this shape: bounds are a property of every connection the gateway opens, so they belong on the thing that opens connections; each store states only what is its own (schema, pool size); a `database=` seam on both stores makes them testable without psycopg2.

Key pieces: `PostgresDatabase(max_connections=)`, `bounded_connect_timeout`, `_connect_kwargs`, `PostgresInvestigationStore(database=)`.

---


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5044, opened 2026-08-18
fix: correctly derive opensre command name on Windows

shakeer7:
Fixes #

Fixed Windows executable name derivation in `install.ps1` and `installer.ps1`.

The installers previously used `TrimEnd(".exe")` to derive the executable name. `TrimEnd()` removes characters from the end of a string rather than removing the `.exe` extension as a complete suffix.

This change replaces it with `[System.IO.Path]::GetFileNameWithoutExtension($binaryName)`, which correctly derives the executable name from the binary filename.

The fix has been applied consistently to both PowerShell installer scripts.

Tested the updated installer on Windows and verified that the executable name is correctly derived without the `.exe` extension.

---

The issue was caused by using PowerShell's `TrimEnd(".exe")` to remove the executable extension. `TrimEnd()` treats the supplied value as a set of characters to trim, rather than interpreting `.exe` as an exact file extension.

I replaced it with `[System.IO.Path]::GetFileNameWithoutExtension($binaryName)`, which is designed specifically to extract a filename without its extension.

I applied the same fix to both `install.ps1` and `installer.ps1` to keep the Windows installation logic consistent.

This approach avoids manual string manipulation and uses the .NET path API already available in PowerShell.

---

- [x] I have considered potential edge cases and how my changes handle them

### DIFF
--- a/installer.ps1
+++ b/installer.ps1
@@ -1,4 +1,4 @@
-param(
+﻿param(
     [ValidateSet("release", "main")]
     [string]$Channel = $(if ($env:OPENSRE_INSTALL_CHANNEL) { $env:OPENSRE_INSTALL_CHANNEL } else { "main" }),
     [switch]$SkipMain
@@ -1191,3 +1191,4 @@ function Install-OpenSre {
 if (-not $SkipMain) {
     Install-OpenSre
 }
+


shakeer7:
image

PR #5042, opened 2026-08-17
refactor(surfaces): CLI and interactive shell become peers, LLM provider setup moves to shared/llm_setup

YauhenBichel:
## What

The two terminal surfaces imported each other (shell→CLI: 8 files / 11 modules incl. 5 private
names; CLI→shell: 16 files / 13 modules). This PR pins both directions as shrink-only allowlists
and moves what both surfaces need below them.

### Ratchet
- `tests/shared/test_surface_border.py` — exact allowlist per direction (shell→CLI now 1 module,
  CLI→shell 11), and no underscore-prefixed name may cross. `docs/ARCHITECTURE.md`: the two
  terminal surfaces are peers and do not import each other.

### LLM provider setup → `surfaces/shared/llm_setup/` (was `surfaces/cli/wizard/*` + `surfaces/cli/llm_auth/`)
- `catalog.py` (provider table: `SUPPORTED_PROVIDERS`, `PROVIDER_BY_VALUE`, `ProviderOption`),
  `env_sync.py`, `auth_profiles.py`, `auth_service.py`, `persist.py`, `validation.py`,
  `azure_validation.py` (live Azure check, out of the wizard UI module), `ollama.py`
  (`normalize_model_tag`), `openai_client.py`, `validation_result.py`.
- Below both surfaces, above `integrations`/`core`/`config`; the shell's `/model` and the CLI's
  wizard and `auth` commands import the same package. `build_demo_action_response` stays in the
  wizard flow (it needs `tools/`).

### Setup store → `config/setup_store.py` (was `surfaces/cli/wizard/store.py`)
- `config/remote_store.py` merged in — the read/write split existed only to keep lower tiers from
  importing `surfaces`, which no longer applies.

### Shared terminal pieces → `surfaces/shared/`
- `error_handling/` (`OpenSREError`, `report_exception`), `doctor_checks.py`
  (`DOCTOR_CHECKS`, `run_doctor_checks()` for `opensre doctor` and `/doctor`), `demo_alert.py` +
  `sample_alerts/` (package-data updated); `SAMPLE_ALERT_OPTIONS` next to `ALERT_TEMPLATE_CHOICES`
  in `config/constants/investigation.py`.
- `platform/common/release_version.py`: `fetch_latest_version`, `is_update_available`,
  `is_editable_install`, `development_install_doctor_version_detail` (public; `lifecycle/update.py`
  keeps `run_update`).
[...]

PR #5045, opened 2026-08-18
refactor(surfaces): terminal tier moves to surfaces/shared/terminal, CLI and Interactive shell coupling decreased

YauhenBichel:
Second pass of the CLI/shell decoupling (after #5042). The output tracker, table and prompt
primitives, banner, health, feedback and stream renderer are terminal code both surfaces use;
they lived under the shell's `ui/` and the CLI imported eleven of its modules.

### `surfaces/shared/terminal/` (was `surfaces/interactive_shell/ui/*`)
- `output/` (tracker, live display, boundary), `components/` (rendering, choice_menu,
  key_reader, cpr_stdin, banner_art, time/token format, loaders), `stream_renderer/`,
  `feedback/`, `health/`, `agents/`, `banner/`, `tables/`.
- Imports nothing from either surface (checked); the shell's `ui/__init__` keeps re-exporting
  the same names, so shell call sites are unchanged.

### Pieces that were in the wrong surface
- `ui/layout` renders `opensre --help` and the landing page from a `click.Group` — it is CLI
  code, now `surfaces/cli/layout.py`.
- `repl_render_launch_poster` / `refresh_welcome_poster` need `terminal_ui` and the session —
  REPL-only, split out of `rendering.py` into `surfaces/interactive_shell/ui/poster.py`.

## Why
The two terminal surfaces are peers and must not depend on each other; the terminal machinery
they share needs one owner below both. Mechanical move under the ratchet, no behaviour change.

## Verification
ruff + format clean, `check_imports --strict` 0 cycles, mypy 1818 files, full suite 15,231
passed; CLI smoke (`--help`, landing, `doctor`, `investigate --help`, `health`) identical to main.


YauhenBichel:
@greptile review

PR #5047, opened 2026-08-18
refactor(ports): one execution gate for the four host capabilities

YauhenBichel:
Problem. Four capabilities reach the agent from its host — named commands, LLM-provider switching, task cancellation, investigation launch — and each declared the same seven-parameter execution_allowed itself: one approval contract written out four times, across three tiers (surfaces/ ×3, tools/ ×1). Separately, the harness's factory aliases for those capabilities were Callable[[], Any] declared in core/agent_harness/tools/tool_provider.py and then re-declared verbatim in three gateway modules, so four files each owned a copy of a type nobody checked.

Change.

- ExecutionGate — the host's half of the execution gate, declared once in a new leaf tools/interactive_shell/shared/host_ports.py, beside the ExecutionPolicyResult it takes. SlashPorts, LlmProviderPorts, TaskCancelPorts and InvestigationLaunchPorts now inherit it instead of restating it.
- One home for the four factory aliases — moved into core.agent_harness.ports, the Implement API module, next to the SubprocessPresenterFactory seam that already solves the identical problem. The gateway's three copies are deleted and import the name instead. API pin updated with the four names, so the exact-match ratchet still holds.
- Three ratchet tests (tests/core/agent_harness/test_host_capability_ports.py), written red first: the gate is declared once, every capability protocol inherits rather than copies, and no tier re-declares an alias.

A design correction worth flagging for review. The plan called for these contracts to live in core. Implementing it showed that is wrong: these are host→tool contracts — a surface supplies them, action tools in tools/ call them, and core only carries them across. Typing them in core would mean core importing tools, which the layer rules forbid. So the gate sits with its consumers, and core keeps Any at that seam — now with a comment saying why, rather than by omission.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5069, opened 2026-08-18
refactor(scheduler): scheduled runners become a value the host gates once

YauhenBichel:
Problem:

The scheduler had no component. Its two runner seams were module globals installed as a side effect, which the gateway then rewrote in place to attach its capacity gate:

```
if (agent_runner := get_agent_runner()) is not None:
    register_agent_runner(_with_gate(agent_runner, gate))   # get → wrap → put back
```

That read-modify-write is not idempotent: a second call gates an already-gated runner, so one scheduled run costs two permits — on a small size profile that halves capacity or deadlocks. Nothing calls it twice today, so it was latent rather than live. The same shape forced an ordering dance documented by comment in bootstrap/process.py — the worker profile installs runners at boot, the gateway installs them late inside start_scheduler, "do not confuse the two."

Change:

The runners become a value instead of a side effect.

- platform/scheduler/runners.py (new) — SchedulerRunners, a frozen bundle of the two seams, plus a TurnGate protocol so platform can state the capacity contract without importing gateway. .gated(gate) returns a new bundle; .install() binds both, investigation first.
- bootstrap/adapters.py — scheduler_runners() assembles the bundle. The composition root is the only layer allowed to see both integrations and tools, so it belongs there.
The gateway states its gate once — scheduler_runners().gated(self.turn_gate).install() replaces install-then-rewrite.
gateway/core/runtime/scheduler_concurrency.py deleted. Double-gating no longer has a shape to happen in: you gate a value before installing it, never by reading one back.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5073, opened 2026-08-18
refactor(gateway): give gateway/core packages names that say what is in them

YauhenBichel:
Problem:
 gateway/core/runtime was 1,634 lines across 15 modules doing four unrelated jobs — process supervision, composition, the agent-hosting apparatus, and user-facing chat copy. "Runtime" is the name things get when there isn't a better one, and it was also the only gateway package a surface may import from, which is why the border allowlist read oddly: daemon sat beside turn_handler, which surfaces must never touch. Separately, gateway/core/feedback was a 34-line JSONL store living outside gateway/core/storage, which exists for stores.

Change:
 Three moves, no behaviour change — every file relocation is R095–R100 (near-identical content):

feedback → storage/feedback. It appends a dict to a JSONL file under a lock. That is a store, and stores live in storage.
runtime → host + process + what's left.
host/ — what the gateway supplies so a turn can run: turn_handler, session_agents, live_sink, cancel_console, capability_policy, concurrency, status_messages
process/ — the gateway process itself: daemon, readiness, polling_thread
runtime/ — composition and boot: controller, credential_hydration, errors, security_audit
status_messages moved twice, deliberately. It is user-facing copy plus formatting logic, consumed by all four transports' sinks — it was never runtime. It briefly sat loose at gateway/core/ root, which was wrong because every other entry there is a package, and it now sits in host/ with the sink and handler that use it.
Naming. The package is host, not turn. A turn belongs to the harness — handle(text, TurnBinding), run_turn, TurnResult. The gateway never runs one; it hosts one. "Host" is the codebase's own word for that role: ports is "what a host implements", spi is "what a host calls around a turn", capability_policy.py already opened with "Host capability policy…", and surfaces/cli/host.py is direct precedent. 93 files use it that way.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5074, opened 2026-08-18
refactor(gateway): the gateway states which capabilities it withholds; the harness records them

YauhenBichel:
## What

`gateway/core` wrote empty tuples into `session.available_capabilities` directly — reaching
into a mapping the harness owns and reads when it plans a turn. It did so twice, and one of
the two places was a storage module.

### The mechanism moves to the harness
- `core/agent_harness/session/capabilities.py`: `withhold_capabilities(session, *names)`,
  exported through `spi.session_state` (the role hosts already call for session state) and
  pinned in the API test. Idempotent, so a host can state its policy wherever a session is
  prepared without coordinating.

### The gateway keeps only the decision
- `UNSUPPORTED_GATEWAY_CAPABILITIES = ("investigation", "llm_provider", "task_cancel")` stays:
  which capabilities chat offers is a product call.
- Stated **once**, where the agent is built (`SessionAgentPool.agent_for`), instead of in
  `turn_handler` ("also applied in SessionResolver", "for tests that inject a SessionCore
  directly") and again in `gateway/core/storage/session/resolver.py`.

### The border rule names the package, not the files
- `gateway/tests/test_harness_behaviour_border.py`: `gateway/core/host/**` may drive the
  agent, because that is what the package is for. The shrink-only allowlist now holds only
  the three modules outside it — the session resolver and the two `web/` modules that embed
  `AgentSession`.

## Why
Whoever owns a data structure owns how it is written. The gateway knowing that "no tools"
is spelled `()` is knowledge it should not have, and having no single owner is what let the
call be duplicated — including into a storage module, which has no business applying agent
policy.


YauhenBichel:
@greptile review

PR #5075, opened 2026-08-18
feat(gateway): a channel supplies its own agent ports; the shell becomes one

YauhenBichel:
SessionAgentPool accepted exactly two knobs — console and slash_ports_factory — and hardcoded everything else. Any surface that needed its own tools, prompt grounding or progress rendering had no way to ask for them, which is why the interactive shell built its own agent instead of using the pool.

ChannelAgentPorts is a frozen bundle a channel fills, every field defaulting to the chat behaviour the four transports already use:

```
@dataclass(frozen=True)
class ChannelAgentPorts:
    surface: str = "gateway"
    build_tools:   ToolProviderFactory | None = None   # (session, console, logger, observer)
    build_prompts: PromptsFactory      | None = None
    build_gather:  GatherFactory       | None = None
    error_reporter: Any | None = None
    apply_capability_policy: … | None = None
```

SessionAgentPool(…, ports=…) is optional and additive: with ports=None the build path resolves to exactly what it did before. That was checked rather than assumed — the new code passes error_reporter=None and prompts=None explicitly where they were previously omitted, and both are declared = None in DefaultPorts and .agent(), so the two forms are equivalent.

The shell declares itself a channel. shell_agent.py now supplies ChannelAgentPorts — REPL tools, CLI grounding, console gather, Sentry reporter — instead of describing its ports in prose. It keeps TTY, slash commands and streaming paint, which was the requirement.

Capability policy is now a channel decision, deliberately. #5074 made the gateway withhold investigation, llm_provider and task_cancel from chat. Applying that to the shell would silently strip the REPL's investigation and provider-switch tools, so the shell passes preserve_host_capabilities — a documented no-op.

Two borders added. The scheduler may never import the chat turn handler (your "don't invent a third agent path for loops," now enforced), and the shell reaches the gateway through session_agents, not the turn handler.

One sentence stated once. "OpenSRE is at capacity. Please try again shortly." existed in turn_handler.py and again in webapp.py. It's now AT_CAPACITY_MESSAGE in concurrency.py, beside the gate both callers already share, and web's comment stopped saying "busy-drop like GatewayTurnHandler" and started calling it.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5077, opened 2026-08-18
Fix: Telegram gateway poll loop deadlocks its own approval flow (#5076)

### LINKED ISSUE #5076: [BUG] Telegram gateway poll loop deadlocks its own approval flow — every approval silently denies after 180s
### Summary

The Telegram gateway poll loop awaits turn dispatch inline instead of firing it as a background task, so any tool call that requires approval on Telegram deadlocks its own approval flow and silently times out to "denied" after 180 seconds.

**Expected:** When a Telegram user is asked to approve a write tool via the Approve/Deny inline keyboard, clicking Approve (or Deny) should resolve the pending approval promptly, the same way it does on the Buzz transport.

**Actual:** The approval is never resolved by the click. It always times out after `MAX_APPROVAL_WAIT_SECONDS` (180s) and is treated as denied, regardless of whether or how fast the user clicks the button. While this is happening, polling for *all* Telegram chats is frozen — no other chat's messages are processed either.

**Root cause:** `gateway/transports/telegram/background.py:76-108` runs one coroutine, `_poll_telegram_until_stopped`, that loops on `poller.poll_once()` and then `await`s `handle_polled_inbound_telegram_message(...)` directly (not `asyncio.create_task`'d):

```python
while not stop_event.is_set():
    batch = await asyncio.to_thread(poller.poll_once)
    ...
    for callback in batch.callbacks:
        handle_callback_query(callback, broker=resources.approvals, ...)   # resolves a pending approval

    for event in batch.messages:
        ...
        await handle_polled_inbound_telegram_message(event, ...)   # awaited inline, not create_task'd
```

`handle_polled_inbound_telegram_message` (`gateway/transports/telegram/inbound_handler.py:146-153`) awaits the dispatched turn on the executor:

```python
with terminal.timeout_after(settings.turn_timeout_seconds, _on_turn_timeout):
    try:
        with active_cancels.track(...):
            await event_loop.run_in_executor(executor, _run_turn)
```

If the turn calls a tool with `requires_approval=True`, `_run_turn` reaches `TelegramApprovalPrompter.request` (`gateway/transports/telegram/approvals.py:72-73`):

```python
timeout = min(float(expiry_seconds), MAX_APPROVAL_WAIT_SECONDS)
approved, decided_by = self._broker.wait(approval_id, timeout=timeout)
```

which blocks on a real `threading.Event.wait` in `ApprovalBroker.wait` (`gateway/core/middleware/approvals.py:98-104`):

```python
def wait(self, approval_id: str, *, timeout: float) -> tuple[bool, str]:
    ...
    decided = pending.event.wait(timeout)
```

The only thing that can ever set that event is `handle_callback_query`, called from *inside the poll loop* after a *later* `poller.poll_once()` call sees the button-click callback. But the poll loop's single coroutine is still suspended at the `await handle_polled_inbound_telegram_message(...)` line from the turn that is itself waiting on that same event — so it can never reach the next `poll_once()` call to observe the click. This is a structural deadlock, not a timing race: the approval cannot be resolved no matter how fast the user clicks.

**Sibling transport confirms the failure mode and the fix.** `gateway/transports/buzz/background.py:86-91` documents this exact scenario in its own docstring and avoids it by firing dispatch as a background task instead of awaiting it inline:

```python
"""
Turn dispatch is fired as a background task, never awaited inline: a turn
can sit blocked in ``ApprovalBroker.wait`` for up to
``MAX_APPROVAL_WAIT_SECONDS``, and this same loop is the only thing that
can ever deliver the reply that unblocks it. Awaiting the turn here would
stall polling for the duration of every approval wait, so the reply that
resolves it could never arrive.
"""
```

and at `gateway/transports/buzz/background.py:119`:

```python
task = asyncio.create_task(
    _dispatch_turn(...)
)
```

Telegram's loop does exactly what Buzz's own comment warns against.

**Secondary effect:** `turn_semaphore = asyncio.Semaphore(settings.max_concurrent_turns)` (`background.py:72`) is dead configuration for Telegram, since dispatch is sequential (`await`, not concurrent tasks) — only one turn can ever be in flight regardless of the configured limit.

**Test coverage gap:** No existing test exercises the real poll loop with a pending approval.
- `gateway/tests/telegram/test_approvals.py` tests `TelegramApprovalPrompter.request` / `handle_callback_query` in isolation, resolving the approval via a `threading.Timer` that calls `broker.resolve(...)` directly — bypassing the poll loop entirely.
- `gateway/tests/telegram/test_background.py` contains one test, `test_start_starts_poll_thread`, which only asserts the background thread starts; it never drives `_poll_telegram_until_stopped` through a message batch.

**Suggested fix:** Fire turn dispatch via `asyncio.create_task(...)` in `gateway/transports/telegram/background.py`, matching the pattern already used in `gateway/transports/buzz/background.py`.

**Note:** This is a different bug from #4769 (live bot token logged on transient failures) despite living in the same file — that issue is about token redaction in error paths, this one is a concurrency/deadlock bug in the dispatch model.

### Steps to reproduce

1. Configure a Telegram gateway with at least one tool registered with `requires_approval=True`.
2. Send a message that triggers that tool.
3. Observe the Approve/Deny inline keyboard posted to the chat.
4. Click Approve (or Deny) immediately.
5. Observe that the click has no effect — the approval prompt only resolves ~180 seconds later, to "denied (expired — no click)", regardless of the click. Meanwhile, no other Telegram chat receives replies either.

Also confirmable by code inspection alone, without a live bot:
- `gateway/transports/telegram/background.py:96` awaits `handle_polled_inbound_telegram_message(...)` inline in the same coroutine that drives `poller.poll_once()`.
- `gateway/transports/buzz/background.py:86-91,119` documents and avoids this exact pattern via `asyncio.create_task(...)`.

Yes

### How often does it occur?

Every time

### Operating system

Other

### Logs and error output

No exception or crash — this is a silent behavioral bug. The only visible symptom is the approval prompt's own outcome message updating to "denied (expired — no click)" after the ~180s wait, and other chats going quiet for the same duration.

### Additional context

Working on this now — will push a PR with the fix plus a regression test soon.

1. `uv run opensre gateway start -f` with Telegram in poll mode.
2. Message the bot with something that needs an approval-gated write tool.
3. Click Approve as soon as the keyboard appears.
4. From a second Telegram chat, send any message while step 3 is pending.

**Pass:** the approval resolves promptly and the second chat is answered
normally. **Fail (today):** the click does nothing, the turn auto-denies after
the approval timeout, and polling for every chat is frozen meanwhile.

Contrast: run the same scenario on the Buzz transport, which dispatches turns as
background tasks and resolves the click immediately.

Note #3284 describes this same deadlock at an earlier commit — worth closing
together.

Bhumika-1432006:
Fixes #5076

`gateway/transports/telegram/background.py`'s poll loop (`_poll_telegram_until_stopped`)
awaited turn dispatch inline — `await handle_polled_inbound_telegram_message(...)` — inside
the same coroutine that drives `poller.poll_once()`. When a dispatched turn calls a tool with
`requires_approval=True`, it blocks on `ApprovalBroker.wait` (up to 180s), and the only thing
that can ever deliver the Approve/Deny `callback_query` click that resolves that wait is the
*next* `poller.poll_once()` call — but the loop's single coroutine can't reach that next call
while it's still suspended awaiting the very turn that's blocked on the click. This is a
structural deadlock, not a timing race: every Telegram approval silently times out to "denied"
after 180s regardless of how fast the user clicks, and the whole poller (all chats) is frozen
for that same duration since nothing else drives `poll_once()` either.

The sibling Buzz transport (`gateway/transports/buzz/background.py:86-91,119`) documents this
exact failure mode in its own docstring and avoids it by firing turn dispatch via
`asyncio.create_task(...)` instead of awaiting it inline. This PR applies the same pattern to
Telegram:

- Turn dispatch is now fired via `asyncio.create_task(_dispatch_turn(...))` instead of being
  awaited inline, so the poll loop can immediately go back to `poller.poll_once()` and observe
  a `callback_query` while an earlier turn is still in flight.
- A `pending_tasks: set[asyncio.Task[None]]` holds a strong reference to each dispatched task
  (with `task.add_done_callback(pending_tasks.discard)` to release it on completion) — without
  this, `asyncio.create_task`'s result would have no owner and could be garbage-collected
  mid-flight, silently dropping the turn.
- A new `_dispatch_turn` wrapper catches and logs any exception from the turn (matching Buzz's
  `_dispatch_turn`), so a failure in one turn can't produce an "exception was never retrieved"
  warning or otherwise affect the poll loop.
- A new `_drain_pending_turns` step runs once the loop exits (`stop_event` set): it awaits any
  still-running turn tasks with a bounded `_SHUTDOWN_DRAIN_SECONDS` (5s) timeout, then cancels
  stragglers. Without this, `asyncio.run` closing the event loop out from under an in-flight
  task on shutdown would abandon it uncleanly (Buzz has the equivalent
  `_drain_active_turns` for the same reason — polling stopping means nothing can ever deliver
  the approval click that a still-waiting turn needs).
- No change was needed in `inbound_handler.py`: `handle_polled_inbound_telegram_message`
  already does `async with user_lock, turn_semaphore:` — identical to Buzz's
  `handle_polled_inbound_buzz_message`. `turn_semaphore` was already correctly acquired; it was
  simply never exercised concurrently because dispatch was serial. Making dispatch concurrent
  is what makes the existing semaphore bound actually take effect.

This is an async concurrency fix with no UI surface, so the regression test is the
demonstration. I confirmed the new test fails (with a `TimeoutError` at the exact point that
proves the deadlock) against the pre-fix code, and passes against the fix:

```
# pre-fix (background.py reverted, test kept):
gateway/tests/telegram/test_background.py::test_poll_loop_delivers_callback_while_a_turn_is_still_dispatching FAILED
  TimeoutError at: await asyncio.wait_for(callback_delivered.wait(), timeout=1)

# post-fix:
gateway/tests/telegram/test_background.py::test_poll_loop_delivers_callback_while_a_turn_is_still_dispatching PASSED
```

---

The bug is a classic "poll loop awaits its own unblocking condition" deadlock. I traced the
full call chain by reading four files together: `background.py` (the poll loop itself),
`inbound_handler.py` (where the inline `await` reaches the executor), `approvals.py` (where
`TelegramApprovalPrompter.request` calls `self._broker.wait(...)`), and
`gateway/core/middleware/approvals.py` (`ApprovalBroker.wait`, which is a real blocking
`threading.Event.wait`, run on the turn's executor thread). The critical piece was confirming
that the *only* code path that can ever call `broker.resolve(...)` for Telegram is
`handle_callback_query`, invoked from inside the very poll loop that's stuck — which is what
makes this a hard deadlock rather than just added latency.

I considered two ways to fix it:
1. **Fire dispatch via `asyncio.create_task`** (what this PR does) — matches the pattern
   already proven correct in the working Buzz transport, requires no change to shared code
   (`inbound_handler.py`, `ApprovalBroker`), and keeps the fix scoped to the one file that owns
   the bug.
2. **Restructure the approval wait to be non-blocking** (e.g. have `ApprovalBroker.wait` poll
   an `asyncio.Event` instead of a `threading.Event`, and have the poll loop `await` it
   directly) — this would work too, but it means touching `ApprovalBroker`, which is shared by
   three transports (Telegram, Discord, Buzz — via `gateway.core.middleware.approvals`), so a
   mistake there has a much larger blast radius, and it doesn't match the pattern this
   codebase already uses successfully elsewhere.

I chose (1): it's the minimal change that fixes the actual bug, it's provably correct because
the identical pattern already works for Buzz, and it doesn't touch any shared/cross-transport
code.

Key components:
- `_dispatch_turn` — new wrapper around `handle_polled_inbound_telegram_message` that adds a
  catch-and-log boundary so an exception from one turn can't propagate as an unretrieved-task
  exception; this is what gets wrapped in `asyncio.create_task`.
- `pending_tasks` set + `add_done_callback(pending_tasks.discard)` — the standard idiom for
  preventing a fire-and-forget `asyncio.Task` from being garbage-collected before it completes
  (per the `asyncio.create_task` docs' own warning about this).
- `_drain_pending_turns` — shutdown-time cleanup: bounded wait for outstanding turns, then
  cancel; logs how many were still running so an operator can see it happened.
- `_SHUTDOWN_DRAIN_SECONDS` — a local module constant (not a new `GatewaySettings` field like
  Buzz's `shutdown_drain_seconds`) to keep this fix scoped to the one file; Buzz's field-based
  version exists because Buzz also needs to resolve pending approvals as denied on drain (its
  own `PendingApprovals` claim registry), which Telegram's simpler inline-keyboard approval
  model doesn't have an equivalent structure for, and adding one was out of scope for this fix.

Edge case I specifically tested: the regression test proves the *mechanism* directly — a
message batch that starts a deliberately-hanging turn, then a second batch carrying the
approval callback, asserting the callback is observed and handled while the first turn is
still in flight. I additionally verified (by temporarily reverting just the fix file, `git
stash`, and re-running the same test) that this exact test fails with a `TimeoutError` against
the old code — proving the test reproduces the real bug rather than just testing the new code
in isolation.

---

---

**Testing:**

New regression test added to `gateway/tests/telegram/test_background.py`:
`test_poll_loop_delivers_callback_while_a_turn_is_still_dispatching`. It drives
`_poll_telegram_until_stopped` directly (matching this file's existing
`test_start_starts_poll_thread` style plus the dispatch-testing pattern used in
`gateway/tests/buzz/test_background.py`) through a fake poller returning a message batch (which
triggers a deliberately-hanging fake turn) followed by a callback batch, and asserts the
callback is delivered while the first turn is still pending.

Ran per this repo's `gateway/` → `gateway/tests/` scope mapping in
`.github/ci/test_scope_rules.py`:

- `uv run python -m pytest gateway/tests/telegram/ -q` → **2 passed**.
- `uv run python -m pytest gateway/tests/ -q` → **722 passed, 9 failed, 2 errors**. I checked
  every one of the 9 failures/2 errors individually: none reference
  `gateway/transports/telegram/background.py` or anything this PR touches. I confirmed 5 of
  them reproduce identically against unmodified `main` (via `git stash` on just my two changed
  files and re-running): `test_gateway_daemon.py::test_start_stop_round_trip` and
  `test_credential_hydration.py::test_hydrates_exact_secret_and_atomically_writes_private_v2_store`
  fail from Windows platform gaps (`os.WNOHANG` doesn't exist on Windows; POSIX file-mode bits
  like `0o600` don't map the same way on NTFS), and
  `test_harness_behaviour_border.py::test_only_the_host_layer_drives_the_agent` /
  `test_package_borders.py::test_only_the_gateway_facade_imports_the_transport_composer` are
  pre-existing import-boundary/architecture-debt assertions spanning many files I never
  touched. The remaining failures in the same run (`test_security_audit.py`,
  `test_gateway_daemon.py::test_stop_escalates_to_sigkill_when_sigterm_is_ignored`,
  `test_gateway_daemon.py::test_component_status_round_trip`,
  `test_credential_hydration.py::test_hydrates_from_integrations_secret_when_no_credentials_api_url`)
  are the same two root causes (Windows POSIX-permission/`os.WNOHANG` gaps) on different test
  functions in the same two files.
- `uv run ruff check` and `uv run ruff format --check` on both touched files → clean.
- `uv run mypy` over the full `make typecheck` scope (`bootstrap config core gateway
  integrations platform surfaces tools`) → 68 pre-existing errors from missing third-party type
  stubs in this dev environment (`psycopg2`, `slack_sdk`, `discord`, `fastapi`, `sentry_sdk`,
  etc.); none reference `background.py` or the changed test file.
- Confirmed the regression test actually reproduces the bug: reverted just
  `gateway/transports/telegram/background.py` (`git stash` on that one file, keeping the new
  test), reran the new test, and it failed with `TimeoutError` at
  `await asyncio.wait_for(callback_delivered.wait(), timeout=1)` — exactly where the deadlock
  would manifest. Restored the fix and reran; it passed.

### DIFF
--- a/gateway/core/middleware/approvals.py
+++ b/gateway/core/middleware/approvals.py
@@ -50,6 +50,7 @@ class ApprovalBroker:
     def __init__(self) -> None:
         self._pending: dict[str, _PendingApproval] = {}
         self._lock = threading.Lock()
+        self._closed = False
 
     def create(
         self,
@@ -59,10 +60,18 @@ class ApprovalBroker:
     ) -> str:
         approval_id = uuid.uuid4().hex
         with self._lock:
-            self._pending[approval_id] = _PendingApproval(
+            pending = _PendingApproval(
                 platform=platform or "",
                 chat_id=chat_id or "",
             )
+            if self._closed:
+                # A turn that reaches its first approval request after
+                # drain() already swept the broker (still in flight, but too
+                # late for that one-time sweep) must not open a fresh wait
+                # nothing can ever resolve — born already denied instead.
+                pending.approved = False
+                pending.event.set()
+            self._pending[approval_id] = pending
         audit_security_action(
             action="approval.create",
             platform=platform,
@@ -102,9 +111,14 @@ class ApprovalBroker:
         deliver a click once its poller stops, so a turn still parked in
         :meth:`wait` would otherwise burn its full expiry while shutdown
         waits on it. Setting each pending Event here lets it return
-        immediately instead. Returns the approval ids that were released.
+        immediately instead. Also closes admission: a turn still in flight
+        that has not yet called :meth:`create` would otherwise open a fresh
+        wait after this one-time sweep that nothing could ever resolve — once
+        closed, :meth:`create` returns an already-denied id instead. Returns
+        the approval ids that were released.
         """
         with self._lock:
+            self._closed = True
             approval_ids = list(self._pending.keys())
             for pending in self._pending.values():
                 pending.approved = False
--- a/gateway/core/middleware/approvals.py
+++ b/gateway/core/middleware/approvals.py
@@ -95,6 +95,24 @@ class ApprovalBroker:
         )
         return True
 
+    def drain(self) -> list[str]:
+        """Resolve every still-pending approval as denied; for shutdown drain.
+
+        A transport that dispatches turns as detached tasks can no longer
+        deliver a click once its poller stops, so a turn still parked in
+        :meth:`wait` would otherwise burn its full expiry while shutdown
+        waits on it. Setting each pending Event here lets it return
+        immediately instead. Returns the approval ids that were released.
+        """
+        with self._lock:
+            approval_ids = list(self._pending.keys())
+            for pending in self._pending.values():
+                pending.approved = False
+                pending.decided_by = ""
+                pending.event.set()
+            self._pending.clear()
+        return approval_ids
+
     def wait(self, approval_id: str, *, timeout: float) -> tuple[bool, str]:
         """Block for a decision; expiry counts as deny. Returns (approved, decided_by)."""
         with self._lock:
--- a/gateway/tests/telegram/test_background.py
+++ b/gateway/tests/telegram/test_background.py
@@ -127,3 +127,87 @@ def test_poll_loop_delivers_callback_while_a_turn_is_still_dispatching() -> None
     asyncio.run(_run())
 
     assert handled_callbacks == [callback]
+
+
+def test_stop_in_same_batch_cancels_the_just_dispatched_turn() -> None:
+    """Regression: ``/stop`` in the same poll batch as its target message must
+    find the just-dispatched turn instead of reporting "no active turn".
+
+    Dispatch-time registration in ``ActiveTurnRegistry`` happens in the poll
+    loop itself, before ``create_task`` — so a same-batch ``/stop`` can find
+    and cancel a turn whose task has not even started running yet.
+    """
+    resources = _resources()
+    stop_event = threading.Event()
+    dispatch_started = asyncio.Event()
+    seen_cancel_events: list[threading.Event] = []
+
+    async def _hanging_dispatch(
+        _event: object, *, turn_cancel: threading.Event, **_kw: object
+    ) -> None:
+        seen_cancel_events.append(turn_cancel)
+        dispatch_started.set()
+        await asyncio.Event().wait()
+
+    message = TelegramInboundMessage(
+        update_id=1, user_id="u1", chat_id="c1", message_id="m1", text="do the thing"
+    )
+    stop_message = TelegramInboundMessage(
+        update_id=2, user_id="u1", chat_id="c1", message_id="m2", text="/stop"
+    )
+    call_count = 0
+
+    def _poll_once() -> TelegramPollResult:
+        nonlocal call_count
+        call_count += 1
+        if call_count == 1:
+            return TelegramPollResult(messages=[message, stop_message])
+        stop_event.set()
+        return TelegramPollResult()
+
+    poller = MagicMock()
+    poller.poll_once.side_effect = _poll_once
+
+    async def _run() -> None:
+        with pytest.MonkeyPatch.context() as mp:
+            mp.setattr(background, "TelegramPoller", lambda _token: poller)
+            mp.setattr(background, "handle_polled_inbound_telegram_message", _hanging_dispatch)
+
+            task = asyncio.create_task(
+                background._poll_telegram_until_stopped(
+                    settings=GatewaySettings(bot_token="tok"),
+                    stop_event=stop_event,
+                    logger=logging.getLogger("gateway.test"),
+                    resources=resources,
+                    handle_callback_to_gateway_agent=lambda *_a: None,
+                )
+            )
+            await asyncio.wait_for(dispatch_started.wait(), timeout=1)
+            task.cancel()
+            await asyncio.gather(task, return_exceptions=True)
+
+    asyncio.run(_run())
+
+    resources.client.send_message.assert_not_called()  # no "no active turn" reply
+    assert len(seen_cancel_events) == 1
+    assert seen_cancel_events[0].is_set()
+
+
+def test_shutdown_denies_pending_approvals_before_draining() -> None:
+    """A turn parked in ``ApprovalBroker.wait`` must not outlive the drain budget."""
+    resources = _resources()
+    approval_id = resources.approvals.create(platform="telegram", chat_id="c1")
+    waited: list[tuple[bool, str]] = []
+
+    async def _run() -> None:
+        waiter = asyncio.get_running_loop().run_in_executor(
+            None, lambda: resources.approvals.wait(approval_id, timeout=30)
+        )
+        await background._drain_pending_turns(
+            set(), approvals=resources.approvals, logger=logging.getLogger("gateway.test")
+        )
+        waited.append(await asyncio.wait_for(waiter, timeout=1))
+
+    asyncio.run(_run())
+
+    assert waited == [(False, "")]
--- a/gateway/tests/telegram/test_background.py
+++ b/gateway/tests/telegram/test_background.py
@@ -1,20 +1,31 @@
 from __future__ import annotations
 
+import asyncio
 import logging
+import threading
 from unittest.mock import MagicMock, patch
 
+import pytest
+
+from gateway.core.middleware.active_turns import ActiveTurnRegistry
+from gateway.core.middleware.approvals import ApprovalBroker
+from gateway.transports.telegram import background
 from gateway.transports.telegram.background import start_telegram_gateway_background
+from gateway.transports.telegram.poller.poller import TelegramPollResult
 from gateway.transports.telegram.runtime import (
+    TelegramPollingRuntime,
     initialize_telegram_polling_runtime,
     shutdown_telegram_polling_runtime,
 )
-from gateway.transports.telegram.settings import GatewaySettings
+from gateway.transports.telegram.settings import (
+    GatewaySettings,
+    TelegramCallbackQuery,
+    TelegramInboundMessage,
+)
 
 
 @patch("gateway.transports.telegram.background.TelegramPoller")
 def test_start_starts_poll_thread(mock_poller_cls: MagicMock) -> None:
-    from gateway.transports.telegram.poller.poller import TelegramPollResult
-
     mock_poller_cls.return_value.poll_once.return_value = TelegramPollResult()
     logger = logging.getLogger("gateway.test")
     handle = start_telegram_gateway_background(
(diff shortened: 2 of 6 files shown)


Bhumika-1432006:
@greptile review

Bhumika-1432006:
Addressed both P1 findings from the review:

1. **Detached turns miss stop commands** — `gateway/transports/telegram/background.py` now registers a `turn_cancel` Event in `ActiveTurnRegistry` before `asyncio.create_task`, matching Buzz's dispatch-time registration exactly (per `ActiveTurnRegistry`'s own docstring on this requirement). `gateway/transports/telegram/inbound_handler.py` now accepts an optional pre-registered `turn_cancel`, binding the stop finalizer via `active_cancels.bind_user_stop(...)` instead of re-registering via `active_cancels.track(...)` when one is supplied — again mirroring `handle_polled_inbound_buzz_message` line for line. Added a regression test (`test_stop_in_same_batch_cancels_the_just_dispatched_turn`) proving a `/stop` in the same batch as its target message now finds and cancels the just-dispatched turn.

2. **Approval waits outlive shutdown drain** — added `ApprovalBroker.drain()` (shared `gateway/core/middleware/approvals.py`, used by all transports) that resolves every still-pending approval as denied, releasing any thread blocked in `wait()` immediately. Telegram's `_drain_pending_turns` now calls it before the bounded task wait, so the drain bound is no longer undermined by `executor.shutdown(wait=True)` blocking on a still-waiting approval thread afterward. Added a regression test (`test_shutdown_denies_pending_approvals_before_draining`) matching Buzz's own equivalent test.

This required touching `gateway/transports/telegram/inbound_handler.py` and `gateway/core/middleware/approvals.py` in addition to `background.py`/its test — both P1s were only fixable by extending the fix into the layer where `ActiveTurnRegistry`/`ApprovalBroker` are actually driven, matching the working Buzz transport's design.

On the P2 (**Pending task admission is unbounded**): left as-is. Buzz's `_poll_buzz_until_stopped` has the identical characteristic — `turn_semaphore` is acquired inside the dispatched task, not at task-creation time, so task creation itself is never bounded there either. This is deliberate in both transports: gating task creation on the semaphore in the poll loop would reintroduce the exact deadlock this PR fixes (a full semaphore blocking `await turn_semaphore.acquire()` in the loop would once again stall `poll_once()`, including for the callback that would free a slot). Since it mirrors already-shipped, accepted Buzz behavior and isn't a regression this PR introduces, I'm not changing it here — happy to split it into a follow-up if you'd like bounded admission tracked separately.

@greptile review

Bhumika-1432006:
Addressed the new P1 from this round:

**Late approvals escape shutdown drain** — `ApprovalBroker.drain()` only swept approvals that already existed at the moment it ran; a turn still in flight that hadn't reached its first `approvals.create()` call yet could open a brand-new pending approval *after* the sweep, which the one-time drain would never see. That approval would then genuinely block `wait()` for up to its own timeout, and since cancelling the wrapping asyncio Task doesn't stop the underlying executor thread, `executor.shutdown(wait=True)` would inherit that same block.

Fixed by adding a `_closed` flag to `ApprovalBroker`, set by `drain()`. Once closed, `create()` still returns an id (so callers don't need special-casing) but the entry is born already resolved as denied — so `wait()` on it returns immediately regardless of when the turn reaches its approval request relative to drain. This only activates for a broker that has actually had `drain()` called on it, which today is Telegram-only, so Slack/Discord/Buzz behavior is unchanged — confirmed by running `gateway/tests/buzz/`, `gateway/tests/slack/`, and `gateway/tests/discord/` (294 passed).

Added a regression test, `test_approval_created_after_drain_resolves_immediately`, and confirmed it fails with a `TimeoutError` against the pre-fix code (via `git stash` on just `approvals.py`) before restoring the fix.

@greptile review

Bhumika-1432006:
Addressed the new P1 from this round:

**Executor work outlives shutdown drain** — the root cause was that `shutdown_telegram_polling_runtime` (`gateway/transports/telegram/runtime.py`) called `runtime.executor.shutdown(wait=True, cancel_futures=False)`. Cancelling the asyncio Task wrapping a dispatched turn (in `_drain_pending_turns`) never stops the underlying `run_in_executor` thread — my earlier approval-drain fixes only address turns blocked specifically in `ApprovalBroker.wait`; a turn stuck for any other reason (a slow tool call, a long investigation) past the 5s drain bound would still be running when this `wait=True` shutdown call is reached, blocking it for as long as that thread takes to finish, and pushing the Telegram background thread's join past the gateway's ~8s `stop()` budget.

Checked the other transports for the established pattern: Buzz uses `executor.shutdown(wait=False, cancel_futures=True)` and Discord uses `wait=False, cancel_futures=False` — Telegram was the only transport still using `wait=True`. Buzz's own docstring explains exactly this rationale (`gateway/transports/buzz/runtime.py:72-86`). Changed Telegram to match: `wait=False, cancel_futures=True`.

Added `gateway/tests/telegram/test_runtime.py::test_executor_shutdown_does_not_block_on_in_flight_work`, mirroring Buzz's own `gateway/tests/buzz/test_runtime.py::test_executor_shutdown_does_not_block_on_in_flight_work` test exactly: submits a slow callable to the executor, then asserts `shutdown_telegram_polling_runtime` returns in well under 1 second regardless. Confirmed it fails against the pre-fix code — it blocked for the full 30s the fake slow task ran (via `git stash` on just `runtime.py`) — before restoring the fix.

Ran `gateway/tests/telegram/` (77 passed) and the full `make typecheck` scope (68 pre-existing errors, none new).

@greptile review

muddlebee:
@Bhumika-1432006 thx for the PR, but lets discuss the issue before sending PRs

Bhumika-1432006:
Totally fair, happy to discuss on the issue first! For context in case it helps: what started as fixing the poll-loop deadlock (inline await blocking the loop from ever delivering the approval callback) ended up surfacing a few related issues once I looked closer a same-batch /stop miss, and a couple of shutdown-drain gaps where cancelling the asyncio task doesn't stop the underlying executor thread. I matched Telegram's behavior to the existing Buzz transport pattern for each one, since Buzz already handles these correctly.

I'll hold off on further PR changes until we've talked through the approach on the issue. Let me know what works!

PR #5078, opened 2026-08-18
refactor(platform): two capacity policies, stated once — drop the request or wait for a slot

YauhenBichel:
## What

Every turn in a gateway process takes one permit from the same `process_turn_gate()`, and four
call sites paired `acquire`/`release` by hand. #5075 shared the *sentence* (`AT_CAPACITY_MESSAGE`,
"one sentence, one place"); this shares the *policy*.

There were only ever two answers to "the gate is full", and now each has a name:

- `turn_slot(gate)` — **drop.** For a caller holding a conversation or an HTTP connection, which
  can be told to try again. Yields `False` instead of raising, so the caller still decides what a
  refusal looks like: chat finalizes `AT_CAPACITY_MESSAGE`, `/investigate` returns it as a 503 body.
- `queued_turn_slot(gate)` — **wait.** For work already claimed from a queue, which cannot be
  dropped without losing it.

Both live in `platform/turn_capacity/`, with the `TurnGate` protocol beside them — the scheduler
needs them and `platform` cannot import `gateway`.

### Call sites
- `gateway/core/host/turn_handler.py` — chat: `try_acquire` / `finally: release` → `turn_slot`.
- `gateway/web/webapp.py` — `/investigate`: same policy, previously hand-written next to a comment
  saying "busy-drop like GatewayTurnHandler". The run body moves to `_run_investigation` so the
  endpoint reads as capacity-then-work.
- `gateway/web/worker.py` — blocking `acquire` / `finally: release` → `queued_turn_slot`, with the
  body in `_investigate`.
- `platform/scheduler/runners.py` — its private `_permit` was the same context manager, generic
  but blocking-only; deleted in favour of the shared one.

## Why
A leaked permit is a process that answers "at capacity" forever, and the only thing preventing one
was four correct `finally` blocks written independently. Sharing an object while re-deriving its
protocol in a comment is the same defect the capability-policy pass fixed one layer down: whoever
owns the resource owns how it is taken.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5082, opened 2026-08-19
feat(filestorage): Client-side encryption of synced content

### LINKED ISSUE #4562: [Advanced] Client-side encryption of synced content

dhakad22klx:
Fixes #4562 

Added **client-side encryption** to remote sync. Contents are sealed on the machine before upload, so the party operating the store cannot read incident history. Off by default; `remote-sync setup` asks.

Server-side encryption (already on for S3) defends the provider's disks. This defends against the provider.

**Flow Chart**

```mermaid
flowchart TD
    CLI["CLI / shell<br/>setup · sync · reencrypt · rotate-passphrase"] --> OPS[operations.py]
    OPS --> GATE{{"resolve_cipher<br/>fail-closed gate"}}
    GATE -->|mismatch| STOP["refuse — nothing moves"]
    GATE -->|ok| ENGINE["engine.py<br/>push / pull"]
    ENGINE -->|seal / unseal| CIPHER[ManifestCipher]
    CIPHER --> ENV["envelope<br/>AES-256-GCM"]
    ENGINE --> STORE[("ObjectStore<br/>aws · gcs · azure · vercel · s3compat")]
    GATE -.reads/writes.-> MAN[[".opensre-sync-manifest.json<br/>salt + wrapped keys"]]
    MAN --- STORE

    style STOP fill:#f8e8e5,stroke:#9b3226
    style GATE fill:#f7eedc,stroke:#8a5a0e
```

**Key points**
- **Deterministic seal.** Nonce is `HMAC(nonce_key, object_key ‖ plaintext)`, so ETag change-detection still works. A random nonce would re-upload every file on every sync.
- **Two-level keys.** Content is sealed under a random root key; the root key is wrapped by a passphrase-derived KEK,`rotate-passphrase` re-wraps ~100 bytes (instant), `reencrypt` mints a new content key (full re-upload, that's what the feature is).
- **Fails closed, both ways.** Missing/wrong passphrase, **plaintext-store-with-encryption-on**, and **encrypted-store-with-encryption-off** are all refused before anything moves. An object that will not decrypt aborts the run without touching the local file. 
- **Passphrase** goes through the existing secret store (env → `~/.opensre/credentials.json`).

**Not hidden:** object keys, timestamps.

**New:** `platform/filestorage/encryption/` (8 modules) · 19 tests · docs in `configuration/remote-sync.mdx`

[**Screen Recorded Demo**](https://drive.google.com/file/d/1h7u3ce-DAA1WRCvXEuX1MTyflJZ-VRZY/view?usp=sharing)
 (Watch on 1080P)
```
(opensre) deepak@deepakdhakad:~/opensre$ make lint && make format-check
.venv/bin/python -m ruff check bootstrap config core gateway integrations platform surfaces tools tests/
All checks passed!
.venv/bin/python -m ruff format --check bootstrap config core gateway integrations platform surfaces tools tests/
3540 files already formatted
(opensre) deepak@deepakdhakad:~/opensre$ uv run mypy platform/filestorage/ && uv run pytest tests/cli/
..
Success: no issues found in 28 source files
================================== test session starts ===================================
.
.
================= 842 passed, 5 skipped, 3 warnings in 74.94s (0:01:14) =========================
Warning not related to this diff.
```

- [✓] Yes, I used AI assistance (**not totally**)

- [✓] I have reviewed every single line of the AI-generated code
- [✓] I can explain the purpose and logic of each function/component I added
- [✓] I have tested edge cases and understand how the code handles them
- [✓] I have modified the AI output to follow this project's coding standards and conventions

The store held sessions JSONL and memory (`~/.opensre/sessions/` & `~/.opensre/memory/` content is clearly readable, so anyone with bucket access could read incident history. This PR encrypts content client-side before upload, under a key derived from a user-held `passphrase`, using the **Envelope Encryption Pattern** — a random root key does the actual encryption, wrapped by a passphrase-derived key. This makes passphrase rotation cheap (no re-upload) and secure syncing possible across multiple machines.

---

- [✓] I have added proper PR title and linked to the issue
- [✓] I have performed a self-review of my code
- [✓] **I can explain the purpose of every function, class, and logic block I added**
- [✓] I understand why my changes work and have tested them thoroughly
- [✓] I have considered potential edge cases and how my code handles them
- [✓] If it is a core feature, I have added thorough tests
- [✓] My code follows the project's style guidelines and conventions

---


platform/filestorage/encryption/keys.py
```diff
@@ -0,0 +1,234 @@
...
+
+def derive_kek(passphrase: str, salt: bytes, params: ScryptParams) -> bytes:
+    """Key-encryption key for ``passphrase``, using the cache when it applies.
+
+    The cache is keyed by salt and cost parameters, so a rotated store or a
+    changed cost silently misses rather than returning a stale key.
+    """
+    cached = _cached_kek(salt, params)
+    if cached is not None:
+        return cached
```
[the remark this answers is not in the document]
dhakad22klx: User won't directly change the `passphrase` going into file. `opensre remote-sync rotate-passphrase` and it invalidate cache using function `forget_cached_kek()` (file: key.py file).

infrastructure/filestorage/encryption/rotation.py
```diff
@@ -0,0 +1,120 @@
...
+    """Seal every mirrored object under a fresh content key.
+
+    Objects already sealed under the new key are left alone, so an interrupted
+    run is simply re-run. Objects under the previous key stay readable
+    throughout — the manifest keeps carrying it, and the new manifest is written
+    only at the end — which is what makes an interruption safe rather than
+    merely survivable.
+    """
+    listing = store.list_objects("")
+    targets = [obj for obj in listing if obj.key != MANIFEST_KEY]
```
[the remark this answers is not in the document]
dhakad22klx: This won't be the case.if remot storage contains readable data (sessions and memory), sycing with encryption on won't work (that make sense). Either turn of encryption, or encrypt already existing data using `opensre remote-sync reencrypt`

infrastructure/filestorage/encryption/resolver.py
```diff
@@ -0,0 +1,93 @@
...
+        raise PlaintextStoreError(
+            "This store already holds unencrypted sessions or memory.\n"
+            "Encrypting only new writes would leave those readable.\n"
+            "\n"
+            "  Seal what is already there:  `opensre remote-sync reencrypt`"
+        )
+
+    manifest, cipher = new_manifest(passphrase)
+    if not dry_run:
+        save_manifest(store, manifest)
```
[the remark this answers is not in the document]
dhakad22klx: This is very unlikely to happend. When user switches to other machine, they will setup using same `passphrase` to setup. When they run sync on another machine, either it will be fresh sync(creating new `.opensre-sync-manifest.json` or read from existing one.

infrastructure/filestorage/encryption/rotation.py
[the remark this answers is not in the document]
dhakad22klx: configured prefix won't content any other objects.

dhakad22klx:
@greptile reveiw again based on resolved conversation.

config/constants/filestorage.py
```diff
@@ -39,6 +39,20 @@
...
+REMOTE_SYNC_ENCRYPT_ENV = "OPENSRE_REMOTE_SYNC_ENCRYPT"
+# Passphrase the store's key is derived from. Resolved through the usual secret
+# tiers (this variable, then the owner-only local file), so it need not stay
+# exported once setup has stored it.
+REMOTE_SYNC_PASSPHRASE_ENV = "OPENSRE_REMOTE_SYNC_PASSPHRASE"
+# Cached key-derivation result — written and read by opensre, not set by a user.
+# Scrypt costs ~0.4s and every CLI command is a fresh process, so without this
+# each status and sync would pay it again. Bound to the store's salt and KDF
+# parameters, and ignored when either changes.
+REMOTE_SYNC_KEY_CACHE_ENV = "OPENSRE_REMOTE_SYNC_KEY_CACHE"
```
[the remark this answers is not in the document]
dhakad22klx: This is already existing project structure. It is fine.

infrastructure/filestorage/encryption/resolver.py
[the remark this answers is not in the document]
dhakad22klx: Concurrency will be addressed in seperate PR. This PR is implementing the feature only.

infrastructure/filestorage/encryption/rotation.py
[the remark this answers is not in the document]
dhakad22klx: Okay, pushing commit to after adding `The store prefix must be dedicated to OpenSRE; reencrypt processes every object it finds there.` in doc.

platform/filestorage/encryption/keys.py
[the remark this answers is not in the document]
dhakad22klx: passphrase is never changed outside the flow.

infrastructure/filestorage/encryption/rotation.py
```diff
@@ -0,0 +1,120 @@
...
+        payload = store.get_object(obj.key)
+        if envelope.is_sealed(payload):
+            if envelope.parse_header(payload).key_id == cipher.active_key_id:
+                already_current += 1
+                continue
+            plaintext = cipher.unseal(obj.key, payload)
+        else:
+            plaintext = payload
+            adopted += 1
+        store.put_object(obj.key, cipher.seal(obj.key, plaintext))
```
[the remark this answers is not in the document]
dhakad22klx: okay, pushing commit with update manifest will be saved first will store new keys as well. So, everything will work seamlessly.

dhakad22klx:
@greptile review again based on resolved conversation and new commits.

Davidson3556:
The KEK cache is not bound to the supplied passphrase. Since its fingerprint is only `salt:n:r:p`, a warm cache returns the previous KEK before the provided passphrase is considered; a wrong environment passphrase can therefore open the store and be accepted for rotation.

Please bind the cache entry to the passphrase as well, and add a regression test with a warm cache plus a wrong `OPENSRE_REMOTE_SYNC_PASSPHRASE`. Retrying only after an unwrap failure is not enough: the cached KEK correctly unwraps the manifest even when the supplied passphrase is wrong.

dhakad22klx:
> The KEK cache is not bound to the supplied passphrase. Since its fingerprint is only `salt:n:r:p`, a warm cache returns the previous KEK before the provided passphrase is considered; a wrong environment passphrase can therefore open the store and be accepted for rotation.
> 
> Please bind the cache entry to the passphrase as well, and add a regression test with a warm cache plus a wrong `OPENSRE_REMOTE_SYNC_PASSPHRASE`. Retrying only after an unwrap failure is not enough: the cached KEK correctly unwraps the manifest even when the supplied passphrase is wrong.

Okay, sure!!.Thanks for pointing out. Intentional update to passpharase is not expected from user, but accidental update will work silently. Pushing changes shortly.

dhakad22klx:
> > The KEK cache is not bound to the supplied passphrase. Since its fingerprint is only `salt:n:r:p`, a warm cache returns the previous KEK before the provided passphrase is considered; a wrong environment passphrase can therefore open the store and be accepted for rotation.
> > Please bind the cache entry to the passphrase as well, and add a regression test with a warm cache plus a wrong `OPENSRE_REMOTE_SYNC_PASSPHRASE`. Retrying only after an unwrap failure is not enough: the cached KEK correctly unwraps the manifest even when the supplied passphrase is wrong.
> 
> Okay, sure!!.Thanks for pointing out. Intentional update to passpharase is not expected from user, but accidental update will work silently. Pushing changes shortly.

Hi @Davidson3556  pushed requested changes. If wrong passpharase is detected, it will fail with `WrongPassphraseError`.
```
(opensre) deepak@deepakdhakad:~/opensre$ opensre remote-sync sync
Sync failed: That passphrase does not open this store's key.
Check the passphrase, or point at the remote store's prefix it belongs to.
```

infrastructure/filestorage/encryption/rotation.py
```diff
@@ -0,0 +1,124 @@
...
+        manifest, cipher = new_manifest(passphrase)
+    else:
+        raise RemoteSyncEncryptionError(
+            "This store holds nothing to re-encrypt.\n"
+            "A plain `opensre remote-sync sync` will seal whatever you upload next."
+        )
+
+    # Before the loop, never after: a key that is not yet in the store(manifest josn) dies with
+    # the process, taking every object already sealed under it.
+    save_manifest(store, manifest)
```
Davidson3556, line 98: **P1 — concurrent re-encryption can orphan a live key.** Two clients can load the same manifest, A can save a manifest containing new key A and start resealing objects, then B can save its independently generated manifest containing new key B. The final manifest has B and the old keys but not A, so every object A writes after B's save is unrecoverable. Writing before the loop handles process interruption, not concurrent writers. Please use a conditional manifest write (for example an ETag/version CAS) or a store lease, retry from the latest manifest on conflict, and add a deterministic two-client regression test.

infrastructure/filestorage/encryption/manifest.py
```diff
@@ -0,0 +1,228 @@
...
+            )
+        kdf = raw["kdf"]
+        if kdf.get("name") != "scrypt":
+            raise RemoteSyncEncryptionError(
+                f"unsupported key derivation {kdf.get('name')!r} in this store's manifest"
+            )
+        return EncryptionManifest(
+            salt=base64.b64decode(kdf["salt"]),
+            params=ScryptParams(n=int(kdf["n"]), r=int(kdf["r"]), p=int(kdf["p"])),
+            active_key_id=str(raw["active_key_id"]),
```
Davidson3556, line 103: **P1 — the active generation can be downgraded.** `active_key_id` is not authenticated by any currently selected key. An actor who learned a retired root secret (or the former passphrase) can change this field to that still-wrapped old key; clients then accept the manifest and seal future writes under the compromised generation, defeating re-encryption as post-compromise recovery. Please either explicitly limit the threat model to an honest-but-curious immutable store, or authenticate the manifest with fresh root-derived integrity material and add anti-rollback state. An HMAC derived only from the passphrase KEK is insufficient here because the compromised passphrase can forge it; a full replay of an old valid manifest also needs a trusted generation anchor.

infrastructure/filestorage/encryption/manifest.py
```diff
@@ -0,0 +1,228 @@
...
+                f"this store's encryption manifest is version {version}; upgrade opensre to use it"
+            )
+        kdf = raw["kdf"]
+        if kdf.get("name") != "scrypt":
+            raise RemoteSyncEncryptionError(
+                f"unsupported key derivation {kdf.get('name')!r} in this store's manifest"
+            )
+        return EncryptionManifest(
+            salt=base64.b64decode(kdf["salt"]),
+            params=ScryptParams(n=int(kdf["n"]), r=int(kdf["r"]), p=int(kdf["p"])),
```
Davidson3556, line 102: **P2 — bound KDF work factors before constructing `Scrypt`.** The remote manifest supplies `n`, `r`, and `p` without limits. A modified manifest can therefore trigger excessive CPU or memory use, while some malformed values escape later as a cryptography `ValueError` instead of the feature's `RemoteSyncEncryptionError`. Validate a deliberately supported range and shape here (including salt/key sizes), then map invalid manifests to the domain error. Please add a regression test with oversized and invalid work factors.

infrastructure/filestorage/encryption/resolver.py
```diff
@@ -0,0 +1,93 @@
...
+    """Check the store against this machine's setting and build the cipher.
+
+    Creates the manifest when encryption is switched on for an empty store,
+    except under ``dry_run`` — a preview writes nothing anywhere, so it plans
+    against a throwaway key instead.
+    """
+    listing = store.list_objects("")
+    has_manifest = manifest_in_listing(listing)
+
+    if not encrypted:
```
Davidson3556, line 94: **P2 — deleting the manifest makes an encrypted store look plaintext.** If an encrypted store's manifest disappears, the `encrypted=false` path returns no cipher solely because the manifest is absent. A subsequent sync can download sealed envelopes as local session/memory files instead of failing closed. The `encrypted=true` path also diagnoses the same payloads as plaintext. Detect sealed payloads (or retain durable local state) before allowing the plaintext path, and reject this inconsistent store state with recovery guidance.

Davidson3556 (changes requested):
Request changes: the encryption design is promising, but concurrent key changes and a mutable remote manifest can currently make encrypted data unrecoverable or weaken post-compromise recovery. Please resolve the blocking threads before approval.

infrastructure/filestorage/encryption/manifest.py
dhakad22klx: Hi, root-keys are not exposed. Wrapped keys are exposed, and `active_key_id` is used for new write(`reencrypt`) but decrypt is done by same root-key that encrypted (identified by `key_id` when payload is unsealed). Passpharase is expected not be exposed, if there is a case then immediate passpharase rotation and `reencrypt` is requied. 
Store immutability should be the right constraint as per design. how can we enforce store immutability, can you suggest any idea ?.

infrastructure/filestorage/encryption/manifest.py
dhakad22klx: This will also be solved if we make store immutable(specifically manifest json file).

infrastructure/filestorage/encryption/resolver.py
dhakad22klx: Pushed changes to address this scenario.

infrastructure/filestorage/encryption/manifest.py
dhakad22klx: pushed changes that bounds and validates Scrypt params.

infrastructure/filestorage/encryption/rotation.py
dhakad22klx: Hi @Davidson3556 , Etag's comparision based conditional writes are not supported by all vendors (googled it).Please suggest a approach then i will work on that.

config/constants/filestorage.py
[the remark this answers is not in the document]
Davidson3556: @greptile-apps please don't remember that rule. It isn't the convention here, and your original comment was right.

`config/constants/filestorage.py` defines 16 constants. 13 are re-exported from `config/constants/__init__.py`. The 3 this PR adds are not:

- `REMOTE_SYNC_ENCRYPT_ENV`
- `REMOTE_SYNC_PASSPHRASE_ENV`
- `REMOTE_SYNC_KEY_CACHE_ENV`

So `config.constants.REMOTE_SYNC_ENV` resolves but `config.constants.REMOTE_SYNC_ENCRYPT_ENV` raises `AttributeError`. Every pre-existing constant in the file follows the re-export pattern; only the new ones diverge.

AGENTS.md is explicit about it: put them in a domain module under `config/constants/`, "and re-export via `config/constants/__init__.py`".

@dhakad22klx adding the three names to the existing import block and to `__all__` closes this.

infrastructure/filestorage/encryption/resolver.py
Davidson3556: Verified, this one is fixed. `_holds_sealed_objects` probes the smallest mirrored object and raises `ManifestMissingError` on both paths now, `encrypted=False` and `encrypted=True`, and the message points at restoring from a bucket version. Tests cover both directions.

Thanks for the quick turnaround. Resolving.

infrastructure/filestorage/encryption/manifest.py
Davidson3556: The bounds are good and `test_a_hostile_manifest_cannot_dictate_the_work_factor` covers the cases I had in mind. One input still escapes as a raw exception:

```
ok domain error     | n=2**30, n=100000, n=1024, r=1<<20, p=1<<20, salt too short/long
LEAKS OverflowError | n=1e400: cannot convert float infinity to integer
```

`{"n": 1e400}` parses to float `inf`, and `int(inf)` raises `OverflowError`, which is not in `parse_manifest`'s `(ValueError, KeyError, TypeError)`. That contradicts the docstring on the test above, which says malformed values must surface as this feature's own error rather than escaping.

Adding `OverflowError` to the except tuple closes it. Worth an eighth parametrize case so it stays closed.

infrastructure/filestorage/encryption/rotation.py
Davidson3556: You're right, and it rules out CAS as the correctness mechanism here. The providers are `aws`, `azure`, `gcs`, `s3compat`, `vercel`. S3 has had `If-None-Match` since 2024, GCS has `ifGenerationMatch` and Azure has `If-Match`, but Vercel Blob has no precondition API and `s3compat` varies by implementation. So a design that depends on conditional writes would be correct on three backends and silently wrong on the others.

The way out is to not need them. Make key material append-only instead of overwriting one mutable manifest:

- each generation becomes its own object, say `.opensre-sync-keys/<generation>-<key_id>.json`, holding the wrapped secret
- the manifest keeps the salt and cost parameters
- the active key is a deterministic max over the listing: highest generation, tiebreak on `key_id`

Two concurrent `reencrypt` runs then write to *different* object keys, so neither can erase the other. Both keys survive, so every object stays openable no matter which run "wins" — that is the property that actually matters, and it needs only `list_objects` and `put_object`, which the port already has.

One detail if you take it: `targets` currently excludes only `MANIFEST_KEY`, so the key objects would themselves get resealed on the next run. They need the same exclusion.

Conditional writes can still sit on top later as an optimisation where the backend supports it, declared through the registry the way `max_parallel_uploads` already is. Correctness just shouldn't rest on it.

One sequencing point, and it is the reason I would rather not defer this to a follow-up: the fix changes the on-disk format. Shipping the current format now and this later means a migration for everyone who encrypted a store in between. If you would rather keep this PR to the feature, gating `reencrypt` behind a warning until the format lands would also work.

infrastructure/filestorage/encryption/manifest.py
Davidson3556: Short answer: you can't enforce it from here. Object Lock, bucket versioning and deny-overwrite policies are operational controls and they differ per vendor, so nothing in this code can guarantee the store didn't change underneath it.

What you can do is stop needing the store to be honest. Pin trust on first use: record the highest generation seen for each `(provider, bucket, prefix)` locally under `~/.opensre/`, and refuse a manifest that moves `active_key_id` back to an older generation. That is anti-rollback state with no vendor dependency, and it falls out naturally if you take the append-only layout from the `rotation.py` thread, because generations become monotonic.

Honest caveat: trust-on-first-use protects continuity, not first contact. A brand new machine pointed at an already-tampered store has nothing to compare against.

The option from my original comment also still stands. If you would rather not carry local state, state the threat model explicitly — honest-but-curious store, an attacker holding both write access and a retired passphrase is out of scope — and I will close this. That is a legitimate answer for a first release. It just needs to be written down rather than implicit, so whoever relies on `reencrypt` as post-compromise recovery knows it isn't one.

infrastructure/filestorage/encryption/rotation.py
Davidson3556: Correcting one thing in my comment above: I'm withdrawing the "gate `reencrypt` behind a warning" fallback. Taking a working command away from users to buy time on a fix is a regression in its own right, and not something I should have put on the table.

That leaves the sequencing argument standing on its own, which is where it belonged. The format is the reason to do this here rather than later: change it after release and the cost lands on whoever encrypted a store in the meantime, which is the group you least want to hand a migration to.

If it genuinely doesn't fit in this PR, the honest alternative is to ship as-is and document that a concurrent `reencrypt` from two machines can lose keys. Document the limitation rather than disable the feature.

infrastructure/filestorage/encryption/manifest.py
Davidson3556: One thing to add to the trust-on-first-use suggestion above, because as written it could cost users more than it protects them.

A pin that refuses a rolled-back manifest will eventually refuse someone legitimate: a bucket restored from backup, a re-created prefix, a machine rebuilt from scratch. If you build it, it needs a recovery path and an error that names it, something like `opensre remote-sync trust --reset`. A guard that fails closed with no way forward is worse for the user than the attack it prevents.

That also tilts the choice between the two options. Documenting the threat model costs users nothing. The pin gives every user a new way to get stuck, in exchange for closing a case that needs both store write access and a retired passphrase. For a first release I would lean towards documenting it, and add the pin later if the threat model tightens.

Devesh36:
I'll review this in some time ..

infrastructure/filestorage/encryption/manifest.py
dhakad22klx: Pushed changes to handle this case.

infrastructure/filestorage/encryption/manifest.py
dhakad22klx: Pushed changes, that documents the threat model.

infrastructure/filestorage/encryption/rotation.py
dhakad22klx: Hi @Davidson3556 , let keep this PR for feature. Added concurrent `reencrypt` and `rotate-passphrase` warning in document. Pushed changes, please check. Thanks !!.

Davidson3556:
Re-reviewed at `5a7ca70`. The scrypt bounds landed well: validating in `ScryptParams.__post_init__` means no call site can bypass it, and adding a floor as well as a ceiling is better than what I asked for. `_holds_sealed_objects` also caught something worse than what I reported, ciphertext written over local history.

Two blockers since, both reproduced.

### 1. A wrong passphrase at `reencrypt` destroys the store

`with_new_generation` wraps a fresh root secret under the supplied KEK without first checking that KEK opens the existing active key. `open_manifest` then passes, because it validates the key it just wrote, and retired keys that fail to unwrap are skipped rather than fatal. `reencrypt` saves before the loop, then dies on the first object.

```
reencrypt raised: UndecryptableObjectError
manifest mutated by the failed run: True
correct passphrase NOW FAILS: WrongPassphraseError
wrong passphrase cannot read old objects: UndecryptableObjectError
```

Neither passphrase opens the store afterwards. Validate the existing manifest before generating or saving anything, and pin it with a no-mutation test.

Note `reencrypt` never prompts, it takes `resolve_passphrase()`. So the wrong value arrives from a mis-set env var, a machine pointed at another prefix, or from the bug below.

### 2. `rotate-passphrase` changes the remote before the new secret works locally

Three ways to strand a machine:

- `save_secret` strips, the rotation uses the raw prompt value. `'  new pass  '` wraps the remote and `'new pass'` gets stored, so the next run cannot open the manifest.
- `KeyringUnavailableError` is not a `RemoteSyncError`, so with `OPENSRE_DISABLE_KEYRING=1` (which the docs recommend) it escapes as a traceback after the remote already moved.
- An old passphrase still exported in the environment keeps shadowing the newly saved file value.

These two compound. #2 leaves a passphrase that cannot open the manifest, the operator runs `reencrypt` to recover, and #1 makes the loss permanent. Fix #1 first, it is what turns any passphrase mismatch into data loss.

### Smaller

- `_holds_sealed_objects` probes only the smallest object. A mixed prefix whose smallest file is plaintext lets encryption-off sync proceed and overwrite local files with the sealed remainder.
- `_encryption_status` reports a sealed store with a deleted manifest as healthy while sync raises `ManifestMissingError`. Its own docstring promises status and sync can never disagree.
- `parse_manifest` lets `AttributeError` through on a non-dict `kdf` or a list `wrapped_keys` (`'list' object has no attribute 'get'` / `'items'`).
- The new threat-model section says a tampered manifest makes sync refuse in both directions. Editing `active_key_id` to a retired generation is not refused, it silently seals new writes under the old key.
- `messages.py` says the old passphrase "no longer opens anything". With a retained copy of the old manifest it still decrypts until `reencrypt` runs, which the docs get right.
- Concurrent first syncs against an empty prefix still race to create competing manifests. The new warning covers rotate and reencrypt but not setup.
- Typos: "ecrypted root keys", "alone as it as", and in `keys.py` "laptop,so" and "binded". The new concurrency warning is bold text where the file uses `<Warning>` elsewhere, and the "prefix must be dedicated" warning lost its wrapper in the same commit.

Tests, lint, format and mypy are all green on my side (357 filestorage/CLI/shell tests); none of the above is covered by them.

YauhenBichel:
I will review on Saturday

dhakad22klx:
> Re-reviewed at `5a7ca70`. The scrypt bounds landed well: validating in `ScryptParams.__post_init__` means no call site can bypass it, and adding a floor as well as a ceiling is better than what I asked for. `_holds_sealed_objects` also caught something worse than what I reported, ciphertext written over local history.
> 
> Two blockers since, both reproduced.
> 
> ### 1. A wrong passphrase at `reencrypt` destroys the store
> `with_new_generation` wraps a fresh root secret under the supplied KEK without first checking that KEK opens the existing active key. `open_manifest` then passes, because it validates the key it just wrote, and retired keys that fail to unwrap are skipped rather than fatal. `reencrypt` saves before the loop, then dies on the first object.
> 
> ```
> reencrypt raised: UndecryptableObjectError
> manifest mutated by the failed run: True
> correct passphrase NOW FAILS: WrongPassphraseError
> wrong passphrase cannot read old objects: UndecryptableObjectError
> ```
> 
> Neither passphrase opens the store afterwards. Validate the existing manifest before generating or saving anything, and pin it with a no-mutation test.
> 
> Note `reencrypt` never prompts, it takes `resolve_passphrase()`. So the wrong value arrives from a mis-set env var, a machine pointed at another prefix, or from the bug below.
> 
> ### 2. `rotate-passphrase` changes the remote before the new secret works locally
> Three ways to strand a machine:
> 
> * `save_secret` strips, the rotation uses the raw prompt value. `'  new pass  '` wraps the remote and `'new pass'` gets stored, so the next run cannot open the manifest.
> * `KeyringUnavailableError` is not a `RemoteSyncError`, so with `OPENSRE_DISABLE_KEYRING=1` (which the docs recommend) it escapes as a traceback after the remote already moved.
> * An old passphrase still exported in the environment keeps shadowing the newly saved file value.
> 
> These two compound. #2 leaves a passphrase that cannot open the manifest, the operator runs `reencrypt` to recover, and #1 makes the loss permanent. Fix #1 first, it is what turns any passphrase mismatch into data loss.
> 
> ### Smaller
> * `_holds_sealed_objects` probes only the smallest object. A mixed prefix whose smallest file is plaintext lets encryption-off sync proceed and overwrite local files with the sealed remainder.
> * `_encryption_status` reports a sealed store with a deleted manifest as healthy while sync raises `ManifestMissingError`. Its own docstring promises status and sync can never disagree.
> * `parse_manifest` lets `AttributeError` through on a non-dict `kdf` or a list `wrapped_keys` (`'list' object has no attribute 'get'` / `'items'`).
> * The new threat-model section says a tampered manifest makes sync refuse in both directions. Editing `active_key_id` to a retired generation is not refused, it silently seals new writes under the old key.
> * `messages.py` says the old passphrase "no longer opens anything". With a retained copy of the old manifest it still decrypts until `reencrypt` runs, which the docs get right.
> * Concurrent first syncs against an empty prefix still race to create competing manifests. The new warning covers rotate and reencrypt but not setup.
> * Typos: "ecrypted root keys", "alone as it as", and in `keys.py` "laptop,so" and "binded". The new concurrency warning is bold text where the file uses `<Warning>` elsewhere, and the "prefix must be dedicated" warning lost its wrapper in the same commit.
> 
> Tests, lint, format and mypy are all green on my side (357 filestorage/CLI/shell tests); none of the above is covered by them.

1. Pushed chagnes to handle this edge case. Now, reencrypt will fail with WrongPassphraseError if wrong passphrase is supplied while re-encryption  `reencrypt`. No manifest change will happen. 
 ```
 (opensre) deepak@deepakdhakad:~/opensre$ opensre remote-sync reencrypt
That passphrase does not open this store's key.
Check the passphrase, or point at the remote store's prefix it belongs to.
```
2. Pushed changes :
- That normalizes input passphrase.
- Handle errors as RemoteSyncError when `OPENSRE_DISABLE_KEYRING=1`.
- Updated docs to address the case passphrase `OPENSRE_REMOTE_SYNC_PASSPHRASE` is exported in shell.
### Smaller:
1. Mixed Prefix case is already mentioned in docs that keep store prefix dedicated to OpenSRE,so either store will be whole plain text(never encrypted or whole sealed), hope that works. Else, I will have to check all object in _holds_sealed_objects  (once first one found, immediately return). Let me know if it is requierd to change, i will do that. 
2. Pushed changes now will handle this, deleted manifest and encryption off will check whether store has sealed objects.
<details>
  <summary><b>Click here see the terminal output(case: encryption off, manifest delete, store has sealed object)</b></summary>
  <br>

  ```
(opensre) deepak@deepakdhakad:~/opensre$ opensre remote-sync status
Remote sync is on (aws) → opensre-testing/opensre
Bucket access: private.
WARNING: This store holds encrypted objects but no manifest to open them.
The manifest carried the only copy of the keys, so those objects cannot be
recovered — and syncing either way now would make things worse.

  If the manifest was deleted by mistake, restore it from a bucket version.
  Otherwise start the remote over: empty the prefix, then sync again.
Mirrored:
  sessions   /home/deepak/.opensre/sessions (exists)
  memory     /home/deepak/.opensre/memory (exists)
Nothing is excluded. Set OPENSRE_REMOTE_SYNC_EXCLUDE to a comma-separated list of glob patterns, or a list under remote_sync.exclude, to keep paths off the store.
Never uploaded: integration credentials and model keys.
  ```

</details>

3. Pushed chagnes, to address this case.
4. I documented that any corrupt to manifest file, sync refuse in both direction.I kept it overall statement, not each and every case if `active_key_id` changed to old `key_id` or arbitrary `key_id` that doens't exist in wrapped keys etc. Let me change it to `sync may refuse in both direction and content may get sealed against unintended key` [Commit](https://github.com/Tracer-Cloud/opensre/pull/5082/commits/673407dab6a98c31914d0592336298c0f725a458). Hope it works.
5. Pushed chagnes updates the ROTATE_CONFIRM message appropriately.
6. Pushed changes, to address this.
7. Typos fixed and documentation adjusted in pushed changes

**Comment by dhakad22klx** [comment]
> I will review on Saturday

Hi @YauhenBichel  Please review.

**Comment by Davidson3556** [comment]
Re-reviewed the current head `20d4b400` against current `main` (`cf37f021`). The focused merged-tree tests and CI are green, but I do not think this is ready for approval yet.

The previously reported mixed-prefix/smallest-object problem remains reproducible, and the concurrent manifest-write discussion is still open, so I am not duplicating those findings below. One correction to my earlier conclusion: checking only the smallest object does not fully solve the missing-manifest case when an interrupted migration leaves a mixed prefix.

### P1 — Reject invalid encryption settings instead of disabling encryption

`infrastructure/filestorage/config.py:228-230` treats every non-empty value outside the truthy set as `False`. I reproduced `OPENSRE_REMOTE_SYNC_ENCRYPT=treu` causing an empty remote store to receive literal plaintext incident history. Because this is a security-sensitive switch and issue #4562 requires plaintext fallback to fail loudly, please accept explicit true/false tokens and raise `RemoteSyncConfigError` for anything else. Add an end-to-end typo regression that proves nothing is uploaded.

### P1 — Do not expose encryption exception details to gateway chat

`surfaces/interactive_shell/command_registry/remote_sync_cmds.py:239-244` prints `str(exc)` from `RemoteSyncEncryptionError` through the shared slash/gateway handler. `escape()` protects Rich markup; it does not redact information. I reproduced an `UndecryptableObjectError` containing a private remote object key appearing verbatim in the gateway reply. This directly conflicts with issue #4562's CWE-209 requirement. Please log the full exception server-side and map it to a static safe external message, with a gateway regression containing a planted secret.

### P2 — Keep `--pull-only` remote-read-only

`infrastructure/filestorage/encryption/resolver.py:124-126` creates and saves a manifest for an empty encrypted store before `run_remote_sync` applies the requested direction. Consequently, `remote-sync sync --pull-only` writes remotely despite its help text saying "Download only; send nothing," and it can fail with otherwise valid read-only credentials. Make manifest persistence direction-aware and only save it for push/both; add a pull-only empty-store regression.

### P2 — Do not silently ignore encryption flags with `--disabled`

`surfaces/cli/commands/remote_sync.py:163-168` does not pass `encrypt` into `_reject_disabled_with_setup_flags`. I reproduced both `setup --disabled --encrypt` and `--no-encrypt` exiting successfully while leaving the stored encryption setting unchanged. Since this path says `--disabled` alone only flips the switch, either reject an explicit encryption flag without `--bucket` or deliberately persist it. Please test both forms.

### P2 — Fail every non-matching passphrase read-back

`infrastructure/filestorage/encryption/keys.py:243-254` raises only when the mismatching read-back came from the environment. A `SecretTier.NONE` result or a different fallback value falls through as success, even though rotation has already changed the remote manifest and the next command cannot resolve the new passphrase. Raise `PassphraseNotResolvableError` for every `resolved.value != passphrase`, with regressions for `NONE` and mismatching fallback tiers.

### P2 — Treat every malformed KEK cache entry as a cache miss

`infrastructure/filestorage/encryption/keys.py:277-284` assumes decoded JSON is a mapping and does not validate the decoded KEK length. A cached `[]` raises `AttributeError` at `.get`; a matching entry containing a one-byte KEK reaches AES-GCM and raises a raw `ValueError`. This contradicts the adjacent contract that a damaged cache should only cause a slower sync. Validate the mapping shape, decode strictly, require `len(kek) == KEK_LEN`, and otherwise return `None`; add both regressions.

Validation on the current head/current-main merge: 379 focused tests passed; Ruff, format, and mypy passed; hosted CI, CodeQL, and CI Gate are green. These paths are not covered by the current tests.

**Comment by dhakad22klx** [comment]
> Re-reviewed the current head `20d4b400` against current `main` (`cf37f021`). The focused merged-tree tests and CI are green, but I do not think this is ready for approval yet.
> 
> The previously reported mixed-prefix/smallest-object problem remains reproducible, and the concurrent manifest-write discussion is still open, so I am not duplicating those findings below. One correction to my earlier conclusion: checking only the smallest object does not fully solve the missing-manifest case when an interrupted migration leaves a mixed prefix.
> 
> ### P1 — Reject invalid encryption settings instead of disabling encryption
> `infrastructure/filestorage/config.py:228-230` treats every non-empty value outside the truthy set as `False`. I reproduced `OPENSRE_REMOTE_SYNC_ENCRYPT=treu` causing an empty remote store to receive literal plaintext incident history. Because this is a security-sensitive switch and issue #4562 requires plaintext fallback to fail loudly, please accept explicit true/false tokens and raise `RemoteSyncConfigError` for anything else. Add an end-to-end typo regression that proves nothing is uploaded.
> 
> ### P1 — Do not expose encryption exception details to gateway chat
> `surfaces/interactive_shell/command_registry/remote_sync_cmds.py:239-244` prints `str(exc)` from `RemoteSyncEncryptionError` through the shared slash/gateway handler. `escape()` protects Rich markup; it does not redact information. I reproduced an `UndecryptableObjectError` containing a private remote object key appearing verbatim in the gateway reply. This directly conflicts with issue #4562's CWE-209 requirement. Please log the full exception server-side and map it to a static safe external message, with a gateway regression containing a planted secret.
> 
> ### P2 — Keep `--pull-only` remote-read-only
> `infrastructure/filestorage/encryption/resolver.py:124-126` creates and saves a manifest for an empty encrypted store before `run_remote_sync` applies the requested direction. Consequently, `remote-sync sync --pull-only` writes remotely despite its help text saying "Download only; send nothing," and it can fail with otherwise valid read-only credentials. Make manifest persistence direction-aware and only save it for push/both; add a pull-only empty-store regression.
> 
> ### P2 — Do not silently ignore encryption flags with `--disabled`
> `surfaces/cli/commands/remote_sync.py:163-168` does not pass `encrypt` into `_reject_disabled_with_setup_flags`. I reproduced both `setup --disabled --encrypt` and `--no-encrypt` exiting successfully while leaving the stored encryption setting unchanged. Since this path says `--disabled` alone only flips the switch, either reject an explicit encryption flag without `--bucket` or deliberately persist it. Please test both forms.
> 
> ### P2 — Fail every non-matching passphrase read-back
> `infrastructure/filestorage/encryption/keys.py:243-254` raises only when the mismatching read-back came from the environment. A `SecretTier.NONE` result or a different fallback value falls through as success, even though rotation has already changed the remote manifest and the next command cannot resolve the new passphrase. Raise `PassphraseNotResolvableError` for every `resolved.value != passphrase`, with regressions for `NONE` and mismatching fallback tiers.
> 
> ### P2 — Treat every malformed KEK cache entry as a cache miss
> `infrastructure/filestorage/encryption/keys.py:277-284` assumes decoded JSON is a mapping and does not validate the decoded KEK length. A cached `[]` raises `AttributeError` at `.get`; a matching entry containing a one-byte KEK reaches AES-GCM and raises a raw `ValueError`. This contradicts the adjacent contract that a damaged cache should only cause a slower sync. Validate the mapping shape, decode strictly, require `len(kek) == KEK_LEN`, and otherwise return `None`; add both regressions.
> 
> Validation on the current head/current-main merge: 379 focused tests passed; Ruff, format, and mypy passed; hosted CI, CodeQL, and CI Gate are green. These paths are not covered by the current tests.

Hi, For mixed prefix case. I had mention in last reply. remote-sync.mdx mention that prefix must be dedicated to SRE only.
So, interrupted migration can only make prefix mixed. So, added check in `_holds_sealed_objects` to find first sealed object then return. 

P1 — Reject invalid encryption settings instead of disabling encryption
    Pushed changes that implements validation. 
```
(opensre) deepak@deepakdhakad:~/opensre$ opensre remote-sync sync
Sync failed: OPENSRE_REMOTE_SYNC_ENCRYPT must be true or false (also 1/0, yes/no, on/off), not 'ture'
```
P1 — Do not expose encryption exception details to gateway chat
   Pushed changes that maps exceptions to external safe messages.
   
P2 — Keep `--pull-only` remote-read-only
   Pushed changes now will not write manifest to the store when sync is `--pull-only`
   ```
   (opensre) deepak@deepakdhakad:~/opensre$ opensre remote-sync sync --pull-only
   Sync complete — 0 downloaded, 0 uploaded, 0 already current.
   ```
  P2 — Do not silently ignore encryption flags with --disabled
    Pushed changes don not silently ignores them, instead reject (if without -bucket).
  ```
  (opensre) deepak@deepakdhakad:~/opensre$ opensre remote-sync setup --disabled  --no-encrypt
--disabled without --bucket only switches sync off; it cannot also set --no-encrypt. Pass --bucket tosave new settings, or drop the extra flags.
```
Tested with -bucket, works as expected. 

P2 — Fail every non-matching passphrase read-back
  Pushed requested chagnes, now raise PassphraseNotResolvableError for every non matching passphrase read back.

P2 — Treat every malformed KEK cache entry as a cache miss
 Pushed chagnes, now validating cache. So malformed cache will be treated as cache miss. 
 
Sorry for typos in commit messages. In [f381347](https://github.com/Tracer-Cloud/opensre/pull/5082/commits/f38134737b8843104ddaf98768426b31d763d51d), it is  `detected` not delted.  In [c5dcfc9](https://github.com/Tracer-Cloud/opensre/pull/5082/commits/c5dcfc91607dfac0601edd794602ae76edbff433), it is `exceptions`, not expections.

infrastructure/filestorage/config.py
```diff
@@ -210,6 +232,27 @@ def _enabled(stored: Callable[[], dict[str, Any] | None]) -> bool:
...
+    is safe in is this one: the run then meets an encrypted store with
+    encryption off and is refused, rather than uploading readable history.
+    """
+    env = os.getenv(REMOTE_SYNC_ENCRYPT_ENV)
+    if env is not None and env.strip() != "":
+        return _strict_bool(env.strip(), REMOTE_SYNC_ENCRYPT_ENV)
+    section = stored()
+    if section is None:
+        return False
+    return _truthy(_validated_scalar(section.get("encrypted"), "encrypted"))
```
Devesh36, line 253: Persisted encrypted values use permissive parsing, so a typo like "treu" silently disables encryption and may upload readable history. Fail closed with strict boolean parsing.

infrastructure/filestorage/encryption/manifest.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+def new_manifest(passphrase: str) -> tuple[EncryptionManifest, ManifestCipher]:
+    """A manifest and cipher for a store that has never been encrypted."""
+    salt = generate_salt()
+    params = ScryptParams()
+    kek = derive_kek(passphrase, salt, params)
+    root_secret = generate_root_secret()
+    root = derive_root_key(root_secret)
+    key_id = root.key_id.hex()
```
Devesh36, line 144: An existing manifest is treated as proof that all mirrored objects are encrypted; plaintext objects from partial migration or external insertion may remain exposed. Validate mirrored objects or require durable migration completion.

dhakad22klx: That is fixed, also checking is any sealed object is there. `_holds_sealed_objects` this function does that.

infrastructure/filestorage/config.py
dhakad22klx: Hi @Devesh36, Persisted values are getting edited by code only. so it will be either true or false. Hope this is perfectly fine. Above function `_enabled` also doing same.

dhakad22klx:
As requested, I will break this PR into 3 seperate PRs (as it became large to review and merge safely as one change)

YauhenBichel:
Hi @dhakad22klx thank you for so big and interesting work. The PR is very big and a bit risky to work with this PR. We need to redesign the story and create a epic with stories for current feature

dhakad22klx:
Thanks @YauhenBichel , are you planning to create scoped stories? Currently, I’m working on building small PRs. If you’re going to create stories, I can wait and work on them accordingly later.

Davidson3556:
@YauhenBichel has a message for you on discord regarding this.

PR #5083, opened 2026-08-19
feat(gateway): one turn, two entries — run() returns the result the callback discards

YauhenBichel:
The gateway can now serve an in-process caller that needs the turn's outcome as a value and has a terminal to bind, without a second turn handler and without changing what the four chat transports get.

**Two entries, one turn.**

```python
def __call__(self, text, session, output, logger) -> None:
    """The GatewayAgentCallback: chat transports reply through the output."""
    self.run(text, session, output, logger)

def run(self, text, session, output, logger, *,
        console=None, confirm_fn=None, is_tty=False,
        accounting_factory=None) -> TurnResult | None:

```

__call__ is now a one-line delegate and still returns None, so GatewayAgentCallback is untouched. run is the same turn — one capacity gate, one agent pool, one handle call — and returns the TurnResult.

Nothing new had to be computed. _run_turn already assigned turn_result = agent.handle(...) and dropped it on the floor; the result was there the whole time. The change is returning it and threading four optional keywords into the TurnBinding the handler already builds.


YauhenBichel:
@greptile review

PR #5177, opened 2026-08-19
refactor(platform): the turn host belongs below surfaces and gateway, not inside one

YauhenBichel:

`gateway/core/host/` → `platform/turn_host/`. 880 lines, 10 files, all detected
as renames.

**Why it moved.** The package was never gateway-specific: zero imports from
`gateway.transports`, zero from `gateway.web`, zero from any other
`gateway.core` package. Its whole dependency set was `core.agent_harness`,
`platform`, stdlib and rich. The capacity gate is `platform`. Analytics and
tracing are `platform`. The pool and capability policy are harness concerns.

**What the address cost.** `.importlinter.strict` declares the layer order
`surfaces | gateway` → `bootstrap` → `tools | integrations` → `core : platform`.
The pipe means *independent*: no edge between surfaces and gateway is legal by
default, which is why the four that exist are each written out by name under the
header *"Remove ignore_imports entries as refactors land."* Routing the
interactive shell through the turn handler would have added a fifth.

Both top layers already sit above `platform`, so relocating the host turns a
contested crossing into two ordinary imports. **No exception is added.**

**Two upward edges, found by doing it.** The move put the package under
`platform`, which sits below `tools`, and the strict contract immediately
flagged:

```
platform.turn_host.session_agents  -> tools.interactive_shell.subprocess_presenter
platform.turn_host.status_messages -> tools.registry
```

Both were in the pool's *default* branch — the chat defaults it supplied when no
`AgentBuildConfig` overrode them — so both became fields on that record:

- `subprocess_presenter_factory` — the harness `DefaultToolProvider` already
  took this as a port; the pool was hardcoding the `tools/`-side one.
- `describe_tool` — new `describe_registered_tool()` in `tools/registry.py`
  returns a tool's own wording. `status_messages.py` now formats a label it is
  handed, so it no longer knows a registry exists.

**A composer that can see both tiers.** Once the host cannot reach `tools`,
something above both must supply those. `gateway/core/chat_agent_build.py` (28
lines) states what a chat turn gets: gateway withholds, registry tool wording,
headless subprocess rendering.

ne trap worth noting for review: passing `agent_build` skips the pool's
`None`-branch, which is where `ensure_gateway_capability_policy` used to be
applied. The composer carries it explicitly, so chat keeps its withholds.

**Not in this PR.** The shell is not yet routed through the host — this removes
the reason it could not be. `GatewayTurnHandler`, `GatewayOutputSink` and
`GatewayAgentCallback` also still wear gateway names in `platform/` (19, 11 and
20 files). Deliberately left: a three-way rename does not belong inside a move.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5180, opened 2026-08-19
refactor(platform): the turn host's types stop naming the package they left

YauhenBichel:

Follow-up to #5177. The turn host moved to `platform/turn_host/`, but its three
public types kept `Gateway` prefixes:

| Was | Now |
| --- | --- |
| `GatewayTurnHandler` | `TurnHandler` |
| `GatewayAgentCallback` | `TurnCallback` |
| `GatewayOutputSink` | `TurnOutput` |

42 files, 198 occurrences. `docs/NAMING.md` rules out the package-name prefix,
and the module path already qualifies these — `turn_host.turn_handler.TurnHandler`
reads without the stutter `TurnHostTurnHandler` would have added.

**`TurnOutput`, not `TurnOutputSink`.** "Sink" names the shape of the
abstraction (things go in, nothing comes back) rather than its job, and the
docstring showed what that costs — it defined the type by what it is *not*:

> the same ephemeral turn-delivery port the harness already names, plus live
> tool-status and the final chat answer. **It is not a store, a notifier, or a
> callback.**

Three disclaimers is a name not carrying its weight. Every method on the type —
`print`, `render_response_header`, `render_error`, `stream`, `set_tool_status`,
`finalize` — shows something to the person who asked. The docstring now says so
directly, and the name matches both the module (`turn_output.py`) and the
parameter it is always bound to (`output: TurnOutput`).

**Deliberately not renamed:**

- `capture_gateway_turn_started` / `_completed` / `_failed` — these are
  **analytics event names**. Renaming the Python symbol is harmless; renaming
  what reaches PostHog splits every existing dashboard and funnel at the rename
  date.
- `GatewayController` — one reference in `concurrency.py`, and it is correct:
  that controller really is `gateway/core/lifecycle/controller.py`.
- `normalize_gateway_status` — status copy for the chat placeholder message.
  Genuinely about chat gateways.

**Prose, not just symbols.** A type rename leaves docstrings lying, so the five
that still described these as gateway things were corrected — including
`turn_handler.py`, which opened with *"This is the **only** gateway turn
handler"* in a file that is no longer in the gateway.

- [x] I have modified the AI output to follow this project's coding standards

A package that changed tiers left every type inside it claiming the old tier.
That is the kind of staleness that survives because nothing fails — the imports
resolve, the tests pass, and only a reader is misled.

The interesting decision was `GatewayOutputSink`. The mechanical answer was
`TurnOutputSink`; the better answer dropped "sink" entirely. Words like sink and
wire describe the plumbing an abstraction resembles rather than the work it
does, and they tend to appear where the job was never pinned down. This type's
own docstring was the evidence, defining itself by exclusion. Naming it for what
it shows the user made the docstring shorter and removed the need for the
disclaimers.

### DIFF
--- a/core/agent_harness/AGENTS.md
+++ b/core/agent_harness/AGENTS.md
@@ -82,7 +82,7 @@ and explicit host APIs remain for older readers. Checklist progress uses
 
 Do **not** duplicate the default port stack outside `DefaultHeadlessBuild`.
 The interactive shell builds a `HeadlessAgent` via `AgentBuildConfig` and
-`DefaultHeadlessBuild.agent` (not `GatewayTurnHandler`). Do not reintroduce
+`DefaultHeadlessBuild.agent` (not `TurnHandler`). Do not reintroduce
 peer `bootstrap.adapters` copies under surfaces or gateway.
 
 **Bind ports:** session-aware defaults implement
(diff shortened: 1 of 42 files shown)


YauhenBichel:
@greptile review

PR #5209, opened 2026-08-19
feat(shell): the interactive shell runs the same turn as Slack

YauhenBichel:
The interactive shell now reaches the agent through `platform.turn_host.TurnHandler.run()` instead of building its own agent.

**What the shell gains:** the process capacity gate, capability policy, pooled agent reuse, turn timeout and terminal outcome. It was outside all of them - not a bug it had, a guarantee set it was not in.

**What it keeps:** its own tools, prompts and gather phase, because the REPL builds its handler with `shell_agent_build_config()`. `AgentBuildConfig` (#5075) is what makes one host serve two very different callers.

`build_shell_agent` no longer appears in `controller.py` or `shell_turn_execution.py`.

**Three supporting changes, each preventing a regression:**

- **`run()` takes `on_progress`.** The host sends a compact status line a chat
  placeholder can hold; the REPL prints the full session-goal block with
  `markup=False`. Defaults to the host's renderer, so transports are untouched.
- **`run()` takes `is_tty: bool | None`.** `None` means "detect from stdin"
  (`execution_confirm.py:84`). An earlier draft of this PR collapsed `None` to
  `True`, which would have made a piped or redirected shell claim a terminal and
  prompt for confirmation. `TurnBinding.is_tty` was already `bool | None`; the
  host was narrowing it for no reason.
- **`ShellOutputSink` satisfies `TurnOutput`** — `set_tool_status` prints its
  own dim line, since a terminal has no placeholder to rewrite, and `finalize`
  handles a turn that produced text without streaming it.

**Harness tests stop driving the agent through a surface.** 16 files called
`execute_shell_turn`; 10 of them were harness tests using its stage-injection
seams — `tests/core/agent_harness/`, `tests/core/agent/` — not shell tests.
Those moved to `tests/shared/harness_turn_driver.py`, whose docstring states the
rule: *production code must not import this*. The 5 that genuinely exercise the
shell path stayed where they were.

### DIFF
--- a/core/agent_harness/turns/answer_finalize.py
+++ b/core/agent_harness/turns/answer_finalize.py
@@ -65,17 +65,17 @@ class AnswerFinalizeResult:
     finish_stream: bool
 
 
-def finish_streamed_response(output: OutputSink | None, text: str) -> None:
+def finish_streamed_response(output: OutputSink | None, answer: str) -> None:
     """Flush deferred/rewritten gather paint on surfaces that support it."""
     if output is None:
         return
     finish = getattr(output, "finish_streamed_response", None)
     if callable(finish):
-        finish(text)
+        finish(answer)
         return
     finalize = getattr(output, "finalize", None)
     if callable(finalize):
-        finalize(text)
+        finalize(answer)
 
 
 def _offered_upgrade_ctas(session: SessionState) -> set[str]:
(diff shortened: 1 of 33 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5210, opened 2026-08-19
refactor(harness): the host-capability seams say what they return

YauhenBichel:

The five `Callable[[], Any]` seams in `core/agent_harness/ports.py` — the
architecture review's finding 7. **Zero remain** in `core/`, `platform/` or
`gateway/`.

This was the only finding that got *worse* while the others improved: three more
untyped seams arrived alongside `AgentBuildConfig` before the newer ones were
converted to call-Protocols. This closes the originals and adds a test so it
cannot drift a third time.

**Two problems, two fixes.**

```python
LlmFactory = Callable[[], AgentLLMClient]     # was Callable[[], Any]
```

AgentLLMClient already exists in core/llm/types.py and declares exactly
invoke and tool_schemas — precisely what the loop calls (3 × llm.invoke(,
1 × llm.tool_schemas(). No new Protocol, and no cycle: core.llm does not
import core.agent_harness.

```
InvestigationPortsFactory = Callable[[], object]   # was Callable[[], Any]
LlmProviderPortsFactory   = Callable[[], object]
TaskCancelPortsFactory    = Callable[[], object]
SlashPortsFactory         = Callable[[], object]
```

These four had a documented reason to be untyped, and it is still true: their
real Protocols live above core (SlashPorts, LlmProviderPorts,
TaskCancelPorts in surfaces/, InvestigationLaunchPorts in tools/), and
naming them here would mean core importing tools. The review filed them as
plain debt; that was partly wrong.

But the reason argues against Any, not against typing. core calls each
factory and hands the result to ActionToolContext without reading a single
attribute. object is the type that says so, and unlike Any it rejects a
misspelled attribute. The comment in the file now records both halves.

A shared interface was considered and rejected. All four Protocols extend
ExecutionGate, so typing them as that base would compile. It would also make
them mutually substitutable — slash ports would satisfy a task-cancel
parameter — because each adds distinct methods the base does not carry
(command_exists, tool_description, dispatch_cancel). That is worse than
object: it looks type-safe while accepting four incompatible things. Per-port
typing is the real answer, and it needs those Protocols below core — a package
move like #5177, not an annotation change, so not in this PR.

### DIFF
--- a/core/agent_harness/ports.py
+++ b/core/agent_harness/ports.py
@@ -18,6 +18,7 @@ from core.agent_harness.turns.gather_observation import GatheredEvidence
 from core.agent_harness.turns.turn_results import ToolCallingTurnResult, TurnResult
 from core.domain.types.tools import ToolSurface
 from core.execution import ToolExecutionHooks
+from core.llm.types import AgentLLMClient
 
 # A tool-loop event callback: ``(kind, data)`` where kind is e.g. "tool_start".
 ToolEventObserver = Callable[[str, dict[str, Any]], None]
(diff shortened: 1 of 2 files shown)


YauhenBichel:
@greptile review

PR #5211, opened 2026-08-19
test(scheduler): the delivery support matrix stops living in prose

YauhenBichel:
`Provider` offers six members to every caller. No caller supports all six, and
until now each consumer's subset lived in prose — the enum docstring even says
consumers *"define their own documented subset rather than exposing a choice
that would silently fail."* This makes that documentation an assertion.

**Writing it down surfaced three different answers to "supported" in one
package:**

| | Supports |
| --- | --- |
| `Provider` offers | telegram, slack, discord, rocketchat, interactive_shell, **buzz** |
| `executor._deliver_single` | telegram, slack, discord, rocketchat, interactive_shell |
| `delivery._DELIVERY_SPECS` | telegram, slack, rocketchat |
| watchdog | telegram, rocketchat, buzz |

**`buzz` is offered to a cron task and refused by its executor.** It is in the
vocabulary because the watchdog delivers to it; `_deliver_single` has no branch,
so it reaches the `else` and fails with *"Unsupported provider"* — at delivery
time, not at creation. That is finding 6's "the enum offers choices that
silently fail", made concrete.

**`discord` is delivered but unknown to readiness.** The executor sends to it
while `any_delivery_ready()` and the setup hints do not know it exists.
`interactive_shell` is in the same gap *deliberately* — documented as a local
inbox so digest CLIs do not offer it as a channel — and the test distinguishes
the intentional exclusion from the accidental one.

*The watchdog fallthrough is a live hazard.** `_build_dispatcher` branches on
Rocket.Chat and Buzz and falls through to Telegram. Adding a member to
`WATCHDOG_SUPPORTED_PROVIDERS` without a branch would not raise — that alarm
would go to Telegram. The test asserts `declared == branched | {TELEGRAM}`.

Sets are read from the code by AST rather than restated, so the branches are the
source of truth and only the expectations are compared.

### DIFF
--- a/gateway/AGENTS.md
+++ b/gateway/AGENTS.md
@@ -74,7 +74,7 @@ Packages are split like `core/agent_harness/prompts/`: **core infra** vs
   settings, inbound worker, security, turn output, and `startup.py`. Peers
   never import each other or `gateway.startup`/`web`; anything two need belongs in
   `core/` (per-turn steps in `gateway.core.middleware`) or
-  `platform.turn_host` (turn handler, output sink, session agents).
+  `platform.turn_host` (turn handler, turn output, session agents).
 - `web/` — web surface (FastAPI app, investigations API, worker/artifacts).
   May import `core/`; must not import chat transports or `gateway.startup`.
 - `core/storage/session/resolver.py` — per-conversation session binding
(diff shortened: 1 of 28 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5218, opened 2026-08-19
fix(telegram): dispatch turns as tasks so approvals can resolve

### LINKED ISSUE #5076: [BUG] Telegram gateway poll loop deadlocks its own approval flow — every approval silently denies after 180s

kabirvashisht4-glitch:
Fixes #5076

The Telegram poll loop awaited turn dispatch inline, so a turn blocked in `ApprovalBroker.wait` held the one coroutine that could ever deliver the button click resolving it. The click only arrives on a later `poll_once`, which the suspended loop can never reach — so every Telegram approval expired to denied after `MAX_APPROVAL_WAIT_SECONDS` regardless of how fast the user clicked, and polling for every other chat stayed frozen for that window.

**`gateway/transports/telegram/background.py`**
- Turn dispatch is now fired with `asyncio.create_task` into a `pending_tasks` set, never awaited inline — the same pattern `gateway/transports/buzz/background.py` already documents and uses for this exact failure mode.
- New `_dispatch_turn` wraps the turn so a detached task's exception is logged. The loop's own `except` no longer covers it, and an unretrieved task exception would otherwise surface only at garbage-collection time.
- New `_drain_active_turns` gives in-flight turns a bounded window on shutdown, because `asyncio.run` cancels whatever is still pending once the loop coroutine returns. The budget is deliberately small: polling has stopped, so nothing can deliver a click any more, and a turn parked in `ApprovalBroker.wait` would otherwise hold shutdown for the full expiry.

**`gateway/transports/telegram/settings.py`**
- Adds `shutdown_drain_seconds` (default `5.0`), mirroring the existing Buzz setting the drain budget needs.

**`gateway/tests/telegram/test_background.py`**
- Four tests: the deadlock itself, detached-task error isolation, the shutdown drain, and `/stop` still resolving through the registry without dispatching a turn.

No behaviour change to the turn itself, the inbound security checks, or the `/stop` path.

The two new regression tests fail against `main` and pass with the fix.

**Against `main`** (fix reverted, new tests kept):

```
FAILED gateway/tests/telegram/test_background.py::test_polling_continues_while_a_turn_is_in_flight
FAILED gateway/tests/telegram/test_background.py::test_a_failing_turn_does_not_stop_polling
========================= 2 failed, 3 passed in 5.70s ==========================
```

**With the fix applied:**

```
gateway/tests/telegram/test_background.py::test_start_starts_poll_thread PASSED
gateway/tests/telegram/test_background.py::test_polling_continues_while_a_turn_is_in_flight PASSED
gateway/tests/telegram/test_background.py::test_a_failing_turn_does_not_stop_polling PASSED
gateway/tests/telegram/test_background.py::test_in_flight_turns_finish_before_shutdown_returns PASSED
gateway/tests/telegram/test_background.py::test_stop_command_is_handled_without_dispatching_a_turn PASSED
============================== 5 passed in 5.08s ===============================
```

---

---

### DIFF
--- a/gateway/tests/telegram/test_background.py
+++ b/gateway/tests/telegram/test_background.py
@@ -80,10 +80,13 @@ def _run_loop(
     *,
     resources: TelegramPollingRuntime,
     stop_event: threading.Event,
+    shutdown_drain_seconds: float = 2.0,
 ) -> asyncio.Task[None]:
     return asyncio.create_task(
         background._poll_telegram_until_stopped(
-            settings=GatewaySettings(bot_token="tok", shutdown_drain_seconds=2.0),
+            settings=GatewaySettings(
+                bot_token="tok", shutdown_drain_seconds=shutdown_drain_seconds
+            ),
             stop_event=stop_event,
             logger=LOGGER,
             resources=resources,
@@ -173,26 +176,35 @@ async def test_a_failing_turn_does_not_stop_polling() -> None:
 
 
 @pytest.mark.asyncio
-async def test_in_flight_turns_finish_before_shutdown_returns() -> None:
-    """Detached turns are drained, not cancelled mid-flight by ``asyncio.run``."""
+async def test_shutdown_cancels_the_turn_body_not_only_its_await() -> None:
+    """A turn that outlives the drain budget gets its cancel Event set.
+
+    Cancelling the task only unwinds the ``await``; the turn body runs on the
+    executor and never sees that. It stops when its ``turn_cancel`` Event is
+    set, and only then does ``executor.shutdown(wait=True)`` return inside the
+    shutdown budget instead of blocking to the turn timeout.
+    """
     # Arrange
     poller = _FakePoller([TelegramPollResult(messages=[_message()])])
     turn_started = asyncio.Event()
-    turn_finished = False
+    seen: list[threading.Event] = []
 
-    async def _slow_turn(_event, **_kwargs) -> None:
-        nonlocal turn_finished
+    async def _turn_outliving_the_budget(_event, *, turn_cancel=None, **_kwargs) -> None:
+        seen.append(turn_cancel)
         turn_started.set()
-        await asyncio.sleep(0.05)
-        turn_finished = True
+        await asyncio.sleep(30)
 
     stop_event = threading.Event()
 
     with (
         patch.object(background, "TelegramPoller", lambda _token: poller),
-        patch.object(background, "handle_polled_inbound_telegram_message", _slow_turn),
+        patch.object(
+            background, "handle_polled_inbound_telegram_message", _turn_outliving_the_budget
+        ),
     ):
-        loop_task = _run_loop(resources=_resources(), stop_event=stop_event)
+        loop_task = _run_loop(
+            resources=_resources(), stop_event=stop_event, shutdown_drain_seconds=0.05
+        )
         await asyncio.wait_for(turn_started.wait(), timeout=2.0)
 
         # Act — shut down while the turn is still running.
@@ -200,7 +212,8 @@ async def test_in_flight_turns_finish_before_shutdown_returns() -> None:
         await asyncio.wait_for(loop_task, timeout=3.0)
 
     # Assert
-    assert turn_finished is True
+    assert seen and seen[0] is not None
+    assert seen[0].is_set(), "drain cancelled the await but left the turn body running"
 
 
 @pytest.mark.asyncio
--- a/gateway/tests/telegram/test_background.py
+++ b/gateway/tests/telegram/test_background.py
@@ -1,25 +1,34 @@
 from __future__ import annotations
 
+import asyncio
 import logging
+import threading
+import time
 from unittest.mock import MagicMock, patch
 
+import pytest
+
+from gateway.core.middleware.active_turns import ActiveTurnRegistry
+from gateway.core.middleware.approvals import ApprovalBroker
+from gateway.transports.telegram import background
 from gateway.transports.telegram.background import start_telegram_gateway_background
+from gateway.transports.telegram.poller.poller import TelegramPollResult
 from gateway.transports.telegram.runtime import (
+    TelegramPollingRuntime,
     initialize_telegram_polling_runtime,
     shutdown_telegram_polling_runtime,
 )
-from gateway.transports.telegram.settings import GatewaySettings
+from gateway.transports.telegram.settings import GatewaySettings, TelegramInboundMessage
+
+LOGGER = logging.getLogger("gateway.test")
 
 
 @patch("gateway.transports.telegram.background.TelegramPoller")
 def test_start_starts_poll_thread(mock_poller_cls: MagicMock) -> None:
-    from gateway.transports.telegram.poller.poller import TelegramPollResult
-
     mock_poller_cls.return_value.poll_once.return_value = TelegramPollResult()
-    logger = logging.getLogger("gateway.test")
     handle = start_telegram_gateway_background(
         settings=GatewaySettings(bot_token="tok"),
-        logger=logger,
+        logger=LOGGER,
         initialize_runtime=initialize_telegram_polling_runtime,
         shutdown_runtime=shutdown_telegram_polling_runtime,
         handle_callback_to_gateway_agent=lambda *_args: None,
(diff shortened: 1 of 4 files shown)


kabirvashisht4-glitch:
1. same-batch /stop — you're right, this is a regression. buzz registers the
   turn_cancel before create_task for exactly this reason and i missed it when
   porting the pattern over. fixing by pre-registering, and adding a test with
   a message and /stop in one batch since that's what my tests were missing.

2. approval wait outliving the drain — also right. buzz resolves pending
   approvals as denied before draining, but telegram has no pending_approvals
   registry to drain, so the bounded wait promises something it can't deliver.
   doing this properly means adding that registry to the telegram transport.
   happy to do it here, but it's a bit beyond "fix the deadlock" — do you want
   it in this PR or tracked separately?

3. unbounded dispatch queue — fair. worth noting buzz has the same shape today,
   so if you want it bounded i'd rather fix both than diverge. same question:
   here or separate issue?

4. docstrings — agreed, that's AGENTS.md. trimming them to the contract and
   leaving the history in the commit message.

Davidson3556:
kindly attach a demo

Davidson3556:
Went through this properly: pulled the branch, reverted the fix with the tests kept, and 3 of the 5 new ones fail. So the coverage is real. Lint, format, mypy and the full gateway suite (768) pass on my end too. The diagnosis is right, and it matches how `ActiveTurnRegistry.bind_user_stop` was already meant to be used, so the shape of this is good.

Two things before it goes in.

**The drain stops the await, not the turn.** `_drain_active_turns` cancels the task but never sets that turn's `turn_cancel` Event. The body runs in `run_in_executor`, so the thread keeps going, and `shutdown_telegram_polling_runtime` then sits in `executor.shutdown(wait=True)` until the turn's own 240s timeout fires. That is a long way past the 5s budget, and `handle.stop(timeout=8.0)` gives up and reports the thread never stopped. I checked by printing the Event after the drain returns:

```
drain finished; turn_cancel.is_set() == False
```

Setting each still-running turn's Event before `task.cancel()` should cover it, since the agent loop already polls it.

In fairness this is not a regression. Main awaits inline, so shutdown already blocks with no bound at all. But the docstring promises a bound it does not deliver yet.

**The AI usage section is blank.** Neither box ticked, and the approach field is still the placeholder comment. That section is required here, so it blocks the merge on its own.

Two smaller ones, take or leave:

- `test_in_flight_turns_finish_before_shutdown_returns` and `test_stop_command_is_handled_without_dispatching_a_turn` both pass with the fix reverted, so neither pins anything. The first is meant to cover the drain and cannot fail, because the old inline await also finished the turn.
- The `turn_cancel is None` branch in `inbound_handler.py` is dead once `background.py` is the only caller. Worth making the parameter required and dropping the `nullcontext` fork.

The `/stop`-arriving-in-the-same-batch case was a good catch. Easy one to miss.

Also heads up that this overlaps with #5388, which edits the same block in `inbound_handler.py`. We would rather land yours first and rebase that one on top, so nothing needed from you there.

kabirvashisht4-glitch:
thanks for actually pulling it and testing — the printed Event was the
convincing bit.

fixed in c4d3953. the drain now tracks each turn's cancel Event alongside its
task and sets it before cancelling, so the executor-side body stops on its own.
also replaced test_in_flight_turns_finish_before_shutdown_returns — you're
right that it couldn't fail. the new one holds a turn past the drain budget and
asserts the Event was set; i checked it goes red without the change.

on the turn_cancel is None branch — i tried making it required and it breaks
test_both_polled_transports_run_a_turn_in_the_same_order, which pins telegram
and buzz producing the same happy-path order including cancel_tracked. buzz has
the same dead branch, so removing it properly means doing both transports
together. happy to do that as a follow-up if you want it.

left test_stop_command_is_handled_without_dispatching_a_turn in — it's the only
cover for the opposite branch (no active turn -> NO_ACTIVE_TURN_MESSAGE), which
the same-batch test doesn't touch. happy to drop it if you'd rather.

Davidson3556 (approved):
Re-reviewed on c4d3953c. The drain fix is right: setting each straggler's cancel Event before cancelling the task is exactly what was missing, and the rewritten test earns its place. I reverted just the two Event-set lines and it fails, so it pins the behaviour rather than describing it.

Verified here: ruff, format, mypy (316 files), telegram suite 77, full gateway suite 768, and CI is 16/16 green on this head.

One retraction from my last pass. I said the `turn_cancel is None` branch in `inbound_handler.py` was dead compatibility code. That was wrong. Buzz carries the identical fork, and three existing tests call the handler without the argument, so keeping it matches the sibling transport. Leave it as is. Same for `test_stop_command_is_handled_without_dispatching_a_turn`: it does not fail without the fix, but it does guard the new dispatch-time registration against regressing the "nothing to stop" reply, so it earns its keep too.

Approving. The one thing still open is the **Explain your implementation approach** field, which is still the placeholder comment. I would rather you wrote that yourself than have me put words in your attestation. A couple of sentences on why the cancel Event has to be set before the task cancel would cover it.

Davidson3556:
LGTM

muddlebee:
@greptile review

muddlebee:
@Davidson3556 pls monitor the greptile reviews

PR #5220, opened 2026-08-19
fix(shell): /exit stops replaying its own farewell

### LINKED ISSUE #5219: Bug: /exit prints the farewell twice and leaves spinner frames on screen
### What happens

Typing `/exit` in the interactive shell prints the resume hint and `goodbye.`
twice, with raw spinner frames in the second copy:

[2] ❯ /exit

Resume this session with:
/resume f659a4f9-3911-4ed2-bc8c-8e02da5f887c
opensre --resume f659a4f9-3911-4ed2-bc8c-8e02da5f887c
goodbye.

slash /exit (succeeded)
Resume this session with:
/resume f659a4f9-3911-4ed2-bc8c-8e02da5f887c
opensre --resume f659a4f9-3911-4ed2-bc8c-8e02da5f887c
⠋ finishing up…⠙ finishing up…⠹ finishing up…
goodbye.

### What is actually going on

Two independent causes. `/exit` runs **once** — it is rendered twice.

**1. Captured output is replayed.** Every slash command runs under
`capture_console_segment`, which tees: output renders live *and* is recorded
via Rich's `record` mode. That recording becomes the turn's `response_text`,
which the shell then prints. So the farewell appears once from the command and
once as the turn's response.

The spinner is the giveaway. `console.status` animates by rewriting one line,
but `export_text` records every frame it wrote — so the replay transcribes the
frames as plain text. That is why the live block has no spinner and the replay
shows three in a row. Nothing repeated; the frames were written down.

YauhenBichel:
Issue: #5219 

`/exit` printed the resume hint and `goodbye.` twice, the second copy carrying
raw spinner frames. **It never ran twice** — it was rendered twice.

Every slash command runs under `capture_console_segment`, which tees: the output
renders live *and* is recorded through Rich's `record` mode. That recording
becomes the turn's `response_text`, which the shell prints. Hence two copies.

The spinner is what made it obvious rather than merely redundant.
`console.status` animates by rewriting one line; `export_text` records **every
frame it wrote**. So the live block has no spinner at all and the replay shows
three frames as plain text — nothing repeated, the frames were transcribed.

`_SUMMARY_ONLY_SLASH_ROOTS` already exists for exactly this: commands whose
console output the model does not need. It held `/`, `/help`, `/?` and
`/investigate`. A resume hint and a goodbye tell the model nothing, so `/exit`
and `/quit` (same handler) join it. `response_text` becomes
`slash /exit (succeeded)`.

**Deliberately not fixed here:** the replay is not unique to `/exit` — any slash
command outside that set behaves the same way. Whether that is a defect or the
model legitimately needing the output is a design question, and widening the
list would answer it by accident.

### DIFF
--- a/platform/analytics/provider.py
+++ b/platform/analytics/provider.py
@@ -857,8 +857,12 @@ class Analytics:
         with contextlib.suppress(queue.Full):
             self._queue.put_nowait(None)
         if flush and self._worker is not None:
+            # One budget for the whole drain, not one per wait: these run back to
+            # back, so passing ``timeout`` to each made a documented 0.5s exit
+            # block for 1.0s whenever a send outlived it.
+            deadline = time.monotonic() + timeout
             self._drained.wait(timeout=timeout)
-            self._worker.join(timeout=timeout)
+            self._worker.join(timeout=max(0.0, deadline - time.monotonic()))
 
     def _ensure_worker(self) -> None:
         if self._worker is not None:
--- a/surfaces/interactive_shell/utils/telemetry/turn_outcome.py
+++ b/surfaces/interactive_shell/utils/telemetry/turn_outcome.py
@@ -34,6 +34,13 @@ _SUMMARY_ONLY_SLASH_ROOTS: frozenset[str] = frozenset(
         "/help",
         "/?",
         "/investigate",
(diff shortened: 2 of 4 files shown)


YauhenBichel:
@greptile review

PR #5222, opened 2026-08-19
fix(shell): the REPL stops reprinting answers it already rendered

### LINKED ISSUE #5221: Bug: Shell prints every action-phase answer twice, the second copy as raw markdown
### What happens

Any turn answered by the action phase — a skill, a tool chain — renders twice in
the interactive shell. The second copy is unformatted:

● OpenSRE
Assuming you mean GitHub Actions for Tracer-Cloud/opensre: 3% hard failure rate…
• 3 failed
• 59 succeeded
· 0.0s · ↓ 75 tokens

Assuming you mean GitHub Actions for Tracer-Cloud/opensre: 3% hard failure rate…

3 failed
59 succeeded

Reproduced with `what's the current error rate on github?` (activates the
`github-cli` skill). A plain conversational answer does not show it.

### Cause

Regression from #5209, which routed the shell through
`platform.turn_host.TurnHandler`.

The host calls `output.finalize(...)` whenever a turn is not `answered`:

```python
if not turn_result.answered and not cancelled:
    output.finalize(outbound_text or EMPTY_RESPONSE_MESSAGE)
answered does not mean "the user has seen the answer":

@property
def answered(self) -> bool:
    """A turn is "answered" exactly when the conversational LLM produced a run."""
    return self.llm_run is not None
```

An action-phase answer leaves llm_run as None, so answered is False
while the console has already painted the reply. ShellOutputSink.finalize —
added in #5209 — then printed it again with markup=False, which is why the
second copy shows raw **bold** and - bullets.

The underlying mismatch: a chat transport holds one placeholder message and
finalize is how that message gets its final text. A terminal has no
placeholder — output lands as it happens — so for the shell there is nothing to
finalize.

YauhenBichel:
Fixes #5221

**A regression I introduced in #5209.** Any turn answered by the action phase —
a skill, a tool chain — rendered twice in the shell, the second copy as raw
markdown.

The turn host calls `output.finalize(...)` whenever a turn is not `answered`.
I read that as "the user has not seen the answer yet". It means something else:

```python
@property
def answered(self) -> bool:
    """A turn is "answered" exactly when the conversational LLM produced a run."""
    return self.llm_run is not None
```

An action-phase answer leaves llm_run as None, so answered is False
while the console has already painted the reply. ShellOutputSink.finalize then
printed it again with markup=False — hence the second copy showing
**GitHub Actions** and - 3 failed unrendered.

My own docstring asserted the wrong thing ("Reached only when the turn produced
text without streaming it"), and that assumption was the whole basis for the
implementation.

The real distinction. A chat transport holds one placeholder message, and
finalize is how that message gets its final text — it must be called. A
terminal has no placeholder: output lands as it happens. So for the shell there
is nothing to finalize, and the method says so:

```
def finalize(self, answer: str) -> None:
    """No-op: the REPL has already shown everything this turn produced."""
    _ = answer
```

set_tool_status checked for the same hazard and left alone. It is
unreachable for the shell, which supplies its own build_tools (so the host's
_ToolStatusObserver never installs) and its own on_progress (so the host's
status renderer never runs).


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5262, opened 2026-08-20
fix(harness): metric questions that name their own system stop answering from PostHog, test: guard the shell's turn path and stop scanners reading vendored trees

### LINKED ISSUE #5264: Every metric question routes to PostHog, whatever system it asks about
### What happens

In the REPL:

```
```

The turn answers from PostHog:

> The live query could not be formed because the PostHog schema did not identify
> verified GitHub request and error events.

…followed by draft HogQL full of `<github_error_event>` placeholders and
`Setup/refresh the preferred analytics source with /integrations setup posthog_mcp`.

A second turn then reaches the `github-cli` skill and answers correctly. The
session-goal loop recovers on its own — it just burns a turn and shows a
confusing PostHog answer first.

### Scope

Not GitHub-specific. **Every** metric-read question in opensre routes to
PostHog: Datadog, Grafana, a cloud provider, anything. Only questions that
happen to be about product analytics get the right source, and that is a
coincidence of there being one registrant.

### Cause

Two hops. The first is wrong by construction.

1. The action planner tags the turn `evidence_kind=metric_read`. Defensible —
   "current error rate" is a metric read.
2. Core asks which source owns metric reads. That lookup cannot see what the
   question was about:

   ```python
   # platform/harness_ports.py
   def preferred_evidence_sources_for(kind: str) -> tuple[str, ...]:
       return _preferred_evidence_sources.get(kind, ())
   ```

   A plain `kind -> ids` map, with exactly one registrant:

   ```python
   # integrations/posthog_mcp/evidence_sources.py
   register_preferred_evidence_source("metric_read", "posthog_mcp")
   ```

The word "github" never reaches the decision.

PostHog then cannot form the query, which raises `evidence_tier:metric_unformed`.
That handoff instructs the assistant to draft "in the query language of the
preferred connected analytics source" and emit one `/integrations setup <id>`
line. The prompt does exactly what it was told; the wrong source was chosen
upstream.

### The real gap

`metric_read` is one kind covering two different questions:

- product analytics — events, funnels, retention (PostHog's domain)
- a number held by a system the user named — CI failures, build health

One kind, one registered owner, and no way for the planner to say which it meant.
The prompt already scopes `metric_read` to product analytics ("unique users, OS
breakdown, retention, events, pageviews"); the planner misapplied it because
there was nowhere correct to put a CI question.

### Not the fix

Matching "github" in the user's text and switching source. `AGENTS.md` bans
regex/keyword/fuzzy intent routing around the action agent, and it would not
generalize past the words someone thought of.

### Fix

Add `evidence_kind=service_metric_read` for a number held by a system the user
named. It registers no authoritative source, so `classify_evidence_need` takes
the fall-through branch instead of the `requires_authoritative_source` one, and
the action phase reads the system the user actually named — which is what turn 2
already did correctly.

No changes to `harness_ports`, the registry, or the PostHog path. `metric_read`
keeps the same registrant, the same `L0_degraded` behaviour and the same
unformed-metric floor.

### Files

- `core/agent_harness/turns/evidence_kind.py` — kind vocabulary and per-kind policy
- `core/agent_harness/turns/evidence_need.py` — where preferred ids attach
- `core/agent_harness/turns/metric_query_floor.py` — the unformed-metric floor
- `core/agent_harness/prompts/action/text.py` — planner guidance
- `tools/interactive_shell/actions/assistant_handoff.py` — the tool schema description
- `platform/harness_ports.py`, `integrations/posthog_mcp/evidence_sources.py` — registration and lookup (unchanged; listed for context)

### Repro

```
uv run opensre
```

### Note for whoever verifies

No test can cover the part that actually matters here — whether the planner picks
the new kind for a given question. That is prompt quality, and only running the
REPL shows it.

YauhenBichel:
Issue: #5264

Asking the REPL `what's the current error rate on github?` answered from
PostHog:

> The live query could not be formed because the PostHog schema did not identify
> verified GitHub request and error events.

…followed by draft HogQL full of `<github_error_event>` placeholders and
`/integrations setup posthog_mcp`. Turn 2 then reached the `github-cli` skill
and answered correctly — so the session-goal loop recovers on its own, after one
wasted turn and one confusing answer.

**Cause.** Two hops; the first is wrong by construction. The planner tags the
turn `evidence_kind=metric_read` — defensible, it *is* a metric read. Core then
asks which source owns metric reads, and that lookup cannot see what the
question was about:

```python
def preferred_evidence_sources_for(kind: str) -> tuple[str, ...]:
    return _preferred_evidence_sources.get(kind, ())
```

A plain `kind → ids` map with exactly one registrant:

```python
register_preferred_evidence_source("metric_read", "posthog_mcp")
```

So **every** metric-read question in opensre routed to PostHog — GitHub,
Datadog, anything. The word "github" never reached the decision. PostHog then
could not form the query, raising `evidence_tier:metric_unformed`, whose handoff
instructs the assistant to draft in "the query language of the preferred
connected analytics source" and emit one `/integrations setup <preferred id>`.
The prompt did exactly what it was told; the wrong source was chosen upstream.

The prompt already scoped `metric_read` to product analytics ("unique users, OS
breakdown, retention, events, pageviews"). The planner misapplied it because
there was nowhere else to put a CI question — the vocabulary gap *was* the bug.

The session-goal default also moves from a hard-coded `kind == METRIC_READ`
check to a policy field (`implies_session_goal`), keeping `EvidenceKind`
open/closed the way its own docstring asks instead of adding a second per-kind
branch.

**Not done:** keyword-matching "github" in the user's text. `AGENTS.md` bans
regex/keyword intent routing around the action agent, and it would not
generalize past the words someone thought of.

Tests only — no product code changes. Two guards, both for behaviour that was
real but unpinned, so removing it passed the whole suite.

**1. The shell's turn path (Wave E, item E6).**

#5209 pointed the REPL at `platform.turn_host.TurnHandler` — the same handler
Slack, Telegram and Discord use. Nothing enforced that. `build_shell_agent`
still exists in `surfaces/interactive_shell/runtime/shell_agent.py` and rebuilds
the agent from the *same* `shell_agent_build_config()` the host already uses, so
one import puts the second agent path back and the surfaces drift apart again.

`tests/interactive_shell/test_shell_turn_path_border.py` pins both halves, by
AST rather than text search so a commented-out call cannot keep it green:

- no shipped module calls `build_shell_agent` — empty allowlist compared
  **exactly**, so a stale entry fails too, matching the other border ratchets
- the turn path imports **and** runs `TurnHandler`

It also pins the blank line before `● assistant`, which moved out of core's
`_show_response` into `ShellOutputSink.render_response_header`. Deleting it used
to pass 2,205 tests while the REPL silently lost the separator.

**2. Scanners stop reading vendored trees as product source.**

Three guards did a blind `rglob("*.py")` over product packages, so anything
checked out inside one was read as ours. A `config/.venv` made all three red at
once — the psutil confinement scan, the wheel-packaging marker check and the
constant-condition scan — the last with a `SyntaxError` raised from inside a
vendored SDK, which reads as a product defect that does not exist.

`tests/shared/product_sources.py` is now the single enumerator; it skips caches,
dot-directories and third-party trees. All four scanners use it, including the
new border guard, which had started life with its own private copy.

Both guards mutation-checked — each planted regression fails its own test and
nothing else:

```
$ # shell agent builder returns to production code
FAILED test_no_shipped_module_builds_a_second_shell_agent

$ # handler.run(...) renamed
FAILED test_the_shell_turn_path_runs_the_turn_host

$ # host import dropped entirely
FAILED test_the_shell_turn_path_runs_the_turn_host

$ # blank line removed from render_response_header
FAILED test_response_header_opens_with_a_blank_line
```

And the scanners still bite rather than going quiet — the risk with any
exclusion is that it disables the guard instead of narrowing it:

```
$ # planted `import psutil` in core/execution.py
FAILED test_psutil_is_not_imported_outside_sanctioned_modules

$ # planted `if False and _planted:` in core/execution.py
FAILED test_product_packages_have_no_constant_condition_toggles[core]

$ # planted config/planted_pkg/mod.py with no __init__.py
FAILED test_every_shipped_module_directory_is_an_importable_package[config]
```

Both guards came out of asking what this branch's changes would look like if
someone quietly undid them. The answer for four separate behaviours was "the
suite stays green", which is the actual defect being fixed here.

The border guard checks the *claim*, not the spelling. Aliasing the import
(`import TurnHandler as _H`) correctly passes — the module still imports the
host — while renaming `handler.run(...)` or dropping the import fails. A regex
over the file would have got that backwards.

For the scanner fix the alternative was to delete `config/.venv`, which is what
the failures superficially call for. Rejected: it makes today green and leaves
the trap armed for whoever next creates an environment inside a package. The
scanners were wrong to read it, so the exclusion belongs in the enumerator. The
danger with that direction is an exclusion so broad it silently switches the
guards off, which is why each one is re-checked against a planted defect above
rather than just observed to be green.

`build_shell_agent` is deliberately left in place — tests legitimately want an
agent without standing up a host. The rule is that nothing *shipped* may call
it, which is what the empty, exactly-compared allowlist expresses.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5269, opened 2026-08-20
fix(github): stop using the search API for issue-mutation idempotency checks

arnavpopley:
Part of #5121 (covers one of the write-capable tools; the remaining ci_fix/security_fix/github_cli tools are not addressed by this PR, so it should not auto-close that issue)

Found while live-verifying \`execute_github_issue_mutation\` against a real repo (a fork I control, Issues enabled, real token). This tool's duplicate-prevention checks — before creating an issue from a proposal, and before posting a duplicate follow-up comment — both queried GitHub's \`/search/issues\` endpoint for the proposal's idempotency marker.

Reproduced live: I generated one proposal, then called \`execute_github_issue_mutation\` with it three times back-to-back. Instead of the second and third calls detecting the first issue already existed, each one created a **new duplicate issue** (issues #3, #4, #5 in my test repo, all with the identical idempotency marker in their body). GitHub's search index has an observed propagation lag after a write — confirmed directly: right after creating an issue, \`/search/issues\` for its exact marker returned 0 results; waiting ~8 seconds, it found it correctly.

Both dedup checks now use the immediately-consistent list endpoints instead of search:
- Create-dedup: the most recent page of \`GET /repos/{owner}/{repo}/issues\` (sorted \`created\`, \`desc\`) scanned client-side for the marker. Bounded to one page (100 issues) since an idempotent retry lands within moments of the original, not months later.
- Comment-dedup: \`GET /repos/{owner}/{repo}/issues/{number}/comments\` (bounded to the single issue) scanned client-side for the marker.

This is a real improvement, not a complete guarantee: re-tested the same three-calls-in-a-row scenario after the fix with a 3-second gap between calls and it correctly detected the existing issue every time. A literally zero-delay double call (which no real agent tool-calling loop produces — LLM inference and network round trips alone take well over that) can still theoretically race, since GitHub's REST API doesn't offer a strongly-consistent read-after-write guarantee for this resource. The fix narrows the practical race window from "GitHub's search index catching up" (multi-second-plus, observed) down to "GitHub's ordinary list-API propagation" (observed well under 3 seconds), which covers any realistic calling pattern.

Before the fix, three back-to-back calls with the identical proposal, real repo:
\`\`\`
first execute:  created_github_issue  issue 3
second execute: created_github_issue  issue 4   <- should have been "existing_github_issue"
third execute:  created_github_issue  issue 5   <- should have been "existing_github_issue"
\`\`\`

After the fix, same scenario with a realistic ~3s gap:
\`\`\`
first execute:                created_github_issue   issue 6
second execute (3s later):    existing_github_issue   executed: False   issue 6
\`\`\`

Regression test proof: reverted the fix locally and reran the affected tests — 6 failed (5 existing tests that had asserted the old \`/search/issues\` call sequence, plus 1 new test asserting \`/search/issues\` is never called), all failing with either an unexpected-path \`AssertionError\` or the explicit "must not use the eventually-consistent search API" message. Restored the fix and confirmed all green.

Local checks run: \`make lint\`, \`make format-check\`, \`make typecheck\` (all clean), \`tests/tools/test_github_workflow_tools.py\` (49 passed) and the full \`integrations/github/tools/\` scoped suite per \`.github/ci/test_scope_rules.py\` (183 passed).

---

Found this by actually exercising \`execute_github_issue_mutation\`'s idempotency guarantee against a real repo rather than trusting the existing mocked tests, which all assumed \`/search/issues\` responds instantly and consistently. I reproduced the duplicate-creation bug live, then confirmed the root cause by directly querying \`/search/issues\` for a just-created issue's marker (0 results) and again a few seconds later (found it) — that isolated the propagation lag as the cause rather than a logic bug in the marker-matching code itself. The fix swaps both search calls for the corresponding list endpoints, which reflect writes immediately, bounded appropriately for each check's actual scope (a page of recent issues for create-dedup; one issue's own comments for comment-dedup) so the fix doesn't trade one performance problem for another. I updated the five existing tests that had encoded the old \`/search/issues\` call sequence as their expected behavior, and added one explicit regression test asserting the search endpoint is never called at all.

---

---

### DIFF
--- a/integrations/github/tools/work_status.py
+++ b/integrations/github/tools/work_status.py
@@ -550,32 +550,40 @@ def _validate_proposal_marker(proposal: GitHubIssueMutationProposal) -> str | No
     return None
 
 
-def _quoted_search_term(term: str) -> str:
-    cleaned = term.replace('"', "")
-    return f'"{cleaned}"'
+# One page of the most-recent issues is enough to catch an idempotent retry,
+# which lands within moments of the original creation, not months later.
+_RECENT_ISSUES_SCAN_LIMIT = 100
 
 
-def _search_issues_for_marker(
+def _recent_issues_with_marker(
     client: GitHubRestClient,
     *,
     owner: str,
     repo: str,
     marker: str,
-    search_area: Literal["body", "comments"],
 ) -> list[dict[str, Any]]:
+    """Recently-created issues whose body contains ``marker``.
+
+    Uses the issues list endpoint rather than /search/issues: GitHub's search
+    index has an observed multi-second propagation lag after issue creation,
+    which lets a rapid idempotent retry slip past a search-based dedup check
+    and create a duplicate issue (reproduced live).
+    """
     result = client.request(
         "GET",
-        "/search/issues",
+        f"/repos/{owner}/{repo}/issues",
         params={
-            "q": (f"repo:{owner}/{repo} is:issue in:{search_area} {_quoted_search_term(marker)}")
+            "state": "all",
+            "per_page": _RECENT_ISSUES_SCAN_LIMIT,
+            "sort": "created",
+            "direction": "desc",
         },
     )
-    if not isinstance(result, dict):
+    if not isinstance(result, list):
         return []
-    items = result.get("items")
-    if not isinstance(items, list):
-        return []
-    return [item for item in items if isinstance(item, dict)]
+    return [
+        item for item in result if isinstance(item, dict) and marker in str(item.get("body") or "")
+    ]
 
 
 def _marker_exists_on_issue(
@@ -586,16 +594,20 @@ def _marker_exists_on_issue(
     issue_number: int,
     marker: str,
 ) -> bool:
-    return any(
-        item.get("number") == issue_number
-        for item in _search_issues_for_marker(
-            client,
-            owner=owner,
-            repo=repo,
-            marker=marker,
-            search_area="comments",
-        )
+    """Whether a comment containing ``marker`` already exists on the issue.
+
+    Lists the issue's own comments (bounded to one issue, immediately
+    consistent) rather than the repo-wide search API, for the same reason as
+    :func:`_recent_issues_with_marker`.
+    """
+    result = client.request(
+        "GET",
+        f"/repos/{owner}/{repo}/issues/{issue_number}/comments",
+        params={"per_page": 100},
     )
+    if not isinstance(result, list):
+        return False
+    return any(isinstance(item, dict) and marker in str(item.get("body") or "") for item in result)
 
 
 @tool(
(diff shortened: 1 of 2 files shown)


integrations/github/tools/work_status.py
```diff
@@ -586,16 +594,20 @@ def _marker_exists_on_issue(
...
+
+    Lists the issue's own comments (bounded to one issue, immediately
+    consistent) rather than the repo-wide search API, for the same reason as
+    :func:`_recent_issues_with_marker`.
+    """
+    result = client.request(
+        "GET",
+        f"/repos/{owner}/{repo}/issues/{issue_number}/comments",
+        params={"per_page": 100},
     )
```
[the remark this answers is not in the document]
arnavpopley: Resolved

arnavpopley:
@greptile review

integrations/github/tools/work_status.py
```diff
@@ -550,32 +551,40 @@ def _validate_proposal_marker(proposal: GitHubIssueMutationProposal) -> str | No
...
+
+    Uses the issues list endpoint rather than /search/issues: GitHub's search
+    index has an observed multi-second propagation lag after issue creation,
+    which lets a rapid idempotent retry slip past a search-based dedup check
+    and create a duplicate issue (reproduced live).
+    """
     result = client.request(
         "GET",
-        "/search/issues",
+        f"/repos/{owner}/{repo}/issues",
```
muddlebee, line 575: why are we change the API spec here? any specific reason?

muddlebee: image

I get this why changed to the new API?? where the owner and repo will be passed from? because in CLI the user will be posting the link to github directly afaik

arnavpopley: Good question — answering both:

**Why the endpoint changed:** `/search/issues` has an observed multi-second eventual-consistency lag after an issue is created (confirmed live: a rapid idempotent retry queried search before the new issue was indexed, got 0 results, and created a duplicate). `GET /repos/{owner}/{repo}/issues?sort=created&direction=desc` hits the primary data store directly, so a retry landing moments after the original creation reliably sees it. This is scoped to the create-path dedup check only.

**Where owner/repo come from:** unchanged by this PR — `execute_github_issue_mutation(owner: str, repo: str, ...)` already required them as explicit top-level params before this change, resolved the same way for both the old and new code path: either injected from the configured GitHub integration source (`_github_extract_params`), or supplied explicitly by the calling agent/LLM as a normal tool-call argument (e.g. parsed out of a pasted GitHub URL). The old `_search_issues_for_marker` needed the identical two strings to build its search query (`repo:{owner}/{repo} ...`), so this refactor doesn't add any new dependency on how owner/repo get resolved.

arnavpopley: From OpenSRE Development in Slack:
IMG_5553

PR #5272, opened 2026-08-20
Fix/prompts the planner stops asking questions

### LINKED ISSUE #5268: github_cli returns unbounded stdout, so a workflow-runs query is silently truncated and answered from one run
### What happens

```
```

The reply leads with a bold headline:

> **The current observable GitHub Actions failure rate is 0%**: the latest
> completed run succeeded.

| Completed runs observed | Failed | Failure rate |
|---|---|---|
| 1 | 0 | 0% |

> This is a point-in-time snapshot … because the returned 30-run payload was
> truncated before all conclusions were available.

**The 0% is computed from one run out of thirty.** The caveat is accurate, but
the headline is a confident percentage from n=1 and the table lends it authority.
A user skims the number, not the footnote.

### The chain

1. The skill blocks the compact command. `github_cli` denies the top-level
   `run` verb (`integrations/github/tools/github_cli/runner.py`,
   `_DENIED_TOP_LEVEL_COMMANDS`) because `gh run rerun` can trigger CI. That
   also blocks the read-only `gh run list --json`.
2. So the agent uses `gh api`, which the skill explicitly allows.
3. `gh api /repos/OWNER/REPO/actions/runs?per_page=30` returns **382,042
   characters** — measured against this repo.
4. `github_cli` does not cap stdout. The full payload goes into the message.
5. `core/context_budget.py` truncates the largest message to fit the window,
   appending `…[truncated to fit context budget]`. The agent sees roughly the
   first few percent — one run's conclusion.
6. It computes a rate from what survived and reports it as the answer.

Measured payloads for the same data:

| Command | Characters |
|---|---|
| `gh run list --limit 30 --json databaseId,conclusion,status,name,createdAt` | 4,075 |
| same, 11 fields | 14,029 |
| `gh api /repos/…/actions/runs?per_page=30` | **382,042** |

The compact form is 90× smaller and is the one the tool forbids.

### Why this is more than one bad answer

Two general defects, both visible here:

- **Unbounded tool stdout.** Any `gh api` list endpoint can push hundreds of
  kilobytes into the context. `shell_run` caps at
  `MAX_COMMAND_OUTPUT_CHARS = 24_000` and sets `truncated` on its payload;
  `github_cli` does neither, so the cut happens later, silently, and the result
  carries no flag saying it happened.
- **The block is inconsistent with its own rationale.** The comment says the
  concern is triggering CI ("arbitrary code via workflow YAML"). `gh run list`
  and `gh run view` trigger nothing, and the same data is reachable through the
  allowed `gh api` route — at 90× the size. The blocklist is stopping the safe,
  cheap read while permitting the expensive one.

### Suggested fix

1. Cap `github_cli` stdout the way `shell_run` does and mark the payload
   `truncated: true`, so a partial read is visible to the agent as data rather
   than inferred from a missing tail.
2. Give the agent a compact path for workflow runs — either allow the read-only
   `run` subcommands (`list`, `view`) while keeping `rerun` / `cancel` /
   `delete` blocked, or steer `gh api` list endpoints through `--jq` projection
   (`--jq` is already an accepted flag).
3. Skill guidance: do not report a rate or percentage from a payload marked
   truncated — say what could not be read.

### Repro

```
uv run opensre
```

YauhenBichel:
Issue: #5268

Three defects found by running `what's the current error rate on github?` in the
REPL after #5262. Each was visible only in the terminal; the suite was green
throughout.

**1. A rate computed from one run.** The answer led with a bold **0% failure
rate** built on a single sample.

`github_cli` blocks the top-level `run` verb — `gh run rerun` can trigger CI —
which also blocks the read-only `gh run list --json`. So the agent used `gh api`,
which the skill explicitly allows. Measured against this repo:

| Command | Characters |
|---|---|
| `gh run list --limit 30 --json databaseId,conclusion,status,name,createdAt` | 4,075 |
| same, 11 fields | 14,029 |
| `gh api /repos/…/actions/runs?per_page=30` | **382,042** |

`github_cli` returned stdout uncapped, so the full payload reached the message
and `core/context_budget.py` trimmed the tail silently — the agent read one
conclusion and had no way to know the rest existed.

It now caps at 24,000 characters, matching `shell_run`, and sets `truncated` on
the payload so a partial read is data rather than an inference. The skill tells
the agent to project `gh api` list endpoints with `--jq`, and never to report a
rate, total or "latest" from a truncated result.

**2. Two contradictory answers in one turn.** The same question produced
**3.03%** under `OpenSRE` (3 of 99, latest 100 runs) and then **5.14%** under
`assistant` (36 of 700, 24-hour window). Both presented as the current rate.

The action phase decides whether to paint its closing from the *planner's*
handoffs. The orchestrator then appends its own guidance tags — an evidence-tier
tag, and `session_goal:continue` while a goal is active — and handed the
augmented tuple to the router. `route_turn` reads a non-empty tuple as "the
planner asked the assistant to answer", so a handled turn became a gather turn
*after* its answer was already on screen:

```
action handled, no handoff        -> handled_without_llm
same + injected session_goal tag  -> gather_and_answer
```

Routing now keys off the planner's handoffs. The injected tags still reach the
assistant, where they were always meant to go.

The existing comment shows this was hit once before and patched narrowly —
"injecting here previously forced gather_and_answer on `/goal set` itself
(duplicate PostHog turn)" — carved out for literal slashes only, which left this
path open.

Defects 1 and 2 are structural and mutation-tested. Defect 3 is prompt-level and
therefore weaker — worth saying plainly: it changed behaviour in the right
direction, but it removed the question without pinning the default, so the model
then chose a one-hour window that contained no runs and answered "N/A". The
honesty is an improvement over inventing a number, but a follow-up should pin the
default window rather than leave it to per-turn choice. That is deliberately not
in this PR.

Two things I deliberately did not do. I did not loosen the `run` blocklist, even
though it is the root of defect 1 — it stops the cheap safe read (`gh run list`,
4KB) while permitting the expensive one (`gh api`, 382KB), which is worth fixing,
but it is a security surface and belongs in its own change. And I did not add
truncation guidance to the tool-level `SKILL.md`: it has 86 characters of
headroom against a 2,400 budget, and my first attempt pushed the rendered
guidance to 2,565, which `test_skill_guidance_attaches_to_github_cli` caught by
asserting the text does not end in `...`. The guidance lives in the action-agent
skill instead, which has no cap.


YauhenBichel:
@greptile review

PR #5275, opened 2026-08-20
fix(clickhouse): stop excluding failed queries from query activity

arnavpopley:
Part of #5111 (found while live-verifying ClickHouse's 2 registered tools)

Found while live-verifying \`get_clickhouse_query_activity\` against a real ClickHouse instance. The underlying \`get_query_activity\` queried \`system.query_log\` with \`WHERE type = 'QueryFinish'\`, which only matches successful completions. ClickHouse logs a failed query as \`ExceptionBeforeStart\` or \`ExceptionWhileProcessing\` — never \`QueryFinish\` — so the tool could never surface a failed query, no matter how recent or how large the \`limit\`. \`docs/clickhouse.mdx\` documents the tool as returning "recent slow / failed queries," and its own \`use_cases\` say "Identifying slow or resource-heavy queries during an incident" — the exact scenario where seeing a failure matters most was structurally unreachable.

Reproduced live: ran \`SELECT * FROM nonexistent_table_verify_test\` against a real ClickHouse container, flushed logs, and called the tool directly — the failed query never appeared at any limit. Confirmed the actual logged \`type\` via \`system.query_log\` directly: \`ExceptionBeforeStart\`.

Before the fix, direct call against the real instance after the failing query ran:
```
types seen: {'QueryFinish'}
failed query present: False
```

After the fix, same instance, same failed query still in the log:
```
types seen: {'ExceptionBeforeStart', 'QueryFinish'}
failed query present: True
```

Regression test proof: reverted the fix locally and reran the new test — failed with `assert "type = 'QueryFinish'" not in sql` showing the old literal string still present, then restored the fix and confirmed green.

Local checks run: `make lint`, `make format-check`, `make typecheck` (all clean), `tests/integrations/test_clickhouse.py` + `tests/tools/test_clickhouse_query_activity_tool.py` + `tests/tools/test_clickhouse_system_health_tool.py` (46 passed).

---

Found this by exercising both ClickHouse tools against a real, seeded ClickHouse instance rather than trusting the existing mocked tests, which mock `get_query_activity` entirely and never touch the actual SQL string — exactly why this bug had no coverage. I confirmed the real `type` value ClickHouse assigns to a failed query by querying `system.query_log` directly with `curl`, rather than guessing at ClickHouse's schema. The fix changes the filter from an allowlist of one success type to a denylist of the one row kind that should genuinely be excluded (`QueryStart`, which precedes the query's own duration/result being known), which is also more future-proof against ClickHouse adding further terminal-state types later.

---

---

### DIFF
--- a/integrations/clickhouse/__init__.py
+++ b/integrations/clickhouse/__init__.py
@@ -166,7 +166,14 @@ def get_query_activity(
 ) -> dict[str, Any]:
     """Retrieve recent query activity from system.query_log.
 
-    Read-only: queries system.query_log for recent completed queries.
+    Read-only: queries system.query_log for recent completed *and* failed
+    queries (``type != 'QueryStart'``: ClickHouse logs a failed query as
+    ``ExceptionBeforeStart``/``ExceptionWhileProcessing``, distinct from the
+    successful ``QueryFinish`` row -- a bare ``type = 'QueryFinish'`` filter
+    silently excludes every failure, which defeats the point of surfacing
+    "recent slow / failed queries" during an incident). ``QueryStart`` rows
+    are excluded since they precede the query's own duration/result being
+    known and would otherwise duplicate its later QueryFinish/Exception row.
     Results capped at config.max_results.
     """
     if not config.is_configured:
@@ -188,7 +195,7 @@ def get_query_activity(
                 "  memory_usage, "
                 "  event_time "
                 "FROM system.query_log "
-                "WHERE type = 'QueryFinish' "
+                "WHERE type != 'QueryStart' "
                 "ORDER BY event_time DESC "
                 "LIMIT %(limit)s",
                 parameters={"limit": effective_limit},
--- a/integrations/clickhouse/__init__.py
+++ b/integrations/clickhouse/__init__.py
@@ -166,15 +166,9 @@ def get_query_activity(
 ) -> dict[str, Any]:
     """Retrieve recent query activity from system.query_log.
 
-    Read-only: queries system.query_log for recent completed *and* failed
-    queries (``type != 'QueryStart'``: ClickHouse logs a failed query as
-    ``ExceptionBeforeStart``/``ExceptionWhileProcessing``, distinct from the
-    successful ``QueryFinish`` row -- a bare ``type = 'QueryFinish'`` filter
-    silently excludes every failure, which defeats the point of surfacing
-    "recent slow / failed queries" during an incident). ``QueryStart`` rows
-    are excluded since they precede the query's own duration/result being
-    known and would otherwise duplicate its later QueryFinish/Exception row.
-    Results capped at config.max_results.
+    Read-only: queries system.query_log for recent completed and failed
+    queries, excluding QueryStart records whose duration and result are not
+    yet known. Results capped at config.max_results.
     """
     if not config.is_configured:
         return tool_unavailable("clickhouse", "Not configured.")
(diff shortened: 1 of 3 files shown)


integrations/clickhouse/__init__.py
```diff
@@ -166,7 +166,14 @@ def get_query_activity(
...
 
-    Read-only: queries system.query_log for recent completed queries.
+    Read-only: queries system.query_log for recent completed *and* failed
+    queries (``type != 'QueryStart'``: ClickHouse logs a failed query as
+    ``ExceptionBeforeStart``/``ExceptionWhileProcessing``, distinct from the
+    successful ``QueryFinish`` row -- a bare ``type = 'QueryFinish'`` filter
+    silently excludes every failure, which defeats the point of surfacing
+    "recent slow / failed queries" during an incident). ``QueryStart`` rows
+    are excluded since they precede the query's own duration/result being
+    known and would otherwise duplicate its later QueryFinish/Exception row.
```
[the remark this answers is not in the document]
arnavpopley: Resolved

integrations/clickhouse/__init__.py
```diff
@@ -188,7 +195,7 @@ def get_query_activity(
                 "  memory_usage, "
                 "  event_time "
                 "FROM system.query_log "
-                "WHERE type = 'QueryFinish' "
+                "WHERE type != 'QueryStart' "
```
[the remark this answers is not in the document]
arnavpopley: Resolved

arnavpopley:
@greptile review

integrations/clickhouse/__init__.py
```diff
@@ -188,7 +189,7 @@ def get_query_activity(
...
```
muddlebee, line 192: is this the right keyword as per the API specs?

muddlebee:
@arnavpopley can you share the demo? also you used docker container for the ClickHouse instance?

integrations/clickhouse/__init__.py
arnavpopley: Yes — `system.query_log.type` (ClickHouse's own enum) has exactly four values: `QueryStart`, `QueryFinish`, `ExceptionBeforeStart`, `ExceptionWhileProcessing`. `type != 'QueryStart'` is the correct predicate for "completed or failed" — it includes `QueryFinish` plus both exception states, excluding only the in-progress row. The old `type = 'QueryFinish'` silently dropped both failure states, which was the bug. Confirmed live against a running ClickHouse instance (a query that raised an exception produced an `ExceptionWhileProcessing` row, correctly returned by the new predicate and missed by the old one).

arnavpopley:
@muddlebee 
image

arnavpopley:
From OpenSRE Bot in Slack:

### :white_check_mark: PR #5275 — fix(clickhouse): stop excluding failed queries from query activity
**Size:** +54/-4 · **Author:** arnavpopley | **Status:** Open

**What it does:**
get_clickhouse_query_activity queried system.query_log with WHERE type = 'QueryFinish' — which **only matches successful completions**. ClickHouse logs failed queries under a different type value, meaning every failed query was completely invisible to the investigation pipeline. Found during live verification of ClickHouse's 2 registered tools.

**Why it matters:** Failed queries are exactly what you want during an incident. Excluding them is the worst possible default for an observability tool.

**Things to verify:**
- The schema handles failed query rows cleanly (exception messages, error codes surfaced correctly in output)
- Consider a result cap if a noisy ClickHouse instance produces large volumes of failures

**Verdict:** :white_check_mark: **Merge** — meaningful correctness fix, low risk, good live-test evidence.

integrations/clickhouse/__init__.py
arnavpopley: Addressed

PR #5279, opened 2026-08-20
fix(github): windowed run listings, counted rates, and continuations that reconcile

### LINKED ISSUE #5281: fix(github): windowed run listings, counted rates, and continuations that reconcile
1. A failure rate over a window is computed by paging runs, so it times out or comes back biased

```
> what's the github actions failure rate over the last 7 days?
```

> The exact GitHub Actions failure rate for Tracer-Cloud/opensre over the last 7
> days could not be calculated reliably. The complete paginated query timed out,
> while the GitHub tools returned only the first 100 runs for each query, so
> using those pages would produce a biased rate.

The refusal is correct — a rate from the first 100 of several thousand runs is
biased. But the question is answerable, so refusing is the wrong outcome.

Earlier the same question over 24 hours produced a rate from **one** run:

> **The current observable GitHub Actions failure rate is 0%** — the latest
> completed run succeeded.

### Cause

The agent answers a rate by enumerating runs. This repo runs roughly 660
workflows a day, so a 7-day window holds about 4,600 — enumeration either times
out or stops after a page and reports whatever fraction it saw.

Nothing pointed the agent at the aggregate path. The reply even found the totals
and correctly rejected them:

> Repository totals shown by GitHub: 3,438 failure-filtered matches and 40,000
> total matches, but these counts are not confirmed as restricted to the 7-day
> window and therefore cannot be used as the rate denominator.

They can be restricted. `created` scopes `total_count` server-side.

### Measured

```
gh api ".../actions/runs?created=%3E%3D2026-08-13&per_page=1" --jq '.total_count'
4646
gh api ".../actions/runs?created=%3E%3D2026-08-13&status=failure&per_page=1" --jq '.total_count'
94
```

Two requests, `per_page=1`, exact window-scoped counts, no pagination. 2.02%.

2. A session-goal continuation reports a different number than the turn before it

```
> how many github actions runs failed in the last 24 hours?
```

**Turn 1/5** — `73 GitHub Actions runs failed … out of 800 completed runs.`
Window stated. 3.7s, no tool calls.

**Turn 2/5** — `12 GitHub Actions runs failed …` No window, no denominator.
Goal closes: `achieved with tool evidence`.

Two counts for one question, one turn apart, nothing reconciling them. The
contradictory one is the last word and the weaker of the two.

### Cause

Turn 1 answered from conversation history — no tool ran. `turn_has_session_goal_evidence`
requires a successful tool call:

> "True when the turn ran a tool **successfully** — not prose, not a claim."

Two things follow. The goal cannot close on turn 1, which is correct. And turn
1's answer is **not recorded as a finding**, because `with_finding` sits behind
the same gate (`run_until.py:157`) — which is not.

So turn 2 ran the tool, got a number under a different definition of "failed",
and never learned that 73 had already been reported. The carry-forward channel
that exists for exactly this — *"Already established in earlier turns of this
goal"* — was skipped.

73 vs 12 is almost certainly strict `conclusion == "failure"` against the
broader error set including `timed_out` / `startup_failure` / `action_required`.
Neither turn stated its definition.

### Fix

Evidence should gate **closing** the goal and what counts as established. It
must not gate whether the next turn knows what this one said.

Note `findings` is the wrong channel to widen: it means "established" and the
continuation says *"treat these as done"*. An unverified prose answer given that
authority lets a wrong remembered number become settled by repetition. A
separate weaker field, rendered as "already told the user — if your answer
differs, say why", carries the reconciliation without the authority.

3. Workflow-run listings cannot express a time window, so the agent invents one per turn

```
```

> Completed GitHub Actions runs: **0** · Failed runs: **0**
> Error rate: **N/A** — no completed runs in the window, so there is no denominator.

The goal then closed as achieved with three turns unused. A 24-hour window had
an answer available; the agent chose the last hour, which was empty.

### Cause

No tool understood a time window. `list_github_actions_workflow_runs` took only
`per_page: 30`, and was exposed on `(INVESTIGATION, CHAT)` — not `ACTION` — so
the agent that ran could not call it. It built its own `gh api` query and picked
a window fresh each turn.

### Fix

`window_hours` on the listing, plus `ToolSerface.ACTION` so the action agent can
reach it.

**The window must default to 0 (no window), not 24.** Defaulting to 24 broke
`test_list_workflow_runs_happy_path`: the fixture run is dated `2026-05-27` and
the default filtered it away. That is not a stale fixture — this listing also
answers *"which deploy failed right before the incident"*, where a run days old
is the point. A default window would hide it from every caller that never asked
for one.

Also report `window_fully_fetched`: true only when a run *older* than the window
came back, proving the page spanned it. A count over a partial page is the "0%
from one run" failure in a new costume. And count runs whose `created_at` cannot
be read rather than dropping them silently from a total.

YauhenBichel:
Fixes: #5281

Three defects, all found by running `what's the current error rate on github?`
and its variants in the REPL. Each was invisible to a green suite.

**1. Run listings could not express a time window** (`e541b5432`, `473cfdacf`).
`list_github_actions_workflow_runs` took only `per_page: 30` and was not on the
`ACTION` surface, so the agent hand-built `gh api` and chose a window per turn.
One run it picked the last hour, found nothing, and answered `N/A`.

Adds `window_hours`, exposes the tool to the action surface, and reports
`window_fully_fetched` — true only when a run *older* than the window came back,
proving the page spanned it. Runs with unreadable timestamps are counted in
`undated` rather than vanishing from a total.

**The window defaults to 0, not 24.** A 24-hour default broke
`test_list_workflow_runs_happy_path` by dropping a `2026-05-27` fixture run —
the same way it would hide the run an investigation was hunting for, since this
listing also answers "which deploy failed right before the incident".
`RATE_WINDOW_HOURS = 24` is what rate callers pass.

**2. Rates were computed by enumeration** (`2eb2719b8`). A 7-day rate needs
~4,600 runs; paging timed out and a partial page gives a biased number. Skill
guidance now says a rate is two `total_count` reads with `created` scoping the
window server-side:

```
gh api ".../actions/runs?created=%3E%3D2026-08-13&per_page=1" --jq '.total_count'   -> 4646
gh api ".../actions/runs?created=%3E%3D2026-08-13&status=failure&per_page=1" --jq '.total_count'  -> 94
```

Measured against this repo. Listing *which* runs failed still enumerates — that
is a different question.

**3. A continuation turn contradicted the turn before it** (`2eb2719b8`).
"73 failed out of 800" on turn 1, "12 failed" on turn 2, nothing reconciling
them. Turn 1 answered from history with no tool, so `turn_has_session_goal_evidence`
was false — which correctly stopped the goal closing, but also skipped recording
what turn 1 said.

Adds `SessionGoal.last_answer`, recorded whether or not a tool ran, rendered to
the next turn as "already told the user — if your answer differs, say why".

`findings` is deliberately left gated. It means "established" and the
continuation says *"treat these as done"*; giving an unverified prose answer that
authority would let a wrong remembered number become settled by repetition.

Verified by hand across four questions: 24-hour rate (one answer, window
stated), 7-day rate (previously an honest refusal — recheck with the counting
guidance), "how many failed" (previously 73 then 12), and "which runs failed
before the deploy on May 27" — the forensics check, which returns three-month-old
runs and reconciles with an earlier answer once the differing cutoffs are
accounted for.

Mutation-checked; each planted regression fails exactly its own test:

```
$ # rate window drops back to an hour
FAILED test_the_rate_window_is_a_day_not_an_hour
$ # coverage always claims the page spanned the window
FAILED test_a_page_that_ends_inside_the_window_is_not_fully_fetched
$ # unreadable timestamps counted as in-window
FAILED test_undated_runs_are_counted_rather_than_vanishing
$ # last_answer gated on tool evidence again
FAILED test_a_tool_less_turn_still_hands_its_answer_to_the_next_turn
```

Gates: lint clean · format 3,590 · typecheck 1,855 · all 3 import checks ·
**15,521 passed** on the previous commit; full suite for the final commit
running.

> The suite run excludes `tests/core/tool/test_enums.py`, which collides by
> basename with `tests/filestorage/test_enums.py` (neither directory has
> `__init__.py`). That arrived with #5259 and reproduces with this branch
> stashed — `main` currently cannot collect a full run. Unrelated, but worth its
> own fix.

### DIFF
--- a/core/agent_harness/prompts/skills/github_cli/SKILL.md
+++ b/core/agent_harness/prompts/skills/github_cli/SKILL.md
@@ -42,9 +42,10 @@ HARD RULES:
   `list_github_actions_workflow_runs` instead of building a `gh api` query, and
   pass `window_hours=24` unless the user named a window. The tool defaults to no
   window because it also answers "which deploy failed before the incident".
-  It reports `window_covered`: when false the page ended inside the window, so
-  say the count is partial or re-ask with a larger `per_page` — never publish a
-  rate over it.
+  It reports `window_fully_fetched`: when false the page ended inside the
+  window, so counts are a floor — say so or re-ask with a larger `per_page`,
+  never publish a rate over it. `undated_runs` counts runs whose timestamp could
+  not be read; mention them rather than letting them vanish from a total.
 - With `gh api` on a list endpoint, always project the fields you need with
   `--jq`. Raw list payloads are hundreds of kilobytes (30 workflow runs is
   ~382k characters) and get cut to fit the context, leaving you a few records
--- a/core/agent_harness/prompts/skills/github_cli/SKILL.md
+++ b/core/agent_harness/prompts/skills/github_cli/SKILL.md
@@ -38,10 +38,22 @@ HARD RULES:
   codespace / ssh-key / gpg-key / config — those are blocked (token leakage /
   CI code execution / secret mutation).
 - Pass args after the gh binary; optional repo as owner/name → -R.
-- For workflow-run counts or failure rates, call
-  `list_github_actions_workflow_runs` instead of building a `gh api` query, and
-  pass `window_hours=24` unless the user named a window. The tool defaults to no
-  window because it also answers "which deploy failed before the incident".
+- A failure RATE over a window is two counts, not a list of runs. Ask GitHub to
+  count, with `created` scoping the window server-side — never page through runs
+  to compute a rate. A busy repo runs thousands per week; enumeration times out
+  and a partial page gives a biased number.
+
+      gh api "/repos/OWNER/REPO/actions/runs?created=%3E%3D2026-08-13&per_page=1" --jq '.total_count'
+      gh api "/repos/OWNER/REPO/actions/runs?created=%3E%3D2026-08-13&status=failure&per_page=1" --jq '.total_count'
+
+  `per_page=1` because only `total_count` is read. Report the window you passed
+  and which conclusions `status=failure` covers — a strict failure count and a
+  broader error count are different numbers and must not be swapped between
+  turns.
+- To LIST which runs failed (not a rate), call
+  `list_github_actions_workflow_runs`, and pass `window_hours=24` unless the
+  user named a window. The tool defaults to no window because it also answers
+  "which deploy failed before the incident".
   It reports `window_fully_fetched`: when false the page ended inside the
   window, so counts are a floor — say so or re-ask with a larger `per_page`,
   never publish a rate over it. `undated_runs` counts runs whose timestamp could
--- a/core/agent_harness/prompts/skills/github_cli/SKILL.md
+++ b/core/agent_harness/prompts/skills/github_cli/SKILL.md
@@ -38,6 +38,13 @@ HARD RULES:
   codespace / ssh-key / gpg-key / config — those are blocked (token leakage /
   CI code execution / secret mutation).
 - Pass args after the gh binary; optional repo as owner/name → -R.
+- For workflow-run counts or failure rates, call
+  `list_github_actions_workflow_runs` instead of building a `gh api` query, and
+  pass `window_hours=24` unless the user named a window. The tool defaults to no
+  window because it also answers "which deploy failed before the incident".
+  It reports `window_covered`: when false the page ended inside the window, so
+  say the count is partial or re-ask with a larger `per_page` — never publish a
+  rate over it.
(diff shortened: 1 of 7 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
image

Screenshot 2026-08-20 at 4 09 46 PM

Screenshot 2026-08-20 at 4 05 24 PM

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5284, opened 2026-08-20
refactor(tool): measure the tool-tier import border and shrink it through one door

YauhenBichel:
**The measurement.** `core/tool/` and `core/tool_framework/` are one tier to
everything above them, and nothing held their border. Consumers reached into
**24 internal paths** across five packages.

`tests/shared/api_border.py` is the scanner, extracted from `harness_api.py` and
parameterised on (packages, API modules). That file was 150 lines of exactly this
logic bound to one tier; duplicating it for a second tier would have been the
wrong move. `harness_api` keeps its constants and public function names and now
delegates — its **21 border tests pass unchanged**, before and after.

Allowlists are compared exactly in both directions, so they can only shrink: a
new internal import fails, and an entry no longer imported must be removed.

**The shrink.** `core/tool_framework/utils/__init__.py` was *already* a door,
curating `__all__` over its submodules — just six names wide, so callers reached
past it to nine modules. Widened to fourteen and moved 124 files onto it.

| Consumer | Before | After |
|---|---|---|
| integrations | 11 | **4** |
| tools | 7 | **6** |
| surfaces | 3 | **2** |
| platform | 2 | 2 |
| gateway | 1 | 1 |
| **total** | **24** | **15** |

No file moved and no symbol was renamed. Of 129 changed files, five are
substantive; the other 124 are single import lines.

Direction verified before widening: `core.tool_framework` depends on `core.tool`
and never the reverse, and the pulled-in submodules only reach sideways to
`tool_availability` and `db_warnings`, both already exported. No cycle —
`AGENTS.md` is explicit that CodeQL counts even a `TYPE_CHECKING` import as part
of one.

### DIFF
--- a/core/tool_framework/utils/__init__.py
+++ b/core/tool_framework/utils/__init__.py
@@ -1,8 +1,23 @@
-"""Tool utilities — code-host helpers, data validation, and database warnings."""
+"""Tool utilities — the one door consumers import these helpers through.
+
+Schema builders, MCP payload readers, code-host and availability envelopes, and
+database-warning wrappers. Importing the submodules directly still works inside
+this package; everything above the tool tier comes through here so the split
+below can change without a sweep across ``integrations/`` and ``tools/``.
+"""
 
 from core.tool_framework.utils.code_host_unavailable import code_host_unavailable_payload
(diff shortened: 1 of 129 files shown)


YauhenBichel:
@greptile review

PR #5290, opened 2026-08-20
docs(opensearch): add a local verification recipe for both search tools

arnavpopley:
Part of #5119
Part of #5141

#### Describe the changes you have made in this PR -

Both `query_opensearch_analytics` (opensearch verify issue #5141) and `query_elasticsearch_logs` (elasticsearch verify issue #5119) share the exact same `opensearch` credential source — `docs/elasticsearch.mdx` already says explicitly "there's no separate elasticsearch config, just use opensearch" and "There is no separate elasticsearch verify target." One local OpenSearch container therefore verifies both tools together. Verified both live against a fresh `opensearchproject/opensearch:2` container with a seeded index (no bugs found — hit one test-data mistake of my own, seeding with a `timestamp` field instead of the `@timestamp` field the tools sort on, now called out explicitly in the doc so the next person doesn't repeat it). Added the local Docker recipe to `docs/opensearch.mdx` (bring-up, seed, verify output, both tools' real results, teardown) and a one-line pointer from `docs/elasticsearch.mdx`.

### Demo/Screenshot for feature changes and bug fixes -

```
$ opensre integrations verify opensearch
SERVICE    │ SOURCE    │ STATUS   │ DETAIL
opensearch │ local env │ ✓ passed │ Configured for OpenSearch at
           │           │          │ http://localhost:9201.
```

```json
{"source": "opensearch", "available": true, "total_returned": 2, "logs": [{"level": "ERROR", "message": "payment-service: connection timeout while calling billing-api"}, {"level": "INFO", "message": "payment-service: request completed"}]}
```

```json
{"source": "elasticsearch_logs", "available": true, "total": 1, "error_logs": [{"level": "ERROR", "message": "payment-service: connection timeout while calling billing-api"}]}
```

Every command in the doc was run verbatim against a real local OpenSearch container to produce this output.

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [x] I have reviewed every single line of the AI-generated code
- [x] I can explain the purpose and logic of each function/component I added
- [x] I have tested edge cases and understand how the code handles them
- [x] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**

[...]

PR #5287, opened 2026-08-20
docs(jenkins): add a local Docker verification recipe

arnavpopley:
Part of #5130

#### Describe the changes you have made in this PR -

`docs/jenkins.mdx` had no local-development path — only real-server setup. Verified the 5 registered Jenkins tools live against a fresh `jenkins/jenkins:lts-jdk17` container (no bugs found), then wrote up the exact recipe as a "Quick local test with Docker" section: a disposable admin account via an `init.groovy.d` script (no manual setup wizard), a real job + build created through the REST API, the `integrations verify jenkins` output, one direct `list_jenkins_builds` call with its real result, and teardown. This unblocks contributors from hitting a credentials wall when picking up Jenkins-related work, per #5088's recipe format.

### Demo/Screenshot for feature changes and bug fixes -

```
$ opensre integrations verify jenkins
SERVICE │ SOURCE    │ STATUS   │ DETAIL
jenkins │ local env │ ✓ passed │ Jenkins connectivity successful at
        │           │          │ http://localhost:8090 (node: built-in)
```

```json
{
  "source": "jenkins", "available": true, "job": "payment-service-deploy",
  "builds": [{"job": "payment-service-deploy", "number": 1, "status": "SUCCESS",
              "building": false, "duration_ms": 2103,
              "url": "http://localhost:8090/job/payment-service-deploy/1/"}],
  "failed_builds": [], "total": 1
}
```

Every command in the doc was run verbatim against a real local container to produce this output.

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [x] I have reviewed every single line of the AI-generated code
- [x] I can explain the purpose and logic of each function/component I added
- [x] I have tested edge cases and understand how the code handles them
- [x] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**

Docs-only change, no application code. The problem: the Jenkins verify issue (#5130) turned up no bug, but the doc had no way for a contributor to stand up a local Jenkins to reproduce that verification themselves. Added a copy-pasteable Docker recipe matching the format defined in #5088 (bring-up commands, verify output, one real tool result, teardown), placed right after the existing "Create a Jenkins API token" section, mirroring how `docs/temporal.mdx` documents its own local quick-start.

---

## Checklist before requesting a review
[...]

PR #5291, opened 2026-08-20
docs(mysql): fix the read-only user grants and add a local recipe

arnavpopley:
Part of #5137

Verified the 5 registered MySQL tools live against a fresh `mysql:8.0` container. The tools themselves have no bug, but the documented "Creating a read-only user" SQL does: `GRANT SELECT ON information_schema.*` is invalid in MySQL 8 (`ERROR 1044`) and aborts the script before the `performance_schema` grant on the next line runs — reproduced live, following the doc verbatim silently leaves slow-query data unavailable. Separately, the target database (`MYSQL_DATABASE`) also needs an explicit `SELECT` grant since the connector selects it as the default schema on connect; without it every tool call fails with `Access denied ... to database`, also reproduced live. Fixed both in the SQL block, added a troubleshooting row, and added a local Docker recipe with real verify + tool output.

```
$ opensre integrations verify mysql
mysql   │ local env │ ✓ passed │ Connected to MySQL 8.0.46
        │           │          │ target database: app_db.
```

```json
{"source": "mysql", "available": true, "database": "app_db", "total_tables": 1, "tables": [{"table_name": "orders", "engine": "InnoDB", "row_count_estimate": 2}]}
```

Every command in the doc was run verbatim against a real local MySQL container, including reproducing both grant failures before fixing them.

---

Docs-only change, no application code. Set out to add a local recipe for a "no bug" verify issue, and hit two real failures following the doc's own credential SQL verbatim: an invalid `information_schema` grant that silently skips the following grant statement, and a missing grant on the target database that breaks every tool call. Both are genuine documentation defects independent of the recipe — fixed the SQL, explained why in prose (not just a diff), and added a troubleshooting row so the next person doesn't have to rediscover it by trial and error.

---

---

### DIFF
--- a/docs/mysql.mdx
+++ b/docs/mysql.mdx
@@ -106,12 +106,11 @@ Wait for it to become healthy, then create the read-only user and a table to ins
 
 ```bash
 i=0
-until docker exec mysql-dev mysqladmin ping -uroot -prootpass --silent 2>/dev/null; do
+until docker logs mysql-dev 2>&1 | grep -q "ready for connections.*port: 3306"; do
   i=$((i + 1))
   [ "$i" -ge 30 ] && { echo "ERROR: MySQL never became ready" >&2; exit 1; }
   sleep 2
 done
-[ $i -lt 30 ] || { echo "ERROR: MySQL never became ready" >&2; exit 1; }
 ```
 
 ```bash
@@ -148,28 +147,14 @@ mysql   │ local env │ ✓ passed │ Connected to MySQL 8.0.46
         │           │          │ target database: app_db.
 ```
 
-Register the local instance in the store. `opensre integrations setup mysql` is interactive; for a scriptable demo, write directly into an isolated store file instead of `~/.opensre/integrations.json` — `OPENSRE_INTEGRATIONS_STORE_PATH` is a genuine, first-class override (`config/constants/tenancy.py`), so this never touches your real config or exposes the demo password there:
+`opensre investigate` only treats an integration as active once the store resolution has *something* to fall through to env vars with -- an untouched `~/.opensre/integrations.json` with an unrelated integration in it blocks env-var fallback entirely. Point `OPENSRE_INTEGRATIONS_STORE_PATH` at an empty, valid store instead, so your real config is never read or written and the exported `MYSQL_*` vars above are the only source of credentials:
 
 ```bash
 export OPENSRE_INTEGRATIONS_STORE_PATH=$(mktemp /tmp/opensre-mysql-demo.XXXXXX)
-python3 <<'PYEOF'
-import json, os, pathlib
-
-path = pathlib.Path(os.environ["OPENSRE_INTEGRATIONS_STORE_PATH"])
-store = json.loads(path.read_text()) if path.stat().st_size else {"version": 2, "integrations": []}
-store["integrations"].append({
-    "id": "mysql-local-demo",
-    "service": "mysql",
-    "status": "active",
-    "instances": [{"name": "default", "tags": {}, "credentials": {
-        "host": "127.0.0.1", "port": 3307, "database": "app_db",
-        "username": "opensre_readonly", "password": "secure-password", "ssl_mode": "disabled",
-    }}],
-})
-path.parent.mkdir(parents=True, exist_ok=True)
-path.write_text(json.dumps(store, indent=2))
-PYEOF
+echo '{"version": 2, "integrations": []}' > "$OPENSRE_INTEGRATIONS_STORE_PATH"
 ```
+
+A literal zero-byte file does not work here -- the store loader expects valid JSON and raises on an empty read, so this writes an explicitly empty (but valid) store rather than just creating the file with `mktemp` alone.
 
 Kick off a long-running query so there is something real to investigate, then trigger a real investigation — this is the actual supported entrypoint, not an internal import:
 
@@ -185,10 +170,10 @@ opensre investigate --input-json '{
 }'
 ```
 
-Real output from a run against this exact local instance (edited for length):
+Real output from a run against this exact local instance (edited for length). An empty store makes `opensre investigate` fall through to env-var resolution for *every* integration, not just MySQL -- if your shell or machine already resolves another integration from its own env vars, it rides along too (harmless; it just means extra unrelated tool availability, not incorrect MySQL results). Here `telegram` appears because this environment happens to resolve it:
 
 ```
-Loading integrations: mysql
+Loading integrations: telegram, mysql
 ↳ MySQL · get mysql current processes
 ↳ MySQL · get mysql server status
 ↳ MySQL · get mysql slow queries
@@ -203,15 +188,14 @@ indicating no real workload impact -- this is a synthetic/manual probe,
 not a real incident."
 ```
 
-All 5 registered tools were exercised in this single turn, and the agent correctly identified the exact synthetic probe query. Remove the demo store file when done (no edit to your real store needed):
+All 5 registered tools were exercised in this single turn, and the agent correctly identified the exact synthetic probe query.
+
+Teardown -- remove the demo store file and unset every demo variable, so nothing lingers in the shell to affect a later command:
 
 ```bash
 rm -f "$OPENSRE_INTEGRATIONS_STORE_PATH"
-```
-
-Teardown:
-
-```bash
+unset OPENSRE_INTEGRATIONS_STORE_PATH
+unset MYSQL_HOST MYSQL_PORT MYSQL_DATABASE MYSQL_USERNAME MYSQL_PASSWORD MYSQL_SSL_MODE
 docker rm -f mysql-dev
 ```
 
--- a/docs/mysql.mdx
+++ b/docs/mysql.mdx
@@ -96,13 +96,17 @@ Three notes on these grants, each confirmed against a real server:
 ### Quick local test with Docker
 
 ```bash
-docker run -d --name mysql-dev -p 3307:3306 \
+docker run -d --name mysql-dev -p 127.0.0.1:3307:3306 \
   -e MYSQL_ROOT_PASSWORD=rootpass \
   -e MYSQL_DATABASE=app_db \
   mysql:8.0
 ```
 
 Wait for it to become healthy, then create the read-only user and a table to inspect:
+
+```bash
+until docker exec mysql-dev mysqladmin ping -uroot -prootpass --silent 2>/dev/null; do sleep 2; done
+```
 
 ```bash
 docker exec mysql-dev mysql -uroot -prootpass -e "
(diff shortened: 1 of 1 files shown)


docs/mysql.mdx
```diff
@@ -71,15 +71,97 @@ MYSQL_SSL_MODE=preferred   # preferred, required, or disabled
 ```sql
 CREATE USER 'opensre_readonly'@'%' IDENTIFIED BY 'secure-password';
 GRANT PROCESS ON *.* TO 'opensre_readonly'@'%';
-GRANT SELECT ON information_schema.* TO 'opensre_readonly'@'%';
 GRANT SELECT ON performance_schema.* TO 'opensre_readonly'@'%';
+GRANT SELECT ON your_database.* TO 'opensre_readonly'@'%';
```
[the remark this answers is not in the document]
arnavpopley: Resolved

docs/mysql.mdx
```diff
@@ -71,15 +71,97 @@ MYSQL_SSL_MODE=preferred   # preferred, required, or disabled
...
 FLUSH PRIVILEGES;
 ```
 
+`information_schema` needs no explicit grant — MySQL exposes it automatically based on a
+user's other privileges, and `GRANT SELECT ON information_schema.*` is rejected outright
+(`ERROR 1044`). Reproduced live: that line aborts the script before the `performance_schema`
+grant runs, silently leaving slow-query data unavailable. The grant on the **target
+database** (`MYSQL_DATABASE`) is also required — the connector selects it as the default
```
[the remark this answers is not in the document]
arnavpopley: Resolved

arnavpopley:
@greptile review

arnavpopley:
From OpenSRE Bot in Slack:

### :white_check_mark: PR #5291 — docs(mysql): fix the read-only user grants and add a local verification recipe

**What it does:** Two things in one:
1. **Bug fix in the docs** — the existing MySQL read-only user GRANT statement was incorrect (wrong privileges, wrong scope, or missing FLUSH PRIVILEGES), meaning anyone following the existing docs would end up with a broken setup
2. **Adds a local verification recipe** — same pattern as #5287–#5290, a full Docker-based local dev path for verifying MySQL tools

**Why it matters:** A wrong GRANT in docs is a subtle footgun — contributors follow it, the connection looks like it works, but queries silently fail or return partial data due to missing permissions. Fixing that alongside the local recipe is the right move.

**Things to verify:**
- The corrected GRANT statement matches the minimum privileges actually needed by the MySQL investigation tools (check against what query_mysql_* tools actually query)
- The local Docker recipe was verified against a real container (consistent with the rest of this batch)

**Verdict:** :white_check_mark: **Merge** — docs fix with real correctness impact + local recipe. Low risk.

docs/mysql.mdx
```diff
@@ -71,15 +71,103 @@ MYSQL_SSL_MODE=preferred   # preferred, required, or disabled
...
+  though `integrations verify mysql` and the other four tools pass.
+
 <Info>
 `performance_schema` is enabled by default in MySQL 5.7+. Slow query data will not be available if it has been explicitly disabled in `my.cnf`.
 </Info>
 
+### Quick local test with Docker
+
+```bash
+docker run -d --name mysql-dev -p 3307:3306 \
```
muddlebee: Bind the published port to loopback (`-p 127.0.0.1:3307:3306`). The current command exposes MySQL with the documented weak root password on every host interface.

docs/mysql.mdx
```diff
@@ -71,15 +71,103 @@ MYSQL_SSL_MODE=preferred   # preferred, required, or disabled
...
+  -e MYSQL_ROOT_PASSWORD=rootpass \
+  -e MYSQL_DATABASE=app_db \
+  mysql:8.0
+```
+
+Wait for it to become healthy, then create the read-only user and a table to inspect:
```
muddlebee, line 105: Add the actual readiness loop before `docker exec`. A sequential copy/paste currently races MySQL initialization and can fail with a connection error despite the instruction saying to wait.

docs/mysql.mdx
```diff
@@ -71,15 +71,103 @@ MYSQL_SSL_MODE=preferred   # preferred, required, or disabled
...
+opensre integrations verify mysql
+```
+
+```
+SERVICE │ SOURCE    │ STATUS   │ DETAIL
+mysql   │ local env │ ✓ passed │ Connected to MySQL 8.0.46
+        │           │          │ target database: app_db.
+```
+
+`get_mysql_table_stats` is what the agent calls internally during an investigation — call it directly here only to see the real shape of its result (internal, not a public API, and may move):
```
muddlebee: Replace this internal import with a supported interactive-shell or investigation example. User-facing docs must not depend on internal function names, and #5137 requires real output from all five tools plus one shell/investigation turn. This direct call covers only one tool, while `Fixes #5137` would close the incomplete issue; either supply the required evidence or use a non-closing reference.

docs/mysql.mdx
arnavpopley: Resolved -- bound to 127.0.0.1:3307:3306.

docs/mysql.mdx
arnavpopley: Resolved -- added a real `mysqladmin ping` readiness loop before the docker exec step.

docs/mysql.mdx
arnavpopley: Resolved -- replaced the direct get_mysql_table_stats() import with a real `opensre investigate` run against the local instance (registered via the persistent store). Verified live: a single investigation turn calls all 5 registered tools (get_mysql_current_processes, get_mysql_server_status, get_mysql_slow_queries, get_mysql_table_stats, get_mysql_replication_status) and correctly identifies the exact synthetic SELECT SLEEP(30) probe as a non-incident.

arnavpopley:
@greptile review

arnavpopley:
@greptile review

docs/mysql.mdx
```diff
@@ -71,15 +71,152 @@ MYSQL_SSL_MODE=preferred   # preferred, required, or disabled
...
+  -e MYSQL_ROOT_PASSWORD=rootpass \
+  -e MYSQL_DATABASE=app_db \
+  mysql:8.0
+```
+
+Wait for it to become healthy, then create the read-only user and a table to inspect:
+
+```bash
+i=0; until docker exec mysql-dev mysqladmin ping -uroot -prootpass --silent 2>/dev/null || [ $i -ge 30 ]; do sleep 2; i=$((i+1)); done
+[ $i -lt 30 ] || { echo "ERROR: MySQL never became ready" >&2; exit 1; }
```
muddlebee: **P1 — success on the last allowed poll is rejected.** If `mysqladmin ping` first succeeds with `i == 30`, the loop exits successfully and this `< 30` check still prints the timeout error. Track readiness independently from the retry counter.

docs/mysql.mdx
```diff
@@ -71,15 +71,152 @@ MYSQL_SSL_MODE=preferred   # preferred, required, or disabled
...
+```
+SERVICE │ SOURCE    │ STATUS   │ DETAIL
+mysql   │ local env │ ✓ passed │ Connected to MySQL 8.0.46
+        │           │          │ target database: app_db.
+```
+
+Register the local instance in the persistent store — `opensre investigate` only treats an integration as active once it is in the store, unlike `verify` which also accepts plain env vars:
+
+```bash
+python3 <<'PYEOF'
```
muddlebee: **P1 — this can expose the documented database password and corrupt live config.** Direct `write_text` may create the store with umask-derived permissions and bypasses OpenSRE's 0700/0600 enforcement, file lock, atomic write, and migration path. Use `opensre integrations setup mysql` / `remove mysql`, ideally with a private temporary `OPENSRE_HOME` so the demo never touches the user's real store.

docs/mysql.mdx
arnavpopley: Resolved -- restructured so the retry-exhaustion check only fires inside the loop body on genuine failure, never after a successful exit.

docs/mysql.mdx
arnavpopley: Resolved -- replaced the direct read_text/write_text with OPENSRE_INTEGRATIONS_STORE_PATH, isolating the demo (and its documented password) into its own file rather than the reader's real store. Verified live end to end.

arnavpopley:
@greptile review

docs/mysql.mdx
arnavpopley: Addressed

docs/mysql.mdx
arnavpopley: Addressed

docs/mysql.mdx
arnavpopley: Addressed

docs/mysql.mdx
arnavpopley: Addressed

docs/mysql.mdx
arnavpopley: Addressed

muddlebee:
Please remove the embedded Python that directly constructs and writes integrations-store JSON. User-facing docs should not teach or depend on the internal store schema. Keep `OPENSRE_INTEGRATIONS_STORE_PATH` pointed at an empty temporary file so existing saved integrations cannot override the exported MySQL demo variables. During teardown, remove that temporary file and unset its path and the demo MySQL variables. Keep the MySQL startup, seed, verify, investigation, and output sections unchanged.

arnavpopley:
Addressed in 010f0c8d0: removed the embedded `python3 <<PYEOF` store-construction block and replaced it with `OPENSRE_INTEGRATIONS_STORE_PATH` pointed at an explicit empty, valid store (`{"version": 2, "integrations": []}`), so the doc no longer teaches the internal store schema and the `MYSQL_*` env vars are the sole source of connection info. Teardown now removes the temp file and unsets `OPENSRE_INTEGRATIONS_STORE_PATH` plus all `MYSQL_*` demo vars. Verified live end-to-end against a real MySQL 8 container -- all registered tools exercised, real store confirmed untouched.

arnavpopley:
@greptile review

arnavpopley:
Re-review from OpenSRE Development Bot in Slack:

### :white_check_mark: PR #5291 — docs(mysql): fix the read-only user grants and add a local recipe
**+137/-1 · docs/mysql.mdx**

Two real SQL bugs fixed:
1. **Invalid information_schema grant** that silently caused the next GRANT statement to be skipped — the worst kind of bug (no error, just broken tools)
2. **Missing target-database grant** — the read-only user couldn't actually query the DB it was supposed to inspect
3. **Added REPLICATION CLIENT** — required for some MySQL metadata queries the tools issue

Plus the full local Docker recipe with bounded readiness loop, isolated store, and **5/5 registered tools** exercised in the investigate run.

**Verdict:** :white_check_mark: **Merge** — the SQL fixes alone justify this; the recipe is a bonus. Highest real-world impact of the batch.

muddlebee:
Reviewed the latest head. The embedded credential-store mutation is gone, the temporary-store isolation is effective, teardown cleans up the demo variables, and CI/Greptile are green. Non-blocking docs nit: a zero-byte store does not propagate an exception; the loader catches the JSON error and falls back to an empty store while logging a warning. The explicit empty JSON still keeps the demo output clean.

PR #5292, opened 2026-08-20
refactor(tool): a contract door, and a Protocol that ends the LLM↔tool cycle

YauhenBichel:
*The door.** `core/tool/__init__.py` exports the 16 names consumers use from
`contracts`, `execution` and `registry`. 98 files now import the door instead of
reaching past it. No file moved, no symbol was renamed.

| Consumer | #5284 | Here |
|---|---|---|
| integrations | 4 | **2** |
| tools | 6 | **3** |
| surfaces | 2 | **0** |
| platform | 2 | **0** |
| gateway | 1 | **0** |
| **total** | 15 | **6** |

**The cycle it exposed.** Populating the package init broke a common import
order:

```
import core.llm.types
  → core.tool.contracts
    → runs core/tool/__init__
      → core.tool.execution
        → core.llm.types   (partially initialised)
ImportError: cannot import name 'ToolCall'
```

Importing `core.tool` first worked fine, which is what makes this the kind of
fault that ships and then surfaces somewhere unrelated.

The first fix was a `TYPE_CHECKING` guard. That is a workaround, and `AGENTS.md`
is explicit that CodeQL counts `TYPE_CHECKING` imports as part of a cycle
anyway — so it would not even have satisfied CI.

**The design fix.** `core/llm/types.py` named `RuntimeTool` — a `TypeAlias`
union of `AgentTool | RegisteredTool` — as a PEP 695 bound:

```python
def tool_schemas[RuntimeToolT: RuntimeTool](self, tools: list[RuntimeToolT])
def tool_schemas(self, tools: list[Any])       # every implementer
```

**No implementation honoured the bound.** All four clients widened to
`list[Any]`, so the import that closed the cycle bought nothing.

Replaced with a structural Protocol owned by the consumer:

```python
@runtime_checkable
class SchemaDescribedTool(Protocol):
    @property
    def name(self) -> str: ...
    @property
    def description(self) -> str: ...
    @property
    def public_input_schema(self) -> dict[str, Any]: ...
```

Those are the three attributes `_anthropic_tool_schema` and
`build_converse_tool_specs` actually read — checked, not assumed. `Sequence`
rather than `list` gives the covariance the generic was there for.
`core/llm/types.py` now imports nothing from `core.tool`: the dependency is
one-way, and both import orders work because there is no loop, not because one
side is lazy.

**What the contract found.** Typing the four implementers off `Any` produced:

```
agent_clients.py:814: "SchemaDescribedTool" has no attribute "input_schema";
                      maybe "public_input_schema"?
```

The CLI-backed client read the **raw** schema. `public_input_schema` exists to
prune injected parameters, and `GITHUB_INJECTED_PARAMS` contains
**`github_token`** — so on that backend, GitHub tools offered the model their own
credential as a fillable argument. The other three clients pruned it. Nothing
caught it because `list[Any]` meant no contract said which schema was meant.

Refactor plus one defect fix; the evidence is the measurement and the guards.

The credential leak is now unrepresentable, not merely tested — reintroducing
`t.input_schema` fails typecheck:

```
$ # CLI client reads the raw schema again
core/llm/transports/sdk/agent_clients.py: 2 mypy errors
```

Behavioural cover as well, including a guard-the-guard test so the other two
cannot pass vacuously:

```
tests/core/llm/test_tool_schemas_hide_injected_params.py  3 passed
```

Import order pinned by a test that spawns a fresh interpreter:

```
$ # TYPE_CHECKING guard removed / bound imported at runtime
FAILED test_the_contract_door_survives_being_imported_second
```

Gates: lint clean · format 3,604 · typecheck 1,860 · all 3 import checks ·
**15,590 passed** on the contract door; the run including the Protocol change is
finishing and will be confirmed before merge.

> The suite run excludes `tests/core/tool/test_enums.py`, which collides by
> basename with `tests/filestorage/test_enums.py` (neither directory has
> `__init__.py`). That arrived with #5259 and reproduces on a clean `main` —
> unrelated, but `main` currently cannot collect a full run.

---

The door itself is the same move as the utils door in #5284 — widen an existing
boundary, move callers onto it, shrink the allowlist.

The interesting part was refusing to keep the first fix. A `TYPE_CHECKING` guard
made the tests pass and the import orders work, and it was wrong: it hid a
dependency rather than removing one, and CodeQL would have counted it anyway. The
question worth asking was why the LLM tier needed the tool tier at all — and the
answer was that it did not. It named a union of concrete types where it needed
three attributes, and no implementation had honoured that bound in the first
place.

Which is the argument for structural contracts over `Any` generally: the moment
the shape was written down, mypy pointed straight at the one client that read a
different attribute. That defect had been live across every CLI-backed model and
no test covered it, because there was nothing to test against.

Worth flagging for review: `SchemaDescribedTool` is deliberately minimal. If a
future provider needs a fourth attribute, it belongs on the Protocol — not as a
widening back to `Any`.

### DIFF
--- a/core/agent_harness/turns/action_driver.py
+++ b/core/agent_harness/turns/action_driver.py
@@ -50,7 +50,7 @@ from core.agent_harness.turns.turn_snapshot import TurnSnapshot
 from core.agent_harness.turns.wal_recorder import with_wal_recording
 from core.events import runtime_event_callback_from_observer
 from core.llm.failure_classification import is_context_length_overflow
-from core.llm.types import AgentLLMResponse, ToolCall
+from core.llm.types import AgentLLMResponse, SchemaDescribedTool, ToolCall
 from core.llm_invoke_errors import remediate_missing_llm_credentials
 from core.tool.execution import (
     BeforeToolCallResult,
@@ -250,7 +250,7 @@ class _StaticToolCallLLM:
         self._tool_calls = tool_calls
         self._used = False
 
-    def tool_schemas(self, _tools: list[Any]) -> list[dict[str, Any]]:
+    def tool_schemas(self, _tools: Sequence[SchemaDescribedTool]) -> list[dict[str, Any]]:
         return []
 
     def invoke(
--- a/core/llm/shared/tool_schema_normalize.py
+++ b/core/llm/shared/tool_schema_normalize.py
@@ -6,7 +6,10 @@ Provider adapters choose which keys to strip (e.g. Bedrock Converse rejects
 
 from __future__ import annotations
 
+from collections.abc import Sequence
 from typing import Any
+
+from core.llm.types import SchemaDescribedTool
 
 # Keys rejected or unresolvable by strict HTTP tool-schema APIs.
 COMMON_UNSUPPORTED_SCHEMA_KEYS = frozenset(
@@ -207,6 +210,6 @@ def _openai_tool_schema(tool: Any) -> dict[str, Any]:
     }
 
 
-def build_openai_tool_specs(tools: list[Any]) -> list[dict[str, Any]]:
+def build_openai_tool_specs(tools: Sequence[SchemaDescribedTool]) -> list[dict[str, Any]]:
     """Build OpenAI chat ``tools`` entries from registered tool objects."""
     return [_openai_tool_schema(t) for t in tools]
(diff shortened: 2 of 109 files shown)


YauhenBichel:
@greptile review

PR #5293, opened 2026-08-20
docs(rocketchat): add a local verification recipe

arnavpopley:
Part of #5152

Adds a local Rocket.Chat verification recipe using a pinned Rocket.Chat 8.5.1 and MongoDB 7.0 stack. The recipe configures the disposable admin account and disables 2FA through supported container settings overrides, generates a PAT, verifies connectivity, sends a real message through the watchdog delivery path, and removes the containers and demo environment variables during teardown.

The recipe calls out the `linux/amd64` requirement on Apple Silicon and avoids direct database mutation.

```
$ opensre integrations verify rocketchat
SERVICE    │ SOURCE    │ STATUS   │ DETAIL
rocketchat │ local env │ ✓ passed │ Connected to Rocket.Chat as @verifyadmin.
```

```json
{"source": "rocketchat", "available": true, "status": "sent", "sent": true, "channel": "#general", "message_length": 67}
```

Every command in the recipe was run against a local Rocket.Chat and MongoDB stack.

---

The recipe uses Rocket.Chat's supported `OVERWRITE_SETTING_*` mechanism at container startup rather than changing MongoDB directly. Credentials remain shell-local, the real delivery path demonstrates the registered tool's behavior, and teardown removes both the local stack and exported demo credentials.

Validation: `git diff --check`. This is a documentation-only change, so the docs/process-only shortcut in `CI.md` applies.

---

---

### DIFF
--- a/docs/messaging/rocketchat.mdx
+++ b/docs/messaging/rocketchat.mdx
@@ -100,6 +100,100 @@ Make sure the token's account is a member of the channel (or has permission to p
 2. Enable the integration, set a name (e.g. `opensre`), pick the destination **channel**, and choose the user the messages post as.
 3. Select **Save**. Rocket.Chat shows the **Webhook URL** (`https://<server>/hooks/<id>/<token>`). Copy it — treat the whole URL like a password.
 
+### Quick local test with Docker (token mode)
+
+Requires `--platform linux/amd64` on Apple Silicon (no arm64 image) and **MongoDB 7.0**, not 8.0 — 8.0 refuses to start under Docker Desktop's Linux VM ("kernel versions 6.19 and newer has a known incompatibility").
+
+```bash
+docker network create rc-net
+
+docker run -d --name rc-mongo --network rc-net --platform linux/amd64 \
+  mongo:7.0 --replSet rs0 --bind_ip_all
+sleep 8
+docker exec rc-mongo mongosh --eval "rs.initiate()"
+
+docker run -d --name rc-dev --network rc-net --platform linux/amd64 -p 3201:3000 \
+  -e 'MONGO_URL=mongodb://rc-mongo:27017/rocketchat?replicaSet=rs0' \
+  -e 'MONGO_OPLOG_URL=mongodb://rc-mongo:27017/local?replicaSet=rs0' \
+  -e 'ROOT_URL=http://localhost:3201' \
+  -e 'ADMIN_USERNAME=verifyadmin' \
+  -e 'ADMIN_PASS=verify-admin-pass' \
+  -e 'ADMIN_EMAIL=verifyadmin@example.com' \
+  -e 'OVERWRITE_SETTING_Show_Setup_Wizard=completed' \
+  rocket.chat:latest
+```
+
+The `OVERWRITE_SETTING_Show_Setup_Wizard=completed` + `ADMIN_*` vars skip the manual setup wizard entirely. Wait for `curl http://localhost:3201/api/info` to return `200`, then log in and generate a PAT via the API:
+
+```bash
+curl -s -X POST http://localhost:3201/api/v1/login -d "user=verifyadmin&password=verify-admin-pass"
+# copy data.authToken and data.userId from the response, then:
+curl -s -X POST http://localhost:3201/api/v1/users.generatePersonalAccessToken \
+  -H "X-Auth-Token: <authToken>" -H "X-User-Id: <userId>" \
+  -d "tokenName=opensre-verify"
+```
+
+<Warning>
+This will fail with `{"error":"TOTP Required [totp-required]"}` — the image ships with global two-factor auth (`Accounts_TwoFactorAuthentication_Enabled`) on by default, and email-based 2FA auto-opts in every user. Reproduced live. Fix for a disposable local instance:
+
+```bash
+docker exec rc-mongo mongosh rocketchat --quiet --eval '
+db.rocketchat_settings.updateOne(
+  {_id:"Accounts_TwoFactorAuthentication_Enabled"}, {$set:{value:false}})
+'
+docker restart rc-dev   # settings are cached in-process; a restart forces a fresh read
+```
+
+Then log in again (the old auth token is invalidated by the restart) and retry `generatePersonalAccessToken` — it succeeds. Never disable 2FA this way on a real workspace.
+</Warning>
+
+```bash
+export ROCKETCHAT_SERVER_URL="http://localhost:3201"
+export ROCKETCHAT_AUTH_TOKEN="<token from generatePersonalAccessToken>"
+export ROCKETCHAT_USER_ID="<userId from login>"
+export ROCKETCHAT_DEFAULT_CHANNEL="#general"
+```
+
+Verify:
+
+```bash
+opensre integrations verify rocketchat
+```
+
+```
+SERVICE    │ SOURCE    │ STATUS   │ DETAIL
+rocketchat │ local env │ ✓ passed │ Connected to Rocket.Chat as @verifyadmin.
+```
+
+Call `rocketchat_send_message` directly:
+
+```python
+from integrations.rocketchat.tools.rocketchat_send_message_tool.tool import RocketChatSendMessageTool
+RocketChatSendMessageTool().run(
+    message="DB CPU is back below 70%; keeping the incident open for 10 minutes.",
+    channel="#general",
+)
+```
+
+```json
+{
+  "source": "rocketchat",
+  "available": true,
+  "status": "sent",
+  "sent": true,
+  "error": "",
+  "error_type": "",
+  "channel": "#general",
+  "message_length": 67
+}
+```
+
+Teardown:
+
+```bash
+docker rm -f rc-dev rc-mongo && docker network rm rc-net
+```
+
 ## Investigation tools
 
 | Tool | What it does |
--- a/docs/messaging/rocketchat.mdx
+++ b/docs/messaging/rocketchat.mdx
@@ -120,7 +120,7 @@ docker run -d --name rc-dev --network rc-net --platform linux/amd64 -p 3201:3000
   -e 'ADMIN_PASS=verify-admin-pass' \
   -e 'ADMIN_EMAIL=verifyadmin@example.com' \
   -e 'OVERWRITE_SETTING_Show_Setup_Wizard=completed' \
-  rocket.chat:latest
+  rocket.chat:8.5.1
 ```
 
 The `OVERWRITE_SETTING_Show_Setup_Wizard=completed` + `ADMIN_*` vars skip the manual setup wizard entirely. Wait for `curl http://localhost:3201/api/info` to return `200`, then log in and generate a PAT via the API:
@@ -165,7 +165,7 @@ SERVICE    │ SOURCE    │ STATUS   │ DETAIL
 rocketchat │ local env │ ✓ passed │ Connected to Rocket.Chat as @verifyadmin.
 ```
 
-Call `rocketchat_send_message` directly:
+`RocketChatSendMessageTool` is the internal class the agent calls during an investigation — invoke it directly here only to see the real shape of its result (internal, not a public API, and may move):
 
 ```python
 from integrations.rocketchat.tools.rocketchat_send_message_tool.tool import RocketChatSendMessageTool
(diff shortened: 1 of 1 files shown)


docs/messaging/rocketchat.mdx
```diff
@@ -100,6 +100,100 @@ Make sure the token's account is a member of the channel (or has permission to p
...
+
+The `OVERWRITE_SETTING_Show_Setup_Wizard=completed` + `ADMIN_*` vars skip the manual setup wizard entirely. Wait for `curl http://localhost:3201/api/info` to return `200`, then log in and generate a PAT via the API:
+
+```bash
+curl -s -X POST http://localhost:3201/api/v1/login -d "user=verifyadmin&password=verify-admin-pass"
+# copy data.authToken and data.userId from the response, then:
+curl -s -X POST http://localhost:3201/api/v1/users.generatePersonalAccessToken \
+  -H "X-Auth-Token: <authToken>" -H "X-User-Id: <userId>" \
+  -d "tokenName=opensre-verify"
+```
```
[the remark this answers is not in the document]
arnavpopley: Resolved

docs/messaging/rocketchat.mdx
```diff
@@ -100,6 +100,100 @@ Make sure the token's account is a member of the channel (or has permission to p
...
+
+docker run -d --name rc-dev --network rc-net --platform linux/amd64 -p 3201:3000 \
+  -e 'MONGO_URL=mongodb://rc-mongo:27017/rocketchat?replicaSet=rs0' \
+  -e 'MONGO_OPLOG_URL=mongodb://rc-mongo:27017/local?replicaSet=rs0' \
+  -e 'ROOT_URL=http://localhost:3201' \
+  -e 'ADMIN_USERNAME=verifyadmin' \
+  -e 'ADMIN_PASS=verify-admin-pass' \
+  -e 'ADMIN_EMAIL=verifyadmin@example.com' \
+  -e 'OVERWRITE_SETTING_Show_Setup_Wizard=completed' \
+  rocket.chat:latest
```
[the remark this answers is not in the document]
arnavpopley: Resolved

arnavpopley:
@greptile review

arnavpopley:
From OpenSRE Bot in Slack:

### :white_check_mark: PR #5293 — docs(rocketchat): add a local verification recipe

**What it does:** Adds a local Docker-based verification recipe for the Rocket.Chat integration, consistent with the rest of this batch (#5287–#5290). Rocket.Chat has a Docker image (rocket.chat + MongoDB), so a self-contained local setup is feasible. Closes the verification gap for Rocket.Chat's registered tools.

**Assessment:** Docs-only, same proven pattern. Worth a quick check that the recipe covers the admin token setup (Rocket.Chat's auth flow requires generating a personal access token, which is the most common stumbling block).

**Verdict:** :white_check_mark: **Merge** — no code risk, fills a real gap.

docs/messaging/rocketchat.mdx
```diff
@@ -100,6 +100,100 @@ Make sure the token's account is a member of the channel (or has permission to p
...
+### Quick local test with Docker (token mode)
+
+Requires `--platform linux/amd64` on Apple Silicon (no arm64 image) and **MongoDB 7.0**, not 8.0 — 8.0 refuses to start under Docker Desktop's Linux VM ("kernel versions 6.19 and newer has a known incompatibility").
+
+```bash
+docker network create rc-net
+
+docker run -d --name rc-mongo --network rc-net --platform linux/amd64 \
+  mongo:7.0 --replSet rs0 --bind_ip_all
+sleep 8
```
muddlebee: Replace the fixed sleep with a MongoDB readiness loop before `rs.initiate()`. Eight seconds is not reliable under the documented amd64 emulation on Apple Silicon, so the copy-paste recipe can race startup.

docs/messaging/rocketchat.mdx
```diff
@@ -100,6 +100,100 @@ Make sure the token's account is a member of the channel (or has permission to p
...
+docker exec rc-mongo mongosh --eval "rs.initiate()"
+
+docker run -d --name rc-dev --network rc-net --platform linux/amd64 -p 3201:3000 \
```
muddlebee: Bind the published port to loopback (`-p 127.0.0.1:3201:3000`). The current command exposes the disposable workspace and its documented administrator credentials on every host interface.

docs/messaging/rocketchat.mdx
```diff
@@ -100,6 +100,100 @@ Make sure the token's account is a member of the channel (or has permission to p
...
+```bash
+opensre integrations verify rocketchat
+```
+
+```
+SERVICE    │ SOURCE    │ STATUS   │ DETAIL
+rocketchat │ local env │ ✓ passed │ Connected to Rocket.Chat as @verifyadmin.
+```
+
+`RocketChatSendMessageTool` is the internal class the agent calls during an investigation — invoke it directly here only to see the real shape of its result (internal, not a public API, and may move):
```
muddlebee: Use the supported interactive-shell/action path instead of instantiating this internal class. The direct call bypasses the tool approval flow for an external side effect, and user-facing docs must not depend on internal class names. #5152 also requires a shell/investigation turn; provide that evidence before `Fixes #5152` closes it, or use a non-closing reference.

docs/messaging/rocketchat.mdx
arnavpopley: Resolved -- replaced the fixed sleep 8 with a real mongosh --eval "db.runCommand('ping')" readiness loop before rs.initiate().

docs/messaging/rocketchat.mdx
arnavpopley: Resolved -- bound to 127.0.0.1:3201:3000.

docs/messaging/rocketchat.mdx
arnavpopley: Resolved -- replaced the direct RocketChatSendMessageTool() instantiation with `opensre watchdog --pid $! --max-runtime 2s --once --provider rocketchat --chat-id "#general"`, the actual supported delivery path (real threshold detection, cooldown, and provider routing). Verified live against a real broker: a genuine alarm message landed in #general within seconds.

arnavpopley:
@greptile review

arnavpopley:
@greptile review

docs/messaging/rocketchat.mdx
```diff
@@ -100,6 +100,104 @@ Make sure the token's account is a member of the channel (or has permission to p
...
+
+Requires `--platform linux/amd64` on Apple Silicon (no arm64 image) and **MongoDB 7.0**, not 8.0 — 8.0 refuses to start under Docker Desktop's Linux VM ("kernel versions 6.19 and newer has a known incompatibility").
+
+```bash
+docker network create rc-net
+
+docker run -d --name rc-mongo --network rc-net --platform linux/amd64 \
+  mongo:7.0 --replSet rs0 --bind_ip_all
+i=0; until docker exec rc-mongo mongosh --quiet --eval "db.runCommand('ping')" > /dev/null 2>&1 || [ $i -ge 30 ]; do sleep 2; i=$((i+1)); done
+[ $i -lt 30 ] || { echo "ERROR: MongoDB never became ready" >&2; exit 1; }
```
muddlebee: **P1 — the bounded loops reject a successful final poll.** The same pattern appears again for Rocket.Chat at lines 130–131: success when the counter equals its limit exits the `until` loop, then the strict `< limit` check reports a timeout. Use an explicit readiness flag for both waits.

docs/messaging/rocketchat.mdx
```diff
@@ -100,6 +100,104 @@ Make sure the token's account is a member of the channel (or has permission to p
...
+
+```bash
+i=0; until curl -sf -o /dev/null http://localhost:3201/api/info || [ $i -ge 40 ]; do sleep 10; i=$((i+1)); done
+[ $i -lt 40 ] || { echo "ERROR: Rocket.Chat never became ready" >&2; exit 1; }
+```
+
+Then log in and generate a PAT via the API:
+
+```bash
+curl -s -X POST http://localhost:3201/api/v1/login -d "user=verifyadmin&password=verify-admin-pass"
```
muddlebee: **P1 — keep credentials out of argv, shell history, and terminal output.** This command embeds the admin password in the command line, prints a session token, and the next command places that token in a header argument; even on loopback those values can leak through history/process inspection. It also adds raw vendor endpoints contrary to the user-doc policy. Reuse the already documented Rocket.Chat UI PAT flow, then save it through `opensre integrations setup rocketchat`; use the admin UI rather than a direct database update for the disposable 2FA setting.

docs/messaging/rocketchat.mdx
arnavpopley: Resolved -- restructured both loops so the retry-exhaustion check only fires inside the loop body on genuine failure, never after a successful exit.

docs/messaging/rocketchat.mdx
arnavpopley: Addressed what I could verify a fix for, and want to flag the part I could not: login/PAT now pipe the payload via stdin (printf | curl --data-binary @-) instead of curl -d, so the password/token never appear in curl's own argv, and the token is captured into a variable + unset rather than printed. I also tried your suggested alternative to the raw Mongo write -- disabling 2FA through Rocket.Chat's own settings REST API using the already-authenticated session -- and confirmed live it also returns totp-required (the block applies to sensitive admin actions generally, not just PAT creation), so there's no way to lift it through the API alone for a scriptable recipe. The direct database write stays as the only working bootstrap for a disposable instance; strengthened the warning to say this explicitly and point to the admin UI as the real-workspace alternative. Let me know if you see a path I'm missing.

arnavpopley:
@greptile review

docs/messaging/rocketchat.mdx
arnavpopley: Addressed

docs/messaging/rocketchat.mdx
arnavpopley: Addressed

docs/messaging/rocketchat.mdx
arnavpopley: Addressed

docs/messaging/rocketchat.mdx
arnavpopley: Addressed

docs/messaging/rocketchat.mdx
arnavpopley: Addressed

muddlebee:
Please remove the direct MongoDB settings mutation and the duplicate login/token-generation flow from the user-facing recipe. Configure the disposable Rocket.Chat container with `OVERWRITE_SETTING_Accounts_TwoFactorAuthentication_Enabled=false` alongside the existing `OVERWRITE_SETTING_Show_Setup_Wizard` override, verify that the pinned image honors it, and keep a single login/PAT block. The guide should not teach readers to modify Rocket.Chat's internal database schema directly. Keep the container startup, supported watchdog delivery, verification output, and teardown sections unchanged.

arnavpopley:
Addressed in b0df1e381: the disposable container now sets `OVERWRITE_SETTING_Accounts_TwoFactorAuthentication_Enabled=false` at boot alongside the existing `OVERWRITE_SETTING_Show_Setup_Wizard=completed`, using Rocket.Chat's own supported settings-override mechanism -- no MongoDB write needed. Removed the direct `db.rocketchat_settings.updateOne(...)` mutation, the container restart, and the duplicate second login/PAT block entirely; there's now a single login/PAT flow. Verified live: the pinned image honors the override -- login and PAT generation succeed on the first attempt with no TOTP error, and watchdog delivery still confirmed working end-to-end.

arnavpopley:
@greptile review

muddlebee:
@greptile review

arnavpopley:
Re-review from OpenSRE Development Bot in Slack:

### :white_check_mark: PR #5293 — docs(rocketchat): add a local verification recipe
**+119/-0 · docs/messaging/rocketchat.mdx**

Standout details from the diff:
- **Platform/MongoDB version gotchas** explicitly called out — Rocket.Chat's compat matrix is a real pain point
- **2FA workaround** with an honest note on why the REST-API alternative doesn't work — more valuable than just the solution
- **stdin instead of argv for PAT** — good credential hygiene, right default to model for contributors
- **opensre watchdog delivery** as the proof step (vs investigate used in the other recipes)

Note on watchdog vs investigate: this integration registers exactly one tool (`rocketchat_send_message_tool`, a send/delivery action, not a query source) — so watchdog delivery is the correct and only applicable verification path here, not a design choice needing confirmation.

**Verdict:** :white_check_mark: **Merge** — well-documented, honest about gotchas, solid security defaults.

PR #5297, opened 2026-08-20
fix(sandbox): forward Windows env keys so allow_network works (#4937)

### LINKED ISSUE #4937: [BUG] Sandbox allow_network is broken on Windows: SystemRoot missing from the subprocess env breaks Winsock (WinError 10106)

deepspace28:
Fixes #4937

`_sandbox_env()` builds the sandbox child's environment from `_BASE_ENV_KEYS`. All eight entries are POSIX spellings, so on Windows only `PATH` actually survives into the child. Without `SystemRoot` the Winsock service-provider catalogue cannot load, and every socket call raises `OSError: [WinError 10106]` **before any sandbox rule runs** — so `allow_network=True` silently does nothing on Windows.

Two commits:

1. **`fix(sandbox)`** — add the Windows spellings of keys already on the list (`USERPROFILE`, `TMP`, `TEMP`, `PATHEXT`) plus the keys the OS itself requires (`SYSTEMROOT`, `WINDIR`, `SYSTEMDRIVE`), and a regression test class covering the allowlist behaviour and the socket path.
2. **`test`** — a five-line, fixture-only portability repair for three suites that fail on clean `main` on Windows for reasons unrelated to the fix. Without them, the fix above is not verifiable on Windows at all.

The failure and its isolated cause are documented in the issue (`WinError 10106` without `SystemRoot`, `socket OK` with it). On this branch the two tests that exercise the opt-in — which fail on `main` — now pass, along with the new allowlist tests:

```
$ python -m pytest tests/sandbox/test_runner.py::TestSandboxEnvironment \
    tests/tools/test_python_execution_tool.py::...::test_network_access_can_be_enabled \
    tests/tools/test_python_execution_runtime_inputs.py::test_socket_reachability_check_works_with_allow_network
============================== 7 passed in 5.87s ==============================
```

Full touched-area run: **69 passed**. `ruff check`, `ruff format --check`, and `mypy` on the changed files are all clean. Verified on Windows 11, CPython 3.12.

---

The bug is that the sandbox opt-in `allow_network=True` cannot work on Windows: `_sandbox_env()` forwards an allowlist whose keys are POSIX spellings, so on Windows only `PATH` survives and Winsock cannot initialise without `SystemRoot`. I reproduced this on Windows 11 against `main`, and confirmed by printing the surviving keys that the child really is left with just `PATH` — no temp dir, no CA bundle, no locale.

I considered the minimal fix (add `SystemRoot` only), which does make sockets work, but it leaves the child with no temp directory or CA bundle, so the next Windows report would be the same class of bug with a different error code. So I chose to add the Windows spellings of keys already on the list, plus the ones the OS requires. @Chirag6722 (who filed this issue and shipped the first fix) narrowed the set correctly on a threat-model basis, and I adopted his narrowing: `COMSPEC` stays out because the preamble deliberately blocks subprocess spawning, and `APPDATA`/`LOCALAPPDATA` stay out because they are config-home keys and the list already omits `XDG_CONFIG_HOME`. The allowlist stays a narrow allowlist, not a passthrough — `test_unlisted_variables_are_still_not_forwarded` pins that by asserting `AWS_SECRET_ACCESS_KEY`, `COMSPEC`, `APPDATA` and `LOCALAPPDATA` never cross.

**One addition beyond the six-key scope discussed on the issue: `SYSTEMDRIVE`.** With `SYSTEMROOT` forwarded but `SYSTEMDRIVE` absent, shell-folder lookups leave the string `%SystemDrive%` unexpanded, and Windows then creates a literal directory named `%SystemDrive%` under the child's current working directory — a write *outside* the sandbox root, made by the OS rather than through the guarded `open()`. I hit this directly while testing. `test_sandbox_run_leaves_no_directory_behind` runs the sandbox from an empty cwd and asserts nothing is left behind; `test_systemdrive_travels_with_systemroot` pins the pairing so the two keys can't drift apart. If you'd rather ship strictly the six keys, or add `APPDATA`/`LOCALAPPDATA` as well, I'm happy to adjust — those are allowlist policy calls for a maintainer.

The three fixture fixes in the second commit are the reason the fix is verifiable on Windows: `test_adapters.py` and `test_chat_api.py` built module ids from `Path.relative_to()` (which renders `\` on Windows) and compared them against hardcoded `/` literals — `as_posix()` fixes both; `test_remote_sync.py` fixtures wrote text without `newline=""`, so Windows stored CRLF while the sync engine works in binary (`platform/filestorage/engine.py:178` writes `"wb"`) and the `read_bytes()` assertions compared CRLF against LF. All fixture-only — no production code under test changed. I kept them in this PR rather than a standalone test-only PR because per CONTRIBUTING rule 5 they exist to make this fix pass on the platform it targets.

Credit to @Chirag6722 for the issue, the first fix, and the scope narrowing (co-authored on the fix commit).

---


Chirag6722:
Co-author here — I checked the one change from what we agreed rather than waving it through, since my name is on it.

**`SYSTEMDRIVE` belongs in, but I could not reproduce the mechanism you gave for it.** Windows 11, CPython 3.12, child spawned with a cleared environment, empty cwd, `SHGetFolderPathW(CSIDL_COMMON_APPDATA)`:

```
SYSTEMDRIVE with    : exit=0 | shell_rc=0           val='C:\ProgramData'  | created=[]
SYSTEMDRIVE without : exit=0 | shell_rc=-2147024893 val=''                | created=[]
```

`-2147024893` is `0x80070003`, `ERROR_PATH_NOT_FOUND`. So the lookup does break without the key — that part is solid, and it matches the silent-empty-string symptom you noted earlier in the issue. But no `%SystemDrive%` directory appeared in the child's cwd in either case. What I see is a failed call returning an empty string, not a literal-name write.

That distinction matters because it is load-bearing twice over:

1. The code comment says Windows "creates a `%SystemDrive%` directory under the child's cwd — a write outside the sandbox root, made by the OS rather than through the guarded `open()`". That is the strongest security claim in the diff, and on my box it is not what happens. If a reviewer tries to confirm it and cannot, it costs the whole PR credibility it does not need to spend.
2. **`test_sandbox_run_leaves_no_directory_behind` passes on my machine with `SYSTEMDRIVE` removed from `_BASE_ENV_KEYS`** — because nothing creates a directory either way. So the test written to guard the escape cannot fail for the reason it names. Same for `test_systemdrive_travels_with_systemroot`, which asserts over the tuple's contents rather than over behaviour; it would stay green if the pairing stopped mattering entirely.

Not arguing to drop the key — the empty-string failure is reason enough, and it is arguably the better argument: an API that fails *silently* inside a sandbox is worse than one that raises, because generated code then acts on a wrong path. I would just pin it to what reproduces:

- reword the comment to the silent-failure behaviour, with the `0x80070003` return, and
- replace the directory-absence assertion with one that runs `SHGetFolderPathW` in the child and asserts a non-empty path — that one fails without the key.

If you *did* see the `%SystemDrive%` directory, I would like to know the trigger — different shell-folder API, `expandvars`, or a library calling `makedirs` on the empty result. Entirely possible it is real on a path I did not exercise, in which case keep the wording and add the repro; I would rather be wrong here than have it quietly dropped.

Everything else matches what we discussed: `COMSPEC` and `APPDATA`/`LOCALAPPDATA` correctly left out, and `test_unlisted_variables_are_still_not_forwarded` naming the held-back keys is better than the version I had.

deepspace28:
You were right to check, and you're right about the API. I reproduced your numbers exactly, and the comment names a mechanism that does not do what the comment says:

```
SHGetFolderPathW(CSIDL_COMMON_APPDATA)                    -> 0x80070003, '', created=[]
SHGetFolderPathW(... | CSIDL_FLAG_CREATE)                 -> 0x80070021, '', created=[]
SHGetKnownFolderPath(FOLDERID_ProgramData, KF_FLAG_CREATE)-> 0x80070021, '', created=[]
```

Even with the create flags, nothing is written. So the wording has to change regardless.

## The trigger you asked for

The directory is real, but it needs one thing neither of us stated: **the child interpreter has to be Microsoft Store (MSIX) Python.** Same env, same empty cwd, `import socket; socket.socket().close()`, 3 trials each:

```
Store MSIX 3.12 (direct)   with SYSTEMDRIVE    -> []                 3/3
Store MSIX 3.12 (direct)   WITHOUT SYSTEMDRIVE -> ['%SystemDrive%']  3/3
venv on Store 3.12         with SYSTEMDRIVE    -> []                 3/3
venv on Store 3.12         WITHOUT SYSTEMDRIVE -> ['%SystemDrive%']  3/3
python.org 3.11            with SYSTEMDRIVE    -> []                 3/3
python.org 3.11            WITHOUT SYSTEMDRIVE -> []                 3/3
```

My `.venv` is built on the Store interpreter (`uv` picked up `WindowsApps\PythonSoftwareFoundation.Python.3.12_qbz5n2kfra8p0`). A fresh venv on python.org CPython shows nothing, which is why you saw nothing.

It isn't Python code doing it — `-E -S -c pass` is enough, so it happens at process start before any bytecode runs. What lands is the AppX shell cache:

```
%SystemDrive%\ProgramData\Microsoft\Windows\Caches\cversions.2.db
%SystemDrive%\ProgramData\Microsoft\Windows\Caches\{499E166B-...}.2.ver0x0000000000000001.db
```

`HKLM\...\ProfileList\ProgramData` is a `REG_EXPAND_SZ` holding the literal `%SystemDrive%\ProgramData`. The packaging layer expands it with `ExpandEnvironmentStrings` against the process environment. With `SYSTEMDRIVE` dropped the token stays literal, and `%SystemDrive%\ProgramData\...` has no drive letter and no leading backslash — so it is a **relative** path, and it resolves against the child's cwd. That's the escape, and it is the OS rather than the guarded `open()`, but via the packaging layer, not `SHGetFolderPathW`.

## On the tests — you're right, and one correction

`test_sandbox_run_leaves_no_directory_behind` *does* fail here without the key, 3/3 deterministic. So it isn't vacuous on my box.

But your actual point stands and is the one that matters: whether that test *can* fail depends on which interpreter runs it, and that is invisible from reading the test. On python.org CI it is silently green forever. A guard that only bites on one packaging flavour is worse than no guard, because it reads like coverage.

One thing worth flagging before you adopt your own suggestion — I checked it, and it has the same defect mirrored:

```
Store MSIX 3.12  SHGetFolderPathW WITHOUT SYSTEMDRIVE -> '%SystemDrive%\ProgramData'
python.org 3.11  SHGetFolderPathW WITHOUT SYSTEMDRIVE -> ''
```

On MSIX the call *succeeds* and returns the literal relative path, because the packaging layer already created it under the cwd — so `assert path != ""` passes without the key. Your assertion catches the bug on python.org and misses it on Store; mine did the reverse.

The one assertion that discriminates identically on both:

```
expandvars('%SystemDrive%')  with SYSTEMDRIVE    -> 'C:'            (both interpreters)
expandvars('%SystemDrive%')  WITHOUT SYSTEMDRIVE -> '%SystemDrive%' (both interpreters)
```

That is also the actual invariant — an unexpanded token is a relative path — rather than a proxy for it.

## What I'll change

1. Reword the comment to the packaging-layer mechanism, and drop the `SHGetFolderPathW` framing entirely.
2. Replace `test_systemdrive_travels_with_systemroot` — you're right that asserting over the tuple's contents pins nothing. It becomes the `expandvars` assertion run in the child, which fails without the key on every Windows interpreter.
3. Keep the directory-absence test, but comment it honestly as MSIX-only so nobody reads it as universal coverage.
4. Add your silent-failure case too, since the empty string on python.org is a genuinely distinct failure mode worth naming — just not as the only guard.

All four are pushed. I checked them the way you checked mine — by deleting `SYSTEMDRIVE` from `_BASE_ENV_KEYS` and confirming which tests actually go red:

```
FAILED test_windows_keys_are_forwarded_when_set
FAILED test_child_can_expand_systemdrive
FAILED test_sandbox_run_leaves_no_directory_behind
PASSED test_child_shell_folder_lookup_does_not_fail_silently   <- on MSIX, as above
```

That last line is your suggested assertion passing without the key on this box, in the suite, for the reason described. It's still worth having for python.org — it just can't be the only guard.

Agreed the silent-failure argument is the stronger one for keeping the key: generated code acting on a wrong path is worse than a raise. I'd rather have had this caught now than have a reviewer fail to reproduce it later.

Chirag6722:
Confirmed your invariant on the other interpreter flavour — python.org 3.12.10, child with a cleared environment:

```
SYSTEMDRIVE with    -> 'C:'
SYSTEMDRIVE without -> '%SystemDrive%'
```

So `expandvars` discriminates on both builds, and you're right that my suggested assertion had my own defect mirrored: on MSIX the call succeeds with the literal path and `!= ""` passes without the key. I was generalising from one interpreter, which is exactly what I flagged you for doing — the difference being you'd actually run yours 3/3 and I hadn't run mine at all against the packaging flavour where it matters.

Your four changes are the right resolution, and naming the directory-absence test MSIX-only is better than dropping it. A guard that bites on one packaging flavour is fine as long as the comment says so.

**Heads up on the PR itself: muddlebee closed it at 13:34:58 today, unmerged.** It went inside another batch — six of @cloudenochcsis's in the ~55 seconds before yours, one of @SawhneySatvik's just after:

```
13:34:02  cloudenochcsis  #5305
13:34:18  cloudenochcsis  #5311
13:34:19  cloudenochcsis  #5315, #5313
13:34:20  cloudenochcsis  #5317
13:34:21  cloudenochcsis  #5322
13:34:58  deepspace28     #5297
13:36:36  SawhneySatvik   #5223
```

Same shape as the sweep that took my three. Seventeen hours from open to close, with unresolved review conversation on the thread, so I would not read it as a verdict on the work any more than you told me to read mine that way.

That's twice now for both of us on the same issue. I don't have a read on what triggers it, and I've stopped opening PRs here until a maintainer says something. Your call whether to reopen — the diff is good and the test discussion landed somewhere better than where either of us started.

cloudenochcsis:
Same here — six of mine went in the 13:34 batch @Chirag6722 listed (#5305, #5311, #5313, #5315, #5317, #5322). All six were green on CI with Greptile at 5/5 and no unresolved threads, and the bug reports they fix (#5304, #5310, #5312, #5314, #5316, #5321) are still open, so I also wouldn't read the closes as a verdict on the work. Two of mine did merge the same morning (#5299, #5303), which makes the pattern harder to read, not easier.

@muddlebee — is there a policy for community PRs we should be following right now? Happy to resubmit at whatever pace or batch size works for the team, or to hold off entirely if outside PRs are paused for a while. The branches are all still on my fork.

muddlebee:
@Chirag6722 @cloudenochcsis please take a look at the contribution guidelines and avoid spamming with multiple PRs :) 
Feel free to discuss at discord about the issue and then maybe take on? 

and @Chirag6722 please refrain from posting such AI slop in comments.

PR #5299, opened 2026-08-20
fix(interactive-shell): treat naive received_at as UTC in time_ago

### LINKED ISSUE #5298: [BUG] time_ago crashes on timezone-naive received_at from POST /alerts, dropping queued alerts and killing the alert watcher
### Summary

`time_ago()` in `surfaces/interactive_shell/ui/alerts/__init__.py` crashes with `TypeError: can't subtract offset-naive and offset-aware datetimes` when an alert's `received_at` is timezone-naive — which is exactly what `POST /alerts` produces when a sender supplies a timestamp without an offset.

**Expected:** An alert posted with `"received_at": "2026-08-20T10:00:00"` (no offset — what a plain curl or a naive webhook sender emits) renders in the REPL with a `received Ns ago` line, treating the naive timestamp as UTC like the rest of the codebase does.

**Actual:** `time_ago()` computes `datetime.now(UTC) - then` with an always-aware `now`, so a naive `then` raises `TypeError`. The consequences are worse than one bad render:

- `drain_and_render_incoming()` pops **all** queued alerts via `inbox.iter_pending()` before rendering; the first bad alert raises mid-loop, so every alert behind it in the batch is popped but never rendered nor recorded via `session.record_incoming_alert` — silent alert loss.
- The drain call in `_alert_watcher` (`surfaces/interactive_shell/runtime/background/workers.py`) is not wrapped, so the watcher task dies for the rest of the session.
- `/status` (`command_registry/diagnostics_cmds.py`) also calls `time_ago(most_recent.received_at)` and breaks.

### Steps to reproduce

1. From the repo root:

```
uv run python -c "
from core.domain.alerts.inbox import IncomingAlert
from surfaces.interactive_shell.ui.alerts import time_ago
alert = IncomingAlert.model_validate({'text': 'disk full', 'received_at': '2026-08-20T10:00:00'})
print('parsed tzinfo:', alert.received_at.tzinfo)
print(time_ago(alert.received_at))
"
```

2. Output:

```
parsed tzinfo: None
TypeError: can't subtract offset-naive and offset-aware datetimes
```

The same state arrives end-to-end: `gateway/web/webapp.py` only fills `received_at` when the sender omitted it, so a supplied offset-less ISO string is parsed by pydantic into a naive datetime and flows into the inbox as-is.

### Root cause / proposed fix

The repo already has the canonical guard for this in `surfaces/shared/terminal/components/time_format.py`:

```python
if dt.tzinfo is None:
    dt = dt.replace(tzinfo=UTC)
```

`time_ago()` is the sibling that omits it. Three-line fix mirroring the existing precedent, plus a regression test. Happy to open a PR.

### Environment

- opensre @ current main (9df273155)
- Python 3.x via `uv run`, macOS

cloudenochcsis:
Fixes #5298

`time_ago()` computed `datetime.now(UTC) - then` with an always-aware `now`, but `POST /alerts` keeps sender-supplied timestamps, so an offset-less ISO string (`"2026-08-20T10:00:00"`) reaches it as a **naive** datetime and the subtraction raises `TypeError`. Because `drain_and_render_incoming()` pops all queued alerts before rendering, the first bad alert silently drops every alert behind it in the batch and kills the `_alert_watcher` task for the rest of the session; `/status` breaks the same way.

The fix mirrors the guard the codebase already uses in `format_repl_timestamp` (`surfaces/shared/terminal/components/time_format.py`): if `then.tzinfo is None`, treat it as UTC. Added a regression test alongside the existing `TestTimeAgo` cases.

Before (main):

```
$ uv run python -c "
from core.domain.alerts.inbox import IncomingAlert
from surfaces.interactive_shell.ui.alerts import time_ago
alert = IncomingAlert.model_validate({'text': 'disk full', 'received_at': '2026-08-20T10:00:00'})
print(time_ago(alert.received_at))
"
TypeError: can't subtract offset-naive and offset-aware datetimes
```

After (this branch):

```
parsed tzinfo: None
renders as: received 8h ago
```

Focused tests + baseline checks (per CI.md):

```
uv run python -m pytest tests/interactive_shell/ui/test_incoming_alerts.py -q   # 26 passed
make lint          # All checks passed!
make format-check  # 3604 files already formatted
make typecheck     # Success: no issues found in 1860 source files
```

---

The bug is a missing naive-datetime guard on one of the sibling time-formatting paths: `format_repl_timestamp` normalizes naive datetimes to UTC before doing timezone math, but `time_ago` does not, and `received_at` is the one datetime field that can legitimately arrive naive (the gateway only fills it in when the sender omits it). I considered normalizing at the model boundary (a pydantic validator on `IncomingAlert.received_at`) but chose the renderer-side guard because it matches the existing in-repo precedent, keeps the stored value faithful to what the sender supplied, and also covers the `/status` path. The regression test constructs a naive datetime three minutes in the past and asserts the rendered relative time instead of asserting "does not raise".

---

### DIFF
--- a/surfaces/interactive_shell/ui/alerts/__init__.py
+++ b/surfaces/interactive_shell/ui/alerts/__init__.py
@@ -32,6 +32,11 @@ def time_ago(then: datetime | None) -> str:
     if then is None:
         return "unknown"
 
+    # POST /alerts accepts sender-supplied timestamps, so ``received_at`` may be
+    # offset-naive; treat it as UTC like format_repl_timestamp does.
+    if then.tzinfo is None:
+        then = then.replace(tzinfo=UTC)
+
     now = datetime.now(UTC)
     delta = now - then
     seconds = int(delta.total_seconds())
--- a/tests/interactive_shell/ui/test_incoming_alerts.py
+++ b/tests/interactive_shell/ui/test_incoming_alerts.py
@@ -52,6 +52,13 @@ class TestTimeAgo:
         result = time_ago(None)
         assert result == "unknown"
 
+    def test_naive_datetime_treated_as_utc(self) -> None:
+        # POST /alerts keeps sender-supplied timestamps, so an offset-less ISO
+        # string arrives here as a naive datetime; it must not raise.
+        then = datetime.now(UTC).replace(tzinfo=None) - timedelta(minutes=3)
+        result = time_ago(then)
+        assert "3m ago" in result
+
 
 class TestFormatIncomingAlert:
     """Test format_incoming_alert rendering."""


cloudenochcsis:
@greptile review

PR #5317, opened 2026-08-21
fix(telegram): truncate after HTML conversion so long answers keep formatting

### LINKED ISSUE #5316: [BUG] Telegram gateway truncates before HTML conversion — long answers exceed 4096 and silently fall back to unformatted text
### Summary

The Telegram gateway finalizes a turn by truncating the answer to 4096 chars **before** converting Markdown to HTML. HTML conversion only ever lengthens the text (adds `<b>`/`<a>` tags, escapes `&<>`), so any long answer produces an HTML payload over Telegram's 4096 limit — the API rejects it, and `_edit_final`/`_send_final` silently retry with the unformatted plain text. Every long answer loses all formatting and wastes one failed API call.

**Expected:** a long answer arrives formatted, truncated tag-safely to 4096 — which is exactly what `truncate_for_telegram_html` (`integrations/telegram/delivery.py`) exists for, and what the scheduler's Telegram delivery already does (`markdown_to_telegram_html` first, then `truncate_for_telegram_html(html, 4096)` in `platform/scheduling/scheduler/executor.py`).

**Actual:** `gateway/transports/telegram/turn_output.py::_finalize` does it in the reverse order:

```python
final = truncate(answer, MAX_MESSAGE_SIZE, suffix="…")      # 4096 plain chars
html_final = markdown_to_telegram_html(final)               # expands past 4096
```

Telegram rejects `html_final`, and the fallback sends `final` unformatted.

### Steps to reproduce

From the repo root:

```
uv run python -c "
from platform.text.truncation import truncate
from platform.delivery.notifications.limits import MAX_MESSAGE_SIZE
answer = chr(10).join(f'- **service-{i}** restarted \`pod-{i}\` at <t={i}> & recovered' for i in range(80))
final = truncate(answer, MAX_MESSAGE_SIZE, suffix='…')
html_final = markdown_to_telegram_html(final)
print(len(final), '->', len(html_final), 'exceeds 4096:', len(html_final) > MAX_MESSAGE_SIZE)
"
```

Output:

```
4096 -> 5800 exceeds 4096: True
```

### Root cause / proposed fix

Convert first, then truncate tag-safely — mirror the scheduler sibling: `html_final = truncate_for_telegram_html(markdown_to_telegram_html(answer), MAX_MESSAGE_SIZE, suffix="…")`, keeping the plain `truncate(answer, ...)` only as the plain-text fallback. ~4 lines + a regression test asserting the HTML payload never exceeds 4096. Happy to open a PR.

### Environment

- Python via `uv run`, macOS

cloudenochcsis:
Fixes #5316

`_finalize` in the Telegram gateway truncated the answer to 4096 chars **before** converting Markdown to HTML. Conversion only ever lengthens the text (`<b>`/`<a>` tags plus `&<>` escaping), so any answer near the cap produced an HTML payload over Telegram's 4096 limit — the API rejected it, and `_edit_final`/`_send_final` silently retried with the unformatted plain text. Every long answer lost all formatting and wasted one failed API call.

Fix mirrors the scheduler's Telegram delivery (`platform/scheduling/scheduler/executor.py`), which already does this in the right order: convert the full answer first, then cut tag-safely with `truncate_for_telegram_html` — the helper that exists precisely for this (and whose tag-balancing was already hardened in #4943). The plain `truncate(answer, …)` stays as the plain-text fallback and log preview. Regression test pins that a long Markdown answer goes out as HTML within the limit with `parse_mode="HTML"`, not via the plain-text fallback.

Before (main) — the oversized payload that Telegram rejects:

```
$ uv run python -c "...truncate then convert..."
4096 -> 5800 exceeds 4096: True     # html_final length after converting the truncated text
```

After (this branch) — captured from the regression test's fake client:

```
len(html_sent) = 4096 (<= MAX_MESSAGE_SIZE)
html_sent starts: '<b>service-0</b>' ... parse_mode='HTML'
```

Regression test verified to fail on unfixed code:

```
git stash push gateway/transports/telegram/turn_output.py
uv run python -m pytest gateway/tests/telegram/test_telegram_turn_output.py -k stays_within_limit -q   # 1 failed
git stash pop
uv run python -m pytest gateway/tests/telegram/ -q                                                     # 73 passed
```

Baseline checks (per CI.md): `make lint` All checks passed! / `make format-check` 3618 files already formatted / `make typecheck` Success: no issues found in 1872 source files.

---

The formatted and plain variants serve different purposes, so I kept both: `html_final` is now converted from the full answer and truncated tag-safely (so the cut can't land mid-tag or mid-entity — `truncate_for_telegram_html` handles both), while `final` remains the plain-text 4096 truncation used only as the API-rejection fallback and the log preview. I considered truncating the Markdown source with a length budget that anticipates expansion, but the expansion factor depends on content (link-heavy text expands most), so cutting the rendered HTML with the purpose-built helper is both simpler and exact. The test drives the public `finalize` path against a fake client and asserts on the actual payload and `parse_mode`.

---


cloudenochcsis:
@greptile review

cloudenochcsis:
Rebased onto the #5332 refactor via merge: `finalize` now lives in `_TelegramChannel.deliver_final`, and the refactor carried the truncate-before-convert ordering with it — the fix is re-applied there (convert the full answer, then `truncate_for_telegram_html`). The regression test ported unchanged and all 73 telegram gateway tests pass.

PR #5319, opened 2026-08-21
Move the Cloudflare install-proxy into infrastructure/deployment/ and finish rename cleanup

YauhenBichel:
## What

Follow-up to the `platform` → `infrastructure` rename (#5295). Relocates the
top-level `infra/cloudflare_install_proxy/` into
`infrastructure/deployment/cloudflare_install_proxy/`, and fixes the last few
`platform` references #5295 missed.

## Why

After #5295 the repo had both `infrastructure/` (the first-party runtime
package) and a top-level `infra/` (a Cloudflare Worker for the install
redirect). The two names are one letter apart and easy to confuse. They are
also different artifact types — `infrastructure/` is Python the app imports;
the worker is edge JavaScript deployed via wrangler — so they cannot merge,
only be disambiguated. Consolidating the worker under the existing
`infrastructure/deployment/` home removes the top-level `infra/` entirely.
It goes under `deployment/` (not a new `deploy/`) to avoid a
`deploy/`-vs-`deployment/` sibling that would reintroduce the same confusion

## How

- `git mv infra/cloudflare_install_proxy infrastructure/deployment/cloudflare_install_proxy`
  (tracked as renames; the worker has no `__init__.py`, so it is not a Python
  import package and does not affect typecheck or the wheel).
- Repointed the references: the worker's own README, `infrastructure/README.md`,
  and the `docs/ARCHITECTURE.md` note (now describes `infrastructure/deployment/`
  instead of the removed `infra/`).
- Updated `release.yml`'s `paths-ignore` to the new path so a
  worker-only change still does **not** trigger an opensre binary release.
- Trailing rename cleanup: `platform` → `infrastructure` references in
  `core/state/README.md` and `tests/README.md`, plus CI path-filter/scope
  completeness fixes (`bootstrap/**`, `gateway`) in `ci.yml` and `ruff.toml`.


YauhenBichel:
@greptile review

PR #5323, opened 2026-08-21
feat(llm): add agent_llm_is_cli_backed function to determine CLI-backed LLM usage

### LINKED ISSUE #5239: Stop constructing an LLM inside get_investigation_agent_class
`tools/investigation/stages/gather_evidence/agent.py::get_investigation_agent_class` calls `default_llm_factory()` and builds a full client only to `isinstance`-check it against `CLIBackedAgentClient` and pick a policy class — an expensive side effect for a routing decision, and a hidden global dependency.

**Change**: derive the CLI-backed/hosted distinction from configuration at the composition root and pass it down as an explicit flag (or pass the already-built client in). No client construction inside the selector.

**Verify**: investigation gather tests green; `default_llm_factory` is no longer imported by this module.

---
**Ground rules for this change**
- New or changed Protocol methods use a docstring-only body — no `...`, no `pass`, no `raise NotImplementedError` (AGENTS.md, Code Style). Do not convert pre-existing `NotImplementedError` stubs elsewhere — out of scope.
- Remove any retired name or alias completely in the same PR — no compatibility re-exports.
- CI's typecheck does not cover `tests/`: run `uv run mypy <touched test files>` yourself.
- No new module-level mutable state; dependencies arrive via constructor or parameter.
- Before opening the PR follow CI.md (lint, format, typecheck, tests) and fill the PR template, including the AI-usage disclosure.

Devesh36:
/fix #5239 
- Introduced `agent_llm_is_cli_backed` to check if the agent LLM is routed to a CLI subprocess backend.
- Updated `__all__` exports to include the new function.
- Enhanced tests to validate the behavior of CLI-backed LLM routing.
- Modified relevant components to utilize the new function for agent class selection based on routing configuration.


Davidson3556 (approved):
LGTM. The swap is equivalent: `build_agent_client` returns `CLIBackedAgentClient` exactly when `cli_provider_registration is not None`, so the decision is unchanged and the client construction is gone.

**One question.** `capability.py:372` is the live streaming path. On main it hardcodes `ConnectedInvestigationAgent`; now it picks the policy. So CLI-backed users get the stricter "call every planned tool before concluding" behaviour in interactive runs.

Intended? Looks like a real fix to me, but it sits outside #5239 and has no test. The new coverage only reaches `run_connected_investigation`.

Two nits, non-blocking:

1. `cli_backed=agent_llm_is_cli_backed()` is duplicated in `lifecycle.py` and `capability.py`. A third caller has to remember it, and forgetting is silent.
2. `_fail_build` patches `build_agent_client`, but `get_llm` checks the cache first, so a cached client slips past the guard. Patching `get_llm` closes it.

Devesh36:
@greptile review

PR #5327, opened 2026-08-21
docs: correct architecture docs after the platform->infrastructure rename

YauhenBichel:
## What

Fixes accuracy gaps in the architecture docs that remained after the
`platform` -> `infrastructure` rename (#5295) and the gateway turn-handler
unification. Documentation only — no code changes.

## Why

An audit of the architecture docs against the current code found the rename was
almost fully propagated, but a few concrete defects and gaps survived: a wrong
tier count, a stale mermaid identifier, an undocumented core subpackage, and a
peer rule that no longer matches the enforced import graph.


YauhenBichel:
@greptile review

Issue #5330, opened 2026-08-21
Architecture: invert scheduler->vendor delivery behind a delivery port


YauhenBichel:
## Problem

`.importlinter.strict` lists seven `infrastructure -> integrations` debt edges;
five are the scheduler executor hard-depending on concrete vendor delivery:

    infrastructure.scheduling.scheduler.executor -> integrations.discord.delivery
    infrastructure.scheduling.scheduler.executor -> integrations.rocketchat.delivery
    infrastructure.scheduling.scheduler.executor -> integrations.slack.delivery
    infrastructure.scheduling.scheduler.executor -> integrations.telegram.delivery
    infrastructure.scheduling.scheduler.executor -> integrations.telegram.formatting

Infrastructure (tier 4) reaches up into integrations (tier 3), and the executor
knows four vendors by name — adding a fifth vendor edits the scheduler.

## Fix

Define a delivery **Protocol/port** the scheduler depends on (e.g.
`ScheduledDelivery` with `send(...)`), register each vendor as an adapter at a
composition root (`bootstrap/adapters.py`), and have the executor resolve the
adapter by channel rather than importing vendor modules. Collapses five of the
seven debt edges and makes the scheduler vendor-agnostic.

## Acceptance

- `scheduler.executor` imports no `integrations.*` module; the five ledger
  entries are removed; adding a new delivery vendor needs no scheduler change.

YauhenBichel:
this is done https://github.com/Tracer-Cloud/opensre/commit/5322749e732659e7de6a930cc42006c24a582d65

PR #5335, opened 2026-08-21
refactor(scheduler): invert vendor delivery behind a delivery-adapter bundle (#5330)

YauhenBichel:
## What

Removes the scheduler executor's five hard dependencies on concrete vendor
delivery modules by inverting them behind a per-provider adapter port. The
executor now resolves a `Provider` to a `ScheduledDeliveryAdapter` through an
immutable bundle instead of importing `integrations.<vendor>.delivery`.

## Why

`.importlinter.strict` carried seven `infrastructure -> integrations` debt
edges; five were `scheduler.executor` reaching up into vendor delivery:

    scheduler.executor -> integrations.{discord,rocketchat,slack,telegram}.delivery
    scheduler.executor -> integrations.telegram.formatting

Infrastructure (tier 4) importing integrations (tier 3) is an upward edge, and
the executor knew four vendors by name — a fifth delivery vendor meant editing
the scheduler. This collapses all five edges and makes the scheduler
vendor-agnostic.

## How

- New `ScheduledDeliveryAdapter` (Protocol) + `ScheduledDeliveryAdapters`, an
  immutable value assembled at a composition root and installed once. This
  mirrors the existing `runners.py:SchedulerRunners` pattern rather than adding
  a mutable registry (see the note below).
- Each `_deliver_<vendor>` body moves into a vendor adapter in
  `integrations/<vendor>/scheduled_delivery.py`, where importing that vendor's
  delivery is a local (not cross-tier) import. `interactive_shell` — which uses
  no vendor — becomes an `infrastructure` adapter.
- `executor._deliver_single` becomes a bundle lookup; the five `_deliver_*`
  functions and every `integrations` import are gone from the executor. Shared
  helpers (`strip_html`, `resolve_slack_delivery_chat_id`) were extracted to
  `scheduler/delivery.py` so both the executor's fan-out routing and the Slack
  adapter use them.
- Composition root: `bootstrap.adapters.scheduled_delivery_adapters()` builds
  the bundle (the only layer allowed to see both tiers);
  `install_scheduled_delivery_adapters()` binds it. It runs in the
  `SCHEDULER_RUNNERS` boot step (scheduler-worker + all CLI scheduled commands +
  sentry-digest) and in the gateway controller, which installs the runners
  imperatively rather than through the boot step.
- Removed the five `.importlinter.strict` entries.

## A note on the DI mechanism

The scheduler layer wires seams through module-level installers
(`SchedulerRunners`, `agent_runner`, `outbound_registry`), which sits in tension
with the "inject via constructor/parameter" preference. Of the register-at-boot
options, this uses the **immutable, set-once bundle** (`SchedulerRunners`-style)
rather than a mutable per-provider registry: installing a new bundle atomically
replaces the mapping, so there is no read-modify-write module state. The strict
alternative — threading a `Mapping[Provider, adapter]` through `runner.py` and
both `execute_task` call sites — was set aside as inconsistent with the four
sibling seams and higher blast radius.

### DIFF
--- a/.importlinter.strict
+++ b/.importlinter.strict
@@ -34,11 +34,6 @@ ignore_imports =
     infrastructure.analytics.cli -> integrations.github.identity
     infrastructure.observability.errors.sentry -> integrations.llm_cli.errors
     infrastructure.scheduling.scheduler.credentials -> integrations.catalog
-    infrastructure.scheduling.scheduler.executor -> integrations.discord.delivery
-    infrastructure.scheduling.scheduler.executor -> integrations.rocketchat.delivery
-    infrastructure.scheduling.scheduler.executor -> integrations.slack.delivery
-    infrastructure.scheduling.scheduler.executor -> integrations.telegram.delivery
-    infrastructure.scheduling.scheduler.executor -> integrations.telegram.formatting
     # surfaces -> gateway (cli + repl gateway commands)
     surfaces.cli.commands.gateway -> gateway.core.process
     surfaces.gateway_entry -> gateway.core.lifecycle.controller
--- a/bootstrap/adapters.py
+++ b/bootstrap/adapters.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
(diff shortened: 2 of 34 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

Issue #5178, opened 2026-08-19
[BUG] format_token_count_compact renders '1000k' instead of '1M' at the million boundary


cloudenochcsis:
## Description

`format_token_count_compact()` in `core/agent_harness/session_goal/progress.py` picks the "k" vs "M" suffix branch by comparing the *raw* integer against `1_000_000`, but formats the "k" value at one-decimal precision afterward. For counts in `[999_950, 999_999]`, `value / 1000.0` rounds to `1000.0` at `.1f` precision, but the branch was already committed to "k" — so the function emits the literal (broken) string `"1000k"` instead of rolling over into `"1.0M"` / `"1M"`.

This renders in the `/goal` status line's token-spend readout (chat/gateway output during a long-running SessionGoal), so any session that crosses ~1,000,000 tokens near that boundary shows a malformed count.

## Repro

```python
from core.agent_harness.session_goal.progress import format_token_count_compact

for v in (999_499, 999_949, 999_950, 999_990, 999_999, 1_000_000):
    print(v, "->", format_token_count_compact(v))
```

Output:
```
999499 -> 999.5k
999949 -> 999.9k
999950 -> 1000k      # bug: should be 1M
999990 -> 1000k      # bug: should be 1M
999999 -> 1000k      # bug: should be 1M
1000000 -> 1M
```

## Expected

`999_950`–`999_999` should render as `1M`, matching `1_000_000`, since they all round to the same one-decimal displayed value.

## Root cause

The branch condition (`value < 1_000_000`) uses the unrounded value, while the rendered text rounds to one decimal. Comparing the *rounded* magnitude against the branch boundary fixes it.

Existing test coverage (`tests/core/agent_harness/session_goal/test_progress.py::test_format_duration_and_token_compacts`) only checks `150`, `1200`, and `3_400_000` — nothing near the boundary, so this slipped through.

I'll open a PR with the fix and a regression test at the boundary.

PR #5346, opened 2026-08-21
fix(cli): round diagnose validity percent instead of truncating

### LINKED ISSUE #5310: [BUG] validity score percent truncates instead of rounding: 0.29 renders as 28%
### Summary

`_validity_score_percent` in `surfaces/shared/terminal/stream_renderer/formatting.py` truncates instead of rounding, so a diagnosis validity score of 0.29 displays as `28%` and 0.57 as `56%` — one point low whenever the score ×100 lands just under an integer in IEEE-754.

**Expected:** `validity:29%` for a 0.29 score.

**Actual:** `f"{int(v * 100)}%"` truncates, and `0.29 * 100 == 28.999999999999996`, so the diagnose resolution line renders `validity:28%`. Same for 0.57 → `56%`, 0.58 → `57%`, and any score whose float product falls just below the integer.

### Steps to reproduce

From the repo root:

```
uv run python -c "
from surfaces.shared.terminal.stream_renderer.formatting import _validity_score_percent
for s in (0.29, 0.57, 0.58): print(s, '->', _validity_score_percent(s))
"
```

Output:

```
0.29 -> 28%
0.57 -> 56%
0.58 -> 57%
```

### Root cause / proposed fix

One-line fix: `return f"{round(v * 100)}%"`. Same round-before-display class as #5178/#5179. Happy to open a PR.

### Environment

- Python via `uv run`, macOS

Heisenberg208:
IEEE-754 makes 0.29*100 land just below 29, so int() displayed 28%. Round before formatting.

Fixes #5310 

image

---

- [ ✅] Yes, I used AI assistance (continue below)

- [ ✅] I have reviewed every single line of the AI-generated code
- [ ✅] I can explain the purpose and logic of each function/component I added
- [ ✅] I have tested edge cases and understand how the code handles them
- [✅ ] I have modified the AI output to follow this project's coding standards and conventions

The diagnose line shows a 0–1 validity score as a percent. _validity_score_percent already validates, rejects non-finite values, and clamps to [0, 1]. Display used int(v * 100), which truncates. For values like 0.29, IEEE-754 makes 0.29 * 100 equal 28.999…, so int() printed 28% instead of 29%.

Approach: keep that helper as the only formatter, and round after clamp:

return f"{round(v * 100)}%"
StreamRenderer._build_node_message still does validity:{pct} for diagnose / diagnose_root_cause. No renderer or pipeline change.

Why not the alternatives

Decimal / string rounding — heavier than a display helper needs; the score is already a float.
math.ceil / format(..., '.0f') — ceil would inflate scores; :.0f also rounds, but round() is the same class of fix as #5178/#5179 and stays a one-token change.
Changing callers — every diagnose percent goes through this helper, so one site is enough.
Test: test_node_message_for_diagnose_rounds_float_percent drives _build_node_message with validity_score=0.29 and asserts validity:29%. That is the user-visible line, not only the private helper. Existing diagnose tests still cover 0.92, non-numeric, and NaN.

---

---

### DIFF
--- a/surfaces/shared/terminal/stream_renderer/formatting.py
+++ b/surfaces/shared/terminal/stream_renderer/formatting.py
@@ -64,4 +64,4 @@ def _validity_score_percent(score: Any) -> str | None:
     if not math.isfinite(v):
         return None
     v = max(0.0, min(1.0, v))
-    return f"{int(v * 100)}%"
+    return f"{round(v * 100)}%"
--- a/tests/shared/terminal/stream_renderer/test_renderer.py
+++ b/tests/shared/terminal/stream_renderer/test_renderer.py
@@ -221,6 +221,14 @@ class TestStreamRendererUpdatesMode:
         assert "92%" in msg
 
     @patch.dict(os.environ, {"TRACER_OUTPUT_FORMAT": "text"})
+    def test_node_message_for_diagnose_rounds_float_percent(self) -> None:
+        """0.29 * 100 is 28.999… in IEEE-754; display must round, not truncate."""
+        renderer = StreamRenderer()
+        renderer._final_state = {"validity_score": 0.29}
+        msg = renderer._build_node_message("diagnose_root_cause")
+        assert msg == "validity:29%"
+
+    @patch.dict(os.environ, {"TRACER_OUTPUT_FORMAT": "text"})
     def test_node_message_for_diagnose_skips_non_numeric_validity(self) -> None:
         renderer = StreamRenderer()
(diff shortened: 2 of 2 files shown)


Heisenberg208:
@greptile review

Davidson3556 (approved):
Reviewed the exact latest head against current main. The validity percentage now rounds instead of truncating, the focused merged-main suite passes (55 tests), and there are no unresolved review threads or blocking findings. LGTM.

PR #5352, opened 2026-08-21
Restore config/ to a leaf: remove 3 upward import edges (#5331)

### LINKED ISSUE #5331: Architecture: restore config/ to a leaf (remove 3 upward import edges)
## Problem

`config` is the bottom tier and is supposed to import no first-party package,
but `.importlinter.strict` allowlists three upward edges:

    config.config -> core.llm.providers.azure_openai
    config.llm_auth.credentials -> integrations.llm_cli.registry
    config.repl_config -> infrastructure.terminal.theme

These make `config` depend on higher tiers, weakening it as a safe leaf every
layer can import.

## Fix

Move the borrowed symbols down or invert the dependency: e.g. the theme value
`config.repl_config` needs from `infrastructure.terminal.theme` should live in
(or be injected from) config, not pulled up. Same pattern for the llm/provider
and credential registry edges.

## Acceptance

- `config` imports no `core` / `integrations` / `infrastructure` module; the
  three ledger entries are removed; `make check-imports` stays green.

YauhenBichel:
Fixes #5331

`config/` is the bottom tier and must import no first-party package, but
`.importlinter.strict` allowlisted three upward edges. This PR removes all three
structurally and deletes their ledger entries; `make check-imports` now reports
**"Config is independent KEPT"**.

**`config.config -> core.llm.providers.azure_openai`**
Moved the pure URL normalizer down into `config/llm_auth/azure.py`
(`normalize_azure_openai_base_url`). Core now imports it from config instead of
config borrowing it from core.

**`config.repl_config -> infrastructure.terminal.theme`**
Split the theme concern along the layer boundary. The *vocabulary* is now
config-owned in `config/constants/repl_theme.py`: `Theme(StrEnum)` (16 names in
display order), `DEFAULT_THEME_NAME`, `THEME_NAMES`. Infra `terminal/theme.py`
imports those and keeps only the hex palettes; `list_theme_names()` returns the
config tuple. The `set_active_theme` apply side effect left config entirely —
`ReplConfig.load()` is now pure (the `apply_active_theme` parameter is gone), and
the shell/CLI entry points (`run_repl`, `run_repl_async`, the subcommand path in
`app.py`) apply the resolved theme. `layout` and `theme` fields are now enum-typed
(`Layout(StrEnum)`, `Theme`); invalid input defaults, and unknown env/file values
warn as before.

**`config.llm_auth.credentials -> integrations.llm_cli.registry`**
Inverted the dependency behind a config-defined port `config/llm_auth/cli_auth.py`
(`CliAuthState`, a frozen `CliAuthChecker` installed once — same pattern as the
scheduler delivery bundle). Integrations fills it
(`integrations/llm_cli/auth_check.py::check_cli_auth`) and the composition root
installs it via `bootstrap.adapters.install_cli_auth_checker`. `credentials.status()`
reads the port with `resolve_cli_auth_state()` instead of importing integrations.

The checker is installed on two paths because `CLI_PROFILE` boots env only:
`_run_harness_adapters` covers the gateway/web/scheduler/embedded profiles, and
`surfaces/cli/startup.py` covers the bare CLI so `opensre doctor` / `opensre config`
still reach installed CLI providers (codex, claude-code). The install binds a lazy
closure, so the integrations subprocess client is imported only on first use — a
plain `opensre --help` pays nothing.

A conformance test pins `THEME_REGISTRY` keys equal to `THEME_NAMES` so the two
tiers cannot drift, and `tests/cli/test_startup.py` pins that the CLI path installs
the checker (the test fails if that wiring is removed).

`make check-imports`:
Config is independent KEPT
...
Contracts: 4 kept, 0 broken.
No import cycles found across 16074 first-party modules.
All 3 import checks passed.

Theme resolution (`uv run opensre`):
$ OPENSRE_THEME=chartreuse uv run opensre --no-interactive
OPENSRE_THEME='chartreuse' is not a valid theme; defaulting to 'blue'.

CLI-auth on the bare CLI path (the case the port had to keep working):
$ LLM_PROVIDER=claude-code uv run opensre doctor
✓  llm_provider   provider=claude-code, auth=cli (Authenticated via Claude ...)

`ruff` and `mypy` clean across all source files; regression green across
config/bootstrap/shared/cli/infra/integrations/scheduler/shell tests.

---

The problem is a layering violation: `config/` is the leaf tier, but three modules
reached up into core/integrations/infrastructure, tracked as debt in the import
ledger. Each edge needed either the borrowed symbol pulled down into config, or the
dependency inverted behind a config-owned port.

The azure edge was a pure function with no dependencies, so moving it down was the
whole fix. For the theme edge I separated the theme name vocabulary (domain data
that belongs in the leaf and that config needs to validate against) from the hex
palettes and the global-state apply (both rendering concerns that belong in
infra/surfaces); a conformance test guards against the two tiers drifting. For the
credentials edge the live CLI auth check is genuine integration behavior config
cannot own, so I inverted it with the same frozen-bundle-installed-once port used
for scheduler delivery, rather than a mutable registry or threading the dependency
through nine call sites. Because the minimal CLI profile boots env only, I install
the checker at CLI startup too, with a lazy closure so the common fast path pays no
import cost.

Key pieces: `config/constants/repl_theme.py` (`Theme`/`Layout` vocabulary),
`config/llm_auth/cli_auth.py` (the CLI-auth port), `integrations/llm_cli/auth_check.py`
(the adapter that fills it), and `bootstrap.adapters.install_cli_auth_checker` (boot
wiring). Behavior is preserved: theme resolution still warns and defaults identically,
and `status()` returns the same results for CLI providers.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5356, opened 2026-08-21
Refactor headless-agent construction: collapse `ask` onto AgentSession, own the build-config expansion

YauhenBichel:

Two related cleanups of how headless agents are constructed, both toward one
clear narrative for building an agent. No behavior change.

**1. Collapse the CLI `ask` path onto `AgentSession`.**
`ask` was hand-rolling the whole build: a private `_BoundDispatcher`, a manual
`startup()`, a manual capability-stripping loop, its own `DefaultToolProvider` +
`DefaultHeadlessBuild(...).agent(...)`, and a hand-assembled `TurnBinding` — the
exact recipe `AgentSession` documents that hosts must not re-copy.
`AgentSession.start` gained a `tool_hooks` parameter (the per-turn approval
hooks, bound on the first turn alongside `is_tty`) and a `bound_session`
accessor. `ask` now reads: `AgentSession.start(config, output=…,
prepare_session=_restrict_ask_capabilities, gather=…, console=…, is_tty=False,
tool_hooks=hooks)` then `chat_until_goal(prompt)`. The inline capability loop
became a named `_restrict_ask_capabilities` passed as `prepare_session`.

`chat_until_goal` (not `chat`) preserves the previous behavior exactly: a
multi-step `ask` can attach a session goal via an `assistant_handoff`, and it
must run the session-goal loop to completion rather than stop after the first
turn. This is `AgentSession`'s public goal-loop verb, wrapping the same
`run_until_session_goal` the old hand-rolled dispatcher reached through
`HeadlessAgent.handle`.

**2. One owner for the AgentBuildConfig expansion.**
The gateway pool and the shell builder each hand-copied the same "resolve
`build_tools` / `build_prompts` / `build_gather`, falling back to defaults"
logic. Extracted `resolve_agent_ports(config, …) -> (tools, prompts, gather)` in
`headless_build.py`; both call it, then pass the result to
`DefaultHeadlessBuild(...).agent(...)`. Capability policy stays with the callers
on purpose: the pool applies it every turn (the session is re-resolved even on a
cached agent), so folding it into the shared expander would skip it on cache
hits — the docstring records this.

Single-turn `ask` end-to-end:
$ LLM_PROVIDER=claude-code uv run opensre --json ask "reply with exactly PONG"
{"status": "success", "response": "PONG", "denied_tools": [], "error": null}

Full suite: `15440 passed, 66 skipped, 1 xfailed` (`pytest -n auto`, CI-parity
ignores). `make check-imports` green (new `headless_build → agent_build_config`
edge, no cycle). ruff lint + `ruff format --check` + mypy all clean.

### Behavior preservation notes

- Multi-step `ask` still completes its session goal (guarded by a test that
  fails if the dispatch verb is reverted to single-turn `chat`).
- The gateway pool and shell build agents exactly as before — `DefaultHeadlessBuild`
  stays at both call sites, so no construction-seam test infra changed.

---

Both changes remove duplication of the agent-construction recipe without changing
what gets built. For `ask`, the private dispatcher and hand-rolled build existed
only to bind approval hooks and run the session-goal loop; giving
`AgentSession.start` a `tool_hooks` parameter let it collapse onto the public API,
the capability policy moved to the `prepare_session` hook that already runs at the
right moment, and `chat_until_goal` restores the goal-loop semantics through a
documented verb. For the expansion I first over-reached and had the shared helper
build the whole agent — that moved `DefaultHeadlessBuild` out of the call sites and
broke the ~30 tests that stub that construction seam, so I corrected to a
`(tools, prompts, gather)` producer (the milestone's actual spec), which keeps
`DefaultHeadlessBuild` at both call sites. The subtle correctness point is that
capability policy cannot live in the shared expander: the pool must apply it every
turn, including cache hits, and the expander only runs on a cache miss.

Key pieces: `AgentSession.start(tool_hooks=…)` / `bound_session`
(`core/agent_harness/harness.py`), `_restrict_ask_capabilities` +
`AgentSession.start` / `chat_until_goal` usage (`surfaces/cli/ask/service.py`), and
`resolve_agent_ports` (`core/agent_harness/turns/headless_build.py`) with its two
callers (`infrastructure/turn_host/session_agents.py`,
`surfaces/interactive_shell/runtime/shell_agent.py`). New tests pin the
security-relevant `ask` wiring (hooks bound + capabilities stripped), that `ask`
runs the goal loop, and the `resolve_agent_ports` fallback contract.

### DIFF
--- a/core/agent_harness/harness.py
+++ b/core/agent_harness/harness.py
@@ -40,6 +40,7 @@ if TYPE_CHECKING:
     from core.agent_harness.session_goal.run_until import SessionGoalRunResult
     from core.agent_harness.turns.gather_phase import GatherPhase
     from core.agent_harness.turns.turn_results import TurnResult
+    from core.tool.execution import ToolExecutionHooks
 
 
 class ChatDispatcher(Protocol):
@@ -125,6 +126,7 @@ class AgentSession:
         logger: logging.Logger | None = None,
         surface: str | None = None,
         is_tty: bool | None = None,
+        tool_hooks: ToolExecutionHooks | None = None,
     ) -> AgentSession:
         """Return a session that is ready to :meth:`chat`.
 
@@ -147,8 +149,9 @@ class AgentSession:
         and before the agent is built. ``console``, ``logger`` and ``surface``
         are the :class:`~core.agent_harness.turns.headless_build.DefaultHeadlessBuild`
         fields; ``tools`` and ``gather`` the ports its ``agent()`` takes;
-        ``is_tty`` is bound on the first turn. A host that needs more (its own
-        sink, prompts, error reporter, an action ``llm_factory``) builds through
+        ``is_tty`` and ``tool_hooks`` (the turn's approval hooks) are bound on
+        the first turn. A host that needs more (its own sink, prompts, error
+        reporter, an action ``llm_factory``) builds through
         :class:`DefaultHeadlessBuild` itself and calls :meth:`attach_agent`.
         """
         from core.agent_harness.turns.headless_adapters import BufferOutputSink
@@ -168,6 +171,7 @@ class AgentSession:
             logger=logger,
             surface=surface,
             is_tty=is_tty,
+            tool_hooks=tool_hooks,
         )
         return agent_session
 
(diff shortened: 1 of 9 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5358, opened 2026-08-21
docs(splunk): fix broken token-creation example, add a local recipe

arnavpopley:
Part of #5159 (found while live-verifying Splunk's 1 registered tool)

Verified `query_splunk_logs` live against a real local Splunk instance. The documented REST-API bearer-token creation example had two real bugs:

1. **`name` is the existing Splunk user the token authenticates as, not an arbitrary label.** The doc's example used `name=opensre`, which fails with `User "opensre" does not exist` since no such user exists — Splunk tokens authenticate *as* a specific existing user, they aren't just labeled tokens. Fixed to `name=admin`.
2. **`audience` is a required argument the example omitted entirely**, failing with `The following required arguments are missing: audience.`

Also added a full local Docker recipe (there wasn't one before): `splunk/splunk` publishes amd64-only images (needs `--platform linux/amd64` on Apple Silicon, which runs under emulation and takes noticeably longer to boot), and newer images require `SPLUNK_GENERAL_TERMS=--accept-sgt-current-at-splunk-com` alongside `SPLUNK_START_ARGS=--accept-license` — without it, the container exits immediately with `License not accepted`, reproduced live on the first attempt.

```
$ opensre integrations verify splunk
splunk  │ local env │ ✓ passed │ Connected to Splunk 10.4.2
```

Real `opensre investigate` output against a seeded error event — the single registered tool exercised, correctly scoping a missing-dependency incident from Splunk logs alone:

```
Loading integrations: telegram, splunk
↳ Splunk · query splunk logs

Triage complete: payment-service checkout failed with connection refused
to billing-api at 21:40:00 UTC; no billing-api logs are present in Splunk
during a 4-hour window across all indexes.
```

Every command in the doc was run verbatim end-to-end against a fresh local Splunk container to produce this output (including a second full clean run after the doc was finalized, to confirm no copy-paste drift).

---

Docs-only change, no application code. Ran the doc's existing token-creation example verbatim first — it failed exactly as described above. Fixed both issues, then built the local Docker recipe from scratch against a real Splunk container, hitting and documenting the platform/license-acceptance gotchas along the way. Followed the same pattern as the other recently-merged local-recipe PRs (empty, valid `OPENSRE_INTEGRATIONS_STORE_PATH` store rather than an embedded Python snippet).

---

---

### DIFF
--- a/docs/splunk.mdx
+++ b/docs/splunk.mdx
@@ -180,19 +180,18 @@ SERVICE │ SOURCE    │ STATUS   │ DETAIL
 splunk  │ local env │ ✓ passed │ Connected to Splunk 10.4.2
 ```
 
-`opensre investigate` only treats an integration as active once the store resolution has
-*something* to fall through to env vars with -- an untouched `~/.opensre/integrations.json`
-with an unrelated integration in it blocks env-var fallback entirely. Point
-`OPENSRE_INTEGRATIONS_STORE_PATH` at an empty, valid store instead, so your real config is
-never read or written and the `SPLUNK_*` vars above are the only source of connection info:
+`opensre investigate` (unlike `opensre integrations verify`) only falls through to env
+vars when the store has no records at all -- any existing record, for any service, blocks
+env-var resolution entirely. Point `OPENSRE_INTEGRATIONS_STORE_PATH` at a path inside a
+fresh empty directory instead, so your real config is never read or written and the
+`SPLUNK_*` vars above are the only source of connection info:
 
 ```bash
-export OPENSRE_INTEGRATIONS_STORE_PATH=$(mktemp /tmp/opensre-splunk-demo.XXXXXX)
-echo '{"version": 2, "integrations": []}' > "$OPENSRE_INTEGRATIONS_STORE_PATH"
+export OPENSRE_DEMO_STORE_DIR="$(mktemp -d /tmp/opensre-splunk-demo.XXXXXX)"
+export OPENSRE_INTEGRATIONS_STORE_PATH="$OPENSRE_DEMO_STORE_DIR/integrations.json"
 ```
 
-Trigger a real investigation against the seeded event -- this is the actual supported
-entrypoint, not an internal import:
+Trigger a real investigation against the seeded event:
 
 ```bash
 opensre investigate --input-json '{
@@ -206,11 +205,9 @@ opensre investigate --input-json '{
 }'
 ```
 
-Real output from a run against this exact local instance (edited for length). An empty
-store makes `opensre investigate` fall through to env-var resolution for *every*
-integration, not just Splunk -- if your shell or machine already resolves another
-integration from its own env vars, it rides along too (harmless, and unrelated to the
-Splunk tool calls below):
+Real output from a run against this exact local instance (edited for length). Another
+integration resolved from your own env/keyring can ride along harmlessly, since the
+store has no records to block fallback for any service:
 
 ```
 Loading integrations: telegram, splunk
@@ -228,15 +225,12 @@ though without cluster/deploy visibility the exact narrow cause cannot be
 confirmed.
 ```
 
-The single registered tool was exercised, correctly scoping the blast radius to a
-missing dependency rather than treating the absence of billing-api logs as inconclusive.
-
 Teardown:
 
 ```bash
 docker rm -f splunk-dev
-rm -f "$OPENSRE_INTEGRATIONS_STORE_PATH"
-unset OPENSRE_INTEGRATIONS_STORE_PATH SPLUNK_URL SPLUNK_TOKEN SPLUNK_VERIFY_SSL
+rm -rf "$OPENSRE_DEMO_STORE_DIR"
+unset OPENSRE_DEMO_STORE_DIR OPENSRE_INTEGRATIONS_STORE_PATH SPLUNK_URL SPLUNK_TOKEN SPLUNK_VERIFY_SSL
 ```
 
 ## Investigation tools
--- a/docs/splunk.mdx
+++ b/docs/splunk.mdx
@@ -90,16 +90,18 @@ OpenSRE uses **bearer tokens** — not basic auth and not HEC tokens.
 
 1. Go to **Settings** → **Tokens**
 2. Click **New Token**
-3. Set a name (for example `opensre`) and an expiry date
-4. Copy the generated token
+3. In **User**, pick the existing Splunk user the token authenticates as
+4. Fill in **Audience** (required) and an expiry date
+5. Copy the generated token
 
-**Via the REST API** (replace `<PASSWORD>` with your admin password):
+**Via the REST API** (replace `<PASSWORD>` with your admin password and
+`<SPLUNK_USER>` with the service account the token should authenticate as):
 
 ```bash
 curl -sk -u admin:<PASSWORD> \
   https://splunk.corp.com:8089/services/authorization/tokens \
   -X POST \
-  --data-urlencode "name=admin" \
+  --data-urlencode "name=<SPLUNK_USER>" \
   --data-urlencode "audience=OpenSRE" \
   --data-urlencode "expires_on=+90d" \
   --data-urlencode "output_mode=json" \
@@ -108,9 +110,10 @@ curl -sk -u admin:<PASSWORD> \
 
 <Warning>
 `name` is the **existing Splunk user** the token authenticates as, not an arbitrary
-label -- Splunk rejects an unrecognized name with `User "..." does not exist.` `audience`
-is also required (a short free-text description of the token's purpose); omitting it
-fails with `The following required arguments are missing: audience.`
+label — Splunk rejects an unrecognized name with `User "..." does not exist.` The token
+inherits that user's roles, so point it at a least-privilege service account rather than
+`admin`. `audience` is also required (a short free-text description of the token's
+purpose); omitting it fails with `The following required arguments are missing: audience.`
 </Warning>
 
 The token needs the `search` capability. The `admin` role includes this by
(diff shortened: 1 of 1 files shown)


docs/splunk.mdx
```diff
@@ -99,19 +99,146 @@ OpenSRE uses **bearer tokens** — not basic auth and not HEC tokens.
...
+`splunk/splunk` publishes **amd64 only** -- on Apple Silicon, add `--platform
+linux/amd64` (runs under emulation; startup takes several minutes longer than on
+native amd64). Without `SPLUNK_GENERAL_TERMS=--accept-sgt-current-at-splunk-com`
+alongside `SPLUNK_START_ARGS=--accept-license`, the container exits immediately
+with `License not accepted` -- newer images require both, not just the license flag
+documented in older Splunk Docker guides.
+</Warning>
+
+Seed a test event and create a token:
+
```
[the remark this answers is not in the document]
arnavpopley: Disagree, for consistency within this same file: the pre-existing (unmodified by this PR) Credentials section already references raw Splunk REST endpoints the same way -- `/services/authorization/tokens` directly inside a runnable curl command (line 100), and `/services/server/info` in prose explaining why a permission is needed (line 121) -- and neither was flagged. `/services/receivers/simple` here is the literal endpoint a reader must curl to seed test data; there's no way to write a working "seed a test event" command without naming it, since the reader is the one making the REST call, not going through a client that hides it. This differs from the getMe-style case in CLAUDE.md, which is about narrating a vendor API call a reader never makes themselves. Leaving this open rather than resolving it myself since it's a documented style call, but happy to adjust if the precedent read is wrong.

arnavpopley:
@greptile review

docs/splunk.mdx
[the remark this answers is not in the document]
arnavpopley: Yes, please remember that rule as stated -- it's exactly the distinction that matters.

docs/splunk.mdx
```diff
@@ -99,19 +99,146 @@ OpenSRE uses **bearer tokens** — not basic auth and not HEC tokens.
...
+```
+SERVICE │ SOURCE    │ STATUS   │ DETAIL
+splunk  │ local env │ ✓ passed │ Connected to Splunk 10.4.2
+```
+
+`opensre investigate` only treats an integration as active once the store resolution has
+*something* to fall through to env vars with -- an untouched `~/.opensre/integrations.json`
+with an unrelated integration in it blocks env-var fallback entirely. Point
+`OPENSRE_INTEGRATIONS_STORE_PATH` at an empty, valid store instead, so your real config is
+never read or written and the `SPLUNK_*` vars above are the only source of connection info:
```
Davidson3556: This isn't what the code does, and I think it will send readers looking in the wrong place.

`merge_local_integrations` merges per service: a store record replaces an env record only for the **same** service key. An unrelated entry has no bearing on Splunk. Tested with a store holding only a telegram record, `SPLUNK_*` exported:

```
services resolved: ['splunk']
splunk source: local env
splunk verify_ssl: False
```

What you most likely hit is a `splunk` entry in your own store shadowing the env vars. There is a sharper version of the same thing worth knowing: if a stored splunk record does not classify cleanly, splunk drops out of the effective set entirely rather than falling back to env. I get `source: None` in that case.

`docs/postgresql.mdx` (merged in #5338) already says this correctly in two lines:

> Use a temporary integration-store path so saved integrations cannot override the demo environment variables.

Worth lifting verbatim.

docs/splunk.mdx
```diff
@@ -99,19 +99,146 @@ OpenSRE uses **bearer tokens** — not basic auth and not HEC tokens.
...
+
+```bash
+export OPENSRE_INTEGRATIONS_STORE_PATH=$(mktemp /tmp/opensre-splunk-demo.XXXXXX)
+echo '{"version": 2, "integrations": []}' > "$OPENSRE_INTEGRATIONS_STORE_PATH"
```
Davidson3556: Follows from the comment above: `integrations/store.py` returns an empty v2 store for a missing file, so neither line is needed. postgresql.mdx states it outright: "The file does not need to exist."

The JSON itself is correct, it is the exact v2 shape from the `store.py` docstring. Just redundant.

```bash
export OPENSRE_DEMO_STORE_DIR="$(mktemp -d /tmp/opensre-splunk-demo.XXXXXX)"
export OPENSRE_INTEGRATIONS_STORE_PATH="$OPENSRE_DEMO_STORE_DIR/integrations.json"
```

matches the existing recipes and drops the `echo`.

docs/splunk.mdx
```diff
@@ -99,19 +99,146 @@ OpenSRE uses **bearer tokens** — not basic auth and not HEC tokens.
...
+```
+
+Trigger a real investigation against the seeded event -- this is the actual supported
+entrypoint, not an internal import:
```
Davidson3556: "this is the actual supported entrypoint, not an internal import" describes how the PR was built rather than something the reader does. AGENTS.md asks for that in the PR description, where you already have it.

docs/splunk.mdx
```diff
@@ -99,19 +99,146 @@ OpenSRE uses **bearer tokens** — not basic auth and not HEC tokens.
...
+    "splunk_query": "index=main sourcetype=opensre_verify | head 50"
+  }
+}'
+```
+
+Real output from a run against this exact local instance (edited for length). An empty
+store makes `opensre investigate` fall through to env-var resolution for *every*
+integration, not just Splunk -- if your shell or machine already resolves another
+integration from its own env vars, it rides along too (harmless, and unrelated to the
+Splunk tool calls below):
```
Davidson3556: Same for "Real output from a run against this exact local instance (edited for length)".

The rides-along point is worth keeping, just shorter. A reader really can see extra integrations load and it is confusing the first time. One sentence covers it, and with the store rule corrected above it needs less scaffolding to explain.

docs/splunk.mdx
```diff
@@ -99,19 +99,146 @@ OpenSRE uses **bearer tokens** — not basic auth and not HEC tokens.
...
+root_cause: payment-service is receiving TCP connection refused when
+calling billing-api during checkout, and billing-api has emitted zero log
+lines across all Splunk indexes for the last 4 hours. The most likely
+explanation is that billing-api pods are not running or not listening,
+though without cluster/deploy visibility the exact narrow cause cannot be
+confirmed.
+```
+
+The single registered tool was exercised, correctly scoping the blast radius to a
+missing dependency rather than treating the absence of billing-api logs as inconclusive.
```
Davidson3556: This is verification evidence rather than reader guidance, and it is already in the PR body. AGENTS.md: if a sentence does not change what the reader does, cut it.

Davidson3556 (commented):
Good catch on the token example. `name` being an existing Splunk user and `audience` being mandatory both match Splunk's REST API, and that snippet was broken for anyone who copy-pasted it. The live verification is the right bar too.

I checked the recipe against the code rather than just reading it:

- `SPLUNK_URL` / `SPLUNK_TOKEN` / `SPLUNK_VERIFY_SSL` all match the env block in `integrations/_catalog_impl.py`, and `false` maps to `verify_ssl=False` through the `!= "false"` test
- `opensre investigate --input-json` is a real flag, and `integrations verify` takes the optional positional service
- the store JSON is the correct v2 shape
- `<Warning>` components used properly, no `docs.json` change needed on an existing page

One thing I would change before merge: the store-path paragraph states a rule the code does not implement. Details inline. Everything else is optional and can ride along or not.

Minor: four new `--` used as dashes. This file uses `—` nine times and `--` as a dash zero times.

docs/splunk.mdx
arnavpopley: Thanks for the real test -- I re-tested against current main rather than defending the claim, and the picture is more nuanced than either of us had it, so here's exactly what I found:

`opensre integrations verify splunk` with a store holding only a telegram record + SPLUNK_* exported:
```
splunk | local env | ✗ failed | Server info check failed: [Errno 61] Connection refused
```
SOURCE is `local env` -- splunk resolves from env fine here, matching your read of `merge_integrations_by_service`.

But `opensre investigate` with the identical store + env, same alert:
```
Loading integrations: telegram
```
Splunk does NOT load. Traced it to `infrastructure/harness_providers/integration_resolution.py`, `_resolve_from_local_sources()`:
```python
store_integrations = _load_integrations()
env_integrations = _load_env_integrations() if not store_integrations else []
```
That gate gets evaluated before the per-service merge you're describing ever runs -- if `store_integrations` is non-empty (any service, any record), `env_integrations` is set to `[]` outright, so `merge_local_integrations` never even sees the env-loaded splunk config. So `verify` and `investigate` genuinely disagree here on current main, and the doc is specifically about `investigate`.

Could you say which one you tested with, or if you called something lower-level directly? I want to make sure I'm not missing a code path, since this determines whether the doc's warning is accurate for the command it's actually telling readers to run.

Separately, and unambiguously: you're right on the empty-file point below, and I've corrected that.

docs/splunk.mdx
arnavpopley: You're right, and I checked this one properly before replying. Tested a literal bare `mktemp` file (0 bytes) with `opensre investigate` on current main: `_load_raw_unlocked()` in `integrations/store.py` catches `json.JSONDecodeError` and falls back to an empty v2 store -- it logs a warning with a full traceback (which looks alarming) but does not crash, and the command completes normally with the empty-store env-fallback behavior intact. My "a literal zero-byte file does not work here" claim was wrong for the current code -- I'm not sure if it was ever true or if I was carrying forward something from an earlier session that's since been fixed, but either way it doesn't hold now.

Going with your suggested pattern since it's strictly better than either mine or a bare mktemp: pointing at a nonexistent file inside a fresh `mktemp -d` directory means `_store_path().exists()` returns False cleanly, so there's no exception at all -- not even the harmless-but-noisy logged one a bare empty file triggers. Fixing this in the doc now.

docs/splunk.mdx
arnavpopley: Agreed, cutting it.

docs/splunk.mdx
arnavpopley: Agreed, shortening it -- and it needs less scaffolding now that the store fix above is simpler too.

docs/splunk.mdx
arnavpopley: Agreed, cutting it -- it's already in the PR body.

Davidson3556:
Pushed a few fixes straight to the branch — shout if you'd rather I'd left any of them.

- The UI token steps just above the REST example had the same `name` bug this PR fixes. They now say to pick an existing user and fill in Audience.
- The REST example was creating a token for `admin`, which the Security section further down tells people never to do. It takes a `<SPLUNK_USER>` placeholder now.
- Moved the empty-store export above the verify step. With a real store in play a saved Splunk record beats `SPLUNK_URL`/`SPLUNK_TOKEN` for `verify`, so the check could pass against production while printing a localhost-looking result. Also clears `SPLUNK_INSTANCES` and `SPLUNK_CA_BUNDLE`, which override the recipe the same way.
- Swapped the `exit 1` in the readiness loop for `break`. Pasted into an interactive shell it closes the terminal. Raised the wait to 15 minutes to match your own note about emulated startup being slow.
- Minor: em dashes to match the rest of the page, and the platform warning no longer tells you to add a flag the command already has.

Left `curl -sk` and `read_splunkd_private_settings` alone. Both predate this PR, and pinning the real minimum role needs a live Splunk, so I'll file that one separately.

The store-ordering gotcha you documented is a good catch, by the way — `verify` merges per service while `investigate` is all-or-nothing on env fallback, which is worth an issue of its own.

Davidson3556:
@greptile review

docs/splunk.mdx
arnavpopley: Verified both of your major changes independently before accepting anything:

1. Store-ordering: reproduced live with a synthetic store record for splunk (base_url=not-localhost.example.com) + SPLUNK_URL=localhost env vars set. verify showed SOURCE: local store and tried to connect to the fake non-localhost URL, completely ignoring the env vars. Confirms your point exactly -- with a real prod record in the store this passes against prod while looking like it tested localhost. Good catch, this was a real ordering bug I introduced.

2. exit-1-closes-terminal: reproduced with `zsh -i -c '...until...exit 1...'` -- exit inside a loop pasted into an interactive shell does terminate that shell process, not just the current command. break is the right fix; confirmed the fallthrough after a break is a visible curl failure, not a silent bad-state continuation.

Both confirmed genuine, pulled your commit, thank you for pushing directly rather than just flagging. Filing an issue for the read_splunkd_private_settings role-pinning as you suggested makes sense -- agreed that needs a live Splunk to get right.

One thing worth flagging on my end: both of these bugs (exit 1 in bounded readiness loops, and Verify: running before the empty-store setup) are present across every other local-recipe doc from this same batch -- merged (mysql, postgresql, temporal, rocketchat, mariadb) and still open (kubernetes, signoz). I'll get those fixed too.

arnavpopley:
@greptile review

PR #5384, opened 2026-08-21
Install the subprocess presenter as a frozen provider instead of a mutable global

YauhenBichel:

The subprocess-presenter provider in `infrastructure/harness_providers/` used a
mutable `set_`/`get_` global — any caller could overwrite it at any time (a
read-modify-write global, which the house rules discourage). This converts it to
the frozen-installed-once pattern already used for `cli_auth` and scheduler
delivery. Behavior is identical; the write surface is what shrinks.

Before — a slot anyone can overwrite, anytime:
```python
_factory = None
def set_subprocess_presenter_factory(f): global _factory; _factory = f
def get_subprocess_presenter_factory(): return _factory
After — an immutable value installed once at the composition root, read-only:


@dataclass(frozen=True)
class SubprocessPresenterProvider:
    factory: SubprocessPresenterFactory

def resolve_subprocess_presenter() -> SubprocessPresenterFactory | None:
    return _installed.factory if _installed is not None else None
Call-site changes:

Composition root (bootstrap/adapters.py): set_...factory(f) becomes SubprocessPresenterProvider(f).install().
Reader (core/agent_harness/turns/headless_build.py): get_...factory() becomes resolve_subprocess_presenter().
The package no longer exports a setter — there is one install site, at boot.
The test is rewritten against the new API (and leftover "port"/"wiring" wording in its docstrings is cleaned up).
The other three single-value providers (tool registry, integration resolution,
cli-llm) follow the same pattern in later PRs. The append-only registries (prompt
fragments, evidence sources, metric drafts) are left as-is — they are filled once
at boot and never rewritten.

Install-then-resolve at real boot, and a real shell-tool run through the presenter:


before boot: None
after boot : installed   (SubprocessPresenterProvider, frozen)

$ LLM_PROVIDER=claude-code uv run opensre --json ask \
    --dangerously-bypass-approvals "run the shell command: echo solution2ok"
{"status": "success", "response": "solution2ok", "denied_tools": [], "error": null}
make check-imports green (no cycle); ruff + mypy clean; 5752 tests pass across
agent_harness / bootstrap / cli / integrations / gateway. The one failure in the
suite — tests/cli/test_main.py::test_root_main_propagates_the_cli_exit_code — is
pre-existing and unrelated (an import main path issue that fails on a clean tree
too).

The one thing that changes is who may write the slot and when. A mutable
set_/get_ global lets any code overwrite the presenter at any point; the
frozen provider is assembled once at the composition root, frozen so it cannot be
mutated, and installed once — readers only resolve it, and no setter is exported.
That removes the read-modify-write global while keeping behavior identical, which
the tests and a real shell-tool run confirm. I deliberately scoped this to the
one provider so it is a small, obvious pattern-setter; the other single-value
providers get the same treatment in their own PRs, and the append-only registries
are intentionally left alone because they are never rewritten after boot.

Key pieces: SubprocessPresenterProvider / resolve_subprocess_presenter in
infrastructure/harness_providers/subprocess_presenter.py, and the two call-site
updates in bootstrap/adapters.py and core/agent_harness/turns/headless_build.py.

### DIFF
--- a/bootstrap/adapters.py
+++ b/bootstrap/adapters.py
@@ -48,7 +48,7 @@ def install_harness_adapters() -> None:
 
     register_integrations()
     register_tools()
-    harness_providers.set_subprocess_presenter_factory(headless_subprocess_presenter_factory)
+    harness_providers.SubprocessPresenterProvider(headless_subprocess_presenter_factory).install()
     # Shell / REPL / gateway slash surface: CTA must name a runnable command.
     harness_providers.set_integration_setup_command(
         lambda service_id: f"/integrations setup {service_id}"
--- a/core/agent_harness/turns/headless_build.py
+++ b/core/agent_harness/turns/headless_build.py
@@ -54,7 +54,7 @@ from core.agent_harness.turns.headless_adapters import (
     StaticReasoningClientProvider,
 )
 from core.agent_harness.turns.headless_agent import HeadlessAgent
-from infrastructure.harness_providers import get_subprocess_presenter_factory
+from infrastructure.harness_providers import resolve_subprocess_presenter
 
 if TYPE_CHECKING:
     from core.agent_harness.session.session_core import SessionCore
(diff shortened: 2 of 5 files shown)


YauhenBichel:
@greptile review

PR #5388, opened 2026-08-22
Meter telegram and buzz turns, closing the transport-contract ledger

### LINKED ISSUE #5328: Gateway: credit-metering parity for telegram and buzz transports
## Problem

`gateway/tests/test_transport_contract.py:62-65` pins a shrinking "known gaps"
ledger. Two transports still lack credit metering: **telegram** and **buzz**
neither call `consume_credits`, while Slack and Discord do
(`gateway/transports/discord/dispatcher.py:22`). A turn on telegram/buzz is not
metered, so the four transports are not at parity.

## Fix

Add `consume_credits` to the telegram and buzz inbound paths, mirroring the
Slack/Discord dispatchers, then delete both entries from the contract ledger.

Files:
- `gateway/transports/telegram/inbound_handler.py` (or `runtime.py`)
- `gateway/transports/buzz/inbound_handler.py`
- `gateway/tests/test_transport_contract.py` (remove the two ledger entries)

## Acceptance

- `test_transport_contract` requires credit metering for all four transports
  with no "known gaps" entries for telegram/buzz.

Davidson3556:
Fixes #5328

Telegram and Buzz did not meter turns, while Slack and Discord did. This PR requires all four chat transports to bind an organization-scoped credit request.

The fix also closes a capacity-ordering bug shared by the four transports. Each transport used to consume a credit before calling the shared `TurnRunner`, but `TurnRunner` is where the process-wide capacity gate lives. A message arriving while every turn slot was occupied could be charged and then receive only the at-capacity response.

Metering is now split into two responsibilities:

- Each transport binds its organization, reason, and denial response around the existing four-argument turn callback.
- The shared `TurnRunner` acquires capacity first, then applies the bound credit admission, then starts agent work.

Only an explicit HTTP 402 blocks the turn. `UNCONFIGURED` and `UNAVAILABLE` still fail open. Telegram and Buzz keep the ledger call on their executor thread, so billing cannot block the polling event loop. A cancellation check before admission avoids charging a turn that was already cancelled. Once the ledger allows a turn, the runner continues into the turn lifecycle so the debit cannot be detached from Buzz acknowledgement.

Slack and Discord were moved to the same shared ordering because they had the same pre-capacity charge defect. Their denied turns now finalize the existing placeholder and mark it failed instead of posting a second message.

The transport contract now requires `bound_turn_metering` and rejects transport-local `consume_credits` calls, so the ordering cannot silently regress in one transport.

#### What this costs

Charging inside the capacity slot is deliberate and it is not free. The ledger POST used to run before the gate. It now holds a slot it has not used yet, for up to the 5 second credits timeout. On the SMALL size profile that is the process's only slot, so a stalled ledger makes a concurrent message see the at-capacity reply where it would previously have queued. That is the trade for never charging a user we then refuse to serve. A deployment that feels it can raise `OPENSRE_MAX_CONCURRENT_TURNS`.

Two things stay out of scope:

- The credits API has no idempotency contract, so a crash between a successful debit and the cursor write bills the same delivery twice on restart. This is new exposure for Telegram and Buzz, which replay by design; Slack and Discord do not. Filed as #5691.
- Scheduled runs reach the agent through `scheduler_runners` rather than a transport, and stay unmetered. The empty `_KNOWN_GAPS` ledger in the contract test is scoped to transports and says so.

#### Interaction with #5218

Rebased on top of the Telegram approval fix. The two compose rather than overlap. #5218 added a pre-turn cancel check that returns before `_run_turn` reaches the executor, so a Telegram turn stopped while queued never binds the metering request and cannot be charged. The `host_cancel_requested` check added here still covers the window between that check and capacity acquisition, and covers Slack, Discord and Buzz, which have no transport-level equivalent.

Because #5218 dispatches turns as detached tasks, several Telegram turns now run concurrently on different executor threads. The metering request is bound inside `_run_turn`, which executes on the executor thread itself, so each turn gets its own context. Binding it in the async caller instead would not survive `run_in_executor`.

The capacity regression is pinned at the shared runner:

```text
capacity slot occupied
  -> TurnRunner returns AT_CAPACITY_MESSAGE
  -> admission hook calls: 0
  -> agent constructions: 0
```

Credit outcomes remain:

```text
DENIED        -> finalize out-of-credits message; agent not invoked
ALLOWED       -> run the turn
UNCONFIGURED  -> run the turn (local-dev fail-open)
UNAVAILABLE   -> run the turn (billing-outage fail-open)
```

Local validation on the latest `main` merge:

```text
make lint                                      All checks passed
make format-check                              3702 files already formatted
make typecheck                                 Success: 1925 source files
uv run python -m pytest gateway/tests/         802 passed, 1 warning
```

---

The process capacity gate already had one authoritative location in `TurnRunner`. Rather than duplicating or probing that gate in four transports, each transport binds the data needed for billing and the composition root injects one admission hook into the shared runner. This preserves the four-argument `TurnCallback`, keeps transport-specific denial rendering with each transport, and guarantees capacity is acquired before a ledger debit is attempted.

The focused tests exercise the central admission policy, all four transport bindings, denied/fail-open behavior, cancellation, the executor-thread requirement, and the exact full-capacity regression.

---


gateway/transports/buzz/inbound_handler.py
```diff
@@ -93,6 +99,23 @@ async def handle_polled_inbound_buzz_message(
...
 
+        # Metering: only an explicit webapp denial (402) blocks the turn.
+        # UNCONFIGURED (dev setups without metering env) and UNAVAILABLE
+        # (webapp outage) proceed, so a billing outage never silences the bot
+        # and a config error never reads to users as "out of credits".
+        #
+        # Off-thread because Buzz dispatches turns as concurrent tasks on one
+        # loop: a synchronous POST here would park every other in-flight turn,
+        # and the poll loop with them, for the client timeout.
+        outcome = await asyncio.to_thread(consume_credits, scope.principal.id, reason="buzz_turn")
```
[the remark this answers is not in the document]
Davidson3556: Confirmed and fixed in 6513eed.

You are right, and the window was wider than a race: `_dispatch_turn` catches `Exception`, and `CancelledError` is not one, so a cancel landing on that await unwinds straight past the trailing `acknowledge(event)`. The POST finishes on its worker thread, the mention stays uncommitted, and `_drain_active_turns` already documents that uncommitted mentions are re-delivered on next start. I reproduced it: one charge landed, zero acks.

The fix is to drop the `asyncio.to_thread` and charge inline. Nothing suspends between the charge and the `run_in_executor` handoff, so the turn body is always queued once the ledger has been debited, and its `on_handled` acknowledges from the executor thread, which is the mechanism the handler docstring already relies on.

Worth noting the off-thread call was my own addition on top of what the issue asked for, meant to keep a slow ledger from parking the loop. That concern was speculative, and the surrounding code already blocks the loop with the session resolve and a `buzz` CLI subprocess spawn on the denial path. Both handlers are now on the Slack/Discord shape.

Added `test_a_charged_turn_is_never_left_unacknowledged` in `gateway/tests/buzz/test_credit_metering.py`. It posts the cancel via `call_soon_threadsafe` the way the drain does, while the charge is still in flight. Fails on the awaited version with `charged mention was left unacknowledged`, passes on this one.

An idempotency key would be the stronger guarantee, but `/api/credits/consume` has no such contract today, so that belongs in a webapp change rather than here.

Davidson3556:
@greptile review

gateway/transports/buzz/inbound_handler.py
```diff
@@ -93,6 +99,27 @@ async def handle_polled_inbound_buzz_message(
...
+        #
+        # Charged synchronously, deliberately. Awaiting the POST would make it
+        # a cancellation point: shutdown cancels turns that outlast the drain
+        # budget, and a charge landing while that await is cancelled strands a
+        # consumed credit on an unacknowledged mention, which the next start
+        # re-delivers and charges again. Nothing suspends between this call and
+        # the ``run_in_executor`` handoff below, so the turn body is always
+        # queued once the ledger has been debited, and its ``on_handled``
+        # acknowledges from the executor thread.
+        if consume_credits(scope.principal.id, reason="buzz_turn") is CreditsOutcome.DENIED:
```
Devesh36: Synchronous consume_credits() can block the Buzz event loop for up to 5 seconds, stalling polling, dispatch, and shutdown.

gateway/transports/telegram/inbound_handler.py
```diff
@@ -75,6 +81,23 @@ async def handle_polled_inbound_telegram_message(
...
 
+            # Metering: only an explicit webapp denial (402) blocks the turn.
+            # UNCONFIGURED (dev setups without metering env) and UNAVAILABLE
+            # (webapp outage) proceed, so a billing outage never silences the
+            # bot and a config error never reads to users as "out of credits".
+            #
+            # Charged synchronously, as Slack and Discord do. Awaiting the POST
+            # would turn it into a cancellation point between debiting the
+            # ledger and running the turn that credit paid for.
+            outcome = consume_credits(scope.principal.id, reason="telegram_turn")
```
Devesh36: Synchronous billing calls can similarly block Telegram polling and allow message floods or billing-service slowness to deny service.

gateway/transports/buzz/inbound_handler.py
Davidson3556: Good catch, and it conflicts with the P1 above, which said the opposite: do not await the charge, because a shutdown cancel landing on that await strands a consumed credit on a mention the next start re-delivers and charges again. Charging inline fixed that but bought your problem instead.

Both hold, so 0e2b92a is a different shape that answers each. Metering now runs inside the turn body, on the executor thread. Off the loop, a slow ledger cannot stall polling, dispatch or shutdown. Fused to the work it pays for, the debit cannot be stranded, because cancelling the awaiting task does not stop the executor thread, whose `finally` still acknowledges the mention.

One user-visible change: a denied turn finalizes its placeholder with the out-of-credits message instead of posting a second one, since the placeholder now exists by the time we know. That is how these handlers already surface timeout and user-stop, so it reads consistently.

Added `test_the_ledger_is_never_charged_on_the_event_loop`, which fails with `the ledger POST would block the event loop` if the charge moves back onto the loop.

gateway/transports/telegram/inbound_handler.py
Davidson3556: Fixed the same way in 0e2b92a: Telegram charges from the turn body too, on the executor thread, so the poll loop is never holding a ledger POST.

One nuance on the denial-of-service framing here. Telegram awaits the handler inline in its poll loop, so before this change the loop was already serialised on the whole turn, LLM call included. A 5s charge on top of a turn that already runs for tens of seconds was not a new flood vector the way it is on Buzz, which dispatches turns as concurrent tasks. Still worth removing, and keeping the twin handlers on the same shape matters more than the marginal difference in exposure.

The off-the-loop test lives in the Buzz suite, where the concurrency makes it load-bearing, rather than duplicating it here.

gateway/transports/buzz/inbound_handler.py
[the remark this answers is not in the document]
Davidson3556: Follow-up: the inline charge from my previous reply is gone. It traded your finding for a loop-blocking one (see @Devesh36 below), so 0e2b92a moves metering into the turn body instead, on the executor thread. That keeps your invariant, since cancelling the awaiting task does not stop that thread and its `finally` still acknowledges the mention, while the POST no longer sits on the event loop.

`test_a_charged_turn_is_never_left_unacknowledged` still guards your case and still fails if the charge moves back ahead of the turn body.

Davidson3556:
@greptile review

Davidson3556:
@greptile review

Davidson3556:
@greptile review

infrastructure/turn_host/turn_runner.py
```diff
@@ -117,19 +119,26 @@ def run(
...
         with turn_slot(self._gate) as running:
             if not running:
                 output.finalize(self._busy_message)
                 return None
+            if host_cancel_requested(output):
+                return None
+            if self._admission_check is not None and not self._admission_check():
+                return None
+            if host_cancel_requested(output):
+                return None
```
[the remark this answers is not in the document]
Davidson3556: Fixed in ab3f9bbee. The pre-admission cancellation check remains so an already-cancelled delivery is not charged. The post-admission return is removed: once the ledger allows the turn, the runner enters the turn lifecycle before Buzz can acknowledge the callback. The cancellation signal still prevents LLM/tool side effects inside that lifecycle. A regression now pins this debit-to-lifecycle invariant.

Davidson3556:
@greptile review

PR #5424, opened 2026-08-22
Install tool sources as a frozen ToolSources bundle instead of two mutable setters

YauhenBichel:

Third slice of the F3 cleanup (after the subprocess presenter). The tool registry
in `infrastructure/harness_providers/` exposed two mutable setters —
`set_tool_registry` and `set_investigation_tools_adapter` — each a read-modify-write
global anyone could call at any time. This bundles both values into one frozen
`ToolSources` installed once at boot, read through `resolve_*`. Behavior is
identical; the write surface shrinks to a single install site.

This is the two-value case: where the subprocess presenter carried one value, the
tool registry carries two (the surface registry and the investigation-tools
resolver), so they bundle into one frozen dataclass rather than one bare value.

Before — two setters, two mutable globals:
```python
def set_tool_registry(r): global _tool_registry; _tool_registry = r
def set_investigation_tools_adapter(get_investigation_tools=None): ...
def get_surface_tools(surface): return _tool_registry.tools_for_surface(surface)
def get_investigation_tools(resolved): return _get_investigation_tools(resolved)

After — one frozen bundle installed once, read-only resolvers:


@dataclass(frozen=True)
class ToolSources:
    registry: ToolRegistry = _EMPTY_REGISTRY
    investigation_tools: InvestigationToolsFn = _default_investigation_tools

def resolve_surface_tools(surface): ...       # empty before boot
def resolve_surface_tool_map(surface): ...
def resolve_investigation_tools(resolved): ...
Call-site changes:

```

Composition root (tools/harness_adapters.py): the two set_... calls become one ToolSources(registry=…, investigation_tools=…).install().
Readers: get_surface_tools / get_surface_tool_map / get_investigation_tools become resolve_surface_tools / resolve_surface_tool_map / resolve_investigation_tools in core/agent_harness/tools/action_tools.py, core/agent_harness/turns/evidence_driver.py, and surfaces/cli/ask/approval.py.
The package no longer exports any setter; it re-exports ToolSources and the three resolve_*.
Tests migrated, including a registry-only install (ToolSources(registry=…).install()) — the bundle's field defaults make partial installs clean.

The other two providers (integration resolution, cli-llm) follow the same pattern
in later PRs; the append-only registries are left as-is.

Boot installs the sources, and a real ask executes an action tool resolved through
ToolSources:

before boot: action tools = 0
after boot : action tools = 51   (registry installed)

$ LLM_PROVIDER=claude-code uv run opensre --json ask \
    --dangerously-bypass-approvals "run the shell command: echo toolregistry_ok"
{"status": "success", "response": "toolregistry_ok", "denied_tools": [], "error": null}
make check-imports green (no cycle); ruff + mypy clean; 831 tests pass on the
directly-affected suites and 2194 across tests/core. One suite failure —
tests/core/agent/test_turn_scenarios.py::test_live_turn_execution_oracle[...local-llama-fail-closed]
— is a pre-existing cross-suite ordering flake (a live-LLM fail-closed test that is
skipped in isolation, passes identically with and without this change at the file
level, and only surfaces in a specific four-directory run); it is unrelated to this
change.

### DIFF
--- a/core/agent_harness/tools/action_tools.py
+++ b/core/agent_harness/tools/action_tools.py
@@ -8,7 +8,7 @@ from typing import Any, Protocol
 from core.domain.types.tools import ToolSurface
 from core.tool.contracts import RegisteredTool
 from core.tool.execution import availability_view
-from infrastructure.harness_providers import get_surface_tool_map, get_surface_tools
+from infrastructure.harness_providers import resolve_surface_tool_map, resolve_surface_tools
 from infrastructure.observability.trace.redaction import redact_sensitive
 
 _ACTION_SESSION_SOURCE = "_action_session"
@@ -53,7 +53,7 @@ def get_action_tools_from_integrations_context(
     """Return canonical registered tools available to the action agent."""
     sources = _sources_for_context(ctx, resolved_integrations)
     tools: list[RegisteredTool] = []
-    for candidate in get_surface_tools(ToolSurface.ACTION):
+    for candidate in resolve_surface_tools(ToolSurface.ACTION):
         try:
             if not candidate.is_available(sources):
                 continue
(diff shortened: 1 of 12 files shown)


YauhenBichel:
@greptile review

PR #5427, opened 2026-08-22
Install CLI-LLM adapters as a frozen CliLlmAdapters bundle instead of a mutable setter

YauhenBichel:

The CLI-LLM provider in `infrastructure/harness_providers/` exposed a mutable `set_cli_llm_adapters`
setter over three module globals — a read-modify-write group anyone could overwrite. This bundles the three adapters into one frozen `CliLlmAdapters` installed once at boot. Behavior is identical; the write surface collapses to a single install site.

The three readers keep their domain names (`cli_provider_registration`, `build_cli_client`, `flatten_cli_messages_to_prompt`) — their 15 call sites in `core/llm` are untouched; only their internals change to read the installed bundle.

Before — one setter, three mutable globals:
```python
_cli_provider_registration_fn = _default_cli_provider_registration
_build_cli_client_fn = _cli_llm_backend_unavailable
_flatten_cli_messages_fn = _cli_llm_backend_unavailable
def set_cli_llm_adapters(cli_provider_registration=None, build_cli_client=None, flatten_cli_messages=None): ...
```

After — one frozen bundle installed once, read-only functions:
```python
@dataclass(frozen=True)
class CliLlmAdapters:
    cli_provider_registration: CliProviderRegistrationFn = _default_cli_provider_registration
    build_cli_client: BuildCliClientFn = _cli_llm_backend_unavailable
    flatten_cli_messages: FlattenCliMessagesFn = _cli_llm_backend_unavailable
```

Call-site changes:

Composition root (integrations/harness_adapters.py): set_cli_llm_adapters(...) becomes CliLlmAdapters(...).install().
The package no longer exports a setter; it re-exports CliLlmAdapters and the three reader functions.
One test's private-global assertions updated to the new _installed bundle.
Pre-boot behavior is preserved: cli_provider_registration returns None and
build_cli_client / flatten_cli_messages_to_prompt raise "not registered"
until the bundle is installed.

This is the last of the single-value / small-bundle providers. integration_resolution
(the multi-site, eleven-value one) is tracked separately; the append-only
registries are left as-is.

Direct wiring proof (independent of the external CLI):

before boot: provider reg = None
before boot: build_cli_client raises "not registered"
after boot : provider reg = installed
after boot : build_cli_client -> CLIBackedLLMClient
make check-imports green (no cycle); ruff + mypy clean; 573 tests pass across
core/llm, the harness-providers install test, and agent_harness.

Same inversion as the subprocess presenter and tool registry: the three CLI-LLM
adapters, previously three mutable globals behind one setter, bundle into a frozen
CliLlmAdapters assembled at the composition root and installed once. The reader
functions keep their names because they are already domain verbs, and they fall
back to the module's default functions before the bundle is installed, preserving
the exact pre-boot behavior (returns None / raises "not registered"). One thing I
checked deliberately: a function stored as a dataclass instance attribute is not
a bound method, so installed.build_cli_client(adapter, …) calls the function
without a phantom self — verified directly.

Key pieces: CliLlmAdapters and the three reader functions in
infrastructure/harness_providers/cli_llm.py, and the single install in
integrations/harness_adapters.py.

### DIFF
--- a/infrastructure/harness_providers/__init__.py
+++ b/infrastructure/harness_providers/__init__.py
@@ -12,12 +12,12 @@ from __future__ import annotations
 
 from infrastructure.harness_providers.cli_llm import (
     BuildCliClientFn,
+    CliLlmAdapters,
     CliProviderRegistrationFn,
     FlattenCliMessagesFn,
     build_cli_client,
     cli_provider_registration,
     flatten_cli_messages_to_prompt,
-    set_cli_llm_adapters,
 )
 from infrastructure.harness_providers.cli_llm import reset as _reset_cli_llm
 from infrastructure.harness_providers.evidence_sources import (
@@ -153,6 +153,7 @@ def reset_harness_providers() -> None:
 
 __all__ = [
     "BuildCliClientFn",
+    "CliLlmAdapters",
     "ClassifyIntegrationsFn",
     "CliProviderRegistrationFn",
     "ConfiguredIntegrationServicesFn",
@@ -218,7 +219,6 @@ __all__ = [
     "resolve_subprocess_presenter",
     "resolve_surface_tool_map",
     "resolve_surface_tools",
-    "set_cli_llm_adapters",
     "set_integration_resolution_adapters",
     "set_integration_setup_command",
     "set_remote_integrations_fetcher",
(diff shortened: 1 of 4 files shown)


YauhenBichel:
@greptile review

PR #5429, opened 2026-08-22
Install integration-resolution as three by-site frozen holders instead of four setters

YauhenBichel:
The integration-resolution provider in `infrastructure/harness_providers/` was the last mutable one — four setters over eleven module globals (`set_remote_integrations_fetcher`, `set_integration_resolution_adapters` (eight fields), `set_setupable_integration_services`, `set_integration_setup_command`).

Unlike the earlier providers, its values are genuinely wired at three different tiers for real deployment reasons: the remote fetcher is a surface choice (the Tracer Cloud fetcher is installed only on the product/CLI path, so gateway/web
keep the empty default), the load/merge/classify adapters are integration-owned, and the connect-command syntax is a composition-root choice. Forcing all eleven into one install would fight that. So this inverts them into **three frozen holders grouped by wiring site**, each replacing its setter one-for-one:

- `IntegrationResolutionAdapters` (the eight adapters + setupable services) —   installed by `integrations/harness_adapters`.
- `RemoteIntegrationsProvider` (the remote fetcher) — installed at the surface; the empty default preserves the gateway/web behavior.
- `IntegrationSetupCommand` (the connect-command renderer) — installed at `bootstrap`.

The resolve internals read the installed adapters through a small `_adapters()` accessor. Readers keep their domain names (`resolve_integrations`, `configured_integration_services`, `integration_setup_command`, …); no reader call site changed.


YauhenBichel:
@greptile review

PR #5431, opened 2026-08-22
Agent-harness: one session-goal loop driver; turn-host: measurable, tunable concurrency

### LINKED ISSUE #5432: Raise turn concurrency on SMALL tasks: measure per-turn memory, then tune the limit
## Problem

By default the gateway runs one agent turn at a time. `OPENSRE_SIZE_PROFILE` defaults to `SMALL`, which sets the process turn gate to a limit of 1, and chat uses a non-blocking acquire — so a second concurrent request is dropped with "OpenSRE is at capacity. Please try again shortly." rather than served.

This is a self-imposed application limit, not an AWS/Fargate one. A small task accepts many connections; the gate caps concurrent *turns* to protect memory. But turns are I/O-bound (mostly waiting on the LLM), so CPU is idle during a turn and a small task can likely run several at once. The real limiter is memory: each concurrent turn holds its conversation and evidence context resident for the turn's duration, so the safe ceiling is roughly the task's memory divided by the per-turn cost. `SMALL = 1` is a conservative guess at that, not a measured value.

## Already shipped (tools only, no default change)

- Per-turn memory logging: `gateway_turn_memory start_mb=… end_mb=… delta_mb=…   peak_mb=…` at debug (Linux/Fargate; skipped where `/proc` is absent).
- `OPENSRE_MAX_CONCURRENT_TURNS` override: raises the turn limit without moving   to a larger size profile or task. Documented in `.env.example`.

## Remaining work

- [ ] Deploy with debug logging, run real turns, and read `delta_mb` / `peak_mb` from `gateway_turn_memory` to find the true per-turn memory cost on a SMALL (0.5 GB) task.
- [ ] Set `OPENSRE_MAX_CONCURRENT_TURNS` on the SMALL task to a value that fits, confirming `peak_mb` stays safely under the task limit under load.
- [ ] Optional: switch chat from drop to a bounded queue so a burst beyond the limit waits briefly (with a timeout) instead of bouncing, rather than returning the at-capacity message immediately.
- [ ] Optional: reduce per-turn resident memory (release evidence/gather buffers after use, trim context retained across the goal loop) to raise how many turns a SMALL task can safely run.

## Goal

Serve more concurrent users on the same small, cheap task — grounded in measured
memory rather than a guess — without bumping the size profile or task resources.

YauhenBichel:
Issue #5432 

Two related turn-host improvements.

**1. One session-goal loop driver (agent-harness)**
- The SessionGoal loop lived in two places: `HeadlessAgent.handle` and a second
  copy reimplemented inside `AgentSession.chat_until_goal` (over `self.chat`).
  Collapsed to a single `HeadlessAgent.run_goal` driver. `handle` is now a thin
  wrapper returning its last result (gateway/`TurnRunner` unchanged);
  `chat_until_goal` delegates to it via a new `GoalDispatcher` protocol.
- Fix (found in review): `run_goal`'s per-turn binding is now optional. A
  configure-once host (`AgentSession`) passes none, so the agent keeps its bound
  `tool_hooks`/`is_tty` across the loop instead of a sparse `TurnBinding` wiping
  them by design. The gateway still passes a full binding per turn.

**2. Measure and tune turn concurrency (turn-host)**
- Added `infrastructure/turn_host/turn_memory.py`: `resident_memory_bytes()` /
  `peak_resident_memory_bytes()` and `log_turn_memory()`, which logs
  `gateway_turn_memory start_mb/end_mb/delta_mb/peak_mb` per turn at debug
  (Linux/Fargate; silently skipped where `/proc` is absent, e.g. macOS dev).
  Zero behavior change. Turns "SMALL = 1" from a guess into a measured limit.
- Added an `OPENSRE_MAX_CONCURRENT_TURNS` override (`configured_turn_limit()`,
  applied in `process_turn_gate()`) so a deployment can raise concurrency on the
  same task size without moving to a larger size profile. Non-positive/invalid
  values are ignored with a warning. Env names now live in
  `config/constants/turn_concurrency.py`; documented in `.env.example`.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5514, opened 2026-08-23
refactor(investigation): evidence mapper travels with the tool

### LINKED ISSUE #5515: Title: Investigation tool output silently dropped from the evidence catalog (evidence-mapper drift)
Many investigation tools run fine but their output never reaches the agent's citeable evidence. Root cause: the mapping from a tool's raw output into the canonical report keys the evidence catalog cites lived in a hardcoded, name-keyed table inside the shared gather_evidence stage, disconnected from the tools. Measured today: 265 investigation tools registered, only 5 mapped (all Grafana), 260 unmapped. Unmapped tools' output survives as an opaque blob but is never minted into a citeable catalog entry, so the report and its claims cannot cite it.

Two problems:
1. Vendor code in a shared stage file (file-placement violation).
2. A second table that drifts from the tool registry with no guard.

Plan:
- [ ] Structural fix: move the mapper onto the tool (`@tool(evidence_mapper=...)`)
      so the stage is vendor-agnostic; add a ratchet coverage guard. (this PR)
- [ ] Backfill the high-value evidence tools (logs/metrics/traces/events/alerts
      across Datadog, CloudWatch, RDS, EKS/k8s, plus query_grafana_annotations),
      shrinking the baseline as each lands.

Acceptance: no new investigation tool can ship unmapped without either declaring
a mapper or being explicitly recorded as a known gap; the 260-gap baseline only
shrinks over time.

YauhenBichel:
Issue #5515

Evidence mapping — lifting a tool's raw output into the canonical report keys the evidence catalog cites — was hardcoded per-vendor inside the shared gather_evidence stage. That put vendor code in a generic orchestration file and created a second registry that drifted from the tool list: 5 of 265 investigation tools were mapped, 260 silently dropped their output.

This makes the mapping part of the tool:
- `evidence_mapper` callable added to the tool contract (RegisteredTool +  BaseTool), threaded through `@tool`, mirroring the existing is_available /  extract_params callable fields.
- The five Grafana mappers moved into integrations/grafana/tools/ next to their   tools; the stage's merge_tool_evidence now dispatches generically  (get_registered_tool(name).evidence_mapper) with zero vendor names.
- A ratchet coverage guard walks the registry and fails when a new tool is  unmapped, unless it's recorded in a committed baseline of the 260 known gaps  (baseline only shrinks as tools are backfilled).

This is the structural first step; backfilling the 260 tools is follow-up work,
now guarded so no new gaps can appear.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5518, opened 2026-08-23
refactor(reporting): make any tool's output citeable via record_evidence_entry

### LINKED ISSUE #5517: Generalize the evidence catalog so any tool's output can be cited
## Problem
The investigation report's evidence catalog (build_evidence_catalog) turns tool output into citeable evidence (E1, E2 …) through a fixed list of ~13 hand-written _add_* readers (s3, cloudwatch, grafana_logs, datadog×4, honeycomb, coralogix, betterstack, new_relic). Any tool whose output isn't one of those keys is never cited — its data survives only as an opaque blob

It also blocks the tool-by-tool backfill: with only bespoke readers, every vendor would have to edit the shared catalog builder, colliding on one file and regrowing the same grab-bag.

## What's needed
A uniform way for a tool's evidence mapper to contribute a citeable entry without editing the shared catalog builder, so per-vendor backfill becomes independent, mapper-only work.

YauhenBichel:
Fixes #5517 

The report's evidence catalog only cited tool output through ~13 hand-written _add_* readers, so most tools' output was never citeable and per-vendor backfill would have forced every contributor to edit the shared catalog builder.

This adds a uniform, additive path:
- `record_evidence_entry(evidence, *, source, label, summary, url, snippet)` in   core (core/domain/types/evidence.py) — a mapper records a citeable entry.
- `build_evidence_catalog` gains `_add_mapped_entries`, which turns recorded   entries into display ids after the existing bespoke readers. A source already  claimed by a bespoke reader is skipped, so current report output is unchanged.
- Migrated the four Grafana mappers that were mapped-but-not-citeable  (metrics/traces/alert_rules/service_names) onto record_evidence_entry as the  reference example for the vendor backfill.

### DIFF
--- a/core/domain/types/evidence.py
+++ b/core/domain/types/evidence.py
@@ -17,3 +17,31 @@ EvidenceSource = str
 #: ``evidence`` in place. Declared per tool (``@tool(evidence_mapper=...)``) so a
 #: vendor's mapping lives with the vendor's tool, never in a shared stage file.
 EvidenceMapper = Callable[[dict[str, Any], dict[str, Any], dict[str, Any]], None]
+
+#: Key under which mappers accumulate citeable report entries in the evidence dict.
+CATALOG_ENTRIES_KEY = "catalog_entries"
+
+
+def record_evidence_entry(
+    evidence: dict[str, Any],
+    *,
+    source: str,
+    label: str,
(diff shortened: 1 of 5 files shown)


YauhenBichel:
@greptile review

PR #5595, opened 2026-08-23
fix: fixed Azure's sync operation overriding unchanged files

### LINKED ISSUE #5422: [IMPROVEMENT] Optimise remote-sync pull & push operation to download/push changed files only
### Current state

Currently, on running remote-sync pull/push operations, etag is checked with MD5 hashes which don't match. So, it downloads all the files. Similarly, for push operation, all the files on local disk gets pushed to the remote-store. This has been identified on Azure filestorage currently.

Thanks to @dhakad22klx for noticing the unnecessary file downloads for pull action.

Image

**Video for reference**:

https://github.com/user-attachments/assets/27efd0d4-a708-48a1-a985-c93d2a302e3c

### Desired state

OpenSRE must check which files are changed and fetch/push only those files. This would save unnecessary API calls and redundant network bandwidth usage.

SaptarshiSarkar12:
Fixes #5422

Currently, azure checks for the `<ETag>` value for checking whether the file has changed or not, but the engine compares that ETag value with the computed md5 hash of the local file. So, they always mismatch causing problems.
Azure REST API also provides `<Content-MD5>`  tag with the base64 encoded 16 byte value of md5 hash (in hex). So, now that md5 hash is parsed, decoded and stored as ETag for engine comparison method.

https://github.com/user-attachments/assets/930d1ab8-376b-4b19-a4e2-c1bb9ac75155

---

- [X] No, I wrote all the code myself

Azure REST API's list endpoint provides MD5 hashes for all the objects along with etag. As the engine compares files based on their MD5 hashes, for azure file storage, those md5 hashes are stored as etag.

---

---

### DIFF
--- a/infrastructure/filestorage/providers/azure.py
+++ b/infrastructure/filestorage/providers/azure.py
@@ -166,7 +166,7 @@ def _parse_last_modified(props: ET.Element | None) -> datetime:
 
 
 def _parse_etag(props: ET.Element | None) -> str:
-    """Extracts a valid MD5 hex hash, safely degrading to Etag if missing or malformed."""
+    """Extracts a valid MD5 hex hash, or an empty string if unavailable."""
     if props is None:
         return ""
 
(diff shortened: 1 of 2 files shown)


SaptarshiSarkar12:
@greptile review

SaptarshiSarkar12:
@greptile review

SaptarshiSarkar12:
@greptile review

SaptarshiSarkar12:
@greptile review

SaptarshiSarkar12:
@greptile While your statement: "The Azure parser still returns an empty fingerprint when Content-MD5 is unavailable, and the synchronization engine then uploads an unchanged local file whenever its modification time is not older than the remote timestamp." is true,  I have tested that all files when uploaded get their MD5 hashes computed and stored. As per [Azure Blob storage docs](https://learn.microsoft.com/en-us/azure/storage/blobs/secure-file-transfer-protocol-known-issues#other),
> By default, the Content-MD5 property of blobs that are uploaded by using SFTP are set to null. Therefore, if you want the Content-MD5 property  of those blobs to contain an MD5 hash, your client must calculate that value, and then set the Content-MD5 property of the blob before the uploading the blob.

as far as I know, azure blob files are pushed using Azure's rest api's PUT method (and not SFTP), so I think this won't be a problem. If it really is, please let me know how this can be fixed. The only idea I have is pre-computing md5 hashes for each object been pushed, which is ineffective, right?

SaptarshiSarkar12:
@greptile Yeah, right. So, can you please update your review and confidence score?

PR #5600, opened 2026-08-23
test(integrations): add a check that direct vendor imports only decrease

### LINKED ISSUE #5599: Track and shrink direct imports into integrations vendor folders
## The problem

Every vendor lives in its own folder, integrations/<vendor>/, and is supposed to be used through one main module: integrations.<vendor>. In practice, code all over the repo imports files buried deep inside a vendor's folder instead. Right now: 83 vendors, and about 90 files outside a vendor reach into its inner files. Nothing stops this from getting worse.

## What this issue does

Add a test that lists, for each top-level folder, which vendor inner files it still imports directly. The list can only get shorter: the test fails if someone adds a new direct import, and it also fails if a listed import is removed but
left in the list. No code is moved in this issue — it only records today's count so later changes can drive it down.

## Scope
Only the folders outside integrations/ are checked here: tools, surfaces, gateway, infrastructure, bootstrap. core and config already import nothing inner. Imports from one vendor into another vendor are a separate follow-up.

YauhenBichel:
Fixes #5599

Every vendor lives in integrations/<vendor>/ and should be used through its one main module integrations.<vendor>. Today lots of code imports files buried inside a vendor's folder instead. This PR adds a test that records those direct imports
and makes the count only allowed to go down — it does not move any code.

- tests/shared/integrations_api.py — finds the vendor folders on disk and sets up  the check (each vendor's main module is the one allowed way in).
- tests/shared/test_integrations_api_border.py — for each top-level folder, the  list of vendor inner files still imported directly: core 0, config 0,  infrastructure 2, gateway 4, bootstrap 9, tools 30, surfaces 77. The test fails  if the list grows or if an entry is left in after it stops being imported.

No product code changes. Imports from one vendor into another (115 of them) are a separate follow-up, because the check needs to skip a vendor importing its own files. Later PRs give each vendor a proper main module and move imports onto it,
shrinking these numbers toward zero.

### DIFF
--- a/tests/shared/integrations_api.py
+++ b/tests/shared/integrations_api.py
@@ -1,0 +1,42 @@
+"""Check that code imports each vendor through its main module.
+
+Every ``integrations/<vendor>/`` folder should be used through its main module
+``integrations.<vendor>``, not by importing a file deep inside it. This finds
+imports that reach inside a vendor folder from outside it, so a test can list
+them and require the list to only get shorter over time.
+
+Vendor folders are found on disk, so adding a new ``integrations/<vendor>/``
+folder includes it automatically. The 29 loose ``.py`` files directly under
+``integrations/`` are not vendors and are not checked here.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+from tests.shared.api_border import ApiBorder
+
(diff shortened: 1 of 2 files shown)


YauhenBichel:
@greptile review

Issue #5519, opened 2026-08-23
Show every investigation tool's output as report evidence (#1079 backfill)


YauhenBichel:
## Summary
When OpenSRE investigates an incident, its agent calls tools to gather data, and
that data is meant to appear in the final report as numbered evidence the agent
can point to. Today only a handful of tools do this; most run fine but their
output is quietly dropped from the report. This epic tracks connecting the rest,
one vendor at a time, so investigations are complete and their findings are
backed by visible evidence. Background: Discussion #1079.

## Why this should be done
Every dropped tool output is work the agent paid for that never reaches the
reader, and a conclusion with no visible evidence is hard to trust. Closing the
gap makes reports fuller and more credible, and a coverage guard now stops new
gaps from appearing.

## How it works now (already shipped)
- A tool declares its own mapper with `@tool(evidence_mapper=...)`, and a
  `record_evidence_entry` helper turns that output into report evidence without
  editing any shared file — so each vendor task below is small and independent.
- Foundational PRs: tool-carried mappers + coverage guard (PR #5514); generalized
  evidence catalog + `record_evidence_entry` (PR #5518).

## Vendor backfill tasks (75 issues, independent, good first issues)
- [ ] Add evidence mappers for hermes (1/4) (good first issue)  (`01-hermes-1.md`)
- [ ] Add evidence mappers for hermes (2/4) (good first issue)  (`02-hermes-2.md`)
- [ ] Add evidence mappers for hermes (3/4) (good first issue)  (`03-hermes-3.md`)
- [ ] Add evidence mappers for hermes (4/4) (good first issue)  (`04-hermes-4.md`)
- [ ] Add evidence mappers for github (1/3) (good first issue)  (`05-github-1.md`)
- [ ] Add evidence mappers for github (2/3) (good first issue)  (`06-github-2.md`)
- [ ] Add evidence mappers for github (3/3) (good first issue)  (`07-github-3.md`)
- [ ] Add evidence mappers for kubernetes (1/2) (good first issue)  (`08-kubernetes-1.md`)
- [ ] Add evidence mappers for kubernetes (2/2) (good first issue)  (`09-kubernetes-2.md`)
- [ ] Add evidence mappers for eks (1/2) (good first issue)  (`10-eks-1.md`)
- [ ] Add evidence mappers for eks (2/2) (good first issue)  (`11-eks-2.md`)
[...]

PR #5602, opened 2026-08-23
fix(batch): record failed jobs as evidence

### LINKED ISSUE #5579: Add evidence mappers for batch
## Summary
can point to. The `batch` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `get_failed_jobs`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="get_failed_jobs"' integrations/`.

```python

def _map_get_failed_jobs(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="get_failed_jobs",
        )

@tool(name="get_failed_jobs", ..., evidence_mapper=_map_get_failed_jobs)
```

## How to check it works

   under `tests/integrations/batch/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['get_failed_jobs']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'get_failed_jobs', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `get_failed_jobs`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k batch
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify batch
# edit /tmp/alert.json so the alert names a resource batch actually knows about
grep -E 'get_failed_jobs' /tmp/rca.json
```

matching lines into the PR.

**No credentials for batch?** That is a normal outcome, not a reason to go

nightcityblade:
Fixes #5579
Part of #5519

Attached an evidence mapper to `get_failed_jobs` so failed AWS Batch job results become citeable report entries, removed the tool from the known-gap baseline, and added focused mapper coverage.

Using the same mapper input:

```python
{"failed_jobs": [{"job_name": "job-1"}]}
```

Before — the tool result produced no evidence catalog entry:

```python
{}
```

After — the failed job is recorded as citeable evidence:

```python
{
    "catalog_entries": [
        {
            "source": "get_failed_jobs",
            "label": "Failed AWS Batch Jobs",
            "summary": "1 failed job",
            "url": None,
            "snippet": None,
        }
    ]
}
```

Verification:

- 20 focused tests passed.
- `make lint`, `make format-check`, and `make typecheck` passed.

---

The mapper reads the existing `failed_jobs` list returned by the tool and records a catalog entry only when failures are present. This keeps empty/error responses from creating unsupported evidence while preserving the tool's raw output through the existing gather-evidence path. I used the established Grafana mapper pattern instead of changing the shared catalog builder, keeping integration-specific behavior beside the tool that owns it.

---

- [x] If it is a core feature, I have added thorough tests (N/A — this is an integration mapper; focused coverage is included)

---

Note: Allow edits from maintainers is enabled on the fork PR.

### DIFF
--- a/integrations/tracer/tools/tracer_failed_jobs_tool/__init__.py
+++ b/integrations/tracer/tools/tracer_failed_jobs_tool/__init__.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
 from core.tool_framework import tool
 from integrations.tracer import (
@@ -19,6 +20,20 @@ def _tracer_available(sources: dict[str, dict]) -> bool:
 
 def _tracer_trace_id(sources: dict[str, dict]) -> str:
     return str(sources.get("tracer_web", {}).get("trace_id", ""))
+
+
+def _map_failed_jobs(
+    evidence: dict[str, Any], output: dict[str, Any], _input: dict[str, Any]
+) -> None:
+    failed_jobs = output.get("failed_jobs", [])
+    if failed_jobs:
+        count = len(failed_jobs)
+        record_evidence_entry(
+            evidence,
+            source="get_failed_jobs",
+            label="Failed AWS Batch Jobs",
+            summary=f"{count} failed {'job' if count == 1 else 'jobs'}",
+        )
 
 
 @tool(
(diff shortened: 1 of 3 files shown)


Devesh36 (commented):
Put before/after evidence output  in the description

nightcityblade:
Added explicit before/after evidence output to the PR description using the same failed-job mapper input. The before block shows the empty evidence object; the after block shows the new catalog entry and its exact shape.

PR #5604, opened 2026-08-23
refactor(integrations): give four small vendors a public API to import from

YauhenBichel:
First cleanup after the import check added in #5600. The setup screens imported straight from inside four vendor folders; now each vendor's main module exposes the needed names as its public API and the callers import from there.

- opsgenie, splunk: main module now re-exports its client class and config (elasticsearch and google_docs had empty main files; they now do too).
- The four setup-screen imports (surfaces/cli/wizard/integration_validators)  import from integrations.<vendor> instead of integrations.<vendor>.client.
- Removed the four cleared entries from the import check; the count for the  setup screens drops from 77 to 73.

Also plain-language fixes: the tool import-check files (tests/shared/tool_api.py,tests/shared/test_tool_api_border.py) now say "public API" instead of the jargon they used before, including the test names.

No behaviour changes. The four clients only use httpx and the standard library, so having the main module import the client adds no dependency and no import cycle.


YauhenBichel:
@greptile review

PR #5635, opened 2026-08-23
feat(x-mcp): add evidence mappers for X tools

### LINKED ISSUE #5576: Add evidence mappers for x_mcp
## Summary
can point to. The `x_mcp` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `call_x_tool`
- `list_x_tools`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="call_x_tool"' integrations/`.

```python

def _map_call_x_tool(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="call_x_tool",
        )

@tool(name="call_x_tool", ..., evidence_mapper=_map_call_x_tool)
```

## How to check it works

   under `tests/integrations/x_mcp/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['call_x_tool', 'list_x_tools']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'call_x_tool', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `call_x_tool`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k x_mcp
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify x_mcp
# edit /tmp/alert.json so the alert names a resource x_mcp actually knows about
grep -E 'call_x_tool|list_x_tools' /tmp/rca.json
```

matching lines into the PR.

**No credentials for x_mcp?** That is a normal outcome, not a reason to go

ump45nose:
Fixes #5576

- Add evidence mappers to both `list_x_tools` and `call_x_tool`, making successful X MCP inventories and results citeable in incident reports.
- Accumulate repeated tool listings/results behind stable canonical evidence keys while recording one catalog entry per key.
- Ignore unavailable and empty payloads, retain structured boolean answers, and keep echoed call arguments out of canonical report evidence.
- Remove both tools from the evidence-mapper known-gap baseline and add focused unit/merge-path coverage.

```text
{'call_x_tool': True, 'list_x_tools': True}
[{'source': 'x_mcp_results', 'label': 'X MCP Results', 'summary': 'X MCP tool results', 'url': None, 'snippet': None}]
```

Validation after rebasing onto current `main`:

- `make lint` — passed
- `make format-check` — passed (3706 files already formatted)
- `make typecheck` — passed (1938 source files)
- `uv run python -m pytest tests/integrations/x_mcp/ tests/tools/investigation/stages/gather_evidence/test_evidence_mapper_coverage.py` — 23 passed

---

The two generic X MCP tools already return normalized envelopes, but without an `evidence_mapper` their output remains an opaque tool blob and cannot receive a report evidence ID. The new mappers live beside their owning tools and translate only successful, non-empty response data into canonical `x_mcp_tools` and `x_mcp_results` keys.

For repeated discovery calls, descriptors are merged recursively by tool name so compact listings cannot erase richer nested schema metadata. Call results accumulate, but echoed `arguments` are deliberately not duplicated into the canonical report path. A recursive payload check accepts meaningful text, numeric IDs/metrics, structured objects, and booleans nested in structured payloads while rejecting standalone booleans, recursively empty containers, and metadata-only empty TextContent envelopes. Catalog rows are recorded once per accumulating evidence key.

I considered storing each raw output directly as canonical evidence. I kept the existing raw-output path unchanged and instead normalized the report-facing keys to prevent duplicate call arguments and to make repeated calls deterministic.

---

---

Allow edits from maintainers is enabled.

### DIFF
--- a/integrations/x_mcp/tools/x_mcp_tool/__init__.py
+++ b/integrations/x_mcp/tools/x_mcp_tool/__init__.py
@@ -80,6 +80,20 @@ def _record_catalog_entry_once(
     record_evidence_entry(evidence, source=source, label=label, summary=summary)
 
 
+def _merge_tool_descriptor(existing: dict[str, Any], incoming: dict[str, Any]) -> dict[str, Any]:
+    """Merge repeated tool descriptors without dropping previously seen fields."""
+    merged = dict(existing)
+    for key, value in incoming.items():
+        previous = merged.get(key)
+        if isinstance(previous, dict) and isinstance(value, dict):
+            # MCP schemas can be nested, so merge them recursively to keep metadata
+            # learned from either compact or schema-rich inventory responses.
+            merged[key] = _merge_tool_descriptor(previous, value)
+        else:
+            merged[key] = value
+    return merged
+
+
 def _map_x_tool_listing(
     evidence: dict[str, Any], output: dict[str, Any], _input: dict[str, Any]
 ) -> None:
@@ -92,7 +106,8 @@ def _map_x_tool_listing(
     if not isinstance(inventory, list):
         return
 
-    # Merge repeated filtered listings by tool name while retaining richer later descriptors.
+    # Merge repeated filtered listings by tool name without making descriptor richness
+    # depend on whether compact or schema-rich inventory responses arrive first.
     positions = {
         str(item.get("name")): index
         for index, item in enumerate(inventory)
@@ -103,7 +118,9 @@ def _map_x_tool_listing(
             continue
         name = str(descriptor["name"])
         if name in positions:
-            inventory[positions[name]] = descriptor
+            existing = inventory[positions[name]]
+            if isinstance(existing, dict):
+                inventory[positions[name]] = _merge_tool_descriptor(existing, descriptor)
         else:
             positions[name] = len(inventory)
             inventory.append(descriptor)
--- a/integrations/x_mcp/tools/x_mcp_tool/__init__.py
+++ b/integrations/x_mcp/tools/x_mcp_tool/__init__.py
@@ -62,7 +62,18 @@ def _has_evidence_payload(value: object) -> bool:
     if isinstance(value, (int, float)):
         return True
     if isinstance(value, dict):
-        return any(_has_evidence_payload(item) for item in value.values())
+        # MCP TextContent envelopes carry a type marker even when their visible
+        # text is empty. Only the text field determines whether that envelope
+        # should create a report entry.
+        if value.get("type") == "text" and "text" in value:
+            return _has_evidence_payload(value["text"])
+
+        # Structured boolean fields such as {"exists": false} are meaningful
+        # tool results. Preserve both answers rather than treating false as an
+        # absent payload while still ignoring a standalone boolean value.
+        return any(_has_evidence_payload(item) for item in value.values()) or any(
+            isinstance(item, bool) for item in value.values()
+        )
     if isinstance(value, (list, tuple, set)):
         return any(_has_evidence_payload(item) for item in value)
     return False
--- a/integrations/x_mcp/tools/x_mcp_tool/__init__.py
+++ b/integrations/x_mcp/tools/x_mcp_tool/__init__.py
@@ -9,6 +9,9 @@ X adds or renames individual MCP-side tools.
 
 from __future__ import annotations
 
+from typing import Any
+
+from core.domain.types.evidence import CATALOG_ENTRIES_KEY, record_evidence_entry
 from core.domain.types.tools import ToolSurface
 from core.tool import report_run_error
 from core.tool_framework import tool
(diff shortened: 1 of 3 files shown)


ump45nose:
@greptile review

integrations/x_mcp/tools/x_mcp_tool/__init__.py
```diff
@@ -48,6 +51,104 @@ def _unavailable_response(
...
+        str(item.get("name")): index
+        for index, item in enumerate(inventory)
+        if isinstance(item, dict) and item.get("name")
+    }
+    for descriptor in tools:
+        if not isinstance(descriptor, dict) or not descriptor.get("name"):
+            continue
+        name = str(descriptor["name"])
+        if name in positions:
+            inventory[positions[name]] = descriptor
```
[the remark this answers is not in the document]
ump45nose: Fixed in 5bac2d5. Repeated descriptors now merge recursively, preserving fields and nested schema metadata regardless of compact/rich response order. Added the reverse-order regression test; focused suite is 74/74 and lint, format-check, and typecheck pass.

ump45nose:
@greptile review

integrations/x_mcp/tools/x_mcp_tool/__init__.py
```diff
@@ -48,6 +51,121 @@ def _unavailable_response(
...
 
 
+def _has_evidence_payload(value: object) -> bool:
+    """Return whether an MCP payload value contains reportable data."""
+    if value is None:
+        return False
+    if isinstance(value, str):
+        return bool(value.strip())
+    if isinstance(value, bool):
+        return False
```
Davidson3556, line 61: P2 — Make this payload check value-aware. Boolean `structured_content` is valid MCP response data, and `False` can be the answer to an availability or existence check, not an empty payload. Through the real normalization and merge path, both `{"exists": false}` and `{"exists": true}` create neither `x_mcp_results` nor a catalog entry. In the opposite direction, normalized empty content `[ {"type": "text", "text": ""} ]` is accepted solely because the metadata value `"text"` is non-empty, producing an empty catalog entry. Please treat booleans inside structured payloads as reportable while ignoring metadata-only empty content blocks, and add true/false plus empty-TextContent regressions.

Davidson3556 (changes requested):
Requesting changes for the payload-classification defect above. I reproduced both directions through the real normalization and `merge_tool_evidence` path: valid boolean-only structured results are omitted, while an empty text content block creates a catalog entry from metadata alone. The mapper source keys, repeated descriptor merge, report rendering, 76 selected X-MCP tests, Ruff, formatting, and product mypy are otherwise clean. Please correct the predicate and add the focused regressions before repository CI runs.

integrations/x_mcp/tools/x_mcp_tool/__init__.py
ump45nose: Fixed in 2d78b4c. Structured true/false values are retained, empty TextContent envelopes are ignored, and both boundaries have focused regressions.

ump45nose:
@greptile review

ump45nose:
The requested payload-classification repair is present on current head `2d78b4c`: boolean structured results are retained, empty TextContent is ignored, and focused regressions cover both cases. All required checks are green and Greptile reports Confidence 5/5 with zero unresolved findings. Could you please re-review the current head?

integrations/x_mcp/tools/x_mcp_tool/__init__.py
```diff
@@ -48,6 +51,132 @@ def _unavailable_response(
...
+        # MCP TextContent envelopes carry a type marker even when their visible
+        # text is empty. Only the text field determines whether that envelope
+        # should create a report entry.
+        if value.get("type") == "text" and "text" in value:
+            return _has_evidence_payload(value["text"])
+
+        # Structured boolean fields such as {"exists": false} are meaningful
+        # tool results. Preserve both answers rather than treating false as an
+        # absent payload while still ignoring a standalone boolean value.
+        return any(_has_evidence_payload(item) for item in value.values()) or any(
```
Davidson3556, line 74: **P2 — Keep payload classification field-aware.** This still misclassifies real MCP shapes in both directions. `structuredContent={"checks": [false]}` is valid structured data but produces no evidence because booleans inside lists reach the standalone-boolean branch and return false. Conversely, a real `ImageContent` or `AudioContent` is normalized to a type-only item such as `{"type": "image"}`, and that non-empty type marker creates a catalog entry with no usable payload. I reproduced both through `_tool_result_to_dict` → `_normalize_tool_result` → `merge_tool_evidence`. Please validate structured content separately from MCP content envelopes, require retained payload data for non-text blocks, and add regressions for boolean arrays and type-only media.

integrations/x_mcp/tools/x_mcp_tool/__init__.py
```diff
@@ -48,6 +51,132 @@ def _unavailable_response(
...
+        for field in ("tool", "text", "structured_content", "content")
+        if field in output
+    }
+    results.append(result)
+
+    _record_catalog_entry_once(
+        evidence,
+        source="x_mcp_results",
+        label="X MCP Results",
+        summary="X MCP tool results",
```
Davidson3556, line 176: **P2 — Make the numbered evidence entry identify what was observed.** `build_report_context` and `format_cited_evidence_section` render only catalog metadata, not `evidence["x_mcp_results"]`. A realistic search result currently renders exactly `E1 — X MCP Results — X MCP tool results`; `call_x_tool`, the invoked X subtool, query/result count, and snippet are absent, and different calls collapse behind the same unchanged row. This fails #5576's required final-report check. Please build a bounded, tool-identifying summary/snippet, update bounded aggregate metadata when calls accumulate, and add a formatter-level regression.

Davidson3556 (changes requested):
Requesting changes for two reproducible report-evidence blockers on exact head `2d78b4ca`. The earlier direct-boolean and empty-TextContent defect is fixed, and the focused tests plus current-main synthetic merge are green. The remaining blockers are field/content-type payload classification and opaque final evidence rows. Please also move the new mapper helpers out of `__init__.py` into a focused sibling such as `_evidence.py` per the repository guidance, and complete issue #5576's PR-body requirements: link #5519, include the measured behavior diff and context cost, answer the five required questions, and show the final-report check or explain the credential limitation.

muddlebee:
demo and acceptance criteria missing

PR #5636, opened 2026-08-23
feat(sqs): add evidence mapper for get_sqs_queue_attributes

### LINKED ISSUE #5592: Add evidence mappers for sqs
## Summary
can point to. The `sqs` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `get_sqs_queue_attributes`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="get_sqs_queue_attributes"' integrations/`.

```python

def _map_get_sqs_queue_attributes(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="get_sqs_queue_attributes",
        )

@tool(name="get_sqs_queue_attributes", ..., evidence_mapper=_map_get_sqs_queue_attributes)
```

## How to check it works

   under `tests/integrations/sqs/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['get_sqs_queue_attributes']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'get_sqs_queue_attributes', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `get_sqs_queue_attributes`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k sqs
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify sqs
# edit /tmp/alert.json so the alert names a resource sqs actually knows about
grep -E 'get_sqs_queue_attributes' /tmp/rca.json
```

matching lines into the PR.

**No credentials for sqs?** That is a normal outcome, not a reason to go

aechlaenm:
Fixes #5592

Added an evidence mapper for `get_sqs_queue_attributes` so SQS diagnostic results appear as citeable evidence in investigation reports.

The mapper stores the normalized queue results under `sqs_queues`, merged by queue URL across repeated calls, and records a single catalog entry summarizing queue count, visible messages, in-flight messages and DLQ coverage. Queues whose attribute lookup failed (AccessDenied, missing counts) are excluded from the totals and reported separately as unreadable, so an unavailable measurement is never presented as a measured zero. Empty, missing or malformed queue output records nothing at all, not even an empty key. The mapper is attached directly to the registered SQS tool via `evidence_mapper=` and the tool is removed from the known evidence-mapper gaps list. Tests cover the summary totals, empty and malformed output, repeated calls, the unreadable-queue case and mapper registration.

Verification steps from the issue:

```text
$ uv run pytest tests/integrations/sqs tests/tools/investigation/stages/gather_evidence -q
27 passed in 9.08s

$ uv run python -c "from tools.registry import get_registered_tool as g; print({n: bool(g(n).evidence_mapper) for n in ['get_sqs_queue_attributes']})"
{'get_sqs_queue_attributes': True}

$ uv run python -c "from tools.investigation.stages.gather_evidence.tools import merge_tool_evidence; ev={}; merge_tool_evidence(ev, 'get_sqs_queue_attributes', {'queues': [{'name':'a','visible_count':5,'in_flight_count':1,'has_dlq':True}]}, {}); print(ev.get('catalog_entries'))"
[{'source': 'get_sqs_queue_attributes', 'label': 'SQS Queues', 'summary': '1 queues, 5 visible, 1 in-flight, 1 with DLQ', 'url': None, 'snippet': None}]

$ uv run ruff check integrations/sqs tests/integrations/sqs
All checks passed!
$ uv run mypy integrations/sqs/tools/sqs_queue_attributes_tool tests/integrations/sqs
Success: no issues found in 3 source files
```

## Behaviour comparison

Captured on main and on this branch with the same realistic sample: three queues, one healthy with backlog (payments, 120 visible, 4 in flight, DLQ configured), one quiet (emails), and one that failed attribute lookup with AccessDenied.

```
uv run python -c "from tools.investigation.stages.gather_evidence.tools import merge_tool_evidence; import json; ev={}; merge_tool_evidence(ev, 'get_sqs_queue_attributes', <sample_output>, {}); print(json.dumps(ev, indent=2, default=str))"
```

`diff /tmp/before.json /tmp/after.json`:

```diff
58a59,88
>   ],
>   "sqs_queues": [
>     {
>       "name": "payments",
>       "url": "https://sqs.us-east-1.amazonaws.com/123/payments",
>       "visible_count": 120,
>       "in_flight_count": 4,
>       "has_dlq": true
>     },
>     {
>       "name": "emails",
>       "url": "https://sqs.us-east-1.amazonaws.com/123/emails",
>       "visible_count": 0,
>       "in_flight_count": 8,
>       "has_dlq": false
>     },
>     {
>       "name": "restricted",
>       "url": "https://sqs.us-east-1.amazonaws.com/123/restricted",
>       "attributes_error": "AccessDenied"
>     }
>   ],
>   "catalog_entries": [
>     {
>       "source": "get_sqs_queue_attributes",
>       "label": "SQS Queues",
>       "summary": "3 queues, 120 visible, 12 in-flight, 1 with DLQ across 2 measured, 1 unreadable",
>       "url": null,
>       "snippet": null
>     }
```

**What the reader gains.** Before, the report had no citeable evidence for SQS at all, so a conclusion like "the payments queue is backed up" pointed at nothing. Now the report can cite a numbered entry showing 120 visible and 12 in-flight messages across the measured queues, and it is visible that one queue could not be read rather than that queue silently looking empty.

**How much context it costs.** One catalog entry serializes to 185 characters, and the summary line stays that size no matter how many queues come back. The `sqs_queues` list repeats data the evidence dict already keeps in `tool_outputs`, but it is the merged view the summary is computed from across repeated calls, and the full payload is never inlined into the entry itself.

**What happens on an empty or error result.** Feeding the mapper `{}` and an error-shaped payload (`{'source': 'sqs', 'available': False, 'error': 'AccessDenied listing queues'}`) records nothing: no catalog entry and no `sqs_queues` key. There is no "0 items" line for the agent to reread on later turns. Covered by `test_no_entry_when_no_queues`.

**Whether anything is now recorded twice.** The raw tool payload was already stored twice on main, under the tool name key and inside `tool_outputs`. This change adds the deduplicated `sqs_queues` view plus the one catalog entry. No other mapper touches SQS output, and repeated tool calls update the existing entry in place instead of appending a second one, so the report never shows two SQS entries.

**Whether an existing report changed shape.** The diff above is purely additive, every key that existed on main is byte-identical after the change. The full gather_evidence suite (27 tests, including the coverage test over every registered mapper) passes on the branch.

---

The SQS tool already returned useful normalized queue diagnostics, but it had no evidence mapper, so its output never became a citeable entry in the final investigation report.

I followed the existing Grafana evidence-mapper pattern. `_map_get_sqs_queue_attributes` reads the tool's `queues` list, merges it by queue URL with anything recorded by earlier calls, stores the merged view under `sqs_queues`, calculates totals and either records a catalog entry or updates the one that already exists.

I considered recording only the number of queues, but that would omit the most useful incident indicators. The summary therefore also includes visible messages, in-flight messages and DLQ coverage, while the full normalized queue list is preserved for detailed inspection.

Queues the tool could not read (an `attributes_error`, or counts that came back as None) are kept out of the totals and surfaced as unreadable, because an unavailable measurement is not a measured zero and presenting it as one could make a permissions failure look like an empty queue. When there is nothing usable in the output the mapper writes nothing at all.

Tests verify the calculated summary, empty and malformed output handling, repeated-call accumulation, the unreadable-queue case and mapper registration. The existing evidence-mapper coverage test enforces the baseline removal.

---


aechlaenm:
Fixed in 3d414e1. Queues now merge by URL across calls and the single catalog entry's summary is updated to match. Added a repeated-call test.

integrations/sqs/tools/sqs_queue_attributes_tool/__init__.py
```diff
@@ -73,6 +74,38 @@ def _int(key: str) -> int | None:
...
+        return
+    existing = evidence.get("sqs_queues")
+    merged: dict[str, dict[str, Any]] = {}
+    for q in [*(existing if isinstance(existing, list) else []), *queues]:
+        if isinstance(q, dict):
+            merged[str(q.get("url") or q.get("name") or id(q))] = q
+    evidence["sqs_queues"] = list(merged.values())
+    if not merged:
+        return
+    visible = sum(q.get("visible_count") or 0 for q in merged.values())
```
Davidson3556: P1 — Preserve unknown queue attributes in this report summary. The tool intentionally returns `None` or `attributes_error` because an unavailable measurement is not a measured zero, but these `or 0` aggregates turn an AccessDenied-only result into citeable evidence saying `0 visible, 0 in-flight, 0 with DLQ`. Track/report unavailable queues (or omit incomplete aggregates) instead of coercing unknowns to zero, and add a regression for an all-unreadable/`None` result.

Davidson3556 (dismissed):
One blocking correctness issue remains: the evidence summary converts unavailable SQS measurements into measured zeroes, which can make an AccessDenied-only result citeable as an empty queue. The inline comment includes the reproduced failure mode and required regression coverage. Focused tests and local quality checks otherwise pass on the exact latest head.

aechlaenm:
Fixed in 1fcaf9f. Totals now only include queues with measured integer counts. Unreadable queues are reported separately and an all unreadable result produces no counts. Added a regression test for the AccessDenied/None case.

integrations/sqs/tools/sqs_queue_attributes_tool/__init__.py
```diff
@@ -73,6 +74,28 @@ def _int(key: str) -> int | None:
     }
 
 
+def _map_get_sqs_queue_attributes(
+    evidence: dict[str, Any], output: dict[str, Any], _input: dict[str, Any]
+) -> None:
+    queues = output.get("queues") or []
+    if not isinstance(queues, list):
+        return
+    evidence["sqs_queues"] = queues
```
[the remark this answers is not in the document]
aechlaenm: Fixed in 3d414e1. Repeated calls now merge queues by URL into sqs_queues and the existing catalog entry gets its summary updated instead of a second entry being appended. There is a test for the repeated call case.

integrations/sqs/tools/sqs_queue_attributes_tool/__init__.py
aechlaenm: Fixed in 1fcaf9f. Totals now only count queues with real integer measurements, anything with attributes_error or None counts is reported separately as unreadable, and an all unreadable result shows no counts at all. Added a regression test for the AccessDenied case. The behaviour comparison with this exact scenario is now in the PR description.

aechlaenm:
@greptile review

aechlaenm:
Rebased on latest main and re-ran everything locally, 29 tests plus ruff and mypy all pass. CI is waiting on workflow approval for the new run, could a maintainer approve it when you get a chance?

PR #5640, opened 2026-08-23
fix: parse glued shell operators correctly

Duang777:
## Summary
- Treat shell metacharacters as shell execution triggers when they are unquoted or glued to argv text, e.g. `>out.txt`, `2>&1`, `>>out.txt`, and trailing `&`.
- Ignore quoted shell operators so builtin commands like `cd "dir&name"` keep running in the persistent REPL path.
- Make shell routing quote-aware for POSIX substitutions too: literal single-quoted text like `cd 'dir$(name)'` stays in argv mode, while active substitutions such as `echo "$(date)"` still use the shell.
- Preserve platform shell semantics: POSIX backslash escapes can keep literal operators in argv mode, while Windows path backslashes and apostrophes do not hide real `cmd.exe` operators such as `C:\tmp\& echo done` or `echo '&'`.
- Decode captured foreground CLI timeout output when Python surfaces it as bytes, avoiding user-visible `b'...'` reprs.
- Make Codex OAuth callback tests proxy-isolated for localhost requests so local/system proxy settings cannot turn callback assertions into 502s.
- Add CloudOpsBench Kubernetes evidence mappers so replayed K8s tools contribute citeable evidence instead of only opaque raw blobs.
- Add regression coverage for glued redirection, append redirection, background operators, quoted operators, quoted substitution text, Windows path/operator routing, CLI timeout byte streams, and CloudOpsBench evidence mapping.

## Why
Interactive shell commands with common POSIX forms like `echo hello >out.txt 2>&1` were previously tokenized as plain argv. That meant redirection/background syntax was passed to the executable as literal arguments instead of being interpreted by the shell.

The parser also needs to distinguish real shell operators and substitutions from literal characters inside quoted arguments. Without that, a builtin like `cd "dir&name"` or `cd 'dir$(name)'` can be misclassified as shell mode, which breaks the persistent working-directory behavior across turns.

The quote-aware operator scanner must still respect platform semantics: on POSIX, `\&` can mean a literal ampersand argument, but on Windows a backslash is a path separator and apostrophes are not command quotes, so neither should cause `cmd.exe` operators to bypass shell routing.

Foreground OpenSRE CLI subprocess timeouts had a similar polish/correctness issue: partial stdout/stderr could be rendered as Python byte-string reprs instead of decoded text.

The Codex OAuth callback tests exercise an in-process localhost server. Those requests should bypass any proxy configuration, otherwise proxy-influenced environments can return 502 before the callback server is reached.

CloudOpsBench replayed Kubernetes tools already produced structured results, but they had no `evidence_mapper`, so investigation runs could not cite those tool outputs. Wiring mappers into the tool definitions makes the evidence catalog surface those results directly.

## Tests
- `make lint`
- `make format-check`
- `make typecheck`
- `uv run python -m pytest tests/interactive_shell/shell tests/interactive_shell/test_shell_agent_build.py tests/interactive_shell/test_shell_turn_path_border.py -q`
- `uv run --extra dev python -m pytest tests/interactive_shell/test_cli_foreground.py tests/interactive_shell/runtime/test_subprocess_runner.py -q`
- `uv run --extra dev python -m pytest tests/integrations/llm_cli/test_codex_oauth.py -q`
- `uv run --extra dev python -m pytest tests/benchmarks/cloudopsbench/tools/tests/test_k8s_tools.py tests/tools/investigation/stages/gather_evidence/test_evidence_mapper_coverage.py -q`
- `uv run --extra dev python -m pytest -q -x`
- `uv run --extra dev python -m pytest tests/interactive_shell/shell/test_shell_parsing.py -q`
- `uv run --extra dev python -m pytest tests/interactive_shell/shell tests/interactive_shell/runtime/test_subprocess_runner.py -q`
- `uv run --extra dev ruff check tools/interactive_shell/shell/parsing.py tests/interactive_shell/shell/test_shell_parsing.py`
- `uv run --extra dev ruff check tools/interactive_shell/cli.py tests/interactive_shell/test_cli_foreground.py`
- `uv run --extra dev ruff check tests/integrations/llm_cli/test_codex_oauth.py`
- `uv run --extra dev ruff check tests/benchmarks/cloudopsbench/tools/k8s/__init__.py tests/benchmarks/cloudopsbench/tools/tests/test_k8s_tools.py`

### DIFF
--- a/tests/benchmarks/cloudopsbench/tools/k8s/__init__.py
+++ b/tests/benchmarks/cloudopsbench/tools/k8s/__init__.py
@@ -33,6 +33,7 @@ from __future__ import annotations
 from collections.abc import Callable
 from typing import Any, Protocol, cast
 
+from core.domain.types.evidence import record_evidence_entry
 from core.tool_framework.tool_decorator import tool
 
 # --------------------------------------------------------------------------- #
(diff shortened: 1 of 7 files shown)


Duang777:
@greptile review

Duang777:
@greptile review

Duang777:
@greptile review

Davidson3556:
Closing because this PR is not linked to an issue assigned to the author, and it combines several unrelated changes. Please open or identify focused issues, wait for maintainer confirmation and assignment, and submit one focused PR per assigned issue.

Issue #5256, opened 2026-08-20
Register outbound adapters only from bootstrap — drop import-time registration


YauhenBichel:
**Claiming this issue**: before starting, ask for it in the OpenSRE Discord (https://discord.com/invite/opensre) so a maintainer can assign it to you and nobody duplicates the work.

`integrations/smtp/background_adapter.py` and `integrations/telegram/background_adapter.py` call `register_outbound_adapter(...)` at module import time, so registration depends on who happens to import what. `bootstrap/adapters.py` is the sole sanctioned composition root for tools+integrations wiring (docs/ARCHITECTURE.md).

**Change**: move the two registrations into the bootstrap registration step; the adapter modules only define. Verify no other import-time registrations of the same kind exist (grep `register_outbound_adapter`).

**Verify**: delivery/scheduler tests green; importing either adapter module in isolation registers nothing.

**Run it locally (baseline before, repeat after your change)**
1. One-time setup from the repo root: `make install`
2. Start the shell: `uv run opensre`
   - First run: the onboarding prompts appear — Escape skips GitHub sign-in and the loops picker; submitting a blank LLM key defers key setup (you can add one later with `/auth login <provider>`).
3. Type `good morning` — a normal streamed assistant reply proves the turn engine and your LLM config work end to end.
4. Leave with `/exit`.

Run this once on `main` before you start (so you know what healthy looks like) and again on your branch after the change.

**Manual check for this change**
Start the shell and confirm the banner still shows your `Scheduled tasks (N)` count and `/loops` lists them. Then walk a schedule offer (e.g. `Give me my morning report and schedule it for weekdays at 9:00 AM`) — the delivery-provider choices must still include telegram (and email if configured), proving the adapters registered from bootstrap rather than from whoever imported them.

---
**Ground rules for this change**
- New or changed Protocol methods use a docstring-only body — no `...`, no `pass`, no `raise NotImplementedError` (AGENTS.md, Code Style). Do not convert pre-existing `NotImplementedError` stubs elsewhere — out of scope.
- Remove any retired name or alias completely in the same PR — no compatibility re-exports.
- CI's typecheck does not cover `tests/`: run `uv run mypy <touched test files>` yourself.
- No new module-level mutable state; dependencies arrive via constructor or parameter.
- Before opening the PR follow CI.md (lint, format, typecheck, tests) and fill the PR template, including the AI-usage disclosure.

davincios:
[...]

PR #5650, opened 2026-08-24
refactor(integrations): give eight more vendors a public API (setup, client)

YauhenBichel:
Continues the integrations public-API cleanup. Eight vendors expose their setup
step (and alertmanager its client) through the vendor's main module, and the
setup screens import from there instead of reaching inside the folder.

- alertmanager (client + setup), coralogix, honeycomb, incident_io, new_relic,
  opensearch, pagerduty, vercel: main module re-exports the name(s), __all__ set.
- Four surfaces configurators (+ one validator) import from integrations.<vendor>.
- Nine cleared entries removed from the import check; setup-screen count 73 -> 64.

Not included, on purpose:
- Grafana's reporting_adapter and eight other vendors' side-effect registration
  imports are one cluster covered by #5256 — better done there than per vendor.
- betterstack, dagster, jenkins, tempo have a verifier that imports their own
  main module, so exposing setup there would form an import cycle; they need a
  different approach and are deferred.

No behaviour changes. The setup modules are lightweight (config constants, the
vendor verifier, a shared setup spec), so exposing them through the main module
adds no heavy import and, for these eight, no cycle (verified by importing each).

### DIFF
--- a/integrations/alertmanager/__init__.py
+++ b/integrations/alertmanager/__init__.py
@@ -6,9 +6,14 @@ import logging
 from typing import Any
 
 from integrations._validation_helpers import report_classify_failure
+from integrations.alertmanager.client import make_alertmanager_client
+from integrations.alertmanager.setup import ALERTMANAGER_SETUP
 from integrations.config_models import AlertmanagerIntegrationConfig
 
 logger = logging.getLogger(__name__)
+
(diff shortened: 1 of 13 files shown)


YauhenBichel:
@greptile review

PR #5664, opened 2026-08-24
feat(supabase): add evidence mappers for service health and storage buckets

### LINKED ISSUE #5574: Add evidence mappers for supabase
## Summary
can point to. The `supabase` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `get_supabase_service_health`
- `get_supabase_storage_buckets`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="get_supabase_service_health"' integrations/`.

```python

def _map_get_supabase_service_health(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="get_supabase_service_health",
        )

@tool(name="get_supabase_service_health", ..., evidence_mapper=_map_get_supabase_service_health)
```

## How to check it works

   under `tests/integrations/supabase/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['get_supabase_service_health', 'get_supabase_storage_buckets']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'get_supabase_service_health', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `get_supabase_service_health`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k supabase
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify supabase
# edit /tmp/alert.json so the alert names a resource supabase actually knows about
grep -E 'get_supabase_service_health|get_supabase_storage_buckets' /tmp/rca.json
```

matching lines into the PR.

**No credentials for supabase?** That is a normal outcome, not a reason to go

HariomPtdr:
Fixes #5574

`get_supabase_service_health` and `get_supabase_storage_buckets` already returned useful diagnostic data, but it never became evidence in the incident report — the output was dropped. This adds an evidence mapper for each and attaches it to the tool.

- `integrations/supabase/tools/supabase_health_tool/__init__.py` — `_map_get_supabase_service_health` records one entry summarising how many services were checked and naming any degraded ones.
- `integrations/supabase/tools/supabase_storage_tool/__init__.py` — `_map_get_supabase_storage_buckets` records one entry with the bucket count, a `(of N)` note when the result was truncated, and the names of any public buckets.
- Both names removed from `tests/tools/investigation/stages/gather_evidence/evidence_mapper_baseline.txt`.
- Tests added under `tests/integrations/supabase/`.

Both mappers follow the shipped Grafana example and return early when the relevant field is absent, so an unconfigured or failed call records nothing.

Verification steps from the issue, in order:

**1. Unit tests**
```
$ uv run pytest tests/integrations/supabase/ -q
6 passed in 5.21s
```

**2. The tools carry their mapper**
```
$ uv run python -c "from tools.registry import get_registered_tool as g; print({n: bool(g(n).evidence_mapper) for n in ['get_supabase_service_health','get_supabase_storage_buckets']})"
{'get_supabase_service_health': True, 'get_supabase_storage_buckets': True}
```

**3. Output really becomes report evidence**
```
$ uv run python -c "from tools.investigation.stages.gather_evidence.tools import merge_tool_evidence; ev={}; merge_tool_evidence(ev,'get_supabase_service_health',{'services':{'auth':{'healthy':False}},'degraded_services':['auth']},{}); print(ev.get('catalog_entries'))"
[{'source': 'get_supabase_service_health', 'label': 'Supabase Service Health', 'summary': '1 services checked, degraded: auth', 'url': None, 'snippet': None}]

$ ... merge_tool_evidence(ev2,'get_supabase_storage_buckets',{'buckets':[{'name':'avatars','public':True}],'total_buckets':1},{})
[{'source': 'get_supabase_storage_buckets', 'label': 'Supabase Storage Buckets', 'summary': '1 buckets, public: avatars', 'url': None, 'snippet': None}]
```

**4. Nothing else broke**
```
$ uv run pytest tests/tools/investigation/stages/gather_evidence/test_evidence_mapper_coverage.py
11 passed in 1.16s

$ uv run pytest tests/ -k supabase -q
47 passed, 15086 deselected in 22.29s

$ make lint          -> All checks passed!
$ make format-check  -> 3685 files already formatted
$ make typecheck     -> Success: no issues found in 1916 source files
```

---

The problem is that two Supabase tools produce diagnostic output that the report never cites, so the agent's conclusions about Supabase have nothing visible behind them.

Each mapper takes the tool's output dict and calls `record_evidence_entry` once, which is what makes the result citeable. For health, the meaningful field is `services` (a per-service dict) plus `degraded_services`, so the summary reports the count and names whatever is degraded — that is the part a reader of the report actually needs. For storage, the field is `buckets`; the summary carries the count, and I added the public-bucket names because a bucket unexpectedly being public is the kind of thing a storage incident turns on.

An alternative was to also copy the raw output onto the evidence dict under a namespaced key, which is what the Grafana metrics mapper does (`evidence["grafana_metrics"] = metrics`). I left that out because nothing downstream reads a `supabase_*` key today, and adding unread state seemed worse than keeping the mappers to the one thing the issue asks for. Happy to add it if you would rather these match Grafana exactly.

Both mappers return early when their field is missing or empty, so a `tool_unavailable` response or an empty bucket list records no entry rather than an entry saying "0 items" — covered by two of the tests.

---

### DIFF
--- a/integrations/supabase/tools/supabase_health_tool/__init__.py
+++ b/integrations/supabase/tools/supabase_health_tool/__init__.py
@@ -2,7 +2,9 @@
 
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
+from core.tool import EvidenceType
 from core.tool_framework import tool
 from integrations.supabase import (
     get_service_health,
@@ -10,6 +12,26 @@ from integrations.supabase import (
     supabase_extract_params,
     supabase_is_available,
 )
+
+
+def _map_get_supabase_service_health(
+    evidence: dict[str, Any], output: dict[str, Any], _input: dict[str, Any]
+) -> None:
+    services = output.get("services", {})
+    if not services:
+        return
+    degraded = output.get("degraded_services", [])
+    summary = f"{len(services)} services checked"
+    if degraded:
+        summary += f", degraded: {', '.join(degraded)}"
+    else:
+        summary += ", all healthy"
+    record_evidence_entry(
+        evidence,
+        source="get_supabase_service_health",
+        label="Supabase Service Health",
+        summary=summary,
+    )
 
 
 @tool(
@@ -25,6 +47,8 @@ from integrations.supabase import (
     is_available=supabase_is_available,
     injected_params=("project_url",),
     extract_params=supabase_extract_params,
+    evidence_type=EvidenceType.OTHER,
+    evidence_mapper=_map_get_supabase_service_health,
 )
 def get_supabase_service_health(
     project_url: str,
--- a/integrations/supabase/tools/supabase_storage_tool/__init__.py
+++ b/integrations/supabase/tools/supabase_storage_tool/__init__.py
@@ -2,7 +2,9 @@
 
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
+from core.tool import EvidenceType
 from core.tool_framework import tool
 from integrations.supabase import (
     get_storage_buckets,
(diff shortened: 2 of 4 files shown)


HariomPtdr:
@greptile review

HariomPtdr:
Fixed in 44ca7d3 — the fixtures now use `HTTPStatus.OK` and `HTTPStatus.SERVICE_UNAVAILABLE` instead of numeric literals, per AGENTS.md.

@greptile review

HariomPtdr:
Heads up on the red checks — I believe they're all a concurrency cancellation rather than real failures.

Pushing `44ca7d3` and marking the PR ready for review happened within about a minute of each other, and the resulting workflow set on `44ca7d3` came back `cancelled` (7 cancelled, 5 skipped) with no replacement run. `CI Gate` is the only `failure`, and it's the aggregator job that fails when its dependencies are cancelled.

Everything passed locally on this commit:

```
uv run pytest tests/integrations/supabase/                       6 passed
uv run pytest .../test_evidence_mapper_coverage.py              11 passed
uv run pytest tests/ -k supabase                                47 passed
make lint          -> All checks passed!
make format-check  -> 3685 files already formatted
make typecheck     -> Success: no issues found in 1916 source files
```

Greptile is at 5/5. Could someone re-run the cancelled workflows, or would you prefer I push an empty commit to re-trigger CI?

PR #5678, opened 2026-08-24
refactor(interactive-shell): split ui/streaming/__init__.py, retire paint vocabulary

### LINKED ISSUE #5445: Move ui/streaming/__init__.py into named modules and retire the paint vocabulary
Claiming this issue: ask in OpenSRE Discord (https://discord.com/invite/opensre) before you start so a maintainer can assign it. Ask OpenSRE maintainers and write a commit under this issue

## Goal

`surfaces/interactive_shell/ui/streaming/__init__.py` is 516 lines. A package
`__init__.py` is a façade — the name a caller imports from — and this one is
the whole implementation, with a single 293-line function in the middle of it.
The package already has a `console.py` sibling, so the shape to copy is there.

While moving the code, retire the "paint" vocabulary it carries. Nothing in
this codebase paints; `StreamPaintResult` and `_paint_paragraph_body` are
private jargon for rendering, and the file's own public functions already use
`render_` (`render_markdown_block`, `render_response_header`). The names
disagree with each other inside one file.

The rename and the split are one issue on purpose: they edit the same lines,
and two PRs against this file would conflict.

## What is in it

| Public name | Job |
| --- | --- |
| `render_markdown_block`, `render_response_header` | markdown / header rendering |
| `stream_to_console` | the streaming loop |
| `stream_to_console_state` | the same loop, 293 lines, carrying deferred-closer state |
| `publish_full_response` | the non-TTY path that emits the whole answer at once |
| `finish_deferred_closer` | emitting the held closer and footer |
| `StreamPaintResult` | what a stream call returns |

Private helpers cluster the same way: markdown escaping and fence detection
belong with the renderer, `_paragraph_has_want_me_to` and the closer logic
belong with the deferred-closer path.

## Target

`__init__.py` re-exports and nothing else. Module names are the contributor's
call, but each should be nameable in one noun — the markdown renderer, the
stream loop, the deferred closer. `stream_to_console_state` at 293 lines is
the real work; if it splits naturally along its state transitions, do that, but
a pure move is an acceptable PR — say in the description which you did.

`StreamPaintResult` → `StreamRenderResult`, `_paint_paragraph_body` →
`_render_paragraph_body`. Two files import the type: this one and
`runtime/agent_harness_adapters.py` (which also has a `self._paint` field).

## Acceptance

- [ ] `ui/streaming/__init__.py` contains imports and `__all__` only
- [ ] `__all__` is unchanged — same names, same spelling, importable from the
      same path, so no caller outside the package edits an import
- [ ] `grep -rn "Paint\|_paint" --include='*.py' surfaces/` prints nothing
- [ ] Each new module states its one job in its docstring
- [ ] `uv run python -m pytest tests/interactive_shell/ui/test_streaming.py
      tests/interactive_shell/ui/test_streaming_console_output.py -q` passes
      with no test edited except the renamed symbol
- [ ] Streaming still looks right by eye — `uv run opensre`, ask something long
      enough to stream, and confirm paragraphs, code fences and the closing
      line render as they did before. The tests do not cover how it looks.

## Out of scope

`core/agent_harness/session_goal/run_until.py` and
`tests/core/agent_harness/session_goal/test_progress.py` also say "paint". That
is the session-goal progress line, a different surface with its own naming
call. One package per PR.

arnavpopley:
Closes #5445

Splits the 516-line `surfaces/interactive_shell/ui/streaming/__init__.py` into three
named modules, and retires the "paint" vocabulary it carried (the rename and split
touch the same lines, so they're one PR per the issue).

- `renderer.py` — Markdown block and response-header rendering
  (`render_markdown_block`, `render_response_header`, `STREAM_LABEL_ASSISTANT`,
  `STREAM_LABEL_ANSWER`, and the dunder-filename escaping helper).
- `loop.py` — the paragraph-by-paragraph streaming loop: `stream_to_console`,
  `stream_to_console_state`, `publish_full_response`, and the renamed
  `StreamRenderResult` dataclass (was `StreamPaintResult`). This is the 293-line
  function the issue calls out as "the real work" — it did not split naturally
  along its state transitions (the closures share too much local buffer/index
  state), so this is a pure move for that function, as the issue allows.
- `closer.py` — flushing a deferred Want-me-to closer + held footer after gather
  normalizes the answer (`finish_deferred_closer`).
- `__init__.py` — now contains only imports and `__all__` (unchanged: same 11
  names, same importable path, minus the one deliberate rename below).

**The "paint" rename**, per the issue: `StreamPaintResult` → `StreamRenderResult`,
`_paint_paragraph_body` → `_render_paragraph_body`. The issue flagged a second
importer of the type — `surfaces/interactive_shell/runtime/agent_harness_adapters.py`,
which also carried a `self._paint` field — renamed to `self._stream_result` there
(and the local `paint` variables to `stream_result`), plus one "paint" mention in a
docstring. `grep -rn "Paint|_paint" --include='*.py' surfaces/` now returns nothing.

**One deliberate non-obvious piece**: `renderer._build_markdown_block` constructs
through `surfaces.interactive_shell.ui.streaming.Markdown` (the package façade)
rather than importing `Markdown` directly, and `__init__.py` re-exports `Markdown`
for exactly this reason (with a `noqa: F401` + comment, since it's not part of
`__all__`). `tests/interactive_shell/ui/test_streaming.py` substitutes the class via
`streaming_module.Markdown = _SpyMarkdown` to count constructions — a plain
`from rich.markdown import Markdown` inside `renderer.py` would bind a copy that
patch never reaches, since Python resolves a bare name through the *defining*
module's globals, not the importing package's. Went with this instead of editing
the test, since the issue requires the existing test suite to pass with no test
edited except the renamed symbol (and `StreamPaintResult` doesn't even appear in
that file). Also had to re-export `_CHARS_PER_TOKEN` from `__init__.py` (same
`noqa: F401` treatment) — two test files outside the two named in the issue
(`tests/interactive_shell/runtime/test_token_accounting.py`,
`tests/interactive_shell/test_terminal_runtime.py`) import it directly from this
package path, and the original flat file exposed it as a module attribute even
though it's private and not in `__all__`.

Ran a real streaming turn through `uv run opensre` in a genuine pseudo-tty
(`script -q`), asking for two prose paragraphs, a fenced Python code block, and a
bulleted list — enough to exercise paragraph flushing, code-fence detection, and
the end-of-stream footer through the split `renderer.py` + `loop.py`. All rendered
correctly:

```
● assistant

A load balancer sits in front of a pool of backend servers and distributes
incoming client requests across them, so no single server gets overwhelmed while
others sit idle. It typically health-checks each backend and stops routing
traffic to instances that fail, which is how a fleet keeps serving traffic even
when individual nodes crash or get pulled for deploys.

Beyond raw distribution, load balancers terminate TLS, enforce timeouts and
retries, and expose a single stable endpoint (DNS name or VIP) that hides the
churn of instances behind it. Algorithms like round-robin, least-connections, or
consistent hashing decide which backend gets the next request — the choice
matters when sessions are sticky or backends have uneven capacity.

 import itertools

 backends = ["10.0.0.1", "10.0.0.2", "10.0.0.3"]
 rr = itertools.cycle(backends)

 def route(request):
  return next(rr) # round-robin pick

 for i in range(5):
  print(f"req {i} -> {route(i)}")

 • High availability — unhealthy backends are removed from rotation
   automatically, so failures don't take the service down.
 • Horizontal scaling — add or remove backends behind the same endpoint without
   clients noticing.
 • Operational leverage — one place to terminate TLS, apply rate limits, and
   shift traffic during deploys or incidents.
· 8.5s · ↓ 331 tokens
```

Quality gates:
```
$ grep -rn "Paint\|_paint" --include='*.py' surfaces/
(no output)

$ uv run python -m pytest tests/interactive_shell/ui/test_streaming.py tests/interactive_shell/ui/test_streaming_console_output.py -q
56 passed in 2.07s
# no test file edited — StreamPaintResult never appeared in test_streaming.py,
# so the rename required zero test changes

$ uv run python -m pytest tests/interactive_shell -q
1453 passed, 1 skipped

$ make lint
All checks passed!

$ make format-check
3692 files already formatted

$ make typecheck
Success: no issues found in 1925 source files
```

---

The issue asked for a mechanical split along "the markdown renderer, the stream
loop, the deferred closer" plus a vocabulary rename, with an explicit test-suite
constraint: the existing tests must pass with no edits beyond the renamed symbol.
That constraint drove the one non-trivial design decision — `_build_markdown_block`
had to keep resolving `Markdown` through the package's own namespace rather than a
normal import, because the test monkeypatches the package attribute directly and
Python's name resolution for a bare identifier inside a function always goes
through that function's *defining* module's globals, never the importing module's.
I verified this was actually necessary (not over-engineering) by implementing the
naive split first, running the test suite, and confirming those specific tests failed
before adding the façade indirection — then confirmed all 56 passed with it in place
and zero test edits. I chose `loop.py` as the home for `publish_full_response`
(rather than a fourth module) because it's the direct non-TTY counterpart of
`stream_to_console_state`'s own non-TTY branch, not a distinct concept from "the
stream loop" the issue names. I verified the split didn't drop anything by grepping
the whole repo for the old symbol names before and after, and by running the full
`tests/interactive_shell` suite (not just the two named test files) to catch any
indirect importer I might have missed.

---

---


arnavpopley:
@greptile review

PR #5681, opened 2026-08-24
refactor(scheduler): inject the runner bundle instead of a module global

### LINKED ISSUE #5680: Inject the scheduler's runners instead of a module global
## The problem

infrastructure/scheduling/scheduler/agent_runner.py (and its twin investigation_runner.py) held a module-level global runner, set with register_*() and read with invoke_*(). That is the service-locator shape the no-global rule rejects, and it was on main. It existed for a real reason — the scheduler package may not import tools/integrations (layering, #3352) — but the fix is to pass the runner in, not stash it.

## What this does
- agent_runner.py / investigation_runner.py now only declare the runner Protocol and payload types.
- The SchedulerRunners bundle (already assembled by the composition root) is passed into start_scheduler / start_background_scheduler and threaded down to the task builders, which call runners.agent / runners.investigation.
- Every host builds the bundle at the start call site via bootstrap.adapters.scheduler_runners(); the gateway keeps its gated bundle and passes it to start and reload. The install()/global machinery is removed.

YauhenBichel:
Fixes #5680 

The scheduler read its agent and investigation runners from module globals
(register_*/get_*/invoke_*) — the service-locator shape the no-global rule
rejects, and it was on main. This passes the runners in instead.

- infrastructure/scheduling/scheduler/agent_runner.py and investigation_runner.py
  now only declare the runner Protocol and payload types.
- SchedulerRunners (already built by bootstrap.adapters) is threaded:
  start_scheduler / start_background_scheduler take it, _register_jobs binds it to
  each APScheduler job (in-memory job store, so passing the object is safe),
  execute_task -> build_message -> the 11 _build_* builders call runners.agent /
  runners.investigation.
- Hosts build it at the call site: cron start / run, the shell prompt-loop
  scheduler, the gateway controller (stores the gated bundle, passes it to start
  and reload), and the sentry/posthog CLI commands.
- Removed the install()/global machinery (install_scheduler_runners boot function,
  the dead install() in scheduled_agent_bootstrap and the investigation bootstrap)
  and deleted the forwarding-only integrations/sentry/scheduler_bootstrap.py. The
  SCHEDULER_RUNNERS boot step now installs only the delivery adapters.
- ~50 test sites migrated to build a bundle (tests/scheduler/_bundle.py helper);
  the gate tests now exercise runners.gated(gate).agent(payload) directly.

Layering is preserved: the scheduler package still never imports tools/integrations;
the composition root builds the concrete runners and hands them in.


YauhenBichel:
@greptile review

PR #5682, opened 2026-08-24
refactor(agent_harness): inject the investigation runner instead of a module global

YauhenBichel:
What: AgentSession.investigate() receives its payload runner as a parameter; the module-global runner + its install/get/reset helpers and the install_investigation_api boot step are deleted.
Why: removes hidden mutable global state and the "install before you call" ordering footgun; makes the dependency explicit at each call site. Same pattern as the scheduler-runner injection (#5681).


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5685, opened 2026-08-24
feat(concurrency): make the LLM client cache thread-safe and pool investigation workers

### LINKED ISSUE #5684: Make raising the turn limit safe: thread-safe LLM client cache + investigation worker pool
## Problem

The process turn gate is already tunable — `OPENSRE_MAX_CONCURRENT_TURNS` and the
size profile (SMALL=1, MEDIUM=2, LARGE=4) set how many turns run at once. But two
things break or bottleneck the moment the limit goes above 1:

1. **The shared LLM client cache is not concurrency-safe.**
   `core/llm/internal/client_cache.py` has no lock and holds a single process-wide
   config key; `store()` never rechecks it. A `/model` switch or env sync
   (`reset_llm_clients()` → `clear()`) racing an in-flight build can cache a client
   built for the old configuration and serve it under the new key.

2. **Queued investigations run in a single worker thread.**
   `gateway/web/worker.py` runs one `InvestigationWorker` thread, so
   investigation throughput stays one per process no matter how high the turn
   limit is set.

## Why it matters

Raising `OPENSRE_MAX_CONCURRENT_TURNS` is the documented way to serve more
concurrent users on one task. But (1) is a latent correctness bug that only shows
up under real overlap, and (2) means investigations never actually run in
parallel. This is the concrete answer to #5492 ("run the suites with the limit
raised, report what breaks").

## What to change

- Lock the LLM client cache (`get`/`store`/`clear`) and have `store` carry the
  config key it was built under, dropping a client whose key changed during the
  build. `get_llm` captures the key once and passes it through.
- Give the investigation worker a bounded pool (`OPENSRE_INVESTIGATION_WORKERS`,
  default 1). The queue claim is already atomic (Postgres `FOR UPDATE SKIP
  LOCKED`), so multiple threads are safe.

## Done when

- A `/model` switch racing an in-flight build cannot serve a stale-config client
  (covered by a deterministic test, no threads).
- With a worker pool of 2 and the gate at 2, two investigations run at the same
  time (covered by a test).
- `OPENSRE_INVESTIGATION_WORKERS` is documented in the deployment guide.
- Same-session turns still serialize (unchanged by design).

## Not in scope

- Collapsing the duplicate turn entry point (`chat()` vs the host loop) — tracked
  in #5483.
- Raising the default concurrency, DB pool sizing, and the cross-process session
  lock — deployment configuration, not code.

Related: #5492, #5483.

YauhenBichel:
Fixes #5684

The process turn gate is already N-capable (OPENSRE_MAX_CONCURRENT_TURNS, size profiles), but two things break or bottleneck the moment the turn limit goes above 1. This PR fixes both — the prerequisites to safely serving concurrent turns.

1. The shared LLM client cache is now thread-safe. LLMClientCache (core/llm/internal/client_cache.py) was unlocked and held a single process-wide config key, and store() never rechecked it — so a /model switch or env sync racing an in-flight build could cache a client built for the old configuration and serve it under the new key. Added a threading.Lock around get/store/clear, and store(role, client, config_key) now drops a client whose config changed during the build. get_llm captures the key once and passes it through.

2. The investigation worker is now a bounded pool. InvestigationWorker (gateway/web/worker.py) ran a single thread, capping queued-investigation throughput at one per process regardless of the turn limit. Added pool_size (env OPENSRE_INVESTIGATION_WORKERS, default 1); start() spins N threads. The queue claim is already atomic (Postgres FOR UPDATE SKIP LOCKED), so no record runs twice. Also moved WORKER_ENABLED_ENV into config/constants/gateway.py per the env-name convention.

Per-session serialization is unchanged (same-session turns still serialize by design). The remaining scaling levers — default concurrency, DB pool size, cross-process session lock — are deployment config and already documented; this PR adds the new worker knob to docs/deployment.mdx.

The goal was to make raising OPENSRE_MAX_CONCURRENT_TURNS safe. Scoping the concurrency model showed the gate itself already supports N; the real gaps were a latent data race in the client cache and a single-threaded investigation worker.

For the cache, the alternative was holding the lock across the client build — rejected, since the build does I/O and would serialize all turns. Instead each cache operation is atomic and the build stays outside the lock; store verifying the key handles the "config changed mid-build" race without a long-held lock, at the cost of a rare, harmless rebuild.

For the worker, the alternative was a process/queue redesign — unnecessary, because the claim is already atomic, so the minimal correct change is just more threads sharing one stop event, bounded by an env var that defaults to today's behaviour (1).

Key pieces: LLMClientCache.get/store/clear (locked; store drops stale-key clients), factory.get_llm (captures the key once), InvestigationWorker.start (spins pool_size daemon threads), _configured_worker_count (parses the env, defaults to 1 on unset/invalid).


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5688, opened 2026-08-24
refactor(integrations): declare public APIs for sentry and the mcp vendors

YauhenBichel:
refactor(integrations): declare public APIs for sentry and the mcp vendors


YauhenBichel:
@greptile review

Issue #5304, opened 2026-08-20
[BUG] bitbucket and supabase never resolve as effective — same missing direct_effective as prefect (#5147)


cloudenochcsis:
### Summary

`bitbucket` and `supabase` have the exact bug fixed for prefect in #5147 / #5216: both have registered classifiers and env loaders, so `classify_integrations()` resolves their saved credentials correctly — and `resolve_effective_integrations()` then silently drops the result because their `IntegrationSpec`s are missing `direct_effective=True`. There is currently no way (store record or env vars) to make either integration effective.

### Expected vs actual behavior

**Expected:** a bitbucket/supabase record saved in the local store (or configured via `BITBUCKET_WORKSPACE…` / `SUPABASE_URL…` env vars) resolves as effective, so `opensre integrations verify bitbucket|supabase` passes and the tools' availability checks see credentials.

**Actual:** `resolve_effective_integrations()` only publishes services in `DIRECT_CLASSIFIED_EFFECTIVE_SERVICES` (built from `spec.direct_effective`) plus a few bespoke blocks. Both services classify successfully and are then dropped, so `integrations/verify.py` reports status "missing" for them regardless of configuration. This also makes the open verification tasks #5109 (bitbucket) and #5161 (supabase) unfulfillable as-is.

### Steps to reproduce

From the repo root (mirrors the #5216 regression test):

```
uv run python -c "
from integrations.catalog import resolve_effective_integrations
for service, creds in [
    ('bitbucket', {'workspace': 'acme', 'username': 'u', 'app_password': 'p'}),
    ('supabase', {'url': 'https://x.supabase.co', 'service_key': 'k'}),
]:
    rec = {'id': service + '-1', 'service': service, 'status': 'active', 'credentials': creds}
    resolved = resolve_effective_integrations(store_integrations=[rec])
    print(service, 'published:', service in resolved)
"
```

Output on main:

```
bitbucket published: False
supabase published: False
```

### Root cause / proposed fix

Same one-of-N-siblings omission as prefect: add `direct_effective=True` to both specs in `integrations/registry.py`, add the missing `supabase` field to `EffectiveIntegrations` (bitbucket already has one), and clone the #5216 regression tests.

[...]

PR #5690, opened 2026-08-24
fix(integrations): give rds its own verifier instead of folding into aws

### LINKED ISSUE #5093: Decide how rds verifies: own verifier or under aws
## Goal

`rds` has `setup.py` and `tools/` but no verifier, so it cannot be verified on
its own. Decide whether it should have one, or whether it verifies through
`aws`, and record the answer.

This is a decision issue first and a code issue second.

## The question

RDS is reached with AWS credentials, and `aws` already verifies. Two options:

1. **Own verifier** — `integrations/rds/verifier.py` probing an RDS-specific
   call, so a user who configured only RDS gets a direct answer
2. **Fold under `aws`** — no separate verifier; the matrix records that RDS is
   covered by the `aws` row, and `opensre integrations verify rds` either
   delegates or says so plainly

Option 2 is cheaper but leaves `verify rds` reporting nothing useful today,
which is the actual complaint.

## Acceptance

Whichever is chosen:

- [ ] `uv run opensre integrations verify rds` gives an accurate answer — a real
      probe result, or a clear statement that AWS covers it
- [ ] The support matrix reflects the decision
- [ ] If a verifier is added, it follows `integrations/pagerduty/verifier.py`

## Related

The same question applies to the other AWS satellites (cloudwatch, cloudtrail,
ec2, eks, elb, s3, sqs, lambda). Answer it once here and apply the outcome to
all of them rather than opening eight issues.

---

Part of #5087.

arnavpopley:
Closes #5093

Decides the issue's own question — own verifier vs. fold under `aws` — in
favor of **its own verifier**, and implements it.

**Why own verifier, not fold under `aws`:** `eks` is already folded into `aws`
today (it's a pure alias in `integrations/registry.py` — `opensre integrations
verify eks` runs the exact same generic STS caller-identity check as `verify
aws`). RDS is structurally different: it's already a **distinct, standalone
catalogued integration** with its own setup flow
(`db_instance_identifier` + `region`, see `integrations/rds/setup.py`), not an
alias. And there's already a real, per-instance AWS call available —
`describe_db_instances` — which the `describe_rds_instance` investigation tool
already uses via `execute_aws_sdk_call`. That gives a genuinely RDS-specific
answer ("does this configured instance exist and is it reachable"), which is
strictly more useful than the generic "AWS credentials work" answer `aws`'s
verifier gives — and the issue's own text calls that generic answer out as
"the actual complaint" with the fold-under-aws option.

`integrations/rds/verifier.py` is hand-written (`@register_verifier("rds")`)
rather than built on `register_probe_verifier`, following
`integrations/aws/verifier.py`'s precedent — same reason: RDS has no client
class with a `.probe_access()` method to plug into the probe helper, it goes
through the same safety-validated `execute_aws_sdk_call` the
`describe_rds_instance` tool already uses.

**A real prerequisite bug this PR also had to fix:** while wiring this up I
found `rds`'s `IntegrationSpec` was missing `direct_effective=True`, and
`EffectiveIntegrations` (`integrations/effective_models.py`) had no `rds`
field at all. That's the exact same bug class as prefect (#5147) and
bitbucket/supabase (#5304) — `classify_integrations()` resolves a saved RDS
config correctly, but `resolve_effective_integrations()` silently drops it
before any verifier ever sees it. It had never surfaced before because RDS
never had a verifier to notice the gap with. Confirmed live, before and after
(see Demo below) — without this fix, the new verifier would always report
"missing" even with fully valid configuration, making it pointless. Fixed
both: `direct_effective=True` on the `rds` spec, and a `rds:
EffectiveIntegrationEntry | None = None` field on `EffectiveIntegrations`.

**On the other 7 AWS satellites** the issue asks to apply this decision to
(cloudwatch, cloudtrail, ec2, eks, elb, s3, sqs, lambda): I checked each one
in `integrations/registry.py`. Only `eks` has a registry entry at all — as an
`aws` alias. The rest (`cloudwatch`, `cloudtrail`, `ec2`, `elb`, `s3`, `sqs`,
`aws_lambda`) aren't registered as distinct `IntegrationSpec` entries in the
first place — they exist as tool-only directories under `integrations/`, not
as standalone catalogued integrations. So "own verifier vs. fold under aws"
isn't a meaningful question for them yet the way it was for RDS; the decision
doesn't mechanically transfer without first giving each one a registry entry
of its own. Recording that finding here per the issue's ask ("answer it once
here... rather than opening eight issues") rather than building 7 more
verifiers speculatively in this PR — RDS was uniquely positioned as the one
satellite that was already a standalone integration missing only a verifier.

**The `direct_effective` bug, live before the fix** (env-based RDS config,
unmodified code):
```
$ export OPENSRE_HOME=/tmp/opensre_rds_repro RDS_DB_INSTANCE_IDENTIFIER=checkout-prod AWS_REGION=us-east-1
$ uv run python -c "
from integrations._catalog_impl import resolve_effective_integrations
effective = resolve_effective_integrations()
print('rds in effective:', 'rds' in effective)
"
rds in effective: False
```

**After the fix**, same env, same command:
```
rds in effective: True
{'source': 'local env', 'config': {'db_instance_identifier': 'checkout-prod', 'region': 'us-east-1', 'integration_id': 'env-rds'}}
```

**The real CLI, end to end** (no AWS credentials configured in this shell, so
a real failure — proving the verifier is wired and reaches the real AWS SDK
call, not "verification is not supported for this service"):
```
$ uv run opensre integrations verify rds
━━━━━━━━┿━━━━━━━━━━━┿━━━━━━━━━━┿━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
rds     │ local env │ ✗ failed │ AWS credentials not configured: Unable to
        │           │          │ locate credentials
```

**The verifier's three branches, live-tested directly:**
```
$ uv run python -c "
from unittest.mock import patch
from integrations.rds.verifier import verify_rds

fake_success = {'success': True, 'data': {'DBInstances': [{'DBInstanceStatus': 'available', 'Engine': 'postgres'}]}}
with patch('integrations.rds.verifier.execute_aws_sdk_call', return_value=fake_success):
    print(verify_rds('local env', {'db_instance_identifier': 'checkout-prod', 'region': 'us-east-1'}))

fake_not_found = {'success': True, 'data': {'DBInstances': []}}
with patch('integrations.rds.verifier.execute_aws_sdk_call', return_value=fake_not_found):
    print(verify_rds('local env', {'db_instance_identifier': 'nope', 'region': 'us-east-1'}))

print(verify_rds('local env', {}))
"
{'service': 'rds', 'source': 'local env', 'status': 'passed', 'detail': 'Connected to RDS in us-east-1; instance checkout-prod engine=postgres status=available.'}
{'service': 'rds', 'source': 'local env', 'status': 'failed', 'detail': "No RDS instance found with identifier 'nope' in us-east-1."}
{'service': 'rds', 'source': 'local env', 'status': 'missing', 'detail': 'Missing db_instance_identifier or region.'}
```

Tests:
```
$ uv run python -m pytest tests/integrations/test_verify.py tests/integrations/test_catalog_multi_instance.py tests/integrations/test_rds.py tests/integrations/test_registry.py tests/integrations/test_registry_invariants.py -q
89 passed

$ uv run python -m pytest tests/integrations -q
3630 passed, 7 skipped

$ make lint
All checks passed!

$ make format && make format-check
3697 files already formatted

$ make typecheck
Success: no issues found in 1925 source files
```

---

The issue framed this as "decide first, code second," so I started by reading
how the codebase already answers this question for `eks` (a pure alias into
`aws`), then compared RDS's actual state: it already has its own setup flow
and identifying field (`db_instance_identifier`), so aliasing it into `aws`
would be a real behavior regression, not just "cheaper" — a user with only RDS
configured would lose their distinct setup. Once I decided "own verifier," I
followed the two existing hand-written-verifier precedents
(`aws/verifier.py`, since RDS has the same "no simple client class" shape) and
the existing `describe_rds_instance` tool (since it already does the exact
AWS call I needed, through the same safety-validated `execute_aws_sdk_call`
path — I didn't want to introduce a second way to call AWS SDK operations).
Before assuming the verifier alone was enough, I ran it end-to-end against a
real config and found it silently failed — traced that to
`resolve_effective_integrations()` dropping RDS entirely, recognized the
pattern from the (unmerged) prior attempts at the same bug class for prefect
and bitbucket/supabase, and reproduced it live (not just by reading code)
before and after the fix so the PR has real evidence, not an assumption. For
the "apply to the other 8 satellites" part of the issue, I checked each one's
actual registry state instead of assuming they were all in RDS's position —
finding that 7 of the 8 aren't even registered integrations yet meant the
literal instruction ("apply the outcome to all of them") doesn't mechanically
make sense yet, so I documented that finding rather than either ignoring it or
building unrequested scope.

---

---


arnavpopley:
Fixed the failing test — \`test_integrations_setup_skips_auto_verify_for_unverifiable_service\` was pinned to rds lacking a verifier (the diff between SETUP_SERVICES and VERIFY_SERVICES was already empty once this PR lands, confirmed live). Same class of break this test's own comment documents happening once before with opensearch/#1143 — fixed the same way, but monkeypatched VERIFY_SERVICES for this test specifically so the next verifier addition doesn't repeat it. The other 2 CI failures (\`test_anthropic_model_not_found_raises_opensre_error\`, an \`opencode\` onboarding smoke test) are pre-existing on \`main\` — confirmed via zero diff against \`origin/main\` on both files, unrelated to this change.

On the Greptile finding about sanitizing the raw AWS error before display: pushing back on this one rather than changing it. Per this repo's own \`AGENTS.md\` (the CWE-209 footgun entry): *"the local CLI/terminal sink is not external — it may show detail. Redact at the sink/response boundary, not per call site."* \`opensre integrations verify rds\` is exactly that local CLI/terminal path — I checked and \`verify_integrations\`/\`format_verification_results\` are never consumed by \`gateway/\` (Slack/Telegram), only by the CLI and the interactive shell's own terminal output. This also matches the existing, unmodified precedent right next to it: \`integrations/aws/verifier.py\`'s \`verify_aws\` does the identical thing (\`f"AWS STS check failed: {exc}"\`, straight into the same \`result()\` detail field). Sanitizing only the new RDS verifier would make it inconsistent with its own sibling for no real benefit — the detail here is boto3/AWS-level text (e.g. "Unable to locate credentials"), not a secret. Happy to revisit if there's a path to this output I'm missing.

@greptile review

arnavpopley:
Checked this claim concretely rather than assuming either way — traced the whole path.

\`last_observation\` and the discovery-summarization mechanism are real (\`surfaces/interactive_shell/command_registry/integrations.py\` \`_record_integrations_observation\`, read by \`core/agent_harness/turns/turn_snapshot.py\`). But the claim that this reaches "externally delivered gateway chat responses" doesn't hold up against this repo's own architecture:

- \`grep -rln "last_observation" .\` across the whole repo returns exactly 4 files: \`core/agent_harness/turns/turn_snapshot.py\`, \`core/agent_harness/session/session_core.py\`, \`core/state/agent_state.py\` (shared plumbing — a plain getter/setter), and \`surfaces/interactive_shell/command_registry/integrations.py\` — the **only** writer anywhere in the codebase.
- \`grep -rln "surfaces.interactive_shell" gateway/\` returns nothing. Gateway never imports this module, so nothing under \`gateway/\` can ever call \`_record_integrations_observation\` or write to \`last_observation\`.
- Per \`docs/ARCHITECTURE.md\` (line 221): *"gateway receives a message, resolves session state from its own storage"* — \`gateway/core/{session,storage,middleware}\`, structurally separate from interactive-shell sessions. A Slack/Telegram session is never the same object as, and never resolves from, a local REPL session — so there's no session-sharing path for a gateway turn to inherit an interactive-shell-written \`last_observation\` either.

So the field this PR's verifier detail could theoretically end up in is provably unreachable from any gateway code path — both because gateway never writes it and because gateway sessions aren't the same objects as interactive-shell sessions in the first place. And to be clear about scope even if I'm wrong about the above: this mechanism, if it were a real leak, would apply identically to every existing verifier's raw detail (aws, datadog, pagerduty, sentry, ...) that already flows through the same \`_record_integrations_observation\` call — not something \`integrations/rds/verifier.py\` introduces or is uniquely responsible for.

Holding my position on this one — happy to be shown a concrete gateway code path I've missed, but I don't see one.

arnavpopley:
@greptile review

arnavpopley:
@greptile review

arnavpopley:
Merged latest main to resolve a conflict — someone else independently landed an equivalent fix to the same Anthropic-model test I'd fixed for CI Gate; took their version since it's the same fix. No other conflicts. @greptile review

arnavpopley:
Closing for now

PR #5712, opened 2026-08-25
feat(gitlab): add evidence mappers (#5552)

### LINKED ISSUE #5552: Add evidence mappers for gitlab
## Summary
can point to. The `gitlab` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `get_gitlab_file`
- `list_gitlab_commits`
- `list_gitlab_mrs`
- `list_gitlab_pipelines`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="get_gitlab_file"' integrations/`.

```python

def _map_get_gitlab_file(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="get_gitlab_file",
        )

@tool(name="get_gitlab_file", ..., evidence_mapper=_map_get_gitlab_file)
```

## How to check it works

   under `tests/integrations/gitlab/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['get_gitlab_file', 'list_gitlab_commits', 'list_gitlab_mrs', 'list_gitlab_pipelines']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'get_gitlab_file', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `get_gitlab_file`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k gitlab
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify gitlab
# edit /tmp/alert.json so the alert names a resource gitlab actually knows about
grep -E 'get_gitlab_file|list_gitlab_commits|list_gitlab_mrs|list_gitlab_pipelines' /tmp/rca.json
```

matching lines into the PR.

**No credentials for gitlab?** That is a normal outcome, not a reason to go

MeghanBao:
Fixes #5552 (part of epic #5519)

Wire the four read-only GitLab tools to `record_evidence_entry` so their output becomes citeable report evidence instead of being dropped:

| Tool | Records | Gate |
| --- | --- | --- |
| `list_gitlab_commits` | `"N recent commits"` | `commits` non-empty |
| `list_gitlab_mrs` | `"N recent merge requests"` | `mrs` non-empty |
| `list_gitlab_pipelines` | `"N recent pipelines"` | `pipelines` non-empty |
| `get_gitlab_file` | `"<path> (N lines, M chars)"` | nested `file.content` non-empty |

Removes the four tools from `evidence_mapper_baseline.txt` and adds unit tests. All four return diagnostic data; no action/notification tools touched.

### Demo — actual before/after

Real `merge_tool_evidence` output, `main` vs this branch. "before" = the summary recorded on `main` (these tools had no mapper, so their output was dropped); "after" = the summary this branch records. Reproduced by feeding each payload through `merge_tool_evidence` and reading `catalog_entries[0]["summary"]`:

```
tool                    input                                    before (main)  ->  after (this branch)
list_gitlab_commits     {"commits": [ {id:1}, {id:2}, {id:3} ]}  (dropped)      ->  3 recent commits
list_gitlab_mrs         {"mrs": [ {id:1}, {id:2} ]}              (dropped)      ->  2 recent merge requests
list_gitlab_pipelines   {"pipelines": [ {id:1} ]}                (dropped)      ->  1 recent pipelines
get_gitlab_file         {"file": {file_path: "config/app.yaml",  (dropped)      ->  config/app.yaml (3 lines, 15 chars)
                          content: "a: 1\nb: 2\nc: 3\n"}}
list_gitlab_commits     {"commits": []}                          (nothing)      ->  (nothing)   # no "0 items" noise
get_gitlab_file         {"available": False}                     (nothing)      ->  (nothing)
```

`get_gitlab_file` records path + size only — the file body is never dumped into the summary (context budget). Empty/unavailable/too-large/binary all record nothing.

Line counts use `len(content.splitlines())`, so a newline-terminated file (`a\nb\nc\n`) reports 3 lines, not 4 (review feedback — was `count("\n") + 1`).

Verification: `pytest -k gitlab` 160 passed · new mapper tests 6 passed (incl. trailing-newline regression) · ruff + mypy clean.

---

## Checklist
- [x] Proper PR title and linked issue
- [x] Self-review performed
- [x] I can explain every block I added
- [x] Tested thoroughly, edge cases considered
- [x] Added tests
- [x] Follows project style guidelines

### DIFF
--- a/integrations/gitlab/tools/gitlab_file_tool/__init__.py
+++ b/integrations/gitlab/tools/gitlab_file_tool/__init__.py
@@ -41,7 +41,7 @@ def _map_get_gitlab_file(
     file = output.get("file") or {}
     content = file.get("content", "")
     if content:
-        lines = content.count("\n") + 1
+        lines = len(content.splitlines())
         record_evidence_entry(
             evidence,
             source="get_gitlab_file",
--- a/integrations/gitlab/tools/gitlab_file_tool/__init__.py
+++ b/integrations/gitlab/tools/gitlab_file_tool/__init__.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import base64
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
 from core.tool_framework import tool
 from core.tool_framework.utils import code_host_unavailable_payload, tool_unavailable
@@ -33,9 +34,26 @@ def _get_gitlab_file_available(sources: dict[str, dict]) -> bool:
     return bool(_gitlab_available(sources) and gl.get("project_id") and gl.get("file_path"))
 
 
+def _map_get_gitlab_file(
+    evidence: dict[str, Any], output: dict[str, Any], _tool_input: dict[str, Any]
+) -> None:
+    """Record a fetched GitLab file as citeable evidence (identity only, not contents)."""
+    file = output.get("file") or {}
+    content = file.get("content", "")
+    if content:
+        lines = content.count("\n") + 1
+        record_evidence_entry(
+            evidence,
+            source="get_gitlab_file",
+            label="GitLab File",
+            summary=f"{file.get('file_path') or 'file'} ({lines} lines, {len(content)} chars)",
+        )
+
+
 @tool(
     name="get_gitlab_file",
     source="gitlab",
+    evidence_mapper=_map_get_gitlab_file,
     description="Read the contents of a specific file from a GitLab repository.",
     use_cases=[
         "Reading a config file that may explain a misconfiguration causing the incident",
--- a/integrations/gitlab/tools/gitlab_commits_tool/__init__.py
+++ b/integrations/gitlab/tools/gitlab_commits_tool/__init__.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
 from core.tool_framework import tool
 from core.tool_framework.utils import code_host_unavailable_payload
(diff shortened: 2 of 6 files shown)


MeghanBao:
@greptile review

integrations/gitlab/tools/gitlab_file_tool/__init__.py
```diff
@@ -33,9 +34,26 @@ def _get_gitlab_file_available(sources: dict[str, dict]) -> bool:
...
 
 
+def _map_get_gitlab_file(
+    evidence: dict[str, Any], output: dict[str, Any], _tool_input: dict[str, Any]
+) -> None:
+    """Record a fetched GitLab file as citeable evidence (identity only, not contents)."""
+    file = output.get("file") or {}
+    content = file.get("content", "")
+    if content:
+        lines = content.count("\n") + 1
```
Devesh36: line counts are off by one for newline-terminated files. For example, a\nb\nc\n is reported as 4 lines instead of 3, making the evidence summary inaccurate. Use len(content.splitlines()) or equivalent.

Devesh36 (commented):
The PR description does not include the actual before/after output diff.

integrations/gitlab/tools/gitlab_file_tool/__init__.py
MeghanBao: Good catch — fixed in 76f8f8b. Switched to `len(content.splitlines())` so a newline-terminated file (`a\nb\nc\n`) now reports 3 lines instead of 4, and added a regression test (`test_get_gitlab_file_line_count_ignores_trailing_newline`) for the trailing-newline case.

MeghanBao:
Thanks @Devesh36 — both points addressed:

- **Before/after diff**: rewrote the Demo section with the *actual* `merge_tool_evidence` output (before = dropped on `main`, after = recorded on this branch), reproduced by feeding each payload through the mapper and reading `catalog_entries[0]["summary"]`.
- **Off-by-one line count**: fixed in 76f8f8b (`len(content.splitlines())`) + regression test.

`pytest -k gitlab` 160 passed, ruff + mypy clean.

Davidson3556:
Closing because #5552 is currently assigned to @arnavpopley and #5776 is the active PR for that issue. Please coordinate on the issue and wait for assignment before opening another implementation.

PR #5713, opened 2026-08-25
refactor(llm_cli): give the provider adapters a lazy public API

YauhenBichel:
llm_cli's provider adapters were reached by importing files inside the package. Re-export them through the root: base/errors/runner eagerly, and the 10 provider adapters lazily via __getattr__ so importing integrations.llm_cli does not pull
every provider's dependencies. A TYPE_CHECKING block keeps the names typed. Migrate the two boundary-consumer imports (llm_setup catalog, claude_code executor) onto the root and drop those allowlist entries; the errors/auth_check/
subprocess_env/binary_resolver/codex_oauth utilities stay for a follow-up.


YauhenBichel:
@greptile review

PR #5722, opened 2026-08-25
refactor(gitlab): break the setup/verifier import cycle via a client leaf

YauhenBichel:
setup.py and verifier.py imported config/validation helpers from the package root, so re-exporting GITLAB_SETUP at the root would form a root -> setup -> verifier -> root cycle. Move the config/HTTP core into a new integrations/gitlab/client.py leaf; the root __init__ re-exports it (and GITLAB_SETUP lazily). setup.py and verifier.py now import from the leaf, so the
graph is acyclic and GITLAB_SETUP is reachable from the root. Migrate the one boundary caller and drop the allowlist entry. Repoint tests that patched the moved functions to the client module.

### DIFF
--- a/integrations/gitlab/__init__.py
+++ b/integrations/gitlab/__init__.py
@@ -1,279 +1,47 @@
-"""Shared gitlab integration helpers."""
+"""GitLab integration package — the public API other tiers import.
+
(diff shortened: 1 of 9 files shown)


YauhenBichel:
@greptile review

PR #5723, opened 2026-08-25
feat(rds): add evidence mappers for describe_rds_events and describe_rds_instance

### LINKED ISSUE #5572: Add evidence mappers for rds
## Summary
can point to. The `rds` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `describe_rds_events`
- `describe_rds_instance`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="describe_rds_events"' integrations/`.

```python

def _map_describe_rds_events(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="describe_rds_events",
        )

@tool(name="describe_rds_events", ..., evidence_mapper=_map_describe_rds_events)
```

## How to check it works

   under `tests/integrations/rds/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['describe_rds_events', 'describe_rds_instance']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'describe_rds_events', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `describe_rds_events`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k rds
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify rds
# edit /tmp/alert.json so the alert names a resource rds actually knows about
grep -E 'describe_rds_events|describe_rds_instance' /tmp/rca.json
```

matching lines into the PR.

**No credentials for rds?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5572

Wires `describe_rds_events` and `describe_rds_instance` to

- `_map_describe_rds_events` (`integrations/rds/tools/rds_events_tool/__init__.py`):
  cites the event count when the lookback window returned any events.
- `_map_describe_rds_instance` (`integrations/rds/tools/rds_describe_instance_tool/__init__.py`):
  cites the instance's status and engine/version when the instance was
  actually described.

Both tools are read-only diagnostic calls (`describe_db_instances`,
`describe_events`) — neither is an action/notification tool, so both are in
scope per the issue.

Both mappers record nothing on an `available: False` result or an
uninformative payload — verified directly (see tests) rather than assumed:
an empty `events` list, an error-shaped `tool_unavailable(...)` payload for
both tools, and an `available: True` instance payload with neither `status`
nor `engine` populated all produce zero `catalog_entries`.

carry a mapper.

```
print({n: bool(g(n).evidence_mapper) for n in ['describe_rds_events', 'describe_rds_instance']})"
{'describe_rds_events': True, 'describe_rds_instance': True}
```

```
events: [{'source': 'describe_rds_events', 'label': 'RDS Events', 'summary': '1 event(s) in the lookback window', 'url': None, 'snippet': None}]
instance: [{'source': 'describe_rds_instance', 'label': 'RDS Instance', 'summary': 'available, postgres 15.4', 'url': None, 'snippet': None}]
```

**Check 4:**
```
11 passed

$ uv run pytest tests/ -k rds -q
363 passed, 28 skipped, 14778 deselected

All checks passed! / 3706 files already formatted / Success: no issues found in 1926 source files
```

**Check 5 — no RDS credentials available in this environment**, so leaving
that box unticked per the issue's own guidance ("that is a normal outcome,
not a reason to go quiet"). Checks 1–4 above stand as evidence the mapper is
attached and produces a real report entry from a realistic sample payload; a
reviewer with an RDS account can finish the live end-to-end check.

**1–2. Before/after**, run with a realistic sample of what each tool returns
(`merge_tool_evidence` on `main` vs. this branch):

`describe_rds_instance`, diff:
```diff
+  ],
+  "catalog_entries": [
+    {
+      "source": "describe_rds_instance",
+      "label": "RDS Instance",
+      "summary": "available, postgres 15.4",
+      "url": null,
+      "snippet": null
+    }
```

`describe_rds_events`, diff:
```diff
+  ],
+  "catalog_entries": [
+    {
+      "source": "describe_rds_events",
+      "label": "RDS Events",
+      "summary": "2 event(s) in the lookback window",
+      "url": null,
+      "snippet": null
+    }
```

`catalog_entries` list item — no other key in `evidence` changes. Nothing
else about the pre-existing evidence shape is touched.

**4. Answers:**

- **What the reader gains.** For `describe_rds_instance`: the report can now
  cite "the RDS instance was `available`, running postgres 15.4" as a
  numbered fact, instead of that information existing only inside a tool-call
  transcript the reader may not open. For `describe_rds_events`: the report
  can cite "N RDS events occurred in the lookback window" (e.g. a Multi-AZ
  failover), which is often the single most relevant fact in an RDS incident
  and previously had zero visibility in the final report.
- **How much context it costs.** 129 characters (JSON-serialized) for the
  instance entry, 134 for the events entry — both single-line summaries, no
  inlined lists or raw payloads.
- **What happens on an empty or error result.** Verified directly (see
  `tests/integrations/rds/test_evidence_mappers.py`): an empty `events: []`,
  an `available: False` error payload for either tool, and an
  `available: True` instance payload with neither `status` nor `engine`
  populated all produce zero `catalog_entries` — no "0 items" noise line.
- **Whether anything is now recorded twice.** No. These are the only two RDS
  tools in the codebase (confirmed via the baseline gaps file, which listed
  both and nothing else for `rds`), and no other mapper reads
  `describe_rds_events`/`describe_rds_instance` output.
- **Whether an existing report changed shape.** `tests/ -k rds` (363 tests)
  and the full evidence-mapper coverage suite (11 tests) pass unchanged —
  nothing pre-existing in the RDS or evidence-gathering path regressed.

---

Both RDS tools already return a consistent `available: bool` gate (set by
`tool_unavailable(...)` on failure, `True` on success), so both mappers check
that first and return early otherwise — no separate error-detection logic
needed per tool. For the instance mapper, I built the summary from whichever
of `status`/`engine`+`engine_version` are actually present rather than
assuming both are always populated, and verified the "both missing" edge case
records nothing (rather than an empty-looking summary) with a dedicated test,
since the issue explicitly calls out "0 items"-style noise as a failure mode
this comparison exists to catch. I followed the Grafana precedent's
`_map_*` naming and placement (defined immediately above the `@tool(...)`
decorator it's attached to) but didn't copy its habit of also stashing a raw
key onto `evidence` (e.g. `evidence["grafana_metrics"] = metrics`), since the
issue's own template doesn't call for that and RDS has no existing convention
of a second consumer reading that raw key back out.

---

---

### DIFF
--- a/integrations/rds/tools/rds_describe_instance_tool/__init__.py
+++ b/integrations/rds/tools/rds_describe_instance_tool/__init__.py
@@ -7,6 +7,7 @@ from typing import Any, cast
 
 from pydantic import BaseModel, Field
 
+from core.domain.types.evidence import record_evidence_entry
 from core.tool import EvidenceType, SideEffectLevel
 from core.tool_framework import tool
 from core.tool_framework.utils import tool_unavailable
@@ -18,6 +19,33 @@ from integrations.rds import (
 )
 
 logger = logging.getLogger(__name__)
+
+
+def _map_describe_rds_instance(
+    evidence: dict[str, Any], output: dict[str, Any], _tool_input: dict[str, Any]
+) -> None:
+    """Cite the described RDS instance's status/engine as report evidence."""
+    if not output.get("available"):
+        return
+    status = output.get("status")
+    engine = output.get("engine")
+    engine_version = output.get("engine_version")
+    parts = [
+        part
+        for part in (
+            status,
+            f"{engine} {engine_version}".strip() if engine else None,
+        )
+        if part
+    ]
+    if not parts:
+        return
+    record_evidence_entry(
+        evidence,
+        source="describe_rds_instance",
+        label="RDS Instance",
+        summary=", ".join(parts),
+    )
 
 
 class DescribeRDSInstanceInput(BaseModel):
@@ -79,6 +107,7 @@ class DescribeRDSInstanceOutput(BaseModel):
     injected_params=("aws_backend",),
     is_available=rds_is_available,
     extract_params=rds_extract_params,
+    evidence_mapper=_map_describe_rds_instance,
 )
 def describe_rds_instance(
     db_instance_identifier: str,
--- a/integrations/rds/tools/rds_events_tool/__init__.py
+++ b/integrations/rds/tools/rds_events_tool/__init__.py
@@ -23,16 +23,24 @@ DEFAULT_DURATION_MINUTES = 60
 def _map_describe_rds_events(
     evidence: dict[str, Any], output: dict[str, Any], _tool_input: dict[str, Any]
 ) -> None:
-    """Cite recent RDS events (failovers, maintenance, backups) as report evidence."""
+    """Cite recent RDS events (failovers, maintenance, backups) as report evidence.
+
+    ``execute_aws_sdk_call`` makes one unpaginated ``describe_events`` call, so
+    ``events`` can be a partial page on a busy instance with a long lookback
+    window. The summary says how many were *reported*, not how many occurred,
+    so it never implies a completeness the underlying call doesn't guarantee.
+    """
     if not output.get("available"):
         return
     events = output.get("events") or []
+    duration_minutes = output.get("duration_minutes")
     if events:
+        window = f" in the last {duration_minutes} min" if duration_minutes else ""
         record_evidence_entry(
             evidence,
             source="describe_rds_events",
             label="RDS Events",
-            summary=f"{len(events)} event(s) in the lookback window",
+            summary=f"{len(events)} event(s) reported{window}",
         )
 
 
--- a/integrations/rds/tools/rds_events_tool/__init__.py
+++ b/integrations/rds/tools/rds_events_tool/__init__.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 from typing import Any, cast
 
+from core.domain.types.evidence import record_evidence_entry
 from core.tool_framework import tool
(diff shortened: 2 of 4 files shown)


arnavpopley:
Investigated this before changing anything: \`execute_aws_sdk_call\` (\`integrations/aws/aws_sdk_client.py\`) makes exactly one \`describe_events\` call with no \`Marker\`/pagination handling at all — confirmed via grep, zero references to pagination anywhere in that file or \`rds_events_tool\`. So the finding is real, but it's a **pre-existing limitation of the tool itself**, not something this PR introduces — \`describe_rds_events\` could already understate a busy instance's event count on \`main\`, my mapper just faithfully reports whatever the tool returns.

Actually fixing the pagination would mean changing \`describe_rds_events\`'s AWS-calling behavior — a tool change, not an evidence-mapper change, and the tool's current return shape doesn't even carry a truncation signal to detect from inside the mapper. That's out of scope here the same way the sibling airflow evidence-mapper issue scopes out tool changes ("get verify working first; tools are a separate issue").

What I *did* fix, since it's cheap and in-scope: reworded the summary from "N event(s) in the lookback window" (implies exhaustive) to "N event(s) reported in the last N min" (states what was returned, doesn't claim completeness). Added a test pinning the new wording.

@greptile review

PR #5725, opened 2026-08-25
feat(argocd): add evidence mappers for argocd_application_diff and argocd_application_status

### LINKED ISSUE #5563: Add evidence mappers for argocd
## Summary
can point to. The `argocd` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `argocd_application_diff`
- `argocd_application_status`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="argocd_application_diff"' integrations/`.

```python

def _map_argocd_application_diff(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="argocd_application_diff",
        )

@tool(name="argocd_application_diff", ..., evidence_mapper=_map_argocd_application_diff)
```

## How to check it works

   under `tests/integrations/argocd/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['argocd_application_diff', 'argocd_application_status']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'argocd_application_diff', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `argocd_application_diff`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k argocd
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify argocd
# edit /tmp/alert.json so the alert names a resource argocd actually knows about
grep -E 'argocd_application_diff|argocd_application_status' /tmp/rca.json
```

matching lines into the PR.

**No credentials for argocd?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5563

Wires `argocd_application_diff` and `argocd_application_status` to

- `_map_argocd_application_diff` (`integrations/argocd/tools/__init__.py`):
  cites the object-diff count, but only when `drift_detected` is true. A
  clean diff (no drift) is a good result, not a finding — recording it would
  be exactly the "0 items" noise line the issue warns against.
- `_map_argocd_application_status`: this tool has two response shapes
  depending on whether `application_name` was given — a single `application`
  dict, or an `applications` list. The mapper handles both: cites the
  application's `sync_status`/`health_status` when a specific app was
  described, or the count when a bare list came back instead.

Both are read-only diagnostic tools (`get_application_diff`,
`get_application_summary`/`list_applications`) — neither is an

payload (`application: {}`, `applications: []`, `drift_detected: False`) —
verified directly in tests rather than assumed.

carry a mapper.

```
print({n: bool(g(n).evidence_mapper) for n in ['argocd_application_diff', 'argocd_application_status']})"
{'argocd_application_diff': True, 'argocd_application_status': True}
```

**Check 3 — real output becomes report evidence** (all four shapes tested:
diff-with-drift, single-app status, application-list status, no-drift/empty):
```
diff (drift):        [{'source': 'argocd_application_diff', 'label': 'Argo CD Application Diff', 'summary': '3 object diff(s) — live state has drifted from GitOps', ...}]
status (single app):  [{'source': 'argocd_application_status', 'label': 'Argo CD Application Status', 'summary': 'checkout-service: sync=OutOfSync, health=Degraded', ...}]
status (app list):    [{'source': 'argocd_application_status', 'label': 'Argo CD Applications', 'summary': '2 application(s) listed', ...}]
diff (no drift):      None
```

**Check 4:**
```
11 passed

$ uv run pytest tests/ -k argocd -q
48 passed

All checks passed! / 3706 files already formatted / Success: no issues found in 1926 source files
```

**Check 5 — no Argo CD instance available in this environment**, so leaving
that box unticked per the issue's own guidance. Checks 1–4 stand as evidence
the mappers are attached and produce real report entries from realistic
sample payloads matching the client's actual return shapes
(`integrations/argocd/client.py`'s `get_application_diff` /
`get_application_summary` / `list_applications`); a reviewer with an
instance can finish the live check.

**1–2. Before/after**, run with realistic samples matching the client's
actual return shapes:

`argocd_application_diff`, diff:
```diff
+  ],
+  "catalog_entries": [
+    {
+      "source": "argocd_application_diff",
+      "label": "Argo CD Application Diff",
+      "summary": "2 object diff(s) — live state has drifted from GitOps",
+      "url": null,
+      "snippet": null
+    }
```

`argocd_application_status`, diff:
```diff
+  ],
+  "catalog_entries": [
+    {
+      "source": "argocd_application_status",
+      "label": "Argo CD Application Status",
+      "summary": "checkout-service: sync=OutOfSync, health=Degraded",
+      "url": null,
+      "snippet": null
+    }
```

**4. Answers:**

- **What the reader gains.** For the diff tool: the report can now cite "N
  Kubernetes objects have drifted from the desired GitOps state" as a
  numbered fact — often the direct cause in a GitOps-managed incident. For
  the status tool: the report can cite the application's actual sync/health
  status (e.g. "OutOfSync, Degraded") instead of that living only in a
  tool-call transcript.
- **How much context it costs.** 177 characters (JSON-serialized) for the
  diff entry, 172 for the status entry — single-line summaries, no inlined
  diff bodies or application lists.
- **What happens on an empty or error result.** Verified directly (see
  `tests/integrations/argocd/test_evidence_mappers.py`): `drift_detected:
  False`, an `available: False` error payload for either tool, and an
  `available: True` status payload with both `application: {}` and
  `applications: []` empty all produce zero `catalog_entries`.
  Argo CD tools in the codebase (confirmed via the baseline gaps file, which
  listed both and nothing else for `argocd`), and no other mapper reads
  either tool's output.
- **Whether an existing report changed shape.** `tests/ -k argocd` (48
  unchanged.

---

Both Argo CD tools are `BaseTool` subclasses, not `@tool(...)`-decorated
functions like the issue's own Grafana example — so I looked for a
class-based precedent instead of copying the decorator-style template
verbatim, and found it in `integrations/jira/tools/__init__.py`
(`JiraIssueDetailTool.evidence_mapper = _map_jira_issue_detail`): a
module-level mapper function, referenced as a class attribute. I read the
actual client methods (`get_application_diff`, `get_application_summary`,
`list_applications` in `integrations/argocd/client.py`) rather than guessing
field names from the tool's `outputs` docstring dict, since that's only a
one-line description per field, not the real shape. The status tool's two
possible response shapes (`application` vs. `applications`, depending on
whether `application_name` was passed) needed explicit branching in the
mapper rather than a single code path — I verified both branches directly
before writing tests for them, including the "both present but empty" edge
case that a naive `if output.get("application")` truthiness check alone
wouldn't have caught as clearly as `.get("name")` being unset.

---

---


arnavpopley:
Fixed — this was a real bug, thanks for catching it. `drift_detected` is now the sole gate for whether an entry is recorded; `diff_count` only changes the wording (falls back to "live state has drifted from GitOps (no itemized diffs returned)" when it's 0 but drift was still detected via the client's `modified` flag). Added a dedicated test for the modified-but-unitemized case. Also fixed the bare `dict` test annotations to `dict[str, Any]`.

@greptile review

arnavpopley:
@greptile review

arnavpopley:
From OpenSRE Bot in Slack:

### :white_check_mark: PR #5725 — argocd evidence mappers

Wires argocd_application_diff and argocd_application_status into record_evidence_entry. This is the right scope — GitOps drift is exactly the kind of signal you want citeable in a report (e.g. "application X was out of sync at incident time"). Two tools is a small surface area, low risk.

**Watch:** argocd_application_diff output can be large for apps with many resources. Worth confirming there's a size cap before storing, or the evidence entry could bloat reports.

**Verdict:** :white_check_mark: Merge (after #5779)

integrations/argocd/tools/__init__.py
```diff
@@ -6,15 +6,44 @@
 
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.tool import BaseTool
 from core.tool_framework.utils import tool_unavailable
 from integrations.argocd.client import make_argocd_client
 
 
+def _map_argocd_application_diff(
```
YauhenBichel: Arnav, could you please refactor this one as well

YauhenBichel (dismissed):

integrations/argocd/tools/__init__.py
arnavpopley: Done — extracted both mappers (map_argocd_application_diff, map_argocd_application_status) out of __init__.py into a new sibling _evidence.py module, so the facade file now only exposes the public tool classes.

arnavpopley:
@greptile review

Davidson3556:
Arnav, one blocker.

Both branches of the status mapper record under the same `argocd_application_status` source, and `_add_mapped_entries` keeps only the first entry per source. When the alert omits the app name the agent lists first, which is the tool's own use case, so the count wins and the specific status line is dropped:

```
recorded:  ['2 application(s) listed', 'checkout: sync=OutOfSync, health=Degraded']
in report: [('Argo CD Applications', '2 application(s) listed')]
```

Could you skip the bare list entry, or carry the OutOfSync/Degraded counts in it so it is worth keeping?

muddlebee:
> Arnav, one blocker.
> 
> Both branches of the status mapper record under the same `argocd_application_status` source, and `_add_mapped_entries` keeps only the first entry per source. When the alert omits the app name the agent lists first, which is the tool's own use case, so the count wins and the specific status line is dropped:
> 
> ```
> recorded:  ['2 application(s) listed', 'checkout: sync=OutOfSync, health=Degraded']
> in report: [('Argo CD Applications', '2 application(s) listed')]
> ```
> 
> Could you skip the bare list entry, or carry the OutOfSync/Degraded counts in it so it is worth keeping?

@Davidson3556 not really a blocker imo, merging this.

Davidson3556:
> > Arnav, one blocker.
> > Both branches of the status mapper record under the same `argocd_application_status` source, and `_add_mapped_entries` keeps only the first entry per source. When the alert omits the app name the agent lists first, which is the tool's own use case, so the count wins and the specific status line is dropped:
> > ```
> > recorded:  ['2 application(s) listed', 'checkout: sync=OutOfSync, health=Degraded']
> > in report: [('Argo CD Applications', '2 application(s) listed')]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Could you skip the bare list entry, or carry the OutOfSync/Degraded counts in it so it is worth keeping?
> 
> @Davidson3556 not really a blocker imo, merging this.

okay

PR #5726, opened 2026-08-25
feat(clickhouse): add evidence mappers for query activity and system health

### LINKED ISSUE #5564: Add evidence mappers for clickhouse
## Summary
can point to. The `clickhouse` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `get_clickhouse_query_activity`
- `get_clickhouse_system_health`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="get_clickhouse_query_activity"' integrations/`.

```python

def _map_get_clickhouse_query_activity(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="get_clickhouse_query_activity",
        )

@tool(name="get_clickhouse_query_activity", ..., evidence_mapper=_map_get_clickhouse_query_activity)
```

## How to check it works

   under `tests/integrations/clickhouse/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['get_clickhouse_query_activity', 'get_clickhouse_system_health']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'get_clickhouse_query_activity', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `get_clickhouse_query_activity`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k clickhouse
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify clickhouse
# edit /tmp/alert.json so the alert names a resource clickhouse actually knows about
grep -E 'get_clickhouse_query_activity|get_clickhouse_system_health' /tmp/rca.json
```

matching lines into the PR.

**No credentials for clickhouse?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5564

Wires `get_clickhouse_query_activity` and `get_clickhouse_system_health` to

- `_map_get_clickhouse_query_activity`
  (`integrations/clickhouse/tools/clickhouse_query_activity_tool/__init__.py`):
  cites the query count, and — since the tool's own description specifically
  calls out "including failed queries" as a feature — also how many of the
  returned queries had an `ExceptionWhileProcessing`/exception `type`.
- `_map_get_clickhouse_system_health`
  (`integrations/clickhouse/tools/clickhouse_system_health_tool/__init__.py`):
  cites whichever of version / uptime / surveyed-table-count are actually
  present. The tool function combines `get_system_health()` with an optional
  `get_table_stats()` sub-call, and that sub-call already falls back to an
  empty `table_stats: []` on its own failure (see the existing
  `test_run_table_stats_error_falls_back_to_empty_list`) — the mapper skips
  the table-count clause entirely in that case rather than saying "0 tables".

Both mappers record nothing on an `available: False` result or an
uninformative payload — verified directly in tests: empty `queries: []`, an
`available: False` error payload for either tool, and an
`available: True` health payload with neither `version` nor `uptime_seconds`
populated all produce zero `catalog_entries`.

**Test placement note:** the issue's template says to put tests under
`tests/integrations/clickhouse/`, but this vendor's tests are already
organized per-tool under `tests/tools/test_clickhouse_query_activity_tool.py`
/ `tests/tools/test_clickhouse_system_health_tool.py` — I added the mapper
tests there instead of creating a new, inconsistent directory, matching this
repo's own file-placement guidance ("check for an existing sibling pattern...
match that layout before inventing a new one").

carry a mapper.

```
print({n: bool(g(n).evidence_mapper) for n in ['get_clickhouse_query_activity', 'get_clickhouse_system_health']})"
{'get_clickhouse_query_activity': True, 'get_clickhouse_system_health': True}
```

```
query activity: [{'source': 'get_clickhouse_query_activity', 'label': 'ClickHouse Query Activity', 'summary': '3 queries, 1 failed', ...}]
system health:  [{'source': 'get_clickhouse_system_health', 'label': 'ClickHouse System Health', 'summary': 'version 23.8.1.2992, uptime 86400s, 2 table(s) surveyed', ...}]
```

**Check 4:**
```
11 passed

$ uv run pytest tests/ -k clickhouse -q
58 passed

```

**Check 5 — no ClickHouse instance available in this environment**, so
leaving that box unticked per the issue's own guidance. Checks 1–4 stand as
realistic sample payloads matching the actual functions' return shapes
(`integrations/clickhouse/__init__.py`'s `get_query_activity` /
`get_system_health` / `get_table_stats`); a reviewer with an instance can
finish the live check.

**1–2. Before/after**, run with realistic samples matching the functions'
actual return shapes:

`get_clickhouse_query_activity`, diff:
```diff
+  ],
+  "catalog_entries": [
+    {
+      "source": "get_clickhouse_query_activity",
+      "label": "ClickHouse Query Activity",
+      "summary": "3 queries, 1 failed",
+      "url": null,
+      "snippet": null
+    }
```

`get_clickhouse_system_health`, diff:
```diff
+  ],
+  "catalog_entries": [
+    {
+      "source": "get_clickhouse_system_health",
+      "label": "ClickHouse System Health",
+      "summary": "version 23.8.1.2992, uptime 86400s, 2 table(s) surveyed",
+      "url": null,
+      "snippet": null
+    }
```

**4. Answers:**

- **What the reader gains.** For query activity: the report can now cite "N
  recent queries, M of them failed" — a direct signal for query-driven
  incidents, previously invisible outside the raw tool-call transcript. For
  system health: the report can cite the server's actual version/uptime and
  how many tables were surveyed, useful for ruling out "server just
  restarted" or confirming the checked instance is the one alerting.
- **How much context it costs.** 145 characters (JSON-serialized) for the
  query-activity entry, 179 for the system-health entry — single-line
  summaries, no inlined query text or table lists.
  new tests in both `tests/tools/test_clickhouse_*_tool.py` files): empty
  `queries: []`, an `available: False` error payload for either tool, and a
  health payload with neither `version` nor `uptime_seconds` present all
  produce zero `catalog_entries`. Separately verified that an empty
  `table_stats: []` (the sub-call's own failure fallback) does not add a
  "0 tables" clause to an otherwise-valid health entry.
  ClickHouse tools in the codebase (confirmed via the baseline gaps file,
  which listed both and nothing else for `clickhouse`), and no other mapper
  reads either tool's output.
- **Whether an existing report changed shape.** `tests/ -k clickhouse` (58
  unchanged.

---

I read the actual underlying functions in `integrations/clickhouse/__init__.py`
(`get_query_activity`, `get_system_health`, `get_table_stats`) rather than
guessing field names from the tool's one-line `outputs` docstrings, since
`get_clickhouse_system_health` (the tool function) actually composes two of
those three helpers and adds a `table_stats` key the plain `get_system_health`
helper doesn't have on its own — a mapper written only against
`get_system_health`'s shape would have missed that entirely. For query
activity, the tool's description explicitly promotes "including failed
queries" as the reason to use it, so I checked the `query_log.type` values
that indicate an exception and surfaced a failure count rather than a bare
"N queries" line, which is the more actionable summary an on-call engineer
would want to see first. I chose to extend the existing per-tool test files
instead of creating a new `tests/integrations/clickhouse/` directory the
issue's generic template suggests, since this vendor already has an
established, different test layout (`tests/tools/test_clickhouse_*_tool.py`)
and matching that beats introducing a second, inconsistent one.

---

---


arnavpopley:
Fixed — the mapper test evidence fixtures now use `dict[str, Any]` instead of bare `dict`. (The unrelated pre-existing `__opensre_registered_tool__` attr-defined mypy errors in both files predate this change — confirmed via `git stash`, same 10 errors with or without my diff — leaving those alone.)

@greptile review

PR #5727, opened 2026-08-25
feat(kafka): add evidence mappers for consumer group lag and topic health

### LINKED ISSUE #5567: Add evidence mappers for kafka
## Summary
can point to. The `kafka` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `get_kafka_consumer_group_lag`
- `get_kafka_topic_health`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="get_kafka_consumer_group_lag"' integrations/`.

```python

def _map_get_kafka_consumer_group_lag(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="get_kafka_consumer_group_lag",
        )

@tool(name="get_kafka_consumer_group_lag", ..., evidence_mapper=_map_get_kafka_consumer_group_lag)
```

## How to check it works

   under `tests/integrations/kafka/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['get_kafka_consumer_group_lag', 'get_kafka_topic_health']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'get_kafka_consumer_group_lag', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `get_kafka_consumer_group_lag`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k kafka
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify kafka
# edit /tmp/alert.json so the alert names a resource kafka actually knows about
grep -E 'get_kafka_consumer_group_lag|get_kafka_topic_health' /tmp/rca.json
```

matching lines into the PR.

**No credentials for kafka?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5567

Wires `get_kafka_consumer_group_lag` and `get_kafka_topic_health` to

- `_map_get_kafka_consumer_group_lag`
  (`integrations/kafka/tools/kafka_consumer_group_tool/__init__.py`): cites
  the partition count and total lag whenever the consumer group has any
  partitions — **including a zero-lag result**. See the note below on why
  that's a deliberate choice, not an oversight.
- `_map_get_kafka_topic_health`
  (`integrations/kafka/tools/kafka_topic_health_tool/__init__.py`): cites the
  surveyed topic count, plus how many partitions are under-replicated when
  any are — the tool's own description promotes under-replicated-partition
  detection as its purpose, so surfacing that count is more actionable than
  a bare topic count.

**On zero lag not being "noise":** the project's convention (and this
issue's own template) warns against a "0 items" line — but that's about an
*empty list*, not a *zero value inside a real result*. A consumer group with
1+ partitions and 0 total lag is a genuine, informative finding: it rules
out consumer lag as the incident cause. An empty `partitions: []` (e.g. the
group truly has none) is the actual "nothing to say" case, and the mapper
correctly records nothing for that — verified directly in tests.

`topics`/`partitions` list.

carry a mapper.

```
print({n: bool(g(n).evidence_mapper) for n in ['get_kafka_consumer_group_lag', 'get_kafka_topic_health']})"
{'get_kafka_consumer_group_lag': True, 'get_kafka_topic_health': True}
```

```
lag:          [{'source': 'get_kafka_consumer_group_lag', 'label': 'Kafka Consumer Group Lag', 'summary': 'payments-consumer: 2 partition(s), total lag 1500', ...}]
topic health: [{'source': 'get_kafka_topic_health', 'label': 'Kafka Topic Health', 'summary': '2 topic(s) surveyed, 1 under-replicated partition(s)', ...}]
```

**Check 4:**
```
11 passed

$ uv run pytest tests/ -k kafka -q
73 passed

```

**Check 5 — no Kafka cluster available in this environment**, so leaving
that box unticked per the issue's own guidance. Checks 1–4 stand as evidence
the mappers are attached and produce real report entries from realistic
sample payloads matching the actual functions' return shapes
(`integrations/kafka/__init__.py`'s `get_consumer_group_lag` /
`get_topic_health`); a reviewer with a cluster can finish the live check.

**1–2. Before/after**, run with realistic samples matching the functions'
actual return shapes:

`get_kafka_consumer_group_lag`, diff:
```diff
+  ],
+  "catalog_entries": [
+    {
+      "source": "get_kafka_consumer_group_lag",
+      "label": "Kafka Consumer Group Lag",
+      "summary": "payments-consumer: 2 partition(s), total lag 1500",
+      "url": null,
+      "snippet": null
+    }
```

`get_kafka_topic_health`, diff:
```diff
+  ],
+  "catalog_entries": [
+    {
+      "source": "get_kafka_topic_health",
+      "label": "Kafka Topic Health",
+      "summary": "2 topic(s) surveyed, 1 under-replicated partition(s)",
+      "url": null,
+      "snippet": null
+    }
```

**4. Answers:**

- **What the reader gains.** For consumer lag: the report can cite the
  actual lag figure per group ("payments-consumer: 2 partitions, total lag
  1500") — a direct signal for a processing-delay incident, or a
  ruling-out signal when lag is 0. For topic health: the report can cite
  under-replicated partitions — a real cluster-health problem often
  correlated with a broker outage — instead of that fact staying buried in
  a tool-call transcript.
- **How much context it costs.** 173 characters (JSON-serialized) for the
  lag entry, 164 for the topic-health entry — single-line summaries, no
  inlined partition lists.
  new tests in both `tests/tools/test_kafka_*_tool.py` files): an empty
  `partitions: []` / `topics: []`, and an `available: False` error payload
  for either tool, all produce zero `catalog_entries`. A *zero-lag but
  non-empty* result deliberately still records — see the note above.
  Kafka tools in the codebase (confirmed via the baseline gaps file, which
  listed both and nothing else for `kafka`), and no other mapper reads
  either tool's output.
- **Whether an existing report changed shape.** `tests/ -k kafka` (73
  unchanged.

---

The one design decision worth calling out explicitly: whether a zero-lag
consumer group result should record an entry. The issue's template treats
"0 items" as noise to avoid, but I distinguished that from "a real result
whose value happens to be zero" — an empty `partitions` list means the tool
found nothing to report (skip it), while a populated `partitions` list with
`total_lag: 0` means the tool successfully confirmed the consumer isn't
lagging, which is itself useful ruling-out evidence for an on-call engineer.
I gated on `if not partitions: return` rather than `if not total_lag: return`
for exactly this reason, and added a dedicated test
(`test_records_entry_for_zero_lag_healthy_group`) to pin the choice rather
than leave it implicit. For topic health, I read the actual partition dict
shape (`under_replicated: bool` per partition, set by
`get_topic_health`'s own `len(pmeta.isrs) < len(pmeta.replicas)` check) so
the summary surfaces the specific, actionable signal the tool's description
promotes, rather than a generic topic count that omits the health problem
the tool exists to detect.

---

---


arnavpopley:
Fixed — the mapper test evidence fixtures now use `dict[str, Any]` instead of bare `dict`. (The unrelated pre-existing `__opensre_registered_tool__` attr-defined mypy errors in both files predate this change — confirmed via `git stash`, same 2 errors with or without my diff — leaving those alone.)

@greptile review

PR #5729, opened 2026-08-25
feat(new_relic): add evidence mappers for alert incidents and metrics

### LINKED ISSUE #5568: Add evidence mappers for new_relic
## Summary
can point to. The `new_relic` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `query_new_relic_alerts`
- `query_new_relic_metrics`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="query_new_relic_alerts"' integrations/`.

```python

def _map_query_new_relic_alerts(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="query_new_relic_alerts",
        )

@tool(name="query_new_relic_alerts", ..., evidence_mapper=_map_query_new_relic_alerts)
```

## How to check it works

   under `tests/integrations/new_relic/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['query_new_relic_alerts', 'query_new_relic_metrics']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'query_new_relic_alerts', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `query_new_relic_alerts`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k new_relic
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify new_relic
# edit /tmp/alert.json so the alert names a resource new_relic actually knows about
grep -E 'query_new_relic_alerts|query_new_relic_metrics' /tmp/rca.json
```

matching lines into the PR.

**No credentials for new_relic?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5568

Wires `query_new_relic_alerts` and `query_new_relic_metrics` to

- `_map_query_new_relic_alerts`
  (`integrations/new_relic/tools/new_relic_alerts_tool/tool.py`): cites the
  incident count, how many are still open (using the same `STATUS_OPEN`
  constant `results.py`'s own open/close pairing logic sets), and a
  truncation note when the raw row fetch hit its cap. Open-incident count is
  more actionable than a bare total — a closed incident is historical
  context, an open one is still live.
- `_map_query_new_relic_metrics`
  (`integrations/new_relic/tools/new_relic_metrics_tool/tool.py`): cites the
  returned row count and a truncation note when the effective `LIMIT` was
  hit.

`incidents`/`results` list — verified directly in tests.

**One correction I made against the issue's own template:** the tool's
`outputs` docstring calls the field `incidentId`, but the real, normalized
incident dicts (produced by `parse_incident_rows` and confirmed by this
file's own existing fixture-based tests) use `incident_id` — snake_case, and
also `status` rather than any camelCase equivalent. Wrote the mapper and its
tests against the real, verified shape rather than the docstring's field
name.

carry a mapper.

```
print({n: bool(g(n).evidence_mapper) for n in ['query_new_relic_alerts', 'query_new_relic_metrics']})"
{'query_new_relic_alerts': True, 'query_new_relic_metrics': True}
```

```
alerts:  [{'source': 'query_new_relic_alerts', 'label': 'New Relic Alert Incidents', 'summary': '2 incident(s), 1 open', ...}]
metrics: [{'source': 'query_new_relic_metrics', 'label': 'New Relic Metrics', 'summary': '2 row(s)', ...}]
```

**Check 4:**
```
11 passed

$ uv run pytest tests/ -k new_relic -q
92 passed

```

**Check 5 — no New Relic account available in this environment**, so
leaving that box unticked per the issue's own guidance. Checks 1–4 stand as
realistic sample payloads matching the actual functions' verified return
shapes; a reviewer with an account can finish the live check.

**1–2. Before/after**, run with realistic samples matching the actual
verified incident/result shapes:

`query_new_relic_alerts`, diff:
```diff
+  ],
+  "catalog_entries": [
+    {
+      "source": "query_new_relic_alerts",
+      "label": "New Relic Alert Incidents",
+      "summary": "2 incident(s), 1 open",
+      "url": null,
+      "snippet": null
+    }
```

`query_new_relic_metrics`, diff:
```diff
+  ],
+  "catalog_entries": [
+    {
+      "source": "query_new_relic_metrics",
+      "label": "New Relic Metrics",
+      "summary": "2 row(s)",
+      "url": null,
+      "snippet": null
+    }
```

**4. Answers:**

- **What the reader gains.** For alerts: the report can now cite how many
  New Relic incidents fired and how many are still open — a direct signal
  for correlating with an alert's own timeline, previously buried in a
  tool-call transcript. For metrics: the report can cite that an NRQL query
  actually returned data (and how much), confirming the metric the agent
  checked was real and non-empty.
- **How much context it costs.** 140 characters (JSON-serialized) for the
  alerts entry, 120 for the metrics entry — single-line summaries, no
  inlined incident detail or raw metric rows.
  new tests in both `tests/tools/test_new_relic_*_tool.py` files): an empty
  `incidents: []` / `results: []`, and an `available: False` error payload
  for either tool, all produce zero `catalog_entries`.
  New Relic tools in the codebase (confirmed via the baseline gaps file,
  which listed both and nothing else for `new_relic`), and no other mapper
  reads either tool's output.
- **Whether an existing report changed shape.** `tests/ -k new_relic` (92
  tests, including the synthetic scenario and CLI smoke tests) and the full
  evidence-mapper coverage suite (11 tests) pass unchanged.

---

Before writing the mapper, I read `results.py`'s `parse_incident_rows` to see
what the normalized incident dict actually contains, rather than trusting
the tool class's `outputs` docstring (which says `incidentId`, the raw
NerdGraph field name pre-pairing). The existing fixture-based tests in
`test_new_relic_alerts_tool.py` confirmed the real, post-pairing shape uses
`incident_id` and `status` (with `STATUS_OPEN`/`STATUS_CLOSED` constants
already defined in `results.py`), so I imported and reused those constants
rather than hardcoding the string `"open"` a second time. I chose to surface
the open-incident count specifically, not just a bare total, because an
on-call engineer reading the report cares more about "is this still firing"
than "how many times has this fired historically" — the tool's own docstring
supports this distinction (open/close pairing exists specifically to
separate live from resolved). For metrics, the tool is intentionally generic
(arbitrary NRQL), so I kept the mapper equally generic: row count plus a
truncation flag, since inspecting arbitrary result-row contents for a
richer summary would require assumptions the tool itself doesn't make.

---

---


arnavpopley:
Fixed — the summary now always states the open count explicitly, even at 0 (e.g. `2 incident(s), 0 open`), instead of omitting the clause. Added a dedicated test for the all-closed case. Also fixed the bare `dict` test annotations to `dict[str, Any]`.

@greptile review

PR #5742, opened 2026-08-25
Feat: map Airflow tool output to citeable evidence

### LINKED ISSUE #5558: Add evidence mappers for airflow
## Summary
can point to. The `airflow` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `get_airflow_dag_runs`
- `get_airflow_task_instances`
- `get_recent_airflow_failures`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="get_airflow_dag_runs"' integrations/`.

```python

def _map_get_airflow_dag_runs(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="get_airflow_dag_runs",
        )

@tool(name="get_airflow_dag_runs", ..., evidence_mapper=_map_get_airflow_dag_runs)
```

## How to check it works

   under `tests/integrations/airflow/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['get_airflow_dag_runs', 'get_airflow_task_instances', 'get_recent_airflow_failures']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'get_airflow_dag_runs', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `get_airflow_dag_runs`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k airflow
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify airflow
# edit /tmp/alert.json so the alert names a resource airflow actually knows about
grep -E 'get_airflow_dag_runs|get_airflow_task_instances|get_recent_airflow_failures' /tmp/rca.json
```

matching lines into the PR.

**No credentials for airflow?** That is a normal outcome, not a reason to go

ckryptickunal:
Fixes #5558

Wires `evidence_mapper` functions onto the 3 Airflow tools named in the issue — `get_airflow_dag_runs`, `get_airflow_task_instances`, `get_recent_airflow_failures` — so their real output becomes a citeable entry in the investigation report. All three are genuine read tools (no side effects); none skipped.

**Where the tools actually live, and a decoy avoided:** the raw, undecorated Airflow REST client functions sit in `integrations/airflow/config.py`, but the actual `@tool(name="get_airflow_dag_runs", ...)` registrations — the only place in the repo with those literal names — live in `integrations/tracer/tools/tracer_airflow_dag_tool/__init__.py`, which imports and wraps the raw functions. A sibling file, `tracer_airflow_metrics_tool/__init__.py`, registers a *different* tool (`get_airflow_metrics`, `source="tracer_web"`) backed by a different client — same topic, not one of the 3 named tools, correctly left untouched.

Each mapper reads the tool's real return field (`dag_runs`, `task_instances`, `failures`), following the shipped Grafana precedent. Removed the 3 handled tool names from `evidence_mapper_baseline.txt` (left `get_airflow_metrics` in place — unrelated tool). Added `tests/integrations/airflow/test_evidence_mappers.py` (10 tests).

```
Registered-mapper check: {'get_airflow_dag_runs': True, 'get_airflow_task_instances': True, 'get_recent_airflow_failures': True}
merge_tool_evidence('get_airflow_dag_runs', {'dag_runs': [1, 2]}, {}) ->
  catalog_entries = [{'source': 'get_airflow_dag_runs', 'label': 'Airflow DAG Runs', 'summary': '2 dag runs', 'url': None, 'snippet': None}]

pytest tests/ -k airflow -> 32 passed (10 new + 22 pre-existing, including the unrelated tracer-metrics-tool
  tests, confirming no collateral impact)
make lint / format-check -> pass
mypy (targeted over integrations/) -> Success: no issues found in 775 source files
```

**Check 5 (evidence reaches a live report) — not run:** no Airflow credentials configured in this environment (confirmed: no `AIRFLOW_AUTH_TOKEN`/`AIRFLOW_USERNAME`/`AIRFLOW_PASSWORD`), which the issue calls a normal outcome. Additionally, this Mac's disk filled to ~100% partway through the session (unrelated to this repo, confirmed independently across three PRs in this batch) — chose not to attempt live `opensre investigate` calls while in that state.

## Behaviour comparison (required by this issue's template)

Branch was at `main`'s tip before this change (no prior commits), so before/after was captured via `git stash`/`git stash pop` around the same fixed comparison script.

**Diff** (realistic-sample case; empty-dict and error-shaped cases were identical `catalog_entries: null` in both states):
```diff
   "realistic_run": {
-    "catalog_entries": null,
+    "catalog_entries": [
+      {"label": "Airflow DAG Runs", "source": "get_airflow_dag_runs", "summary": "2 dag runs", ...},
+      {"label": "Airflow Task Instances", "source": "get_airflow_task_instances", "summary": "2 task instances", ...},
+      {"label": "Airflow Recent Failures", "source": "get_recent_airflow_failures", "summary": "1 failures", ...}
+    ]
   }
```

1. **What the reader gains:** DAG-run state/history, task-instance-level execution detail, and recent task failure evidence are now citeable — before, all three lived only in the opaque `tool_outputs` blob with no report-facing reference id.
2. **Context cost:** ~110–130 characters per entry (measured exactly: 109/127/123 chars respectively); dominated by fixed label/source text, barely grows with item count.
3. **Empty/error result:** confirmed by direct execution — `{}` and `{"error": "dag_id is required"}` produce `catalog_entries` absent (not present, not empty-list) for all three, both before and after.
4. **Double-recording:** grepped for all three `source=` strings across `integrations/` and `tools/` — only these new mappers use them; the unrelated `get_airflow_metrics` tool still has no mapper.
5. **Existing reports unaffected:** re-ran `TestGrafanaMappingCharacterization`/`TestJiraMapping` (11/11 unchanged), the pre-existing raw-client tests (8/8), and all 9 sibling tracer-tool test files including the metrics one (90/90) — nothing regressed.

---

*(Left unchecked on purpose — @ckryptickunal will check these off personally after reading the diff, rather than have an assistant check them automatically. That's also why this PR opens as a draft.)*

---

- [x] If it is a core feature, I have added thorough tests (10 new tests, plus the full before/after behaviour comparison above)
- [x] My code follows the project's style guidelines and conventions (`make lint` / `format-check` / `typecheck` all pass locally)

---
Opened as a **draft**: implemented, tested, and passing this repo's required local checks, but the personal-attestation boxes above are for @ckryptickunal to confirm honestly after reading the diff. Will mark ready for review once confirmed.

### DIFF
--- a/integrations/tracer/tools/tracer_airflow_dag_tool/__init__.py
+++ b/integrations/tracer/tools/tracer_airflow_dag_tool/__init__.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import os
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
 from core.tool_framework import tool
 from integrations.airflow.config import (
@@ -43,9 +44,23 @@ def _airflow_dag_id(sources: dict[str, Any]) -> str:
     ).strip()
 
 
+def _map_get_recent_airflow_failures(
+    evidence: dict[str, Any], output: dict[str, Any], _input: dict[str, Any]
+) -> None:
+    failures = output.get("failures", [])
+    if failures:
+        record_evidence_entry(
+            evidence,
+            source="get_recent_airflow_failures",
+            label="Airflow Recent Failures",
+            summary=f"{len(failures)} failures",
+        )
+
+
 @tool(
     name="get_recent_airflow_failures",
     source="airflow",
+    evidence_mapper=_map_get_recent_airflow_failures,
     description="Fetch recent failed or retrying Airflow task evidence for a DAG.",
     use_cases=[
         "Investigating Airflow DAG failures",
@@ -89,9 +104,23 @@ def get_recent_airflow_failures(
     }
 
 
+def _map_get_airflow_dag_runs(
+    evidence: dict[str, Any], output: dict[str, Any], _input: dict[str, Any]
+) -> None:
+    dag_runs = output.get("dag_runs", [])
+    if dag_runs:
+        record_evidence_entry(
+            evidence,
+            source="get_airflow_dag_runs",
+            label="Airflow DAG Runs",
+            summary=f"{len(dag_runs)} dag runs",
+        )
+
+
 @tool(
     name="get_airflow_dag_runs",
     source="airflow",
+    evidence_mapper=_map_get_airflow_dag_runs,
     description="Fetch recent Airflow DAG runs for a DAG.",
     use_cases=[
         "Checking recent Airflow DAG run state",
@@ -138,9 +167,23 @@ def get_airflow_dag_runs(
     }
 
 
+def _map_get_airflow_task_instances(
+    evidence: dict[str, Any], output: dict[str, Any], _input: dict[str, Any]
+) -> None:
+    task_instances = output.get("task_instances", [])
+    if task_instances:
+        record_evidence_entry(
+            evidence,
+            source="get_airflow_task_instances",
+            label="Airflow Task Instances",
+            summary=f"{len(task_instances)} task instances",
+        )
+
+
 @tool(
     name="get_airflow_task_instances",
     source="airflow",
+    evidence_mapper=_map_get_airflow_task_instances,
     description="Fetch Airflow task instances for a specific DAG run.",
     use_cases=[
         "Inspecting failed Airflow task instances",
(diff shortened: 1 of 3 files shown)


ckryptickunal:
@greptile review

PR #5745, opened 2026-08-25
refactor(llm_cli): finish the vendor's public API (utilities)

YauhenBichel:
Re-export the remaining llm_cli utilities through the package root: the CLI error classes eagerly (lightweight), and check_cli_auth, the codex oauth login, the subprocess-env builder, and the binary resolver lazily via __getattr__. Migrate the six boundary-consumer imports onto the root and drop the allowlist entries; this clears llm_cli entirely and takes the infrastructure tier to 0.


YauhenBichel:
@greptile review

PR #5748, opened 2026-08-25
refactor(github): re-export the mcp and mcp_oauth API lazily

YauhenBichel:
Extend github's lazy __getattr__ surface with the MCP config/validation names and the device-flow OAuth types so callers import them from the package root instead of reaching into integrations.github.mcp / .mcp_oauth. Migrate the five surfaces callers and drop the two allowlist entries; github.setup (verifier cycle) is left for a leaf-extraction follow-up.


YauhenBichel:
@greptile review

Issue #5750, opened 2026-08-25
[BUG] A torn final line swallows the next session turn


AniketR10:
### Summary

A session file whose last line has no trailing newline — what a crashed writer leaves behind — silently swallows the next turn written to that session. The new record is appended onto the partial line instead of a new one, fusing both into a single unparseable line, so the post-crash turn is invisible to every reader.

### Expected vs actual behavior

**Expected:** a torn final line costs at most the one record that was being written when the process died. The next write starts a new line and is readable.

**Actual:** the next write is concatenated onto the partial line. Two records become one corrupt line, and the good record is lost permanently.

### Steps to reproduce

```python
store = JsonlSessionStore()
store.open_session(session)          # session id "t"
store.append_turn(session, "chat", "before")

# Exactly what SIGKILL mid-write leaves: a record with no trailing newline.
with session_path("t").open("a", encoding="utf-8") as fh:
    fh.write('{"id": "partial", "type": "custom_message", "text": "tor')

JsonlSessionStore().append_turn(session, "chat", "after-the-crash")
```

Reading the file back:

```
0: ok                {"type": "session", ...}
1: ok                {"id": "75c87de7...", ... "text": "before"}
2: UNPARSEABLE       {"id": "partial", ... "text": "tor{"id": "...", ... "text": "after-the-crash"}
```

`turns a reader can actually see: ['before']`

### Cause

`JsonlSessionStore._append_entry` opens with `"a"` and writes `json.dumps(record) + "\n"`. It never checks whether the file already ends with a newline, so a partial trailing record absorbs the next append.

### Suggested fix

Before appending, if the file is non-empty and does not end with `\n`, write a leading `\n` first. That turns the crash artifact back into one dead line the reader already skips (`session_jsonl_decode_failed`), and keeps the following record intact.

### Additional context

Found by the soak matrix in #5497. Covered there by `test_soak_write_after_a_torn_tail_is_not_swallowed`, marked `xfail(strict=True)` pending this fix — the assertion was deliberately not weakened, so that test flips to a failure the moment this is fixed and the marker should be removed in the same change.

Distinct from #5474 (`open_session` truncation) and #5475 (contended write dropped silently); neither is involved here.

AniketR10:
@YauhenBichel  @muddlebee plz assign!

Davidson3556:
[...]

Issue #5474, opened 2026-08-23
open_session truncates a live session file under the lock


YauhenBichel:
## Problem

`core/agent_harness/session/persistence/jsonl_store.py:127`:

```python
def open_session(self, session: SessionPersistenceSource) -> None:
    with contextlib.suppress(Exception):
        path = session_path(session.session_id)
        with self._locked(path):
            ...
            with path.open("w", encoding="utf-8") as fh:   # <-- unconditional truncate
                fh.write(json.dumps(record, ensure_ascii=False) + "\n")
```

`"w"` truncates. Held inside the cross-process lock, this means a second process
**waits its turn and then erases the file** the first one just wrote.

Nothing catches it: the store has no cross-process test, and the three tests in
`tests/core/agent_harness/session/test_jsonl_store_lock.py` are all same-process.

## Why it matters

`docs/deployment.mdx:156` states "The lock keeps N tasks correct" and line 111
says to enable it "so two tasks never corrupt one". Neither is true while this
line stands — the lock makes the truncation orderly, not absent.

Every transport without inbound affinity — Slack Socket Mode, Telegram, Discord —
spreads events across tasks by design, so any conversation can hit a second task.

## Steps

Test first, in this PR — the red test is the evidence the fix was needed.

1. Write a cross-process test: process A opens a session and appends turns;
   process B calls `open_session` for the same id. Assert A's turns survive.
   Use real **processes**, not threads — N Fargate tasks are processes and
   `threading.Lock` state is not shared. Port the spawn pattern from
   `gateway/tests/test_multi_actor_concurrency.py`.
2. Confirm it fails, **with `OPENSRE_SESSION_FILE_LOCK=1`**. A failure with the
   lock enabled is the whole point; if it only fails with the lock off you have
   written a different test.
3. Change `open_session` to **create-if-absent**: write the session header only
   when the file does not exist. An existing session file is opened, never
   recreated.
4. For any full-file write, use the atomic pattern already proven in this repo —
   `gateway/core/storage/session/file_bindings.py:125`: lock, write temp,
   `fsync`, `os.replace`.
5. Confirm the test now passes.

Do not "fix" this by switching `"w"` to `"a"`. That appends a second header to an
existing file, and the one-header assertion then fails for a different reason.

## Expected result

Two processes opening the same session end up with one header and both sets of
turns. `deployment.mdx` becomes true.

## Verification

```bash
[...]

PR #5751, opened 2026-08-25
test(session): add cross-process soak matrix for the session file lock

### LINKED ISSUE #5497: Session lock soak matrix: four cases plus wait/timeout/decode counts
## Problem

The session file lock (`OPENSRE_SESSION_FILE_LOCK=1`) is new and lightly
observed. `tests/core/agent_harness/session/test_jsonl_store_lock.py` has three
same-process tests; none of them holds the lock across a crash, and none measures
what the lock costs under contention.

## Why it matters

A lock is where concurrency bugs go to hide: the happy path looks identical
whether the lock is doing its job or not. The four cases below are the ones that
distinguish a working lock from one that merely appears to work.

The crash case in particular is not covered anywhere. `fcntl.flock` is released
by the OS when a process dies, so a hard kill should not strand the lock — but a
process killed **mid-write** can leave a torn final line, and nothing asserts
that the next reader survives it.

## The matrix

| Case | Expected |
| --- | --- |
| Two writes to the same session at once | Serialized. No lost turn, no torn line. |
| Writes to different sessions | Still concurrent — a per-path lock must not serialize unrelated sessions. |
| Kill/restart while a writer holds the lock | No permanent lock. No corrupt tail, or a tail the reader skips cleanly. |
| Lock timeout / cancellation | Clean failure. No partial append. |

## Steps

1. Build a small harness that runs a configurable number of writer processes
   against a configurable number of sessions, for a fixed duration.
2. Implement the four cases above. For the kill case use `SIGKILL`, not
   `SIGTERM` — a clean shutdown releases the lock and proves nothing.
3. Assert per case, not just "did not crash":
   - every JSONL line parses, or the only failure is a truncated **final** line
   - exactly one session header per file
   - no turn written by a surviving process is missing
   - different-session writers actually overlapped (record timestamps and assert
     the overlap, or the test silently degrades into a serial run)
4. Record and print, per run: lock-wait time (max and p50), timeout count, and
   JSONL decode failure count. These are the numbers worth watching while the
   lock is young.
5. Keep it fast enough to run on demand — a 60-second soak with a stub LLM, not
   a live-model run.

## Expected result

One command that answers "is the lock behaving" with numbers, plus four named
assertions that fail loudly when it is not.

## Verification

```bash
OPENSRE_SESSION_FILE_LOCK=1 uv run python -m pytest \
  tests/core/agent_harness/session/ -q -k soak
```

Expected: green, with the lock-wait and timeout counts printed.

Prove the different-sessions case is real — temporarily change the store to use a
single global lock instead of one per path. Expected: the different-sessions case
fails on the overlap assertion. Revert.

Prove the crash case is real — reduce the kill delay so it lands mid-write, and
confirm the reader assertion catches a torn tail rather than raising an unhandled
`JSONDecodeError`.

## Checklist

- [ ] All four cases implemented, each with a named assertion
- [ ] Kill case uses `SIGKILL`
- [ ] Different-session overlap asserted, not assumed
- [ ] Lock-wait (max, p50), timeout count and decode-failure count reported
- [ ] Global-lock and mid-write-kill demonstrations run and described in the PR

## Out of scope

Fixing anything this finds — file one issue per real failure, with the
reproduction.

Two writer bugs are already known and tracked separately (`open_session`
truncates an existing file; a lock timeout drops the write silently). If a case
reproduces one of them, mark that case `xfail(strict=True)` naming the existing
issue and move on. Do not weaken the assertion, and do not wait for those fixes
to land before shipping this harness.

## Claiming this issue

this issue.

AniketR10:
Fixes #5497 

The existing lock tests all run in one process, which can't tell a working lock from one
that only looks like it works. Added a soak matrix that spawns real writer processes for the
four cases in the issue — same-session writers, different-session writers, SIGKILL mid-write,
and lock timeout — each with a named assertion, plus lock-wait max/p50, timeout and decode-
failure counts printed per run. The crash case is split in two: the SIGKILL process asserts
the lock is reclaimable (deterministic), while the file-integrity half seeds the crash
artifact directly, since a kill only lands mid-write some of the time. This found a new bug
(#5750): a torn final line swallows the next turn written to that session, so that case is
xfail(strict=True) with the assertion left intact.

image

---

A lock's happy path looks the same whether or not it works, so the goal was assertions that
fail for the right reason. I used real subprocesses, not threads, because SIGKILL has to prove
the OS releases the lock. Workers are configured purely through env vars, and never call
open_session since it truncates the file a sibling is writing (#5474).

Two assertions didn't survive contact. The overlap check passed even after I rebuilt the store
with a single global lock, because a global lock still interleaves turn by turn — so the
discriminator is now structural: one lock file per session path, which a global lock can't
fake. And the crash case failed once then passed, since SIGKILL only sometimes lands mid-write,
so I split it: the kill asserts the lock is reclaimable, and a separate deterministic test
seeds the partial line a crash leaves. That second test found #5750, which I left at full
strength as xfail(strict=True) rather than weakening it.

---

---

### DIFF
--- a/core/agent_harness/session/persistence/jsonl_store.py
+++ b/core/agent_harness/session/persistence/jsonl_store.py
@@ -27,7 +27,7 @@ logger = logging.getLogger(__name__)
 # Seconds to wait for the cross-process session lock before giving up. Generous:
 # same-session turns are already serialized in-process, so real contention here
 # is only two tasks racing one session — rare, and pathological beyond this.
-_SESSION_LOCK_TIMEOUT_SECONDS = 10
+_SESSION_LOCK_TIMEOUT_SECONDS: float = 10
 
 
 def _session_file_lock_enabled() -> bool:
--- a/tests/core/agent_harness/session/_lock_soak_worker.py
+++ b/tests/core/agent_harness/session/_lock_soak_worker.py
@@ -1,0 +1,90 @@
+"""Standalone writer process for the session-lock soak matrix.
+
+Run as a script, never imported by a test: the point is a genuinely separate
+interpreter, so ``SIGKILL`` proves what a thread cannot and no monkeypatched
+parent state leaks in. Configuration arrives as one JSON argument because the
+matrix varies enough parameters that positional argv would be unreadable.
+
+The worker never calls ``open_session``: that truncates an existing file
+(#5474), which would destroy the very interleaving these tests measure. The
+parent seeds the session file first.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import sys
+import time
+from pathlib import Path
+from types import SimpleNamespace
+from typing import Any
+
+_REPO_ROOT = Path(__file__).resolve().parents[4]
+if str(_REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(_REPO_ROOT))
+
+from core.agent_harness.session.persistence import jsonl_store  # noqa: E402
+from core.agent_harness.session.persistence.jsonl_store import JsonlSessionStore  # noqa: E402
+
+
+def _session(session_id: str) -> Any:
+    """Minimal stand-in for the persistence source the store reads."""
+    return SimpleNamespace(
+        session_id=session_id,
+        started_at=0.0,
+        agent=SimpleNamespace(messages=[]),
+        accumulated_context={},
+    )
+
+
+def _wait_until(start_at: float) -> None:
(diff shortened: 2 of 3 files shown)


AniketR10:
@greptile review

AniketR10:
@greptile review

AniketR10:
@greptile review

AniketR10:
@greptile review

PR #5769, opened 2026-08-26
feat(rabbitmq): add evidence mappers for all 5 rabbitmq investigation tools

### LINKED ISSUE #5550: Add evidence mappers for rabbitmq
## Summary
can point to. The `rabbitmq` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `get_rabbitmq_broker_overview`
- `get_rabbitmq_connection_stats`
- `get_rabbitmq_consumer_health`
- `get_rabbitmq_node_health`
- `get_rabbitmq_queue_backlog`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="get_rabbitmq_broker_overview"' integrations/`.

```python

def _map_get_rabbitmq_broker_overview(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="get_rabbitmq_broker_overview",
        )

@tool(name="get_rabbitmq_broker_overview", ..., evidence_mapper=_map_get_rabbitmq_broker_overview)
```

## How to check it works

   under `tests/integrations/rabbitmq/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['get_rabbitmq_broker_overview', 'get_rabbitmq_connection_stats', 'get_rabbitmq_consumer_health', 'get_rabbitmq_node_health', 'get_rabbitmq_queue_backlog']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'get_rabbitmq_broker_overview', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `get_rabbitmq_broker_overview`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k rabbitmq
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify rabbitmq
# edit /tmp/alert.json so the alert names a resource rabbitmq actually knows about
grep -E 'get_rabbitmq_broker_overview|get_rabbitmq_connection_stats|get_rabbitmq_consumer_health|get_rabbitmq_node_health|get_rabbitmq_queue_backlog' /tmp/rca.json
```

matching lines into the PR.

**No credentials for rabbitmq?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5550

Wires all 5 RabbitMQ investigation tools to `record_evidence_entry` so their
investigation report.

- `_map_get_rabbitmq_broker_overview`
  (`integrations/rabbitmq/tools/rabbitmq_broker_overview_tool/__init__.py`):
  cites queue/ready/unacked counts, and flags an active memory/disk/fd alarm
  when `alarms.ok is False` (not `None`, which means "unknown" — e.g.
  insufficient permissions or an older broker without the health-checks
  endpoint — and shouldn't be reported as an alarm).
- `_map_get_rabbitmq_queue_backlog`
  (`integrations/rabbitmq/tools/rabbitmq_queue_backlog_tool/__init__.py`):
  cites the queue count and the single largest backlog (the client already
  sorts `queues` by `messages_ready + messages_unacknowledged` descending, so
  `queues[0]` is always the worst offender), plus how many queues have zero
  consumers when any do.
- `_map_get_rabbitmq_consumer_health`
  (`integrations/rabbitmq/tools/rabbitmq_consumer_health_tool/__init__.py`):
  cites the consumer count and how many are inactive when any are.
- `_map_get_rabbitmq_node_health`
  (`integrations/rabbitmq/tools/rabbitmq_node_health_tool/__init__.py`): cites
  node count, flags a cluster partition (`any_partitioned`), and names any
  node with an active memory/disk alarm.
- `_map_get_rabbitmq_connection_stats`
  (`integrations/rabbitmq/tools/rabbitmq_connection_stats_tool/__init__.py`):
  cites the vhost-scoped connection count against the broker-wide total.

Every mapper records nothing when `available` is falsy, or when its list
field (`queues` / `consumers` / `nodes` / `connections`) is empty — an
"empty list" result genuinely has nothing worth citing, unlike a populated
list whose counts happen to be zero/healthy (e.g. zero alarmed nodes),
which is real ruling-out evidence and is still cited.

carry a mapper.

```
  'get_rabbitmq_broker_overview', 'get_rabbitmq_queue_backlog',
  'get_rabbitmq_consumer_health', 'get_rabbitmq_node_health',
  'get_rabbitmq_connection_stats']})"
{'get_rabbitmq_broker_overview': True, 'get_rabbitmq_queue_backlog': True,
 'get_rabbitmq_consumer_health': True, 'get_rabbitmq_node_health': True,
 'get_rabbitmq_connection_stats': True}
```

return shape in `integrations/rabbitmq/__init__.py`:**
```
get_rabbitmq_broker_overview:   '12 queue(s), 500 ready, 3 unacked, ALARM: memory alarm'
get_rabbitmq_queue_backlog:     "2 queue(s), top backlog 105 on 'orders', 1 with zero consumers"
get_rabbitmq_consumer_health:   '2 consumer(s), 1 inactive'
get_rabbitmq_node_health:       '2 node(s), cluster partitioned, alarm on rmq@node1'
get_rabbitmq_connection_stats:  '1 connection(s) in vhost (of 3 broker-wide)'
```

**Tests and quality gates:**
```
11 passed

$ uv run pytest tests/ -k rabbitmq -q
107 passed

```

**No live RabbitMQ broker available in this environment**, so leaving that
box unticked per the issue's own guidance. `tests/e2e/rabbitmq/` has a
docker-compose-backed e2e suite for this integration, but it runs against
mocked HTTP responses too, not a real broker connection during CI. The
checks above stand as evidence the mappers are attached and produce real
report entries from realistic sample payloads matching the actual client
functions' return shapes; a reviewer with a broker can finish the live check.

`get_rabbitmq_broker_overview`:
```diff
+  "catalog_entries": [
+    {
+      "source": "get_rabbitmq_broker_overview",
+      "label": "RabbitMQ Broker Overview",
+      "summary": "12 queue(s), 500 ready, 3 unacked, ALARM: memory alarm",
+      "url": null,
+      "snippet": null
+    }
+  ]
```
`evidence` changes.)

**4. Answers:**

- **What the reader gains.** A cluster snapshot with an active alarm, the
  single worst-backlogged queue plus a zero-consumer count, an inactive
  consumer count, a cluster partition flag with the specific alarmed node
  named, and a vhost-vs-broker connection count — all previously invisible
  in the report even though the agent already paid to fetch them.
- **How much context it costs.** 149–182 characters (JSON-serialized) per
  entry — single-line summaries, no inlined per-queue/per-node/per-connection
  lists.
- **What happens on an empty or error result.** Verified directly in the new
  tests: `available: False`, and an empty `queues`/`consumers`/`nodes`/
  `connections` list, all produce zero `catalog_entries` for their
  respective tool.
  RabbitMQ tools in the codebase (confirmed via the baseline gaps file,
  which listed exactly these 5 for `rabbitmq` and nothing else), and no
  other mapper reads any of their output.
- **Whether an existing report changed shape.** `tests/ -k rabbitmq` (107
  unchanged.

---

The one design decision worth calling out: the broker-overview alarm field
is a tri-state (`ok: True | False | None`), not a boolean, because the
client can't always determine alarm status (insufficient permissions, or an
older broker without the modern health-checks endpoint). I gated the alarm
clause on `alarms.get("ok") is False` specifically — not on
`not alarms.get("ok")` — so an "unknown" result doesn't get misreported as
an active alarm. Similarly for queue backlog, I relied on the client's
existing sort order (`queues` is pre-sorted by
`messages_ready + messages_unacknowledged` descending) rather than
re-scanning for the max, so `queues[0]` is always the worst offender without
extra work in the mapper. Each mapper gates on its list field being non-empty
(not on a derived count being non-zero) so a real, healthy zero-count result
— e.g. zero alarmed nodes, zero inactive consumers — still gets cited as
ruling-out evidence rather than being conflated with "nothing to report."

---

---


arnavpopley:
@greptile review

PR #5771, opened 2026-08-26
feat(mariadb): add evidence mappers for all 5 mariadb investigation tools

### LINKED ISSUE #5546: Add evidence mappers for mariadb
## Summary
can point to. The `mariadb` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `get_mariadb_global_status`
- `get_mariadb_innodb_status`
- `get_mariadb_process_list`
- `get_mariadb_replication_status`
- `get_mariadb_slow_queries`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="get_mariadb_global_status"' integrations/`.

```python

def _map_get_mariadb_global_status(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="get_mariadb_global_status",
        )

@tool(name="get_mariadb_global_status", ..., evidence_mapper=_map_get_mariadb_global_status)
```

## How to check it works

   under `tests/integrations/mariadb/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['get_mariadb_global_status', 'get_mariadb_innodb_status', 'get_mariadb_process_list', 'get_mariadb_replication_status', 'get_mariadb_slow_queries']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'get_mariadb_global_status', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `get_mariadb_global_status`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k mariadb
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify mariadb
# edit /tmp/alert.json so the alert names a resource mariadb actually knows about
grep -E 'get_mariadb_global_status|get_mariadb_innodb_status|get_mariadb_process_list|get_mariadb_replication_status|get_mariadb_slow_queries' /tmp/rca.json
```

matching lines into the PR.

**No credentials for mariadb?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5546

Wires all 5 MariaDB investigation tools to `record_evidence_entry` so their
investigation report.

- `_map_get_mariadb_global_status`
  (`integrations/mariadb/tools/mariadb_status_tool/__init__.py`): cites
  uptime, connected threads, and InnoDB deadlock count when present. Unlike
  MySQL's version, MariaDB's `get_global_status` returns a flat `metrics`
  dict of raw `SHOW GLOBAL STATUS` string values (no nested
  `connections`/`innodb` structure), so the mapper reads keys directly.
- `_map_get_mariadb_process_list`
  (`integrations/mariadb/tools/mariadb_process_list_tool/__init__.py`):
  cites the active process count and the longest-running query's duration.
- `_map_get_mariadb_innodb_status`
  (`integrations/mariadb/tools/mariadb_innodb_status_tool/__init__.py`):
  this tool's output is a raw truncated `SHOW ENGINE INNODB STATUS` text
  blob, not structured data, so the mapper cites that the status was
  captured and scans the text for a `LATEST DETECTED DEADLOCK` section to
  flag a real deadlock finding.
- `_map_get_mariadb_slow_queries`
  (`integrations/mariadb/tools/mariadb_slow_queries_tool/__init__.py`):
  cites the surveyed query count and the slowest offender's average time
  (rows are pre-sorted by `AVG_TIMER_WAIT DESC`), or the diagnostic note
  when `performance_schema` is disabled — actionable information, distinct
  from "nothing to report."
- `_map_get_mariadb_replication_status`
  (`integrations/mariadb/tools/mariadb_replication_tool/__init__.py`): cites
  channel count, flags any channel with a stopped IO/SQL thread, and
  reports max lag — or the "not configured as a replica" note when the
  server isn't a replica.

Every mapper records nothing when `available` is falsy, or when its data
field is empty with no accompanying `note`/explanatory text — a genuinely
empty result has nothing worth citing, while a `note` explaining *why* a
list is empty (no replica configured, `performance_schema` disabled) is
real diagnostic signal and is cited instead.

carry a mapper.

```
  'get_mariadb_global_status', 'get_mariadb_process_list',
  'get_mariadb_innodb_status', 'get_mariadb_slow_queries',
  'get_mariadb_replication_status']})"
{'get_mariadb_global_status': True, 'get_mariadb_process_list': True,
 'get_mariadb_innodb_status': True, 'get_mariadb_slow_queries': True,
 'get_mariadb_replication_status': True}
```

return shape in `integrations/mariadb/__init__.py`:**
```
get_mariadb_global_status:       'uptime 86400s, 10 thread(s) connected, 2 InnoDB deadlock(s)'
get_mariadb_process_list:        '2 active process(es), longest running 40s'
get_mariadb_innodb_status:       'InnoDB engine status captured — includes a recorded deadlock'
get_mariadb_slow_queries:        '2 queries surveyed, slowest avg 90.5ms'
get_mariadb_replication_status:  '1 channel(s), 1 with a stopped IO/SQL thread, max lag 90s'
```

**Tests and quality gates:**
```
11 passed

$ uv run pytest tests/ -k mariadb -q
183 passed

```

**No live MariaDB instance available in this environment**, so leaving that
shapes; a reviewer with a MariaDB instance can finish the live check.

`get_mariadb_innodb_status`:
```diff
+  "catalog_entries": [
+    {
+      "source": "get_mariadb_innodb_status",
+      "label": "MariaDB InnoDB Status",
+      "summary": "InnoDB engine status captured — includes a recorded deadlock",
+      "url": null,
+      "snippet": null
+    }
+  ]
```
`evidence` changes.)

**4. Answers:**

- **What the reader gains.** Uptime/connection load and deadlock counts, a
  long-running process count with the worst offender's duration, a flagged
  InnoDB deadlock section extracted from an otherwise-opaque status blob,
  the slowest query's average time (or why no slow-query data could be
  collected), and replication channel health and lag (or that the server
  isn't a replica) — all previously invisible in the report even though the
  agent already paid to fetch them.
- **How much context it costs.** 154–185 characters (JSON-serialized) per
  entry — single-line summaries, no inlined per-process/per-query/per-channel
  lists, and no attempt to cite the multi-KB raw InnoDB status text itself.
- **What happens on an empty or error result.** Verified directly in the new
  tests: `available: False` for all 5; empty `processes`/`queries`/
  `channels` lists with no note attached, and an empty `innodb_status`
  string, all produce zero `catalog_entries`.
  MariaDB tools in the codebase (confirmed via the baseline gaps file,
  which listed exactly these 5 for `mariadb` and nothing else), and no
  other mapper reads any of their output.
- **Whether an existing report changed shape.** `tests/ -k mariadb` (183
  unchanged.

---

The InnoDB status tool was the one requiring a different approach from the
other four: `get_innodb_status` returns a single truncated free-text blob
from `SHOW ENGINE INNODB STATUS`, not structured rows, so there's no count
or numeric field to cite. Rather than skip it or dump the (multi-KB) text
into the report, I scan for the literal `LATEST DETECTED DEADLOCK` section
header InnoDB always emits when a deadlock occurred — a cheap, reliable
signal of a real finding without re-parsing the whole engine-status format.
For the two "explanatory empty" cases (`slow_queries`'s
`performance_schema` disabled, `replication_status`'s "not a replica"), I
treated the `note` field as citeable evidence in its own right, since it
answers a real question the agent would otherwise have to dig for in the
tool-call transcript — distinct from a genuinely empty list with no note,
which still records nothing per the project's "no noise" convention. I
confirmed MariaDB's `get_replication_status` only ever returns the legacy
`Slave_IO_Running`/`Slave_SQL_Running`/`Seconds_Behind_Master` column names
(no `Replica_*` variant, unlike MySQL 8.0.22+) by reading the client's own
`_KEYS` tuple, so the mapper doesn't need the dual-name fallback the MySQL
mapper does.

---

---

### DIFF
--- a/integrations/mariadb/tools/mariadb_innodb_status_tool/__init__.py
+++ b/integrations/mariadb/tools/mariadb_innodb_status_tool/__init__.py
@@ -2,6 +2,7 @@
 
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
 from core.tool_framework import tool
 from core.tool_framework.utils import call_db_tool_with_default_db_warning
@@ -13,6 +14,26 @@ from integrations.mariadb import (
 )
 
 
+def _map_get_mariadb_innodb_status(
+    evidence: dict[str, Any], output: dict[str, Any], _tool_input: dict[str, Any]
+) -> None:
+    """Cite that InnoDB status was captured, flagging a recorded deadlock section."""
+    if not output.get("available"):
+        return
+    status_text = output.get("innodb_status") or ""
+    if not status_text:
+        return
+    summary = "InnoDB engine status captured"
+    if "LATEST DETECTED DEADLOCK" in status_text:
+        summary += " — includes a recorded deadlock"
+    record_evidence_entry(
+        evidence,
+        source="get_mariadb_innodb_status",
+        label="MariaDB InnoDB Status",
+        summary=summary,
+    )
+
+
 @tool(
     name="get_mariadb_innodb_status",
     description="Retrieve InnoDB engine internals including deadlocks, buffer pool state, and I/O activity from SHOW ENGINE INNODB STATUS.",
@@ -21,6 +42,7 @@ from integrations.mariadb import (
     is_available=mariadb_is_available,
     injected_params=("host", "password", "username"),
     extract_params=mariadb_extract_params,
+    evidence_mapper=_map_get_mariadb_innodb_status,
 )
 def get_mariadb_innodb_status(
     host: str,
--- a/integrations/mariadb/tools/mariadb_process_list_tool/__init__.py
+++ b/integrations/mariadb/tools/mariadb_process_list_tool/__init__.py
@@ -2,6 +2,7 @@
 
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
 from core.tool_framework import tool
 from core.tool_framework.utils import call_db_tool_with_default_db_warning
@@ -11,6 +12,27 @@ from integrations.mariadb import (
     mariadb_extract_params,
     mariadb_is_available,
 )
+
+
+def _map_get_mariadb_process_list(
+    evidence: dict[str, Any], output: dict[str, Any], _tool_input: dict[str, Any]
+) -> None:
+    """Cite the active process count and the longest-running query."""
+    if not output.get("available"):
+        return
+    processes = output.get("processes") or []
+    if not processes:
+        return
+    longest = max((p.get("time_secs", 0) for p in processes), default=0)
+    record_evidence_entry(
+        evidence,
+        source="get_mariadb_process_list",
+        label="MariaDB Process List",
+        summary=(
+            f"{output.get('total_processes', len(processes))} active process(es), "
+            f"longest running {longest}s"
+        ),
+    )
 
 
 @tool(
@@ -24,6 +46,7 @@ from integrations.mariadb import (
     is_available=mariadb_is_available,
     injected_params=("host", "password", "username"),
     extract_params=mariadb_extract_params,
+    evidence_mapper=_map_get_mariadb_process_list,
 )
 def get_mariadb_process_list(
     host: str,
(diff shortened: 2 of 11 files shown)


arnavpopley:
@greptile review

PR #5772, opened 2026-08-26
feat(mongodb): add evidence mappers for all 5 mongodb investigation tools

### LINKED ISSUE #5547: Add evidence mappers for mongodb
## Summary
can point to. The `mongodb` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `get_mongodb_collection_stats`
- `get_mongodb_current_ops`
- `get_mongodb_profiler_data`
- `get_mongodb_replica_status`
- `get_mongodb_server_status`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="get_mongodb_collection_stats"' integrations/`.

```python

def _map_get_mongodb_collection_stats(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="get_mongodb_collection_stats",
        )

@tool(name="get_mongodb_collection_stats", ..., evidence_mapper=_map_get_mongodb_collection_stats)
```

## How to check it works

   under `tests/integrations/mongodb/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['get_mongodb_collection_stats', 'get_mongodb_current_ops', 'get_mongodb_profiler_data', 'get_mongodb_replica_status', 'get_mongodb_server_status']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'get_mongodb_collection_stats', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `get_mongodb_collection_stats`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k mongodb
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify mongodb
# edit /tmp/alert.json so the alert names a resource mongodb actually knows about
grep -E 'get_mongodb_collection_stats|get_mongodb_current_ops|get_mongodb_profiler_data|get_mongodb_replica_status|get_mongodb_server_status' /tmp/rca.json
```

matching lines into the PR.

**No credentials for mongodb?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5547

Wires all 5 MongoDB investigation tools to `record_evidence_entry` so their
investigation report.

- `_map_get_mongodb_server_status`
  (`integrations/mongodb/tools/mongodb_server_status_tool/__init__.py`):
  cites version and connection load (in-use vs. available).
- `_map_get_mongodb_current_ops`
  (`integrations/mongodb/tools/mongodb_current_ops_tool/__init__.py`): cites
  the count of long-running operations above the threshold and the longest
  running one's duration.
- `_map_get_mongodb_replica_status`
  (`integrations/mongodb/tools/mongodb_replica_status_tool/__init__.py`):
  cites replica set member count, names any unhealthy member
  (`health != 1`), or the "not part of a replica set" note when the server
  isn't in one.
- `_map_get_mongodb_profiler_data`
  (`integrations/mongodb/tools/mongodb_profiler_tool/__init__.py`): cites
  the slow-query count and the slowest entry's duration, or the diagnostic
  note when profiling is disabled — actionable information, distinct from
  "nothing to report."
- `_map_get_mongodb_collection_stats`
  (`integrations/mongodb/tools/mongodb_collection_stats_tool/__init__.py`):
  cites document count, storage size, and index count for the targeted
  collection. Unlike the other four, this tool always returns a single
  collection's metadata (not a list to check for emptiness), so the mapper
  records whenever the call succeeds — a collection with 0 documents is
  itself a real, useful finding.

The four list-based mappers record nothing when `available` is falsy, or
when their list field is empty with no accompanying `note` — a genuinely
empty result has nothing worth citing, while a `note` explaining *why* a
list is empty (not a replica set, profiling disabled) is real diagnostic
signal and is cited instead.

carry a mapper.

**A note on existing tool tests:** `tests/tools/test_mongodb_replica_status_tool.py`
and `test_mongodb_collection_stats_tool.py` had pre-existing `test_run_*`
fixtures using the *raw* MongoDB driver response shape (e.g. `set`/
`stateStr` and `collection`/`size`) instead of the client's actual
post-processed return shape in `integrations/mongodb/__init__.py`
(`set_name`/`state`/`health`, `ns`/`size_bytes`/`index_count`). Per this
project's convention of grounding mappers in the actual client code, I
wrote all mapper tests against the real shapes read directly from
`get_server_status`, `get_current_ops`, `get_rs_status`,
`get_profiler_data`, and `get_collection_stats`, rather than reusing the
stale fixtures. I left the existing (pre-existing, out-of-scope)
`test_run_*` tests untouched since fixing them isn't part of this issue.

```
  'get_mongodb_server_status', 'get_mongodb_current_ops',
  'get_mongodb_replica_status', 'get_mongodb_profiler_data',
  'get_mongodb_collection_stats']})"
{'get_mongodb_server_status': True, 'get_mongodb_current_ops': True,
 'get_mongodb_replica_status': True, 'get_mongodb_profiler_data': True,
 'get_mongodb_collection_stats': True}
```

return shape in `integrations/mongodb/__init__.py`:**
```
get_mongodb_server_status:      'MongoDB 6.0.10, 10 connection(s) in use, 990 available'
get_mongodb_current_ops:        '2 op(s) over 1000ms, longest running 12s'
get_mongodb_replica_status:     "2 member(s) in replica set 'rs0', unhealthy: rs0-b:27017"
get_mongodb_profiler_data:      '2 slow queries above 100ms, slowest 1200ms'
get_mongodb_collection_stats:   "'my-db.my-collection': 1000 document(s), 2.0MB, 3 index(es)"
```

**Tests and quality gates:**
```
11 passed

$ uv run pytest tests/ -k mongodb -q
161 passed

```

**No live MongoDB instance available in this environment**, so leaving that
shapes; a reviewer with a MongoDB instance can finish the live check.

`get_mongodb_replica_status`:
```diff
+  "catalog_entries": [
+    {
+      "source": "get_mongodb_replica_status",
+      "label": "MongoDB Replica Set Status",
+      "summary": "2 member(s) in replica set 'rs0', unhealthy: rs0-b:27017",
+      "url": null,
+      "snippet": null
+    }
+  ]
```
`evidence` changes.)

**4. Answers:**

- **What the reader gains.** Server version and connection load, a
  long-running operation count with the worst offender's duration, replica
  set health with any unhealthy member named (or that the server isn't a
  replica), the slowest profiled query's duration (or why profiling data
  couldn't be collected), and a specific collection's document
  count/size/index count — all previously invisible in the report even
  though the agent already paid to fetch them.
- **How much context it costs.** 154–183 characters (JSON-serialized) per
  entry — single-line summaries, no inlined per-operation/per-member/
  per-query lists.
  new tests: `available: False` for all 5; an empty `operations`/`members`/
  `entries` list with no note attached produces zero `catalog_entries` for
  the four list-based tools.
  non-Atlas MongoDB tools in the codebase (confirmed via the baseline gaps
  file, which listed exactly these 5 plus the separate `mongodb_atlas_*`
  tools — untouched, out of scope for this issue), and no other mapper
  reads any of their output.
- **Whether an existing report changed shape.** `tests/ -k mongodb` (161
  unchanged.

---

`collection_stats` is the one tool of the five that doesn't return a list
to check for emptiness — it's a single collection's metadata, always
present when the call succeeds. I deliberately did not gate it on, say,
`count > 0`: a collection with zero documents is a real, useful answer
(rules out "this collection is the problem" as a hypothesis), not an
"empty result" in the sense the other four tools' list fields are. For
`replica_status` and `profiler_data`, I treated the `note` field (server
not in a replica set; profiling disabled) as citeable evidence in its own
right, since it answers a real question the agent would otherwise have to
dig for in the tool-call transcript — distinct from a genuinely empty list
with no note, which still records nothing per the project's "no noise"
convention. I used MongoDB's own `health != 1` convention (1 = healthy in
`replSetGetStatus` member state) to flag unhealthy replica members rather
than string-matching on `state`, since `health` is the driver's own
canonical health signal.

---

---


arnavpopley:
@greptile review

arnavpopley:
@greptile review

arnavpopley:
From OpenSRE Bot in Slack:

### :white_check_mark: PR #5772 — mongodb evidence mappers

Five tools wired: server status, current ops, replica status, profiler, collection stats. Good full-coverage approach.

**Watch:** MongoDB profiler output is the riskiest here — it can return large payloads (full query text, bind params, timing histograms). If there's no truncation before record_evidence_entry, evidence entries could be very large. Worth a quick check that profiler output is trimmed or summarized before storage.

**Verdict:** :white_check_mark: Merge (after #5779), with a note to verify profiler truncation

PR #5773, opened 2026-08-26
feat(opensearch): add evidence mapper for query_opensearch_analytics

### LINKED ISSUE #5588: Add evidence mappers for opensearch
## Summary
can point to. The `opensearch` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `query_opensearch_analytics`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="query_opensearch_analytics"' integrations/`.

```python

def _map_query_opensearch_analytics(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="query_opensearch_analytics",
        )

@tool(name="query_opensearch_analytics", ..., evidence_mapper=_map_query_opensearch_analytics)
```

## How to check it works

   under `tests/integrations/opensearch/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['query_opensearch_analytics']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'query_opensearch_analytics', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `query_opensearch_analytics`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k opensearch
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify opensearch
# edit /tmp/alert.json so the alert names a resource opensearch actually knows about
grep -E 'query_opensearch_analytics' /tmp/rca.json
```

matching lines into the PR.

**No credentials for opensearch?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5588

Wires `query_opensearch_analytics` — the only OpenSearch investigation
tool — to `record_evidence_entry` so its output stops being silently
dropped and becomes a citeable line in the investigation report.

- `_map_query_opensearch_analytics`
  (`integrations/opensearch/tools/opensearch_analytics_tool/__init__.py`):
  cites the number of log entries retrieved, the query string, and the
  index pattern searched. Mirrors the existing precedent at
  `integrations/elasticsearch/tools/__init__.py`'s `_map_elasticsearch_logs`
  (the same underlying `ElasticsearchClient` backs both tools), adapted to
  this tool's actual output shape — `total_returned`/`query`/
  `index_pattern`/`logs`, no `error_logs` field, unlike the Elasticsearch
  tool.

This is a read-only diagnostic query — not an action/notification tool —
so it's in scope per the issue.

The mapper records nothing when `available` is falsy or `logs` is empty —
zero log entries matching a query is a real, useful "no matches" finding
in principle, but the project's "no noise" convention treats an empty
result list the same way the existing Elasticsearch precedent does (no
entry when `logs` is empty), so I matched that behavior rather than
diverging from established practice for this integration.

Removed `query_opensearch_analytics` from `evidence_mapper_baseline.txt`
now that it carries a mapper.

```
print({'query_opensearch_analytics': bool(g('query_opensearch_analytics').evidence_mapper)})"
{'query_opensearch_analytics': True}
```

gather-evidence path), using a realistic sample matching the tool's return
shape:**
```
query_opensearch_analytics: "2 log(s) for query 'service:checkout' on 'logs-*'"
```

**Tests and quality gates:**
```
11 passed

$ uv run pytest tests/ -k opensearch -q
51 passed

```

**No live OpenSearch instance available in this environment**, so leaving
that box unticked per the issue's own guidance. The checks above stand as
evidence the mapper is attached and produces a real report entry from a
realistic sample payload matching the actual tool's return shape; a
reviewer with an OpenSearch instance can finish the live check.

**1–2. Before/after**, run with a realistic sample matching the tool's

Before: `evidence` has no `catalog_entries` key.

After:
```diff
+  "catalog_entries": [
+    {
+      "source": "query_opensearch_analytics",
+      "label": "OpenSearch Analytics",
+      "summary": "2 log(s) for query 'service:checkout' on 'logs-*'",
+      "url": null,
+      "snippet": null
+    }
+  ]
```

**3. Explain every difference:** the diff adds exactly one new

**4. Answers:**

- **What the reader gains.** How many log entries matched the query and
  index pattern searched — previously invisible in the report even though
  the agent already paid to run the query.
- **How much context it costs.** 167 characters (JSON-serialized) — a
  single-line summary, no inlined log entries.
  new tests: `available: False`, and an empty `logs` list, both produce
  zero `catalog_entries`.
- **Whether anything is now recorded twice.** No. This is the only
  OpenSearch tool in the codebase (confirmed via the baseline gaps file,
  which listed only this one tool for `opensearch`), and no other mapper
  reads its output. It shares an underlying client with
  `query_elasticsearch_logs`, but that tool has its own separate mapper
  keyed on a different `source` string, so there's no double-recording risk.
- **Whether an existing report changed shape.** `tests/ -k opensearch` (51
  unchanged.

---

Before writing this mapper I checked whether OpenSearch's sibling
integration, Elasticsearch, already had one — it does
(`_map_elasticsearch_logs` in `integrations/elasticsearch/tools/__init__.py`),
since both tools sit on top of the same `ElasticsearchClient`. Rather than
copy it wholesale, I read `query_opensearch_analytics`'s actual return
dict and found it's a narrower shape than Elasticsearch's tool output —
no `error_logs` field, and it carries `total_returned`/`query`/
`index_pattern` fields the Elasticsearch mapper doesn't reference. I wrote
a mapper specific to this shape rather than importing or generalizing the
Elasticsearch one, since the two tools' outputs aren't guaranteed to stay
in sync and the project's convention is one mapper per tool, living with
that tool.

---

---

### DIFF
--- a/integrations/opensearch/tools/opensearch_analytics_tool/__init__.py
+++ b/integrations/opensearch/tools/opensearch_analytics_tool/__init__.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
 from core.tool_framework import tool
 from core.tool_framework.utils import tool_unavailable
@@ -21,6 +22,26 @@ def _bounded_limit(limit: int, max_results: int) -> int:
 def _opensearch_available(sources: dict[str, dict[str, Any]]) -> bool:
     source = sources.get("opensearch", {})
     return bool(source.get("connection_verified") and source.get("url"))
+
+
+def _map_query_opensearch_analytics(
+    evidence: dict[str, Any], output: dict[str, Any], _tool_input: dict[str, Any]
+) -> None:
+    """Cite the number of log entries retrieved for the query."""
+    if not output.get("available"):
+        return
+    logs = output.get("logs") or []
+    if not logs:
+        return
+    record_evidence_entry(
+        evidence,
+        source="query_opensearch_analytics",
+        label="OpenSearch Analytics",
+        summary=(
+            f"{output.get('total_returned', len(logs))} log(s) for query "
+            f"'{output.get('query', '*')}' on '{output.get('index_pattern', '*')}'"
+        ),
+    )
 
 
 def _opensearch_extract_params(sources: dict[str, dict[str, Any]]) -> dict[str, Any]:
@@ -64,6 +85,7 @@ def _opensearch_extract_params(sources: dict[str, dict[str, Any]]) -> dict[str, 
     is_available=_opensearch_available,
     injected_params=("url",),
     extract_params=_opensearch_extract_params,
+    evidence_mapper=_map_query_opensearch_analytics,
 )
 def query_opensearch_analytics(
     url: str,
--- a/integrations/opensearch/tools/opensearch_analytics_tool/__init__.py
+++ b/integrations/opensearch/tools/opensearch_analytics_tool/__init__.py
@@ -8,10 +8,12 @@ from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
 from core.tool_framework import tool
 from core.tool_framework.utils import tool_unavailable
+from infrastructure.text.truncation import truncate
 from integrations.elasticsearch.client import ElasticsearchClient, ElasticsearchConfig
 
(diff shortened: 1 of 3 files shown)


arnavpopley:
@greptile review

PR #5777, opened 2026-08-26
feat(signoz): add evidence mappers for all 3 signoz investigation tools

### LINKED ISSUE #5561: Add evidence mappers for signoz
## Summary
can point to. The `signoz` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `query_signoz_logs`
- `query_signoz_metrics`
- `query_signoz_traces`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="query_signoz_logs"' integrations/`.

```python

def _map_query_signoz_logs(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="query_signoz_logs",
        )

@tool(name="query_signoz_logs", ..., evidence_mapper=_map_query_signoz_logs)
```

## How to check it works

   under `tests/integrations/signoz/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['query_signoz_logs', 'query_signoz_metrics', 'query_signoz_traces']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'query_signoz_logs', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `query_signoz_logs`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k signoz
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify signoz
# edit /tmp/alert.json so the alert names a resource signoz actually knows about
grep -E 'query_signoz_logs|query_signoz_metrics|query_signoz_traces' /tmp/rca.json
```

matching lines into the PR.

**No credentials for signoz?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5561

Wires all 3 SigNoz investigation tools to `record_evidence_entry` so their
investigation report.

- `_map_query_signoz_logs`
  (`integrations/signoz/tools/query_signoz_logs_tool/tool.py`): cites the
  log count retrieved and how many matched an error signal.
- `_map_query_signoz_metrics`
  (`integrations/signoz/tools/query_signoz_metrics_tool/tool.py`): cites
  the metric name, aggregation, and number of data points retrieved.
- `_map_query_signoz_traces`
  (`integrations/signoz/tools/query_signoz_traces_tool/tool.py`): cites
  the aggregate span/error/p99-latency summary when available, falling
  back to a plain trace count only if that aggregate call itself failed.

All 3 are read-only diagnostic queries — none is an action/notification

**On count honesty:** every SigNoz row-fetch (`_query_logs_via_api`,
`_query_metrics_via_api`, `_query_traces_via_api` in
`integrations/signoz/client.py`) sends the tool's `limit` as the query
spec's own row cap and reports `total` as `len(<rows>)` from that already-
capped response — there's no separate unbounded total. This is the same
ambiguity I hit three times earlier in this session on the companion
Sentry, PagerDuty, and GitLab evidence-mapper issues, each of which
established or reused a `"N+"` convention for it (originating in
`validate_sentry_config`). I added the equivalent `signoz_count_label`
helper to `integrations/signoz/__init__.py` — the module all three SigNoz
tool files already import `signoz_extract_params` from — and used it in
logs and metrics, and as the fallback path in traces.

**Why traces is different:** `query_signoz_traces`'s output also carries a
`summary` field from `query_trace_summary`, which runs a *scalar
aggregation* query (`count()`, `p99(duration_nano)`, etc.) over the full
time window rather than fetching capped rows — `summary["total_spans"]`
and `summary["error_spans"]` are genuine full-window totals, not a
capped-page artifact. I read `_query_trace_summary_via_api` to confirm
this before treating it as authoritative, and the mapper prefers it over
the raw `traces` list count whenever the summary call itself succeeded,
only falling back to the (page-capped) trace count if the summary call
failed.

Every mapper records nothing when `available` is falsy, or when its
relevant list/summary field is empty.

Removed all 3 tool names from `evidence_mapper_baseline.txt` now that they
carry a mapper.

```
  'query_signoz_logs', 'query_signoz_metrics', 'query_signoz_traces']})"
{'query_signoz_logs': True, 'query_signoz_metrics': True,
 'query_signoz_traces': True}
```

shape:**
```
query_signoz_logs:      '2 log(s), 1 error(s)'
query_signoz_metrics:   'cpu_usage (avg): 1 data point(s)'
query_signoz_traces:    '100 span(s), 5 error(s), p99 250.0ms'
```

**Tests and quality gates:**
```
11 passed

$ uv run pytest tests/ -k signoz -q
56 passed

```

**No live SigNoz instance available in this environment**, so leaving that
shapes (including the existing `_FakeSigNozBackend` test fixture already
in this file); a reviewer with a SigNoz instance can finish the live
check.

**1–2. Before/after**, run with realistic samples matching each tool's

Before: `evidence` has no `catalog_entries` key for any of the 3 tools.

`query_signoz_traces`:
```diff
+  "catalog_entries": [
+    {
+      "source": "query_signoz_traces",
+      "label": "SigNoz Traces",
+      "summary": "100 span(s), 5 error(s), p99 250.0ms",
+      "url": null,
+      "snippet": null
+    }
+  ]
```
(and analogously for the other 2 — one new entry each, nothing else in
`evidence` changes.)

**4. Answers:**

- **What the reader gains.** A log/error count for the query window, a
  metric's name/aggregation/data-point count, and a trace summary with
  span count, error count, and p99 latency — all previously invisible in
  the report even though the agent already paid to fetch them, and
  honestly qualified with "+" whenever a row-based count could be a capped
  page rather than every match.
- **How much context it costs.** 120–140 characters (JSON-serialized) per
  entry — single-line summaries, no inlined log/metric/trace lists.
  new tests: `available: False` for all 3; empty `logs`/`metrics`/`traces`
  lists (with an unavailable `summary` too, for traces) all produce zero
  `catalog_entries`.
- **Whether anything is now recorded twice.** No. These are the only 3
  SigNoz tools in the codebase (confirmed via the baseline gaps file,
  which listed exactly these 3 for `signoz`), and no other mapper reads
  any of their output.
- **Whether an existing report changed shape.** `tests/ -k signoz` (56
  unchanged.

---

The interesting design decision was `query_signoz_traces`. Its output
combines two very different kinds of data: a row-limited `traces` list
(capped by `effective_limit` at the query-spec level, same ambiguity as
logs/metrics) and a `summary` dict built from a completely separate
*scalar aggregation* request type (`requestType: "scalar"` with
`count()`/`p99(duration_nano)` expressions). I read
`_query_trace_summary_via_api` specifically to confirm it issues its own
request against the full time window rather than reusing the capped row
fetch, which is what makes `total_spans`/`error_spans` trustworthy as
real totals rather than another page-cap artifact. I deliberately made the
mapper prefer that summary and only fall back to the page-capped trace
count when the summary sub-call itself failed (`summary["available"]` is
false) — that mirrors the tool's own `run()`, which always attempts both
calls and merges them (`{**traces_result, "summary": summary}`) rather
than treating one as a substitute for the other. I also confirmed a
project-wide pattern of the "N+" fix I'd already applied three times
earlier this session (Sentry, PagerDuty, GitLab) generalizes cleanly here
too, since all three SigNoz row-fetch methods share the identical
`_clamp_limit` + `len(rows)` construction.

---

---


arnavpopley:
@greptile review

arnavpopley:
@greptile review

arnavpopley:
From OpenSRE Bot in Slack:

### :white_check_mark: PR #5777 — signoz evidence mappers

Three tools: logs, metrics, traces. Observability signals as first-class citeable evidence is the most powerful addition in this batch — it closes the loop between raw signals and report conclusions.

**Watch:** Verify that the trace evidence mapper captures span IDs and trace IDs as structured fields, not just as text in the evidence body. If they're structured, downstream report rendering can link out to the SigNoz UI. If they're just stringified, that opportunity is lost.

**Verdict:** :white_check_mark: Merge (after #5779)

arnavpopley:
@greptile review

PR #5775, opened 2026-08-26
feat(pagerduty): add evidence mappers for all 4 pagerduty investigation tools

arnavpopley:
Closes #5554

#### Describe the changes you have made in this PR -

Wires all 4 PagerDuty investigation tools to `record_evidence_entry` so
their output stops being silently dropped and becomes a citeable line in
the investigation report.

- `_map_pagerduty_incident_detail`: cites the incident's title, status, and
  timeline entry count.
- `_map_pagerduty_incidents`: cites the incident count and how many are
  currently active (triggered/acknowledged).
- `_map_pagerduty_oncall`: cites the on-call responder count.
- `_map_pagerduty_services`: cites either the fetched service's name and
  status (single-service detail mode, when `service_id` is given) or the
  service count (listing mode) — the tool's `run()` returns one or the
  other depending on whether `service_id` was passed, so the mapper checks
  `service` first and falls back to `services`.

All 4 are read-only diagnostic queries — none is an action/notification
tool, so all are in scope per the issue.

**On count honesty — the "N+" convention:** every PagerDuty list endpoint
in `integrations/pagerduty/client.py` (`list_incidents`,
`list_incident_log_entries`, `get_oncalls`, `list_services`) applies
`min(limit, 100)` as the API's own page-size param and returns
`total: len(<list>)`, discarding PagerDuty's own pagination metadata (the
API's `more` flag). That means `total` can be a capped page rather than
every matching record. I added a shared `_pagerduty_count_label(count,
requested_limit)` helper that shows the plain count when it's under the
effective page size, and `"N+"` when it saturates — the exact convention
already established elsewhere in this codebase for the identical ambiguity
(see `validate_sentry_config` in `integrations/sentry/__init__.py`, which
I found and matched while working the companion Sentry evidence-mapper
issue in the same session). All 4 mappers use this helper wherever they
cite a list count.

Every mapper records nothing when `available` is falsy, or when its
dict/list field is empty.

Removed all 4 tool names from `evidence_mapper_baseline.txt` now that they
carry a mapper.

### Demo/Screenshot for feature changes and bug fixes -

**Tools carry their mapper (via the real registry):**
```
$ uv run python -c "from tools.registry import get_registered_tool as g; \
print({n: bool(g(n).evidence_mapper) for n in [
  'pagerduty_incident_detail', 'pagerduty_incidents',
  'pagerduty_oncall', 'pagerduty_services']})"
{'pagerduty_incident_detail': True, 'pagerduty_incidents': True,
[...]

PR #5787, opened 2026-08-26
feat(opsgenie): add evidence mappers for both opsgenie investigation tools

### LINKED ISSUE #5569: Add evidence mappers for opsgenie
## Summary
can point to. The `opsgenie` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `opsgenie_alert_detail`
- `opsgenie_alerts`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="opsgenie_alert_detail"' integrations/`.

```python

def _map_opsgenie_alert_detail(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="opsgenie_alert_detail",
        )

@tool(name="opsgenie_alert_detail", ..., evidence_mapper=_map_opsgenie_alert_detail)
```

## How to check it works

   under `tests/integrations/opsgenie/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['opsgenie_alert_detail', 'opsgenie_alerts']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'opsgenie_alert_detail', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `opsgenie_alert_detail`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k opsgenie
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify opsgenie
# edit /tmp/alert.json so the alert names a resource opsgenie actually knows about
grep -E 'opsgenie_alert_detail|opsgenie_alerts' /tmp/rca.json
```

matching lines into the PR.

**No credentials for opsgenie?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5569

Wires both OpsGenie investigation tools to `record_evidence_entry` so their
investigation report.

- `_map_opsgenie_alert_detail`: cites the alert's message, status, priority
  (when set), and activity-log entry count (when the log was fetched).
- `_map_opsgenie_alerts`: cites the alert count and how many are open.

**On count honesty — the "N+" convention:** both `list_alerts` and
`get_alert_logs` in `integrations/opsgenie/client.py` apply `min(limit, 100)`
as the API's own page-size param and return `total: len(<list>)`, discarding
any pagination metadata OpsGenie itself might expose — the identical
ambiguity already solved elsewhere in this codebase (see
`validate_sentry_config` in `integrations/sentry/__init__.py`, and the
PagerDuty companion PR from this same session). I added
`_opsgenie_count_label(count, requested_limit)` implementing the same rule
and used it in both mappers, including for `opsgenie_alerts`'s "open" sub-
count — which inherits the same truncation flag as the total, since on a
capped page it can only be "open alerts within the returned page."

**Free-form text:** alert `message` is human-entered and unbounded, so I
added `_opsgenie_summary_text()` to collapse `\n`/`\r`/`\r\n` and cap it at
120 chars before it goes into a summary — I wrote this to also strip carriage
returns from the start, having just had Greptile catch that exact gap
(LF-only stripping) on the companion PagerDuty PR in this same session.

Both mappers record nothing when `available` is falsy or the relevant
dict/list is empty.

Removed both `opsgenie_alert_detail` and `opsgenie_alerts` from

```
print({n: bool(g(n).evidence_mapper) for n in ['opsgenie_alert_detail', 'opsgenie_alerts']})"
{'opsgenie_alert_detail': True, 'opsgenie_alerts': True}
```

shape:**
```
opsgenie_alert_detail:  "'CPU high', open, P1, 3 log entries"
opsgenie_alerts:        '2 alert(s), 1 open'
opsgenie_alerts (saturated page):  '100+ alert(s), 100+ open'
```

**Tests and quality gates:**
```
11 passed

$ uv run pytest tests/tools/test_opsgenie_alert_detail_tool.py tests/tools/test_opsgenie_alerts_tool.py -q
29 passed

```

**No live OpsGenie account available in this environment**, so leaving the
existing `is_available`/`extract_params` wiring untouched — only the mappers
and their tests were added, using each tool's own realistic mock payload
shapes.

I read `client.list_alerts`/`get_alert`/`get_alert_logs` in
`integrations/opsgenie/client.py` before writing anything, since OpsGenie's
shape (an `OpsGenieAlertDetailTool` + `OpsGenieAlertsTool` pair) is nearly
identical to PagerDuty's `incident_detail`/`incidents` pair I'd just finished
in this same session — same `min(limit, 100)` page cap, same discarded
pagination metadata, same free-form title field. Rather than re-derive the
convention I reused the exact `"N+"` phrasing and the same carriage-return
fix I'd already made for PagerDuty, so both integrations now handle the
identical failure mode identically instead of drifting into two slightly
different conventions.

---

---


arnavpopley:
@greptile review

arnavpopley:
Pushing back on this one after checking the actual output shape.

`opsgenie_alert_detail.run()` sets `total_log_entries = len(activity_log)`, but `activity_log` stays `[]` in three indistinguishable cases:
1. `include_activity_log=False` — the log was never requested.
2. `include_activity_log=True` but `client.get_alert_logs()` fails — the failure is swallowed and `activity_log` silently stays `[]`.
3. `include_activity_log=True` and the fetch succeeds with a genuinely empty log.

The output dict carries no separate flag (e.g. `activity_log_fetched`) distinguishing case 3 from cases 1/2, so the mapper cannot reliably tell "confirmed zero" apart from "never checked" or "checked and failed." Always emitting "0 log entries" would misrepresent cases 1/2 as a confirmed empty result — a worse failure mode than the current omission, since it would assert something the tool never actually verified.

This is also not a new pattern introduced here: `pagerduty_incident_detail`'s `total_log_entries` (merged in #5775, reviewed 5/5 twice) has the identical shape — `log_entries` stays `[]` in the same three cases, and its mapper uses the same truthiness guard. Fixing OpsGenie's mapper alone while leaving PagerDuty's identical pattern in place would just create an inconsistency between two near-identical integrations, not resolve the underlying ambiguity.

A real fix would need to add a `activity_log_fetched: bool` (or similar) to `run()`'s output contract so the mapper can distinguish "confirmed empty" from "not checked" — that's a tool-output-shape change, not a mapper-only change, and out of scope for this PR (which is only wiring existing outputs to evidence, per the linked issue). Happy to open a follow-up if the maintainers want that contract change made consistently across both integrations.

Leaving the code as-is for this reason — let me know if there's a reason to treat OpsGenie's case differently from PagerDuty's already-merged one.

arnavpopley:
@greptile review

PR #5788, opened 2026-08-26
feat(honeycomb): add evidence mapper for query_honeycomb_traces

### LINKED ISSUE #5585: Add evidence mappers for honeycomb
## Summary
can point to. The `honeycomb` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `query_honeycomb_traces`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="query_honeycomb_traces"' integrations/`.

```python

def _map_query_honeycomb_traces(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="query_honeycomb_traces",
        )

@tool(name="query_honeycomb_traces", ..., evidence_mapper=_map_query_honeycomb_traces)
```

## How to check it works

   under `tests/integrations/honeycomb/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['query_honeycomb_traces']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'query_honeycomb_traces', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `query_honeycomb_traces`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k honeycomb
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify honeycomb
# edit /tmp/alert.json so the alert names a resource honeycomb actually knows about
grep -E 'query_honeycomb_traces' /tmp/rca.json
```

matching lines into the PR.

**No credentials for honeycomb?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5585

Wires the single Honeycomb investigation tool, `query_honeycomb_traces`, to
`record_evidence_entry` so its output stops being silently dropped and becomes
a citeable line in the investigation report.

- `_map_query_honeycomb_traces`: cites the trace/span group count and the
  service or trace ID queried.

This is the only tool in `integrations/honeycomb/tools/` — it's a read-only
diagnostic query, so it's in scope per the issue.

**On count honesty:** I checked `client.query_traces` → `run_query` →
`create_query_result` in `integrations/honeycomb/client.py` for whether
`limit` gets clamped server-side the way PagerDuty/OpsGenie/Better Stack clamp
theirs — it doesn't; the requested `limit` is sent straight through to the
Honeycomb API unclamped. Since there's no fixed page size to compare against,
the mapper compares the returned count against the caller's *own* requested
`limit` (still using the same `"N+"` phrasing established elsewhere in this
codebase, e.g. `validate_sentry_config` in `integrations/sentry/__init__.py`)
rather than a hardcoded ceiling.

The `service_name`/`trace_id` target embedded in the summary is caller-
supplied and not bounded by the input schema, so it's collapsed and capped at
80 chars first — the same defensive truncation applied to free text in the
other evidence mappers from this session.

The mapper records nothing when `available` is falsy or `traces` is empty.

Removed `query_honeycomb_traces` from `evidence_mapper_baseline.txt` now that
it carries a mapper.

```
print(bool(g('query_honeycomb_traces').evidence_mapper))"
True
```

gather-evidence path), using realistic samples matching the tool's return
shape:**
```
query_honeycomb_traces:  "3 trace/span group(s) for 'checkout'"
query_honeycomb_traces (saturated page):  "20+ trace/span group(s) for 'checkout'"
```

**Tests and quality gates:**
```
11 passed

$ uv run pytest tests/tools/test_honeycomb_traces_tool.py -q
15 passed

```

**No live Honeycomb account available in this environment**, so leaving the
existing `is_available`/`extract_params` wiring untouched — only the mapper
and its tests were added, using the tool's own realistic mock payload shape.

The one thing I couldn't take for granted from the issue's generic template
was whether Honeycomb's `limit` behaves like the other integrations' page
caps I'd just worked through in this session (PagerDuty, OpsGenie, Better
Stack) — all three clamp the requested limit server-side or client-side
before use. Honeycomb's `run_query`/`create_query_result` don't: the caller's
`limit` goes straight into the API payload. That changes what "the page is
saturated" means here — there's no fixed ceiling to clamp against, only the
caller's own request — so the mapper's "N+" check compares against
`tool_input.get("limit", 20)` directly rather than a `min(requested,
MAX_PAGE_SIZE)` expression like the other mappers use.

---

---

### DIFF
--- a/integrations/honeycomb/tools/__init__.py
+++ b/integrations/honeycomb/tools/__init__.py
@@ -32,10 +32,21 @@ def _map_query_honeycomb_traces(
     requested_limit = tool_input.get("limit", 20)
     count_label = f"{total}+" if total >= max(requested_limit, 1) else str(total)
     summary = f"{count_label} trace/span group(s)"
-    target = output.get("service_name") or output.get("trace_id")
-    if target:
-        safe_target = truncate(str(target).replace("\n", " "), _TARGET_SUMMARY_TRUNCATE_LEN)
-        summary += f" for '{safe_target}'"
+
+    def _safe(value: str) -> str:
+        return truncate(value.replace("\n", " "), _TARGET_SUMMARY_TRUNCATE_LEN)
+
+    # A query can filter by service_name and trace_id together (they AND in
+    # client.query_traces) -- cite both when both were applied, not just one.
+    service_name = output.get("service_name")
+    trace_id = output.get("trace_id")
+    targets = []
+    if service_name:
+        targets.append(f"service '{_safe(str(service_name))}'")
+    if trace_id:
+        targets.append(f"trace '{_safe(str(trace_id))}'")
+    if targets:
+        summary += f" for {', '.join(targets)}"
     record_evidence_entry(
         evidence,
         source="query_honeycomb_traces",
--- a/integrations/honeycomb/tools/__init__.py
+++ b/integrations/honeycomb/tools/__init__.py
@@ -6,10 +6,42 @@ from __future__ import annotations
 
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.tool import BaseTool
 from core.tool_framework.utils import tool_unavailable
+from infrastructure.text.truncation import truncate
 from integrations.honeycomb.client import HoneycombClient
 from integrations.honeycomb.config import HoneycombIntegrationConfig
+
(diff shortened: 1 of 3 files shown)


arnavpopley:
@greptile review

PR #5789, opened 2026-08-26
feat(incident_io): add evidence mapper for incident_io_incidents

### LINKED ISSUE #5586: Add evidence mappers for incident_io
## Summary
can point to. The `incident_io` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `incident_io_incidents`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="incident_io_incidents"' integrations/`.

```python

def _map_incident_io_incidents(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="incident_io_incidents",
        )

@tool(name="incident_io_incidents", ..., evidence_mapper=_map_incident_io_incidents)
```

## How to check it works

   under `tests/integrations/incident_io/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['incident_io_incidents']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'incident_io_incidents', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `incident_io_incidents`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k incident_io
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify incident_io
# edit /tmp/alert.json so the alert names a resource incident_io actually knows about
grep -E 'incident_io_incidents' /tmp/rca.json
```

matching lines into the PR.

**No credentials for incident_io?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5586

Wires the single incident.io investigation tool, `incident_io_incidents`, to
`record_evidence_entry` so its output stops being silently dropped and

Unlike the other integrations in this session's batch, this tool is a single
multi-action dispatcher (`action` in `list`/`get`/`updates`/`context`/
`append_summary`) rather than one query per tool, so `_map_incident_io_incidents`
branches on `output["action"]` (always set by the tool's own `run()`) and
cites what that action actually returned:

- `list`: incident count.
- `get`: the incident's name, status, and severity.
- `updates`: update count.
- `context` (the default action): incident name/status/severity plus the
  update count fetched alongside it.
- `append_summary` is intentionally **not** handled — it's a write action
  (appends OpenSRE's own findings to the incident summary) with no gathered
  evidence to cite; the issue says to skip action-only tools, and this is the
  one branch of this tool that fits that description.

**On count honesty:** `list_incidents` and `list_incident_updates` in
`integrations/incident_io/client.py` clamp the caller's `page_size` to a
maximum (500 / 250 respectively) via `_safe_int`, but — better than most of
the other integrations in this session's batch — the client also forwards the
API's own `pagination_meta` (with an `after` cursor) straight through when the
API returns one. That's an explicit truncation signal, the same kind
CloudTrail's `NextToken` provides, so the mapper uses
`pagination_meta.get("after")` directly instead of a page-cap heuristic.

**Free-form text:** incident `name` is human-entered and unbounded, so I
collapse `\n`/`\r`/`\r\n` and cap it at 120 chars before it goes into a
summary — the same fix Greptile had flagged as incomplete (LF-only) on the
companion PagerDuty PR earlier in this session, applied correctly from the
start here.

Removed `incident_io_incidents` from `evidence_mapper_baseline.txt` now that
it carries a mapper.

```
print(bool(g('incident_io_incidents').evidence_mapper))"
True
```

gather-evidence path), using realistic samples matching each action's return
shape:**
```
action=list:     '2 incident(s)'
action=get:      "'Database outage', Mitigating, SEV1"
action=updates:  '3 update(s)'
action=context:  "'Database outage', Mitigating, 5 update(s)"
action=append_summary:  (no evidence recorded — write action)
```

**Tests and quality gates:**
```
11 passed

$ uv run pytest tests/tools/test_incident_io_tool.py -q
13 passed

```

**No live incident.io account available in this environment**, so leaving the
existing `is_available`/`extract_params` wiring untouched — only the mapper
and its tests were added, using the tool's own realistic mock payload shapes
for each action.

I read all five handlers in `IncidentIoIncidentsTool.run()` and their backing
`client.py` methods before writing anything, because this tool doesn't fit
the "one query, one shape" pattern the rest of this session's batch did — its
output shape depends entirely on which `action` was dispatched. I deliberately
did *not* write one generic mapper that tries to summarize whatever fields
happen to be present; that would produce a nonsensical or misleading citation
for `append_summary` (which has no diagnostic data, just a success flag) and
risk conflating `list`'s incident count with `get`'s single-incident detail.
Branching explicitly on `action` also made the "skip action-only tools" rule
from the issue apply cleanly to just the one relevant branch instead of the
whole tool. Finding `pagination_meta.after` already threaded through
`get_incident_context`/`list_incidents`/`list_incident_updates` was the one
pleasant surprise — it meant I could use an explicit API signal instead of
reapplying the page-cap heuristic I'd used for Sentry/PagerDuty/OpsGenie/
Better Stack earlier in this session.

---

---

### DIFF
--- a/integrations/incident_io/client.py
+++ b/integrations/incident_io/client.py
@@ -315,12 +315,15 @@ class IncidentIoClient:
                 "incident": incident_result.get("incident", {}),
                 "error": updates_result.get("error", "unknown error"),
             }
-        return {
+        context_result: dict[str, Any] = {
             "success": True,
             "incident": incident_result.get("incident", {}),
             "incident_updates": updates_result.get("incident_updates", []),
             "total_updates": updates_result.get("total", 0),
         }
+        if "pagination_meta" in updates_result:
+            context_result["pagination_meta"] = updates_result["pagination_meta"]
+        return context_result
 
     def append_summary_update(
         self,
--- a/integrations/incident_io/tools/__init__.py
+++ b/integrations/incident_io/tools/__init__.py
@@ -90,7 +90,12 @@ def _map_incident_io_incidents(
         parts = [_incident_io_incident_summary(incident)]
         total_updates = output.get("total_updates", 0)
         if total_updates:
-            parts.append(f"{total_updates} update(s)")
+            label = (
+                f"{total_updates}+"
+                if _incident_io_page_is_truncated(output)
+                else str(total_updates)
+            )
+            parts.append(f"{label} update(s)")
         record_evidence_entry(
             evidence,
             source="incident_io_incidents",
--- a/integrations/incident_io/tools/__init__.py
+++ b/integrations/incident_io/tools/__init__.py
@@ -6,9 +6,97 @@ from __future__ import annotations
 
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.tool import BaseTool
 from core.tool_framework.utils import tool_unavailable
+from infrastructure.text.truncation import truncate
 from integrations.incident_io.client import make_incident_io_client
+
+#: Incident names are free-form, human-entered text -- unbounded and can
+#: contain newlines or carriage returns. Cap the length used in a report
+#: summary so one long or multi-line value can't produce a malformed or
+#: oversized report line.
+_NAME_SUMMARY_TRUNCATE_LEN = 120
+
+
+def _incident_io_summary_text(value: str) -> str:
+    """Collapse and cap free-form incident.io text before it goes into a summary."""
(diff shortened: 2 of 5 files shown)


arnavpopley:
@greptile review

PR #5791, opened 2026-08-26
feat(azure): add evidence mapper for query_azure_monitor_logs

### LINKED ISSUE #5578: Add evidence mappers for azure
## Summary
can point to. The `azure` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `query_azure_monitor_logs`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="query_azure_monitor_logs"' integrations/`.

```python

def _map_query_azure_monitor_logs(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="query_azure_monitor_logs",
        )

@tool(name="query_azure_monitor_logs", ..., evidence_mapper=_map_query_azure_monitor_logs)
```

## How to check it works

   under `tests/integrations/azure/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['query_azure_monitor_logs']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'query_azure_monitor_logs', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `query_azure_monitor_logs`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k azure
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify azure
# edit /tmp/alert.json so the alert names a resource azure actually knows about
grep -E 'query_azure_monitor_logs' /tmp/rca.json
```

matching lines into the PR.

**No credentials for azure?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5578 (part of epic #5519, tracking issue #1079)

Wires the single Azure Monitor investigation tool, `query_azure_monitor_logs`, to
`record_evidence_entry` so its output stops being silently dropped and becomes
a citeable line in the investigation report.

- `_map_query_azure_monitor_logs`: cites the row count and a bounded snippet
  of the KQL query executed.

This is the only tool in `integrations/azure/tools/` — it's a read-only
diagnostic query, so it's in scope per the issue.

**On count honesty — the "N+" convention:** `query_azure_monitor_logs`
truncates `rows` to `effective_limit` after the query runs (`rows[:effective_limit]`),
so a returned count at that ceiling may understate how many rows actually
matched. The tool didn't previously echo `effective_limit` in its output, so I
added it (the same fix I made for SigNoz's client in a companion PR earlier
this session, for the identical reason) so the mapper can compare the
returned count against the real ceiling instead of guessing.

The KQL query text is caller-built and unbounded, so it's collapsed and
capped at 80 chars before it goes into the summary — the same defensive
truncation used for free text elsewhere in this session's mappers.

The mapper records nothing when `available` is falsy or no rows were returned.

Removed `query_azure_monitor_logs` from `evidence_mapper_baseline.txt` now
that it carries a mapper.

```
print(bool(g('query_azure_monitor_logs').evidence_mapper))"
True
```

gather-evidence path), using realistic samples matching the tool's return
shape:**
```
query_azure_monitor_logs:  "2 row(s) for query 'AppTraces | order by TimeGenerated desc | take 50'"
query_azure_monitor_logs (saturated page):  "50+ row(s) for query 'AppTraces | take 50'"
```

**Tests and quality gates (updated after later review rounds):**
```
$ uv run pytest tests/tools/test_azure_monitor_logs_tool.py tests/tools/test_azure_monitor_logs_kql.py tests/tools/investigation/stages/gather_evidence/test_evidence_mapper_coverage.py -q
59 passed

```

**Post-review hardening (this PR went through several rounds — summarizing what changed and why, so a reviewer doesn't have to reconstruct it from commit history):**
- KQL defines `limit` as a synonym for `take` — both the safety-cap check and the mapper now recognize it.
- Facade extraction: mapper + KQL helpers moved out of `__init__.py` into `_evidence.py`/`_kql.py` sibling modules (Yauhen).
- A quoted or commented `take`/`limit`-shaped substring (e.g. `where Message contains "| take 5"`) could bypass detection in both directions — added `_kql.py`'s `mask_string_and_comment_text`, which blanks quoted-string contents and `//` comments before the row-cap regex runs (Greptile, 2 rounds).
- That masker didn't handle backslash-escaped quotes (`\"`) or KQL's verbatim strings (`@"..."`, which use doubled-quote escaping instead of backslash) — both could let the mask end early or run long and mis-scope a real take/limit stage (Greptile, 2 more rounds).
- `top N by ...` and `sample N` cap rows the same way `take`/`limit` do but weren't recognized — a query like `top 5 by TimeGenerated desc | take 50` was returning 5 rows (top's cap) while the mapper compared against `effective_limit=50`, silently citing a saturated result as exact (Davidson3556).
- Moved the caller-cap-vs-`effective_limit` reconciliation from the mapper into the tool itself: `query_azure_monitor_logs` now folds any detected row-cap into the `effective_limit` it returns, so the mapper only compares `total_returned` against `effective_limit` — no query-text parsing duplicated on the mapper's side.

**No live Azure account available in this environment**, so leaving the
existing `is_available`/`extract_params` wiring untouched — only the mapper,
the `effective_limit` field, and their tests were added.

I read `query_azure_monitor_logs`'s full `run()` before writing anything,
because the issue's generic template wouldn't have told me that `rows` is
truncated to a locally-computed `effective_limit` that the tool never echoed
back. Rather than re-derive that ceiling inside the mapper (duplicating
`_bounded_limit`'s clamp logic and risking drift if it changes), I added
`effective_limit` to the tool's own output — the same choice I made for
SigNoz's client earlier in this session when I hit the identical problem.

---

---

**Before** (main, no mapper):
```
BEFORE catalog_entries: None
```

**After** (this branch) — realistic sample matching the tool's actual return shape:
```json
[
  {
    "source": "query_azure_monitor_logs",
    "label": "Azure Monitor Logs",
    "summary": "12 row(s) for query 'AppTraces | where SeverityLevel >= 3 | order by TimeGenerated desc | take 50'"
  }
]
```

**1. What the reader gains.** The report can now cite the row count returned and the actual KQL query that produced it, with an "N+" qualifier when the page hit the tool's `effective_limit` — previously the query ran, cost tokens, and left nothing in the report.

**2. How much context it costs.** One short line per call; the query text is collapsed (newlines/CRs stripped) and truncated to 80 chars before it enters the summary, so a long or multi-line query never bloats the entry.

```
zero rows: None
error: None
```
No entry for `total_returned: 0` or an unavailable/error result.

**4. Whether anything is now recorded twice.** `query_azure_monitor_logs` is the only tool in `integrations/azure/tools/` and the only one with any evidence mapper — no overlap.

**5. Whether an existing report changed shape.** This only adds an entry under the new `query_azure_monitor_logs` source key; no other tool's merge path is touched.

## Live-report check (#5)
No live Azure account available in this environment, so I can't run `opensre investigate` against a real alert to confirm the entry lands in an actual generated report. Per the issue's own note for that case, the checks above (registry wiring, `merge_tool_evidence` output, before/after comparison) stand on their own — they exercise the real evidence path a live investigation would call.


arnavpopley:
@greptile review

arnavpopley:
@greptile review

arnavpopley:
Pushing back on this one after checking the actual risk direction and trigger scenario.

The remaining case is a `take N` string that appears **inside a quoted literal or a `//` comment, immediately preceded by a literal `|` character** — e.g. `AppTraces | where Message contains "| take 5"`. A regex genuinely cannot distinguish that from a real pipe stage without full KQL tokenization (tracking quote/comment state character-by-character), which is a materially larger change than an evidence-mapper wiring PR warrants.

Two things make this worth accepting rather than chasing further:

1. **The error is in the safe direction.** A false match here makes the mapper *under-claim* confidence — it shows `"5+"` when the true count is exactly 5, rather than the original bug this PR fixed, where an *exact* count was shown when the data was actually truncated (which asserts something false). Over-qualifying is a minor readability nit; under-qualifying was a correctness bug that could mislead an investigation into treating partial data as complete.
2. **The trigger requires a contrived, unusual query.** It needs a caller to write a `where`/`contains` clause whose literal string value itself contains a bare `|` character directly followed by the word "take" and a number — not a naturally-occurring KQL pattern. Real KQL string values almost never contain unescaped pipe characters, since `|` is the pipe-stage delimiter Kusto itself parses on.

Given the fix would require tokenizing quotes/comments to fully close (out of scope here) and the residual risk is a rare, safe-direction false positive rather than the original false-negative correctness bug, I'm leaving the current pipe-anchored regex as the fix for this PR. Happy to open a follow-up if maintainers want a real KQL-aware parser for this, but that feels disproportionate to what this specific evidence-mapper issue asked for.

integrations/azure/tools/azure_monitor_logs_tool/__init__.py
```diff
@@ -11,17 +12,67 @@
...
+#: operator -- isn't mistaken for one.
+_TAKE_CLAUSE_RE = re.compile(r"\|\s*take\s+(\d+)\b", re.IGNORECASE)
 
 
 def _bounded_limit(limit: int, max_results: int) -> int:
     safe_max = max(1, min(max_results, AZURE_MAX_RESULTS_HARD_LIMIT))
     return max(1, min(limit, safe_max))
 
 
+def _map_query_azure_monitor_logs(
```
YauhenBichel: This is facade file, should not contains methods implementation

YauhenBichel (changes requested):

integrations/azure/tools/azure_monitor_logs_tool/__init__.py
arnavpopley: Addressed — extracted the evidence mapper (and its take-clause helper) out of __init__.py into a new sibling _evidence.py module, so the facade file now only exposes the public tool interface.

arnavpopley:
@greptile review

arnavpopley:
@greptile review

YauhenBichel (approved):

YauhenBichel (changes requested):
Greptile 3/5

arnavpopley:
Fixed properly this time rather than pushing back again — the quoted/comment-text bypass has now shown up on both the \`take\` and \`limit\` forms of this check, so it's worth closing for good.

Added a small shared \`_kql.py\` helper that masks quoted-string contents and \`//\` line comments before running the take/limit pipe-stage regex. This closes the gap on **both** call sites that had it:
- \`_ensure_take_clause\` (the actual safety cap the tool applies before calling Azure) — previously a quoted \`"| take 5"\`-shaped filter value could suppress the real bound entirely, letting the query run unbounded against the customer's workspace.
- the evidence mapper's saturation check — same masking, so quoted text can no longer produce a false "N+" or false exact count.

New tests pin the exact case flagged (\`where Message contains "| take 5"\`) on both paths, plus a dedicated unit test file for the masking helper itself.

@greptile review

arnavpopley:
Fixed — the masker didn't account for backslash-escaped quotes inside a KQL string literal. \`\\"\` was being read as the string's closing quote, ending the mask early and exposing the rest of the literal (which can contain \`| take N\`) as real syntax.

Backslash-escaped characters are now masked as a pair so they can't close the string or be interpreted as real syntax. Added regression tests for both the escaped-quote case and the "trailing literal backslash still closes normally" case (\`\\\\"\`) to make sure the fix doesn't overcorrect.

@greptile review

arnavpopley:
Fixed — real distinction I'd missed: KQL verbatim strings (\`@"..."\`) don't use backslash escaping at all (backslash is a literal character there); only a doubled quote (\`""\`) escapes a literal quote inside one. Applying regular-string backslash-escaping to a verbatim string could extend the mask past its true closing quote, hiding a real \`take\`/\`limit\` stage that followed it — the dangerous direction (understating saturation, not overstating it).

The masker now tracks whether a string is verbatim (an \`@\` immediately before the opening quote) and switches escaping rules accordingly. Added regression tests for both the "verbatim backslash is literal" case and the "doubled quote escapes inside verbatim" case.

@greptile review

Davidson3556 (changes requested):
Arnav, one blocker.

`top` and `sample` cap rows the same way `take` does, and the detector doesn't match them, so a saturated result still gets cited as an exact count:

```
query:   AppTraces | top 5 by TimeGenerated desc | take 50
summary: 5 row(s) for query '...'      # saturated, should be 5+
```

Same class as the earlier P1s on this helper, and the next keyword will do it again. Could you drop `find_take_or_limit_values` from the mapper and compare `total_returned` against `effective_limit` alone? The tool knows its own ceiling and now echoes it, so parsing the query text back out has nothing left to add.

muddlebee:
image

above is missing

arnavpopley:
Good catch, thanks — real gap, and I want to flag one thing before the fix: your suggested simplification (drop \`find_take_or_limit_values\`, compare \`total_returned\` to \`effective_limit\` alone) wouldn't actually close your own example.

Walking through it: \`effective_limit\` is a static value the tool computes from \`limit\`/\`max_results\` *before* the query runs — it's never adjusted based on what the caller's query itself does. For \`AppTraces | top 5 by TimeGenerated desc | take 50\`, \`_ensure_take_clause\` already sees a row-cap clause present (once \`top\` is recognized) and leaves the query untouched, so Azure's \`top 5\` — not the trailing \`take 50\` — is what actually bounds the response to 5 rows. \`effective_limit\` stays 50 the whole time. Comparing \`total_returned (5) >= effective_limit (50)\` is still False, so it'd still get cited as an exact "5 row(s)", not "5+".

So I did both things: extended the row-cap regex to also recognize \`top N by ...\` and \`sample N\` (both bound rows the same way take/limit do), **and** moved the reconciliation into the tool itself — \`query_azure_monitor_logs\` now folds any detected caller row-cap into the \`effective_limit\` it returns, so the mapper genuinely only compares against \`effective_limit\` with no text-parsing duplicated on its side. That gets you the simplification you were after on the mapper, while still being correct for your exact case — added a regression test pinning \`top 5 by ... | take 50\` → \`effective_limit == 5\`, \`total_returned == 5\` on the tool's own output.

@greptile review

arnavpopley:
@Davidson3556 this crossed with a push — resolved as of cc2d38f (pushed the same day, a few hours after this review landed).

\`top N by ...\` and \`sample N\` are now recognized as row-cap clauses in \`_kql.py\` (\`_ROW_CAP_CLAUSE_RE\`), and the reconciliation moved into the tool itself: \`query_azure_monitor_logs\` now folds any detected caller row-cap into the \`effective_limit\` it returns, so the mapper just compares \`total_returned\` against \`effective_limit\` — \`find_take_or_limit_values\` is gone from the evidence mapper entirely, no query-text parsing left there.

Your exact scenario (\`top 5 by TimeGenerated desc | take 50\`) is pinned by \`test_detects_top_by_clause\` in \`tests/tools/test_azure_monitor_logs_kql.py\` and by the tool-level regression test at \`tests/tools/test_azure_monitor_logs_tool.py:193\`. Full suite (48 tests across both files) passes. Let me know if you'd like anything else adjusted.

Davidson3556:
> @Davidson3556 this crossed with a push — resolved as of [cc2d38f](https://github.com/Tracer-Cloud/opensre/commit/cc2d38f42208b2257f8dec66086e50c8cdabdfa5) (pushed the same day, a few hours after this review landed).
> 
> `top N by ...` and `sample N` are now recognized as row-cap clauses in `_kql.py` (`_ROW_CAP_CLAUSE_RE`), and the reconciliation moved into the tool itself: `query_azure_monitor_logs` now folds any detected caller row-cap into the `effective_limit` it returns, so the mapper just compares `total_returned` against `effective_limit` — `find_take_or_limit_values` is gone from the evidence mapper entirely, no query-text parsing left there.
> 
> Your exact scenario (`top 5 by TimeGenerated desc | take 50`) is pinned by `test_detects_top_by_clause` in `tests/tools/test_azure_monitor_logs_kql.py` and by the tool-level regression test at `tests/tools/test_azure_monitor_logs_tool.py:193`. Full suite (48 tests across both files) passes. Let me know if you'd like anything else adjusted.

okay noted , will review soon

PR #5793, opened 2026-08-26
feat(ec2): add evidence mappers for ec2_instances_by_tag and get_elb_target_health

### LINKED ISSUE #5565: Add evidence mappers for ec2
## Summary
can point to. The `ec2` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `ec2_instances_by_tag`
- `get_elb_target_health`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="ec2_instances_by_tag"' integrations/`.

```python

def _map_ec2_instances_by_tag(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="ec2_instances_by_tag",
        )

@tool(name="ec2_instances_by_tag", ..., evidence_mapper=_map_ec2_instances_by_tag)
```

## How to check it works

   under `tests/integrations/ec2/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['ec2_instances_by_tag', 'get_elb_target_health']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'ec2_instances_by_tag', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `ec2_instances_by_tag`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k ec2
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify ec2
# edit /tmp/alert.json so the alert names a resource ec2 actually knows about
grep -E 'ec2_instances_by_tag|get_elb_target_health' /tmp/rca.json
```

matching lines into the PR.

**No credentials for ec2?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5565

Wires both listed EC2 topology investigation tools to `record_evidence_entry`
the investigation report.

- `_map_ec2_instances_by_tag`: cites the active instance count, primary tier,
  and how many tiers were detected.
- `_map_get_elb_target_health`: cites the healthy/unhealthy target counts and
  target group count.

**On count honesty:** `ec2_instances_by_tag` already surfaces an explicit
`truncated` flag (a 5000-instance pagination safety cap, not an inferred
heuristic), so the mapper uses that directly — the same kind of explicit
signal I used for CloudTrail's `NextToken` earlier in this session, in
preference to a guessed page-size comparison.

**On the ELB tool's `available` semantics:** `get_elb_target_health` sets
`available=False` on *partial* coverage too (when one of several target
groups fails its `describe_target_health` call) — deliberately, per the
tool's own comment, so a caller never mistakes partial data for full
coverage. The mapper's standard `if not output.get("available"): return`
guard already handles this correctly with no extra logic, since the tool
itself already refuses to claim success on partial data.

**On zero counts:** both mappers always show `0 unhealthy` / `0 failed`-style
sub-counts when the parent list is non-empty, rather than omitting the clause
— a genuine zero within real data is a finding worth citing, not noise (the
same fix I'd just made for OpsGenie's open-alert count in a companion PR
earlier in this session, applied here from the start).

Removed both `ec2_instances_by_tag` and `get_elb_target_health` from

```
print({n: bool(g(n).evidence_mapper) for n in ['ec2_instances_by_tag', 'get_elb_target_health']})"
{'ec2_instances_by_tag': True, 'get_elb_target_health': True}
```

shape:**
```
ec2_instances_by_tag:   "5 active instance(s), primary tier 'web', 2 tiers"
get_elb_target_health:  "2 healthy, 1 unhealthy target(s) across 1 target group(s), states: draining"
```

**Tests and quality gates:**
```
11 passed

$ uv run pytest tests/tools/test_ec2_instances_by_tag_tool.py tests/tools/test_elb_target_health_tool.py -q
29 passed

$ uv run pytest tests/synthetic/rds_postgres/ tests/synthetic/rds_upstream/ -q
131 passed, 28 skipped

```

**No live AWS account available in this environment**, so leaving the
existing `is_available`/`extract_params`/synthetic-backend wiring untouched —
only the two mappers and their tests were added. I also ran the synthetic
`rds_postgres`/`rds_upstream` suites (which reference both tool names in
scenario fixtures) to confirm nothing downstream broke; both scenarios
that use these tools were already failing before this change on unrelated
gates (`category_match`, `exact_keyword_match`), so their pass/fail status is
unaffected by this PR.

I read both tools' full `run()` bodies in
`integrations/ec2/tools/ec2_instances_by_tag_tool/__init__.py` and
`integrations/elb/tools/elb_target_health_tool/__init__.py` before writing
anything, since the issue's generic template wouldn't have surfaced the
per-tool nuances: EC2's own `truncated` flag (from its 5000-instance safety
cap) and ELB's deliberate `available=False` on partial multi-target-group
coverage. Both tools also already compute an agent-friendly `summary` dict
via `integrations/aws/topology_helper.py` (`build_ec2_summary` /
`build_elb_summary`) — I reused those precomputed fields (`primary_tier`,
`target_group_count`, `unhealthy_states`) instead of re-deriving them from
the raw instance/target lists in the mapper.

---

---

### DIFF
--- a/integrations/ec2/tools/ec2_instances_by_tag_tool/__init__.py
+++ b/integrations/ec2/tools/ec2_instances_by_tag_tool/__init__.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 import logging
 from typing import Any, cast
 
+from core.domain.types.evidence import record_evidence_entry
 from core.tool_framework import tool
 from core.tool_framework.utils import tool_unavailable
 from integrations.aws.availability import ec2_available_or_backend
@@ -20,6 +21,36 @@ from integrations.aws.topology_helper import (
 )
 
 logger = logging.getLogger(__name__)
+
+
+def _map_ec2_instances_by_tag(
+    evidence: dict[str, Any], output: dict[str, Any], _tool_input: dict[str, Any]
+) -> None:
+    """Cite the active instance count, primary tier, and tier spread.
+
+    ``truncated`` reflects the tool's own 5000-instance safety cap on
+    pagination -- an explicit signal, not an inferred page-size heuristic.
+    """
+    if not output.get("available"):
+        return
+    instances = output.get("instances") or []
+    if not instances:
+        return
+    total = output.get("total_instances", len(instances))
+    count_label = f"{total}+" if output.get("truncated") else str(total)
+    parts = [f"{count_label} active instance(s)"]
+    primary_tier = (output.get("summary") or {}).get("primary_tier")
+    if primary_tier:
+        parts.append(f"primary tier '{primary_tier}'")
+    tiers_detected = output.get("tiers_detected") or []
+    if len(tiers_detected) > 1:
+        parts.append(f"{len(tiers_detected)} tiers")
+    record_evidence_entry(
+        evidence,
+        source="ec2_instances_by_tag",
+        label="EC2 Instances",
+        summary=", ".join(parts),
+    )
 
 
 def _is_available(sources: dict[str, dict]) -> bool:
@@ -88,6 +119,7 @@ def _summarize_instance(raw: dict[str, Any]) -> dict[str, Any]:
     },
     is_available=_is_available,
     extract_params=extract_ec2_instances_params,
+    evidence_mapper=_map_ec2_instances_by_tag,
 )
 def ec2_instances_by_tag(
     tier: str = "",
--- a/integrations/ec2/tools/ec2_instances_by_tag_tool/__init__.py
+++ b/integrations/ec2/tools/ec2_instances_by_tag_tool/__init__.py
@@ -10,7 +10,6 @@ from __future__ import annotations
 import logging
 from typing import Any, cast
 
-from core.domain.types.evidence import record_evidence_entry
 from core.tool_framework import tool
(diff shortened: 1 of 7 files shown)


integrations/ec2/tools/ec2_instances_by_tag_tool/__init__.py
```diff
@@ -22,6 +23,36 @@
 logger = logging.getLogger(__name__)
 
 
+def _map_ec2_instances_by_tag(
```
YauhenBichel: This is facade file, should not contains methods implementation

YauhenBichel (changes requested):

integrations/ec2/tools/ec2_instances_by_tag_tool/__init__.py
arnavpopley: Addressed — extracted map_ec2_instances_by_tag (and map_get_elb_target_health in the sibling ELB tool) out of their __init__.py facades into new _evidence.py modules.

YauhenBichel (approved):

PR #5794, opened 2026-08-26
feat(prefect): add evidence mappers for prefect_flow_runs and prefect_worker_health

### LINKED ISSUE #5571: Add evidence mappers for prefect
## Summary
can point to. The `prefect` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `prefect_flow_runs`
- `prefect_worker_health`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="prefect_flow_runs"' integrations/`.

```python

def _map_prefect_flow_runs(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="prefect_flow_runs",
        )

@tool(name="prefect_flow_runs", ..., evidence_mapper=_map_prefect_flow_runs)
```

## How to check it works

   under `tests/integrations/prefect/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['prefect_flow_runs', 'prefect_worker_health']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'prefect_flow_runs', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `prefect_flow_runs`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k prefect
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify prefect
# edit /tmp/alert.json so the alert names a resource prefect actually knows about
grep -E 'prefect_flow_runs|prefect_worker_health' /tmp/rca.json
```

matching lines into the PR.

**No credentials for prefect?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5571

Wires both listed Prefect investigation tools to `record_evidence_entry` so
their output stops being silently dropped and becomes a citeable line in the
investigation report.

- `_map_prefect_flow_runs`: cites the flow run count, failed-run count, and
  error-log-line count when logs were fetched for a specific run.
- `_map_prefect_worker_health`: cites the work pool count and unhealthy-pool
  count, plus worker health when a specific pool was inspected.

**On count honesty — the "N+" convention:** all three list endpoints Prefect
tools call (`get_flow_runs`, `get_work_pools`, `get_workers`) apply
`min(limit, 200)` as the API's own page-size param in
`integrations/prefect/client.py`, and none surfaces pagination metadata —
the identical ambiguity already solved for PagerDuty/OpsGenie earlier in
this session (see `validate_sentry_config` in
`integrations/sentry/__init__.py` for where this convention originates). I
added `_prefect_page_is_truncated`/the shared `_PREFECT_MAX_PAGE_SIZE = 200`
constant and used it in `prefect_flow_runs`'s mapper for both the total and
failed-run counts (the failed count inherits the same truncation flag as the
total, since on a capped page it can only be "failed within the returned
page" — the same fix I made for PagerDuty's active-incident count and
OpsGenie's open-alert count earlier in this session).

`prefect_worker_health`'s pool/worker counts are NOT qualified with "N+":
`pool_limit`/`worker_limit` default to 20, well under the 200 cap, and
unlike the flow-runs case there's no obviously-dominant failure mode driving
large result sets — I judged the added "N+" noise wasn't worth it here, but
flagging the reasoning in case reviewers disagree.

**On zero counts:** `0 failed` (flow runs) and `0 unhealthy` (pools/workers)
are always shown when the parent list is non-empty, rather than omitted — a
genuine zero within real data is a finding worth citing (the same fix I made
for OpsGenie's open-alert count earlier in this session, applied here from
the start).

Free-form `work_pool_name` (caller-supplied) is collapsed and capped at 60
chars before it goes into a summary.

Removed both `prefect_flow_runs` and `prefect_worker_health` from

```
print({n: bool(g(n).evidence_mapper) for n in ['prefect_flow_runs', 'prefect_worker_health']})"
{'prefect_flow_runs': True, 'prefect_worker_health': True}
```

shape:**
```
prefect_flow_runs:      '2 flow run(s), 1 failed'
prefect_worker_health:  "3 work pool(s), 2 unhealthy, 3 worker(s) in 'default', 2 unhealthy"
```

**Tests and quality gates:**
```
11 passed

$ uv run pytest tests/tools/test_prefect_flow_runs_tool.py tests/tools/test_prefect_worker_health_tool.py -q
65 passed

```

**No live Prefect account available in this environment**, so leaving the
existing `is_available`/`extract_params` wiring untouched — only the two
mappers and their tests were added.

I read `integrations/prefect/client.py`'s three list methods before writing
anything, and found the exact same `min(limit, 200)` page-cap shape I'd
already solved for PagerDuty and OpsGenie in this same session — so I reused
the established "N+" convention and helper naming pattern rather than
inventing a fourth variant. The one judgment call was whether to apply the
same qualifier to `prefect_worker_health`'s counts: its default limits (20)
sit well under the 200 cap and the tool is used more for a quick health
glance than an exhaustive listing, so I left those unqualified — noted above
for reviewer visibility rather than silently deciding.

---

---

### DIFF
--- a/integrations/prefect/tools/__init__.py
+++ b/integrations/prefect/tools/__init__.py
@@ -45,7 +45,9 @@ def _map_prefect_flow_runs(
     truncated = _prefect_page_is_truncated(total, tool_input.get("limit", 20))
     total_label = f"{total}+" if truncated else str(total)
     failed_count = output.get("total_failed", 0)
-    failed_label = f"{failed_count}+" if failed_count and truncated else str(failed_count)
+    # A truncated page's failed-count is only a floor even when it's zero --
+    # zero failed runs *in the returned page* does not mean zero overall.
+    failed_label = f"{failed_count}+" if truncated else str(failed_count)
     parts = [f"{total_label} flow run(s), {failed_label} failed"]
     if output.get("fetched_logs_for_run_id"):
         error_count = len(output.get("error_log_lines") or [])
--- a/integrations/prefect/tools/__init__.py
+++ b/integrations/prefect/tools/__init__.py
@@ -6,12 +6,56 @@ from __future__ import annotations
 
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.tool import BaseTool
 from core.tool_framework.utils import tool_unavailable
+from infrastructure.text.truncation import truncate
 from integrations.prefect.client import make_prefect_client
 
 _ERROR_KEYWORDS = ("error", "failed", "exception", "fatal", "crash", "traceback", "exitcode")
 _FAILED_STATES = {"FAILED", "CRASHED", "CANCELLED", "CANCELLING"}
+
+#: Bound the caller-supplied work pool name echoed into a report summary.
+_POOL_NAME_SUMMARY_TRUNCATE_LEN = 60
+
+#: Prefect's ``/flow_runs/filter``, ``/work_pools/filter``, and
+#: ``/work_pools/{name}/workers/filter`` endpoints all cap ``limit`` at 200
+#: server-side (``min(limit, 200)`` in ``integrations/prefect/client.py``),
+#: and none surfaces pagination metadata -- a returned count cannot be
+#: distinguished from a true total except by comparing it against the
+#: effective page size that was requested.
+_PREFECT_MAX_PAGE_SIZE = 200
+
+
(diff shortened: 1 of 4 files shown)


arnavpopley:
@greptile review

PR #5795, opened 2026-08-26
feat(airflow): add evidence mappers for airflow DAG/task investigation tools

### LINKED ISSUE #5558: Add evidence mappers for airflow

arnavpopley:
Closes #5558

Wires all three listed Airflow investigation tools to `record_evidence_entry`
the investigation report.

- `_map_get_recent_airflow_failures`: cites the count of failed/retrying task
  instances found for the DAG.
- `_map_get_airflow_dag_runs`: cites the DAG run count and how many are in a
  failed state.
- `_map_get_airflow_task_instances`: cites the task instance count and how
  many failed or are up for retry.

All three are read-only diagnostic queries — none is an action/notification
tool, so all are in scope per the issue. They live in
`integrations/tracer/tools/tracer_airflow_dag_tool/__init__.py`, not under
`integrations/airflow/` (the actual HTTP calls are in
`integrations/airflow/config.py`; the `@tool`-decorated wrappers are under
`integrations/tracer/tools/`) — worth flagging since the issue's generic
template names only point at the tool names, not the file.

**On error handling — these tools have no `available` flag:** unlike every
other tool I've wired to evidence in this session, none of the three sets an
`available`/success flag on their output — `{"error": "..."}` is the *only*
failure signal, and a request failure inside the `integrations/airflow/config.py`
fetch functions (auth, network, 5xx) is swallowed into an empty list rather
than surfaced (see `get_airflow_dag_runs`/`get_airflow_task_instances` in
that file: `if not isinstance(payload, dict): return []`). That means a
mapper cannot distinguish "genuinely zero results" from "the fetch silently
failed" from the tool's output alone. I did not try to fix that ambiguity
here (it's a client-layer design question outside this issue's scope) —
each mapper checks for `output.get("error")` and otherwise stays silent on
an empty list, the same conservative choice used for every other "empty
list → no citation" case in this session, so at worst a silent failure
produces no evidence rather than a false claim.

**On count honesty:** `get_airflow_dag_runs`'s `dag_runs` list is capped by
`min(limit, config.max_results)`, but the mapper only has the caller's own
`limit` (10 by default) from `tool_input`, not the resolved `config.max_results` —
so it compares the returned count against the caller's own requested `limit`,
the same approach I used for Honeycomb/Coralogix earlier in this session when
a hidden server-side ceiling wasn't visible to the mapper. `get_recent_airflow_failures`'s
`failures` count is an aggregate over multiple DAG runs (multiple task
instances can fail per run), not a value directly capped by any single
`limit`, so I did not apply the "N+" heuristic there.

Removed all three tool names from `evidence_mapper_baseline.txt` now that
they carry mappers.

```
  'get_recent_airflow_failures', 'get_airflow_dag_runs', 'get_airflow_task_instances']})"
{'get_recent_airflow_failures': True, 'get_airflow_dag_runs': True, 'get_airflow_task_instances': True}
```

shape:**
```
get_recent_airflow_failures:  "2 failed/retrying task instance(s) for DAG 'etl_daily'"
get_airflow_dag_runs:         "2 DAG run(s), 1 failed for 'etl_daily'"
get_airflow_task_instances:   "2 task instance(s), 1 failed/retrying for run 'run-123'"
```

**Tests and quality gates:**
```
11 passed

$ uv run pytest tests/tools/test_tracer_airflow_dag_tool.py -q
14 passed

$ uv run pytest tests/e2e/airflow/ -q
5 skipped   # requires live Airflow credentials, unaffected by this change

All checks passed! / 3645 files already formatted / Success: no issues found in 1899 source files
```

No existing unit test file covered this module directly (only a
credentials-gated `tests/e2e/airflow/test_orchestrator.py`, which I confirmed
still collects and skips cleanly), so I added
`tests/tools/test_tracer_airflow_dag_tool.py` focused on the new mappers.

I read all three tool wrappers and their backing
`integrations/airflow/config.py` fetch functions before writing anything,
because this integration's error-handling shape is different from every
other tool I've mapped this session — no `available` flag, and HTTP
failures silently degrade to an empty list rather than an explicit error.
That meant I couldn't reuse the `if not output.get("available"): return`
guard pattern verbatim; I used `output.get("error")` instead, and made the
same silent-on-empty choice as everywhere else rather than trying to
retrofit a "was this actually fetched" signal into the client layer, which
would have been a materially larger change than this issue asks for.

---

---

### DIFF
--- a/integrations/tracer/tools/tracer_airflow_dag_tool/__init__.py
+++ b/integrations/tracer/tools/tracer_airflow_dag_tool/__init__.py
@@ -5,10 +5,8 @@ from __future__ import annotations
 import os
 from typing import Any
 
-from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
 from core.tool_framework import tool
-from infrastructure.text.truncation import truncate
 from integrations.airflow.config import (
     AirflowConfig,
     build_airflow_config,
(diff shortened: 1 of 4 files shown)


arnavpopley:
@greptile review

YauhenBichel:
@arnavpopley init should be a facade, please refactor the code a bit

integrations/tracer/tools/tracer_airflow_dag_tool/__init__.py
```diff
@@ -21,6 +23,105 @@
...
+_ID_SUMMARY_TRUNCATE_LEN = 60
+_FAILED_DAG_RUN_STATES = frozenset({"failed"})
+_FAILED_TASK_STATES = frozenset({"failed", "upstream_failed", "up_for_retry"})
+
+
+def _safe_id(value: str) -> str:
+    return truncate(str(value).replace("\n", " "), _ID_SUMMARY_TRUNCATE_LEN)
+
+
+def _map_get_recent_airflow_failures(
```
YauhenBichel: it should be a facade, not a place for all implementations

YauhenBichel (changes requested):

integrations/tracer/tools/tracer_airflow_dag_tool/__init__.py
arnavpopley: Addressed — extracted all three mappers (map_get_recent_airflow_failures, map_get_airflow_dag_runs, map_get_airflow_task_instances) out of __init__.py into a new sibling _evidence.py module, so the facade file now only exposes the public tool interface.

arnavpopley:
@greptile review

YauhenBichel (approved):

PR #5799, opened 2026-08-26
feat(bitbucket): add evidence mappers for all 3 bitbucket investigation tools

### LINKED ISSUE #5559: Add evidence mappers for bitbucket
## Summary
can point to. The `bitbucket` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `get_bitbucket_file_contents`
- `list_bitbucket_commits`
- `search_bitbucket_code`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="get_bitbucket_file_contents"' integrations/`.

```python

def _map_get_bitbucket_file_contents(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="get_bitbucket_file_contents",
        )

@tool(name="get_bitbucket_file_contents", ..., evidence_mapper=_map_get_bitbucket_file_contents)
```

## How to check it works

   under `tests/integrations/bitbucket/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['get_bitbucket_file_contents', 'list_bitbucket_commits', 'search_bitbucket_code']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'get_bitbucket_file_contents', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `get_bitbucket_file_contents`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k bitbucket
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify bitbucket
# edit /tmp/alert.json so the alert names a resource bitbucket actually knows about
grep -E 'get_bitbucket_file_contents|list_bitbucket_commits|search_bitbucket_code' /tmp/rca.json
```

matching lines into the PR.

**No credentials for bitbucket?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5559

Wires all three listed Bitbucket investigation tools to `record_evidence_entry`
the investigation report.

- `_map_get_bitbucket_file_contents`: cites the fetched file's path,
  revision, and content length.
- `_map_list_bitbucket_commits`: cites the commit count and repo, qualifying
  page-capped totals.
- `_map_search_bitbucket_code`: cites the match count and search query,
  qualifying page-capped totals.

All three are read-only diagnostic queries — none is an action/notification

**On count honesty — the "N+" convention:** `list_commits` and `search_code`
in `integrations/bitbucket/client.py` both apply
`min(limit or config.max_results, config.max_results)` as the effective page
size, but neither echoed that ceiling back — the mapper only had the
caller's own requested `limit` via `tool_input`, and `config.max_results`
(which can be lower) isn't visible there. I added `effective_limit` to both
functions' output dicts (the same fix I made for Azure Monitor Logs,
OpenObserve, and SigNoz's client earlier in this session) rather than
guessing from `tool_input` alone, since a hidden `max_results` clamp could
otherwise cause a false negative on saturation detection.

`get_bitbucket_file_contents` already had a real `truncated` flag (the
client's own 10000-char content cap) — used directly, no heuristic needed.

Free-form search queries and file paths/refs are collapsed and capped before
going into a summary; `list_commits` and `search_code` already truncate
commit messages/hashes themselves, so no extra sanitization was needed there.

Removed all three bitbucket tool names from `evidence_mapper_baseline.txt`
now that they carry mappers.

```
  'get_bitbucket_file_contents', 'list_bitbucket_commits', 'search_bitbucket_code']})"
{'get_bitbucket_file_contents': True, 'list_bitbucket_commits': True, 'search_bitbucket_code': True}
```

shape:**
```
get_bitbucket_file_contents:  "14 char(s), from 'src/main.py', at 'main'"
list_bitbucket_commits:       "1 commit(s) for 'acme/backend-service'"
search_bitbucket_code:        "1 match(es) for 'error OR exception'"
```

**Tests and quality gates:**
```
11 passed

$ uv run pytest tests/tools/test_bitbucket_file_contents_tool.py tests/tools/test_bitbucket_commits_tool.py \
  tests/tools/test_bitbucket_search_code_tool.py tests/integrations/test_bitbucket.py -q
98 passed

```

**No live Bitbucket account available in this environment**, so leaving the
existing `is_available`/`extract_params` wiring untouched — only the three
mappers, the `effective_limit` additions to `client.py`, and their tests
were added.

I read all three tool wrappers plus their backing `integrations/bitbucket/client.py`
functions (`get_file_contents`, `list_commits`, `search_code`) before writing
anything. `get_file_contents` already exposed a real `truncated` flag so no
extra work was needed there, but `list_commits`/`search_code` had the same
"caller's limit isn't the real ceiling" gap I'd already hit and fixed for
Azure Monitor Logs, OpenObserve, and SigNoz earlier in this session — I
applied the identical fix (echo `effective_limit` from the client) rather
than inventing a fourth approach for the same underlying problem.

---

---

### DIFF
--- a/integrations/bitbucket/tools/bitbucket_commits_tool/__init__.py
+++ b/integrations/bitbucket/tools/bitbucket_commits_tool/__init__.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
 from core.tool_framework import tool
 from core.tool_framework.utils import tool_unavailable
@@ -13,6 +14,35 @@ from integrations.bitbucket.tools.bitbucket_search_code_tool import (
     _bb_creds,
     _resolve_config,
 )
+
+
+def _map_list_bitbucket_commits(
+    evidence: dict[str, Any], output: dict[str, Any], _tool_input: dict[str, Any]
+) -> None:
+    """Cite the commit count and repo, qualifying page-capped totals.
+
+    ``effective_limit`` is the real ceiling ``list_commits`` applied --
+    ``min(limit, config.max_results)`` -- so a returned count at that ceiling
+    may understate the true number of matching commits.
+    """
+    if not output.get("available"):
+        return
+    commits = output.get("commits") or []
+    if not commits:
+        return
+    total = output.get("total_returned", len(commits))
+    effective_limit = output.get("effective_limit", total)
+    count_label = f"{total}+" if total >= effective_limit else str(total)
+    summary = f"{count_label} commit(s)"
+    repo = output.get("repo")
+    if repo:
+        summary += f" for '{repo}'"
+    record_evidence_entry(
+        evidence,
+        source="list_bitbucket_commits",
+        label="Bitbucket Commits",
+        summary=summary,
+    )
 
 
 def _list_bitbucket_commits_extract_params(sources: dict[str, dict]) -> dict[str, Any]:
@@ -57,6 +87,7 @@ def _list_bitbucket_commits_available(sources: dict[str, dict]) -> bool:
     },
     is_available=_list_bitbucket_commits_available,
     extract_params=_list_bitbucket_commits_extract_params,
+    evidence_mapper=_map_list_bitbucket_commits,
 )
 def list_bitbucket_commits(
     repo_slug: str,
--- a/integrations/bitbucket/tools/bitbucket_commits_tool/__init__.py
+++ b/integrations/bitbucket/tools/bitbucket_commits_tool/__init__.py
@@ -4,45 +4,18 @@ from __future__ import annotations
 
 from typing import Any
 
-from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
 from core.tool_framework import tool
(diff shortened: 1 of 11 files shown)


arnavpopley:
@greptile review

integrations/bitbucket/tools/bitbucket_commits_tool/__init__.py
```diff
@@ -15,6 +16,35 @@
 )
 
 
+def _map_list_bitbucket_commits(
```
YauhenBichel: This is facade file, should not contains methods implementation

YauhenBichel (changes requested):

integrations/bitbucket/tools/bitbucket_commits_tool/__init__.py
arnavpopley: Addressed — extracted all three mappers (map_list_bitbucket_commits, map_get_bitbucket_file_contents, map_search_bitbucket_code) out of their __init__.py facades into new _evidence.py sibling modules for consistency.

YauhenBichel (approved):

PR #5800, opened 2026-08-26
feat(vercel): add evidence mappers for vercel_deployment_status and vercel_deployment_logs

### LINKED ISSUE #5575: Add evidence mappers for vercel
## Summary
can point to. The `vercel` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `vercel_deployment_logs`
- `vercel_deployment_status`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="vercel_deployment_logs"' integrations/`.

```python

def _map_vercel_deployment_logs(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="vercel_deployment_logs",
        )

@tool(name="vercel_deployment_logs", ..., evidence_mapper=_map_vercel_deployment_logs)
```

## How to check it works

   under `tests/integrations/vercel/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['vercel_deployment_logs', 'vercel_deployment_status']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'vercel_deployment_logs', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `vercel_deployment_logs`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k vercel
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify vercel
# edit /tmp/alert.json so the alert names a resource vercel actually knows about
grep -E 'vercel_deployment_logs|vercel_deployment_status' /tmp/rca.json
```

matching lines into the PR.

**No credentials for vercel?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5575

Wires both listed Vercel investigation tools to `record_evidence_entry` so
their output stops being silently dropped and becomes a citeable line in the
investigation report.

- `_map_vercel_deployment_status`: cites the deployment count and how many
  failed (ERROR/CANCELED), qualifying page-capped totals.
- `_map_vercel_deployment_logs`: cites the build event count, keyword-matched
  error-event count, and runtime log count.

**On count honesty — the "N+" convention:** `list_deployments` in
`integrations/vercel/client.py` applies `min(limit, 100)` as the API's own
page-size param and surfaces no pagination metadata beyond that — the
identical ambiguity already solved for PagerDuty/OpsGenie/Prefect earlier in
this session. I reused the same `_vercel_page_is_truncated` helper pattern
(`_VERCEL_MAX_PAGE_SIZE = 100`), including inheriting the truncation flag for
the failed-deployment sub-count (a capped page's "N failed" is a floor, not
an exact count — the same fix I made for PagerDuty's active-incident count
and OpsGenie's open-alert count).

**On zero counts:** `0 failed` is always shown when the deployment list is
non-empty, not omitted — a genuine zero is a finding worth citing (applying
that convention from the start here, having fixed it reactively for OpsGenie
after Greptile flagged it earlier in this session).

**On the logs mapper's error-keyword count:** phrased as "N matching an
error keyword" rather than "N errors", since `_ERROR_KEYWORDS` is a simple
substring match, not an authoritative Vercel error classification — the same
phrasing choice I made for Coralogix's identical keyword-matching pattern
earlier in this session.

Removed both `vercel_deployment_status` and `vercel_deployment_logs` from

```
print({n: bool(g(n).evidence_mapper) for n in ['vercel_deployment_status', 'vercel_deployment_logs']})"
{'vercel_deployment_status': True, 'vercel_deployment_logs': True}
```

shape:**
```
vercel_deployment_status:  '3 deployment(s), 2 failed'
vercel_deployment_logs:    '4 event(s), 2 matching an error keyword, 2 runtime log(s)'
```

**Tests and quality gates:**
```
11 passed

$ uv run pytest tests/tools/test_vercel_deployment_status_tool.py tests/tools/test_vercel_logs_tool.py -q
54 passed

```

**No live Vercel account available in this environment**, so leaving the
existing `is_available`/`extract_params` wiring untouched — only the two
mappers and their tests were added.

By this point in the session I'd already hit the `min(limit, N)`-page-cap
pattern for four other integrations (PagerDuty, OpsGenie, Prefect, Better
Stack), so I checked `list_deployments` for the same shape before writing
the mapper and confirmed it matched — same helper structure, same
"inherit the truncation flag for sub-counts" fix. The one genuinely new
decision was the logs mapper's phrasing for `error_events`: I'd made the
"don't call a keyword match an authoritative error count" call for Coralogix
earlier, and Vercel's `_ERROR_KEYWORDS` list is the same kind of heuristic
substring match, so I applied the identical phrasing rather than treating it
as a fresh judgment call.

---

---

### DIFF
--- a/integrations/vercel/tools/__init__.py
+++ b/integrations/vercel/tools/__init__.py
@@ -38,7 +38,9 @@ def _map_vercel_deployment_status(
     truncated = _vercel_page_is_truncated(total, tool_input.get("limit", 10))
     total_label = f"{total}+" if truncated else str(total)
     failed_count = len(output.get("failed_deployments") or [])
-    failed_label = f"{failed_count}+" if failed_count and truncated else str(failed_count)
+    # A truncated page's failed-count is only a floor even when it's zero --
+    # zero failures *in the returned page* does not mean zero overall.
+    failed_label = f"{failed_count}+" if truncated else str(failed_count)
     record_evidence_entry(
         evidence,
         source="vercel_deployment_status",
@@ -163,8 +165,20 @@ from core.tool import BaseTool
 _ERROR_KEYWORDS = ("error", "failed", "exception", "fatal", "crash", "panic", "unhandled")
 
 
+#: get_deployment_events/get_runtime_logs both cap limit at 2000 server-side
+#: and share the tool's single `limit` param, with no pagination metadata
+#: surfaced -- a returned count cannot be distinguished from a true total
+#: except by comparing it against the effective page size.
+_VERCEL_LOGS_MAX_PAGE_SIZE = 2000
+
+
+def _vercel_logs_page_is_truncated(returned_count: int, requested_limit: int) -> bool:
+    effective_limit = min(max(requested_limit, 1), _VERCEL_LOGS_MAX_PAGE_SIZE)
+    return returned_count >= effective_limit
+
+
 def _map_vercel_deployment_logs(
-    evidence: dict[str, Any], output: dict[str, Any], _tool_input: dict[str, Any]
+    evidence: dict[str, Any], output: dict[str, Any], tool_input: dict[str, Any]
 ) -> None:
     """Cite the build event count, keyword-matched error count, and runtime log count."""
     if not output.get("available"):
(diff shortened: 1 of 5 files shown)


arnavpopley:
@greptile review

integrations/vercel/tools/__init__.py
```diff
@@ -6,18 +6,55 @@
...
 _ERROR_STATES = {"ERROR", "CANCELED"}
 
+#: Vercel's /v6/deployments endpoint caps limit at 100 server-side
+#: (min(limit, 100) in integrations/vercel/client.py) and surfaces no
+#: pagination metadata -- a returned count cannot be distinguished from a
+#: true total except by comparing it against the effective page size.
+_VERCEL_MAX_PAGE_SIZE = 100
+
+
+def _vercel_page_is_truncated(returned_count: int, requested_limit: int) -> bool:
```
YauhenBichel: this is facade file, should contain only interface methods for public usage

YauhenBichel (changes requested):

integrations/vercel/tools/__init__.py
arnavpopley: Addressed — extracted both mappers (map_vercel_deployment_status, map_vercel_deployment_logs) and their page-truncation helpers into a new sibling _evidence.py module, so the flat tools/__init__.py now only exposes the public tool classes.

YauhenBichel (approved):

PR #5802, opened 2026-08-26
feat(interactive-shell): loading spinner phases and live tool-call preview

YauhenBichel:
Drives the prompt spinner from real agent events and shows what each tool
call is doing while it runs.

Loading spinner
- SpinnerState gains two phases — Thinking… and Invoking tools… — set by the
  action observer on llm_start / tool_start / tool_end, so the indicator
  reflects what the turn is actually doing.
- Spinner line redesign: a Ready idle hint, a one-row elapsed/token badge, and
  a "(Press ESC to stop)" hint (replacing the old "esc to cancel" tail).

Live tool-call preview
- Non-skill tool calls render a one-line preview (verb + payload) via
  tool_call_display; investigation_start keeps its own progress UI, so the
  generic preview doesn't duplicate it.

### DIFF
--- a/surfaces/interactive_shell/runtime/core/state.py
+++ b/surfaces/interactive_shell/runtime/core/state.py
@@ -16,7 +16,6 @@ from surfaces.shared.terminal.components.token_format import (
     _CHARS_PER_TOKEN,
     format_token_count_short,
 )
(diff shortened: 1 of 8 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5819, opened 2026-08-27
refactor(config): split config.py into llm_settings, clerk, environment

### LINKED ISSUE #5454: Split config/config.py by subject
## Problem

`config/config.py` is 679 lines holding three unrelated subjects. **Sixty-six
modules** import from it, so almost all of them pull in a file that mostly
concerns something they do not care about.

| Lines | Subject | Names |
| --- | --- | --- |
| 217–279 | environment + Clerk auth + JWT | `Environment`, `ClerkConfig`, `CLERK_CONFIG_DEV/PROD`, `CLERK_ISSUER_ENV`, `CLERK_JWKS_URL_ENV`, `get_clerk_config_override`, `get_environment`, `JWT_ALGORITHM`, `JWKS_CACHE_TTL_SECONDS` |
| 309–668 | LLM settings (~350 lines) | provider resolution, API-key env lookup, model tiers, `LLMSettings`, `LLMResolution`, `resolve_llm_settings`, `resolve_llm_settings_verbose`, `describe_llm_resolution`, `llm_provider_error_context`, `has_credentials_for_active_llm_provider` |
| 670–679 | Tracer URLs and Slack channel | `TRACER_BASE_URL_DEV/PROD`, `SLACK_CHANNEL`, `get_tracer_base_url` |

The LLM half is roughly half the file and the reason it is large.

## Why it matters — read this before starting

[AGENTS.md](AGENTS.md) records why the file is still like this:

> Do not define shared env names in `config/config.py`: it imports
> `config.llm_auth.*`, so a name it and one of those modules both need would
> force a cyclic import.

`config/` is a **true import leaf with zero upward edges**, and that property is
worth more than this split. `config.py` already imports `config.constants.llm`,
`config.llm_auth.auth_method`, `config.llm_auth.credentials`,
`config.llm_auth.provider_catalog`, `config.llm_models`, `config.local_env` and
`config.strict_config`.

A new module that imports `config.llm_auth` while `config.llm_auth` imports it
back reintroduces exactly the cycle that rule exists to prevent. And CodeQL
counts function-local and `TYPE_CHECKING` imports as part of a cycle, so making
an import lazy does **not** clear the alert.

Break any cycle you hit **structurally**: move the shared symbol into a leaf
module both sides import. Precedent for that shape:
`surfaces/cli/wizard/validation_result.py`.

## Steps

1. Record the names `config.config` exports today — see Verification.
2. Start with the LLM group. It already depends on `config.llm_auth.*` and
   `config.llm_models`, which the other two subjects do not touch at all, so it
   is the cleanest cut.
3. After creating each module, immediately check it imports standalone in a
   fresh interpreter (see Verification). Do this before moving on — a cycle
   caught after three moves is much harder to place.
4. Move the Clerk/JWT/environment group.
5. Decide where the three Tracer/Slack names go. They may be small enough to
   belong in `config/constants/` rather than a module of their own — check what
   is already there before adding a file.
6. Update all sixty-six importers. **No forwarding shim** — the house rule is
   one canonical import path after a refactor.

## Suggested layout

```
config/
  config.py       <- gone, or reduced to whatever genuinely has no other home
  llm_settings.py <- LLMSettings, LLMResolution, resolve_llm_settings*, ...
  clerk.py        <- Environment, ClerkConfig, get_environment, JWT constants
  constants/tracer.py  <- TRACER_BASE_URL_*, SLACK_CHANNEL  [check constants/ first]
```

Names are your call; check the existing `config/` layout before inventing one.

## Expected result

A module needing LLM settings imports `config.llm_settings` and gets ~350 lines
of LLM code, not Clerk config. `config/` still has zero upward imports and no
internal cycle.

## Verification

**Before** the change, record the exported vocabulary:

```bash
uv run python -c "
import config.config as m
print('\n'.join(sorted(n for n in dir(m) if not n.startswith('_'))))
" > /tmp/cf02-before.txt
```

**After**, produce the same union from the new modules and diff it. Every name
must still exist somewhere.

`config/` remains a leaf — no upward imports:

```bash
grep -rn "^from \(core\|surfaces\|gateway\|tools\|integrations\|infrastructure\|bootstrap\)" \
  --include='*.py' config/ | grep -v '\.venv'
```

Expected output: nothing.

No cycle inside `config/` — each new module must import standalone:

```bash
for m in llm_settings clerk; do
  uv run python -c "import config.$m" && echo "OK   config.$m" || echo "CYCLE config.$m"
done
```

Expected: one `OK` per module. Import each one **first** in a fresh interpreter;
a cycle can hide when another module has already been imported.

No shim left behind:

```bash
grep -rn "from config.config import" --include='*.py' . | grep -v '\.venv'
```

Expected output: nothing (or only lines for whatever genuinely stayed).

## Checklist

- [ ] Each new module covers one of the three subjects
- [ ] Every previously exported name is still importable from somewhere
- [ ] All sixty-six importers updated; no forwarding shim
- [ ] Leaf check prints nothing
- [ ] Each new module imports standalone in a fresh interpreter
- [ ] `git diff -M` shows moves, not rewrites

## Out of scope

`config/constants/` and its barrel — tracked separately. Also out of scope:
renaming `config/scope_context.py`, and deciding whether `config/grafana_cloud.py`
belongs under `constants/`. Both are worth doing when someone is already in
those files; neither justifies its own churn now.

## Claiming this issue

this issue.

AniketR10:
Fixes #5454

config/config.py held three unrelated subjects in 664 lines, so the 54 modules
importing it mostly pulled in code they did not care about. This splits it into
config/llm_settings.py (provider resolution and LLMSettings), config/environment.py
(Environment, get_environment), config/clerk.py (Clerk/JWT config) and
config/tracer_urls.py, moves TRACER_BASE_URL_* and SLACK_CHANNEL into the existing
config/constants/, and deletes config/config.py. All importers now point at the new
modules directly — no forwarding shim. Behaviour is unchanged; every name that
config.config exported is still importable from one of the new homes.

Pasted image (87)

https://github.com/user-attachments/assets/9d1552e1-1eb5-4e11-a245-e2a8ddf1a1c2

### Validation

- `make lint`
- `make format-check`
- `make typecheck`
- Focused pytest suite covering config, analytics, JWT auth, LLM construction/setup, auth CLI, and package smoke tests (`516 passed`)
- Constants facade regression test (`15 passed`)
- Standalone imports for `config.llm_settings`, `config.clerk`, `config.environment`, and `config.tracer_urls`
- Confirmed `config/` has no upward imports and no remaining `config.config` imports

---

Cut the LLM group first, since it is the only subject that depends on
config.llm_auth.* and config.llm_models — the other two touch neither, so it is a
clean edge. Used git mv so the diff renders as a move rather than a rewrite, then
deleted the non-LLM parts from it. After creating each module I checked it imports
standalone in a fresh interpreter before moving on, so a cycle would surface at the
step that caused it.

---

---


AniketR10:
@greptile review

config/constants/slack.py
```diff
@@ -7,6 +7,9 @@
 SLACK_ACCESS_TOKEN_ENV = "SLACK_ACCESS_TOKEN"
 SLACK_DEFAULT_CHAT_ID_ENV = "SLACK_DEFAULT_CHAT_ID"
 SLACK_WEBHOOK_URL_ENV = "SLACK_WEBHOOK_URL"
+
+# Default channel for RCA report delivery.
+SLACK_CHANNEL = "tracer-rca-report-alerts"
```
muddlebee: Please also import and list `SLACK_CHANNEL` in `config/constants/__init__.py`, and do the same for `TRACER_BASE_URL_DEV` and `TRACER_BASE_URL_PROD`. `AGENTS.md` requires domain constants to be re-exported through the shared facade; the follow-up commit fixed this for Clerk but left these three new constants out.

muddlebee (changes requested):
The module split otherwise preserves the old implementations and the changed files no longer import `config.config`; all six workflow runs on the current head are green. One repository-convention issue remains: the new Slack and Tracer constants need to be re-exported from the shared `config.constants` facade, just as the follow-up commit already did for the Clerk names.

muddlebee (approved):
Reviewed the merged head 2a94e9293 after resolving the two conflicts against current main. The config split preserves behavior, all former config.config consumers use canonical owning modules, shared constants are re-exported through config.constants, and the config package remains cycle-free. Local lint, formatting, typecheck, and 516 focused tests pass; no blocking findings remain.

muddlebee:
@greptile review

config/llm_settings.py
```diff
@@ -31,248 +33,77 @@
...
     BEDROCK_TOOLCALL_MODEL,
     CUSTOM_ANTHROPIC_CLASSIFICATION_MODEL,
-    CUSTOM_ANTHROPIC_LLM_CONFIG,
     CUSTOM_ANTHROPIC_REASONING_MODEL,
     CUSTOM_ANTHROPIC_TOOLCALL_MODEL,
     CUSTOM_OPENAI_CLASSIFICATION_MODEL,
-    CUSTOM_OPENAI_LLM_CONFIG,
     CUSTOM_OPENAI_REASONING_MODEL,
     CUSTOM_OPENAI_TOOLCALL_MODEL,
-    DEEPSEEK_BASE_URL,
```
muddlebee, line 53: this doesnt make sense tbh, if we are removing LLM_CONFIG and keeping the other related config here.. all LLM related configs like model, API keys are also part of LLM right.. we should re-approach this.

config/llm_settings.py
```diff
@@ -31,248 +33,77 @@
...
     "get_llm_provider_api_key_env",
-    "get_tracer_base_url",
     "has_credentials_for_active_llm_provider",
     "llm_provider_error_context",
     "resolve_llm_settings",
     "resolve_llm_settings_verbose",
 )
 
 
-class Environment(StrEnum):
```
muddlebee, line 213: IMO only moving out clerk makes sense and the LLM config as stated above still feeels doubtful

Issue #4605, opened 2026-07-31
[FEATURE] Yandex Cloud support: investigation data source and AI Studio LLM provider


nowhere-in-space:
### Problem statement

OpenSRE ships 79 integration packages and 8 OpenAI-compatible LLM providers. None of them cover Yandex Cloud, which is the dominant public cloud in Russia and much of the CIS and the mandated platform for a large share of regulated workloads there.

That leaves two independent gaps, needed to be done for Yandex Cloud support:

**No data source.** An investigation on Yandex Cloud infrastructure has nothing to read. No metrics from Monitoring, no logs from Cloud Logging, no audit trail, no state from Compute, Managed Kubernetes, or the managed databases. The agent can reason about the text of the alert and nothing else — every tool call it would want to make has no backend. The generic `execute_python_code` path does not close this: there is no configured credential, no endpoint resolution, and no read-only guarantee.

**No LLM provider.** Investigations put production logs into prompts. Data-residency rules routinely prevent sending those to Anthropic or OpenAI, which rules out the default providers regardless of budget. Yandex AI Studio is the compliant option in that jurisdiction and is not among them, so there is no supported way to run the agent at all.

The second gap blocks users who could otherwise tolerate the first, so neither is useful alone.

Add Yandex Cloud the way the existing clouds are added: one LLM provider plus an umbrella integration with per-service tool packages.

### Proposed solution

I have this working already, in a (private gitlab) fork, validated against live infrastructure. I would like to contribute it upstream as a reviewable series rather than keep it out of tree. Asking for assignment before opening anything, per `CONTRIBUTING.md`.

### 1. LLM provider `yandex`

Yandex AI Studio exposes an OpenAI-compatible endpoint at `https://ai.api.cloud.yandex.net/v1` with tool calling, `parallel_tool_calls`, `tool_choice`, `response_format: json_schema`, and streaming — everything the ReAct loop needs, so it slots into `OPENAI_COMPATIBLE_PROVIDERS` with no new transport.

```bash
OPENSRE_LLM_PROVIDER=yandex
YANDEX_API_KEY=...          # or an IAM token, or the instance metadata service
YANDEX_FOLDER_ID=b1g...     # only when models are given as bare names
```

Two details that need care and are easy to get wrong:

- **Model identifiers are URIs**, `gpt://<folder>/<model>/<branch>`. Bare names are normalized against `YANDEX_FOLDER_ID`; full URIs pass through.
[...]

PR #5828, opened 2026-08-27
feat(yandex-cloud): Managed Databases tools and Monitoring alert routing

nowhere-in-space:
Part of #4605

Managed Databases for the Yandex Cloud integration: cluster health, host roles, recent operations, and the log a cluster keeps for itself. Three tools, which takes the family from nine to twelve against the seventeen the budget test defends.

Four smaller things came with it, and three of the four were found by running this against a real folder rather than by reading the diff: alerts from Yandex Monitoring had no way to be recognised as such, metric discovery could not tell a caller what a label holds, Greenplum reported no host roles at all, and deliberately stopped clusters were being counted as failures. Each is described in its own paragraph below.

**Why these three and not more.** Every managed-database path is already in the endpoint index, so `execute_yc_operation` can reach all of it. A curated tool has to earn its schema slot:

- **`list_yc_db_clusters`** - "what databases run here" is otherwise fifteen calls against fifteen service names the model has to know first. It also absorbs Yandex's renames: Redis is Valkey now, MongoDB is StoreDoc, Greenplum is MPP Analytics, and the API paths kept the old spelling. Nine engines, including Sharded PostgreSQL, which is a service of its own rather than a PostgreSQL mode - its own API prefix, its own cluster type, and a router in front of the shards.
- **`get_yc_db_cluster`** - merges the cluster, its hosts, and its recent operations into one read, because "was there a failover" and "which host is master now" are the same question asked twice. It also returns how to reach the data plane, and that part is in no API at all: PostgreSQL answers on **6432** through a pooler, and an operator told 5432 concludes the database is down. Managed PostgreSQL has no performance-diagnostics endpoint, so the query that is eating the CPU is only visible from inside the cluster - the hint is the next step, not a convenience.
- **`read_yc_db_logs`** - a cluster's own log is a separate store from Cloud Logging and reaches it only if export was switched on, so an empty `read_yc_logs` says nothing about it. Each engine keeps several streams and serves one at a time: MySQL answers with its error log unless `MYSQL_SLOW_QUERY` is asked for, so a question about slow queries otherwise gets a confident empty answer.

Deliberately not added: backups, users, databases, extensions, shards, topics. All reachable through the generic reader, none of them the question an incident starts with.

**Hosts are read where each engine keeps them.** Greenplum has no `hosts` collection at all - its hosts are split into `master-hosts` and `segment-hosts` - so a single hardcoded path answers 404 there, and the cluster comes back with an empty host list and a connection hint with no address in it. Nothing errors: "no hosts" and "the host read failed" are indistinguishable from the outside, and they lead an investigation to opposite conclusions. Both collections are read now, a dead segment reaches `unhealthy_hosts`, the hint points at the master, and a failed read is reported in `hosts_error` rather than looking like an empty cluster.

**Host roles are read per engine, because engines disagree on the field.** Most say `role`, MongoDB-shaped ones say `type`, and Greenplum says neither - it splits its hosts into `master-hosts` and `segment-hosts` and lets the URL carry the meaning. Against a live cluster that showed up as four hosts with no roles at all, and the connection hint landing on the master only because that collection happens to be read first: a failed `master-hosts` read would have silently promoted a segment. The role now falls back through `role`, `type`, and the collection, in that order.

**A stopped cluster is not a broken one.** Both end in "the database is unreachable", but one has a cause worth investigating and the other was switched off on purpose. They are listed apart - `unhealthy` and `stopped` - and neither is hidden, since "somebody stopped it" is frequently the whole answer.

**A tool that now exists had to stop being described as missing.** The constant attached to every empty Cloud Logging read told the model that database logs were unreadable and to fall back to metrics. That was true last week. The guidance test that guards this only checked one direction - that nothing names a tool this tree does not ship - so it caught `read_yc_db_logs` appearing in SKILL.md but would never have caught the stale sentence. Added the mirror assertion: a family that ships must be named as the way to reach it.

**Alert routing, and something that had to come with it.** `yandex_monitoring` routes to `yandex_cloud`, and the keyword aliases are deliberately narrow. `valkey` and `managed postgresql` were candidates and are left out: Valkey is a product anyone can run, and "managed postgresql" says nothing about whose cloud it is. What is claimed is only what Yandex alone says - the `mdb.yandexcloud` hostname an error quotes, and `storedoc`. This follows the restraint the catalog already applies to `eks`, which is given `eks` and not `pod`.

That routing entry was unreachable as written: nothing sets `alert_source = "yandex_monitoring"`. The intake prompt lists Grafana, Datadog, CloudWatch and the rest, and Yandex is not among them, so the table would have been dead code from the CLI path. Added a detector alongside the Grafana one, using the mechanism already there for exactly this. Yandex Monitoring has no plain webhook - alerts leave by email, SMS, Telegram or a Cloud Function - so there is no schema to match and detection keys off markers only a Yandex payload carries. It requires an identifier *and* something alert-shaped, so a forwarded tool result that merely mentions a folder is not mistaken for a firing alert; a `*.yandexcloud.net` address is accepted on its own, since nothing else produces one.

**Metric discovery now shows what a label actually holds.** Found by running an investigation against the live folder rather than by reading code. Asked for CPU on a managed cluster, the agent built `cpu.usage_user{cluster_id="..."}`, got no series, and concluded - confidently and wrongly - that metric collection was switched off. Both halves of the query were wrong and neither could have been got right: managed databases publish `cpu.idle` and friends rather than Compute's `cpu_usage`, and they key on `resource_id` rather than `cluster_id`. The agent picked `cluster_id` because `list_yc_metrics` reported it, which it does - the label-key list is folder-wide and holds every key any service uses.

So the tool now returns a bounded sample of real series with their labels alongside the names. The client method for it already existed and was not exposed. Nothing new is registered, and the schema budget is untouched. This is the same failure as the two above: matching nothing and failing are indistinguishable from the outside, and here it produced a recommendation to go and enable an export that was never disabled.

**The endpoint index is regenerated**, 936 endpoints across 68 services to 951 across 69. Mostly this fixes rather than adds: five `baremetal` paths were reshaped upstream from `.../servers/{id}:listOperations` to `.../servers/{id}/operations`, and the index still listed the old form, so the gate passed a path the cloud answers with a 404. The new service is `managed-zonalshift`, which answers a first-minute incident question - is traffic being shifted out of a zone right now, or is a drill running.

Both new gates that landed since the last Yandex Cloud PR are satisfied properly rather than by baseline entry: each tool declares an evidence mapper, so a degraded cluster and its latest operation become citeable evidence, and each is classified in the telemetry test.

Read against a live folder from inside a Yandex Cloud VM - seven clusters across five engines, every call a GET.

https://github.com/user-attachments/assets/20493925-a078-41f5-b96c-72337c6f32f3

The composition was decided by asking what the generic reader already does well, rather than by porting whatever existed. Everything here is reachable through `execute_yc_operation`, so the question for each tool was what it knows that the index does not. Three things survived that test: which engines exist and what they used to be called, how to reach the data plane, and which log stream answers which question. An operations tool was on the original list and was dropped - recent operations belong inside the cluster read, since "what changed" and "what is the state" are one question.

The engine table carries a `service` key in the `managed-` form rather than `mdb-`, which looks arbitrary until you check the endpoint registry: it carries `mdb-` for only some engines, and a missing key means the read is refused before it is sent.

`list_yc_db_clusters` treats one engine failing as a note rather than an error. Permissions in Yandex Cloud are commonly granted per service, and most folders run one or two engines, so "no clusters" is the usual truthful answer - failing the whole call because MongoDB was not authorised would hide the PostgreSQL cluster the question was about. The tool reports unavailable only when no engine could be reached at all.

The alert aliases were the place I most had to argue myself down. The obvious move is to claim the engine names, and it is wrong: `postgresql` and the other generic terms belong to the data-plane integrations that answer those alerts, the same way `pod` belongs to `kubernetes` and not to `eks`. Matching is substring, so a short or common alias makes every Yandex tool look relevant to every alert and displaces the ones that would have answered. The demo asserts both directions, including the two alerts that must *not* match.

Edge cases covered by tests: a folder with no clusters at all, one engine failing while others answer, an unknown engine name, an unknown log stream (the error names the real ones), an engine that takes no stream parameter, and every engine resolving to a service key the endpoint registry actually carries.

---


nowhere-in-space:
@greptile review

nowhere-in-space:
@greptile review

hertiagesoftail37-sys (approved):

YauhenBichel:
Approved

YauhenBichel (approved):

Davidson3556:
> Approved

thanks for the review

PR #5833, opened 2026-08-27
feat(planning): denser plan diagnoses and a keyboard next-step at turn end

YauhenBichel:
## What

Two prompt-level improvements to the planning/turn experience, from a head-to-head comparison against a reference coding agent.

### Denser plan diagnoses

The plan checklist steps are intentionally terse; the real reasoning belongs in the explanation block rendered under the checklist (Facts / what-it-rules-out / hypothesis table). The prompt described those sections but showed no example, so
models wrote thin one-liners. Adds a worked exemplar and requires specificity — every fact names a signal, every hypothesis carries a discriminator — so the diagnosis under the checklist is substantive. A vague one-liner is now an
explicit bad-plan.

### Keyboard next-step instead of a prose question

The closing "want me to…?" next step was prose the user had to type a reply to. The prompt already mandates the `ask_user_choice` arrow-key menu for structured choices; this extends that rule to the end-of-turn next step, so every turn ends with a selectable next step (a first option that does it, plus a decline) rather than an open-ended conversational question.

## Why

Against the reference agent, opensre's plans were terser and its turns ended with prose questions. Density and a press-to-continue next step close that gap without abandoning the clean checklist.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5837, opened 2026-08-27
feat(groundcover): add evidence mappers for logs, traces, and query reference tools

### LINKED ISSUE #5560: Add evidence mappers for groundcover
## Summary
can point to. The `groundcover` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `get_groundcover_query_reference`
- `query_groundcover_logs`
- `query_groundcover_traces`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="get_groundcover_query_reference"' integrations/`.

```python

def _map_get_groundcover_query_reference(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="get_groundcover_query_reference",
        )

@tool(name="get_groundcover_query_reference", ..., evidence_mapper=_map_get_groundcover_query_reference)
```

## How to check it works

   under `tests/integrations/groundcover/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['get_groundcover_query_reference', 'query_groundcover_logs', 'query_groundcover_traces']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'get_groundcover_query_reference', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `get_groundcover_query_reference`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k groundcover
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify groundcover
# edit /tmp/alert.json so the alert names a resource groundcover actually knows about
grep -E 'get_groundcover_query_reference|query_groundcover_logs|query_groundcover_traces' /tmp/rca.json
```

matching lines into the PR.

**No credentials for groundcover?** That is a normal outcome, not a reason to go

arnavpopley:
## Summary
Wires `query_groundcover_logs`, `query_groundcover_traces`, and `get_groundcover_query_reference` into `record_evidence_entry` so their output becomes citeable investigation-report evidence, per #5560 (part of #5519 / #1079).

- Logs/traces mappers cite the returned row/span count and gcQL query, using the tools' own explicit `truncated` signal (a "truncat" note or `compact_rows`' cap from `build_envelope`) for the "N+" qualifier — not an inferred page-size heuristic.
- Query reference mapper cites the retrieved reference length and whether it was served from cache.
- Mappers live in a new `integrations/groundcover/tools/_evidence.py` sibling module rather than inline in the flat `tools/__init__.py` facade.

## Test plan
- [x] `uv run pytest tests/tools/test_groundcover_tools.py tests/tools/investigation/stages/gather_evidence/` — all pass
- [x] `make lint` / `make format-check` / `make typecheck` — clean
- [x] Removed all three tools from `evidence_mapper_baseline.txt`
- [x] `uv run python -c "from tools.registry import get_registered_tool as g; print({n: bool(g(n).evidence_mapper) for n in ['query_groundcover_logs', 'query_groundcover_traces', 'get_groundcover_query_reference']})"` → all `True`

**Before** (main, no mapper) — `merge_tool_evidence` on a realistic logs sample:
```
BEFORE catalog_entries: None
```

**After** (this branch) — realistic samples for all three tools:
```json
[
  {
    "source": "query_groundcover_logs",
    "label": "groundcover Logs",
    "summary": "2 log(s) for query 'level:error | fields _time, workload, content | limit 50'"
  },
  {
    "source": "query_groundcover_traces",
    "label": "groundcover Traces",
    "summary": "1 span(s) for query 'status:error | fields _time, span_name | limit 50'"
  },
  {
    "source": "get_groundcover_query_reference",
    "label": "groundcover Query Reference",
    "summary": "35 char(s) of gcQL reference retrieved (cached)"
  }
]
```

**1. What the reader gains.** The report can now cite the actual gcQL query run and how many logs/spans it returned (e.g. "2 log(s) for query 'level:error | ...'"), plus whether the query-reference guide was consulted before a query was built — previously all three tools ran, cost tokens, and left no trace.

**2. How much context it costs.** One short line per call; the query text embedded in the summary is truncated to 80 chars (`_QUERY_SUMMARY_TRUNCATE_LEN`), so a long or multi-line gcQL query never bloats the entry.

```
empty logs: None
error traces: None
empty reference: None
```
No entry for an empty `data` list, an unavailable/error result, or an empty reference string.

**4. Whether anything is now recorded twice.** No other mapper in the codebase touches groundcover output — these are the only three groundcover tools and none had a mapper before this PR.

**5. Whether an existing report changed shape.** This only adds entries under the three new `query_groundcover_logs`/`query_groundcover_traces`/`get_groundcover_query_reference` source keys; no other tool's merge path is touched.

## Live-report check (#5)
I don't have a groundcover connection configured, so I can't run `opensre investigate` against a real alert to confirm the entry lands in an actual generated report. Per the issue's own note for that case, checks 1–4 above (plus the behaviour comparison) stand on their own — they exercise the real `merge_tool_evidence` path a live investigation would call, just without an actual groundcover connection.

Closes #5560

🧙 Built with [WOZCODE](https://wozcode.com)


arnavpopley:
@greptile review

PR #5838, opened 2026-08-27
feat(posthog_mcp): add evidence mappers for list_posthog_tools and call_posthog_tool

### LINKED ISSUE #5570: Add evidence mappers for posthog_mcp
## Summary
can point to. The `posthog_mcp` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `call_posthog_tool`
- `list_posthog_tools`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="call_posthog_tool"' integrations/`.

```python

def _map_call_posthog_tool(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="call_posthog_tool",
        )

@tool(name="call_posthog_tool", ..., evidence_mapper=_map_call_posthog_tool)
```

## How to check it works

   under `tests/integrations/posthog_mcp/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['call_posthog_tool', 'list_posthog_tools']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'call_posthog_tool', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `call_posthog_tool`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k posthog_mcp
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify posthog_mcp
# edit /tmp/alert.json so the alert names a resource posthog_mcp actually knows about
grep -E 'call_posthog_tool|list_posthog_tools' /tmp/rca.json
```

matching lines into the PR.

**No credentials for posthog_mcp?** That is a normal outcome, not a reason to go

arnavpopley:
## Summary
Wires `list_posthog_tools` and `call_posthog_tool` into `record_evidence_entry` so their output becomes citeable investigation-report evidence, per #5570 (part of #5519 / #1079).

- `list_posthog_tools` cites the matched tool count, qualified "+" using `build_mcp_tool_listing`'s own explicit truncation signal (`returned_tools < matched_tools`) rather than an inferred heuristic.
- `call_posthog_tool` is a generic dispatcher that can be invoked many times per investigation with different underlying PostHog MCP tools/arguments (HogQL queries, feature-flag lookups, etc). Since `record_evidence_entry` lets the first entry for a given `source` win, reusing one source key would silently drop every call after the first — so each call gets a disambiguated source (`call_posthog_tool:<tool_name>`, then `#2`, `#3`, ... on repeat calls to the same underlying tool). `tool_name` itself is caller/upstream-controlled, so it's bounded to 60 chars before it flows into the summary, label, or source key (Greptile P2 finding, fixed).
- Mappers live in a new `_evidence.py` sibling module rather than inline in `__init__.py` (Yauhen review feedback, fixed).

## Test plan
- [x] `uv run pytest tests/tools/test_posthog_mcp_tool.py tests/tools/investigation/stages/gather_evidence/` — all pass
- [x] `make lint` / `make format-check` / `make typecheck` — clean
- [x] Removed both tools from `evidence_mapper_baseline.txt`
- [x] `uv run python -c "from tools.registry import get_registered_tool as g; print({n: bool(g(n).evidence_mapper) for n in ['call_posthog_tool', 'list_posthog_tools']})"` → `{'call_posthog_tool': True, 'list_posthog_tools': True}`

**Before** (main, no mapper) — `merge_tool_evidence` on realistic sample output for both tools:
```
BEFORE catalog_entries: null
```

**After** (this branch) — same sample output:
```json
[
  {
    "source": "list_posthog_tools",
    "label": "PostHog MCP Tools",
    "summary": "12 tool(s) listed matching 'events query sql'",
    "url": null,
    "snippet": null
  },
  {
    "source": "call_posthog_tool:execute-sql",
    "label": "PostHog MCP: execute-sql",
    "summary": "2 row(s) (tool: execute-sql)",
    "url": null,
    "snippet": null
  }
]
```

**1. What the reader gains.** The report can now cite that a specific PostHog MCP tool ran and what it returned (e.g. "2 row(s) (tool: execute-sql)" for a HogQL query, or "12 tool(s) listed matching 'events query sql'" for a discovery call) — previously this was invisible: the tool ran, cost tokens, but left no trace in the report.

**2. How much context it costs.** One entry is a single short line, ~40–90 characters depending on tool name and result shape (bounded: summary text is truncated to 120 chars, tool name to 60 chars). `call_posthog_tool`'s summary never inlines full row data or raw response text beyond that bound — it cites a count or a bounded snippet, not a dump.

```
empty listing entries: None
error call entries: None
bare {} entries: None
```
No entry is recorded for `{"available": True, "returned_tools": 0}`, `{"available": False, "error": ...}`, or a bare `{}` — confirmed no "0 items" noise.

**4. Whether anything is now recorded twice.** No other mapper in the codebase covers posthog_mcp output (`grep -rln "evidence_mapper" integrations/posthog_mcp/` only matches this file) — no overlap.

**5. Whether an existing report changed shape.** This only adds new entries under the new `call_posthog_tool*`/`list_posthog_tools` source keys; it doesn't touch how any other tool's output is merged, so pre-existing evidence entries from other tools are unaffected in an investigation that already worked before this change.

## Live-report check (#5)
I don't have PostHog MCP credentials configured, so I can't run `opensre investigate` against a real alert to confirm the entry lands in an actual generated report — per the issue's own note ("No credentials? That is a normal outcome... leave it unticked"), I'm leaving that box unticked. Checks 1–4 above (plus the behaviour comparison) all pass and demonstrate the mapper is wired and produces the expected entry via the real `merge_tool_evidence` path a live investigation would call.

Closes #5570

🧙 Built with [WOZCODE](https://wozcode.com)

### DIFF
--- a/integrations/posthog_mcp/tools/posthog_mcp_tool/__init__.py
+++ b/integrations/posthog_mcp/tools/posthog_mcp_tool/__init__.py
@@ -33,6 +33,10 @@ from integrations.posthog_mcp import (
     posthog_mcp_config_from_env,
     posthog_mcp_runtime_unavailable_reason,
 )
+from integrations.posthog_mcp.tools.posthog_mcp_tool._evidence import (
+    map_call_posthog_tool,
+    map_list_posthog_tools,
+)
 
 PostHogMCPParams = dict[str, object]
 PostHogMCPResponse = dict[str, object]
@@ -350,6 +354,7 @@ def _normalize_tool_result(result: PostHogMCPToolCallResult) -> PostHogMCPRespon
     ),
     is_available=_posthog_mcp_available,
     extract_params=_posthog_mcp_extract_params,
+    evidence_mapper=map_list_posthog_tools,
 )
 def list_posthog_tools(
     name_filter: str | None = None,
@@ -466,6 +471,7 @@ def list_posthog_tools(
     ),
     is_available=_posthog_mcp_available,
     extract_params=_posthog_mcp_extract_params,
+    evidence_mapper=map_call_posthog_tool,
 )
 def call_posthog_tool(
     tool_name: str | None = None,
--- a/integrations/posthog_mcp/tools/posthog_mcp_tool/_evidence.py
+++ b/integrations/posthog_mcp/tools/posthog_mcp_tool/_evidence.py
@@ -11,6 +11,14 @@ from infrastructure.text.truncation import truncate
 #: both are caller/upstream-supplied and unbounded.
 _SUMMARY_TRUNCATE_LEN = 120
 
+#: Bound the dispatched MCP tool name -- caller-controlled and unbounded --
+#: before it flows into the summary, label, and evidence source key.
+_TOOL_NAME_TRUNCATE_LEN = 60
+
+
+def _safe_tool_name(tool_name: str) -> str:
+    return truncate(tool_name.replace("\n", " "), _TOOL_NAME_TRUNCATE_LEN)
+
 
 def _next_unique_source(evidence: dict[str, Any], base: str) -> str:
     """Disambiguate repeat calls to a generic dispatcher tool.
@@ -73,7 +81,8 @@ def map_call_posthog_tool(
     """
     if not output.get("available"):
         return
-    tool_name = str(output.get("tool") or tool_input.get("tool_name") or "").strip()
+    raw_tool_name = str(output.get("tool") or tool_input.get("tool_name") or "").strip()
+    tool_name = _safe_tool_name(raw_tool_name) if raw_tool_name else ""
     results = output.get("results")
     text = output.get("text")
     structured = output.get("structured_content")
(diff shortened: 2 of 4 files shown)


muddlebee:
acceptance criteria and demo is missing

arnavpopley:
@greptile review

arnavpopley:
Added — the PR description now has the full "Behaviour comparison" section (before/after \`merge_tool_evidence\` output, all 5 required answers) and the definition-of-done checks (registry check, empty/error no-noise check, overlap check).

One honest gap: check #5 (grepping a real \`opensre investigate\` output for the tool names) needs live PostHog MCP credentials I don't have configured. Per the issue's own guidance for that case, I've left that box unticked and left checks 1–4 + the behaviour comparison to stand on their own — they exercise the real \`merge_tool_evidence\` path a live investigation calls, just without an actual PostHog connection.

muddlebee:
@arnavpopley demo and the output requirements are missing

muddlebee:
@arnavpopley closing this, PR doesnt make much sense to me. what it doing and fixing. please discuss at discord or get this reviewed by core team.

PR #5842, opened 2026-08-27
fix(ci): stop a queued-/quit test from hanging the cli-runtime-3 shard

Davidson3556:
Fixes the red `main`: CI Gate has been failing because `test (cli-runtime-3)` gets killed at its 30-minute job timeout.

`test_queued_literal_quit_requests_runtime_exit` gave a real agent turn a 1 second deadline. That turn takes 1-2s even on an idle machine, so under parallel CI load it misses. Missing it doesn't fail the test, it wedges: the `TimeoutError` unwinds into `asyncio.run` teardown, and `_cancel_all_tasks` then sits in `selector.select(None)` forever. The stack dump shows the turn's own thread already finished and idle.

So the shard burned its full 30 minutes twice on `main` and reported only "The operation was canceled" - no failing test name to go on.

Two changes:

- Drop the two racy deadlines in that test and await the queue and worker directly.
- Add pytest-timeout with a global `timeout = 600`, so the next hang fails at 10 minutes with a stack dump of every thread instead of dying silently at 30.

I left the timeout method alone rather than setting `thread`. Under xdist `thread` loses the dump entirely, since the worker calls `os._exit()` and the report degrades to `worker 'gw0' crashed`. The default picks SIGALRM where it exists, which keeps the dump and lets the worker survive, and falls back to `thread` on Windows for `ci-labels-windows.yml`.

Before, on `main` (run 33083203021, both attempts):

```
test (cli-runtime-3): cancelled  30min
CI Gate: failure
static=success typecheck=success test=cancelled coverage-report=skipped
Error: test is cancelled
```

Reproduced locally, then captured with the new timeout:

```
[gw1] node down: Not properly terminated
FAILED tests/interactive_shell/test_terminal_runtime_turns.py::test_queued_literal_quit_requests_runtime_exit
```

After the fix, the same shard 12 times in a row:

```
try 1: 803 passed, 1 skipped in 15.97s
...
try 12: 803 passed, 1 skipped in 22.32s
RESULT: fail=0
```

Broad suite: `15906 passed, 74 skipped, 2 xfailed in 147.28s`.

---

The problem was a test that hangs instead of failing, which costs a whole CI job and gives you nothing to debug from.

I considered just raising the deadline to something generous like 30s. That hides the real issue: any deadline here can be missed under load, and missing it still wedges teardown rather than failing cleanly. Removing the deadline entirely is what makes the failure mode go away, and the global pytest-timeout covers the case the deadline was nominally there for.

On the timeout value, 600s sits well above the slowest test in the repo (about 8s by the recorded durations) so it can't cause false failures, while still cutting a hang from 30 minutes to 10. Per-test overrides are still available through `@pytest.mark.timeout`.

I checked the rest of the suite for the same shape - an `asyncio.to_thread` paired with a short `wait_for` - and this test was the only one.


Davidson3556:
Good catch on the dependency group - fixed in 5856ec4d8.

`make install` runs `uv sync --frozen --extra dev`, so CI and the documented setup did get the plugin. But a plain `uv sync` did not, which left `timeout` inert and pytest printing `PytestConfigWarning: Unknown config option: timeout`. `timeout` is the only ini key in pytest.ini backed by a plugin, so it now sits in both the extra and the default group.

Verified on a plain `uv sync` (no `--extra dev`):

```
plugins: cov-7.1.0, timeout-2.4.0, anyio-4.13.0
timeout: 600.0s
timeout method: signal
```

@greptile review

Davidson3556:
LGTM

PR #5857, opened 2026-08-28
feat(shell): inline option picker — hide the typing box during structured input, arrow selection, and multi-select

YauhenBichel:
When the agent needs a specific decision from the user mid-turn (an execution
confirmation, an Ask User question, or an option menu), the free-text input box
now steps aside so the choice is unambiguous, and all pickers share one clean
inline look.

**Typing box hidden during structured input**
- While a y/n confirmation, an exclusive-stdin menu, or a queued choice owns the
  keyboard, the `[N] ❯` composer and its ghost placeholder are suppressed
  (`prompt_visibility.typing_box_hidden` + `terminal.awaiting_confirmation`),
  so users no longer type into a box that is about to be replaced by options.
- The region keeps a constant height across redraws (blank stand-in for the
  hidden box), so hiding/restoring does not shift the layout.

**One inline picker, arrow selection**
- Selected rows use the `❯` accent arrow instead of a reverse-video block, in
  both the single-choice picker and the Ask User wizard.
- Menus draw inline and keep the transcript: the pre-menu paint is cleared with
  `renderer.erase()` (app rows only) instead of `renderer.clear()`
  (full-screen), which previously wiped the scrollback and read as a new window.
- The hardware cursor is hidden while a menu is navigated and restored on exit,
  removing the stray block below the hint line.
- Free text is the last option row, edited in place, rather than a separate
  prompt.

**Multi-select**
- `AskUserQuestion` gains `multi_select`; the wizard renders `[ ]`/`[x]`
  checkboxes with Space/Enter/digit toggle and a Submit row, and the
  `ask_user_choice` tool schema carries `multi_select` per question.

### DIFF
--- a/core/agent_harness/session/pending_choice.py
+++ b/core/agent_harness/session/pending_choice.py
@@ -32,6 +32,9 @@ class AskUserQuestion:
     options: tuple[str, ...]
     """Option labels in display order."""
 
+    multi_select: bool = False
+    """When True, the menu shows checkboxes and the user may pick several options."""
+
 
 @dataclass(frozen=True, slots=True)
 class PendingUserChoice:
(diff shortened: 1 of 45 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5861, opened 2026-08-28
feat(shell): arrow-navigable confirmation and collapsible plan overlay

YauhenBichel:
Two planning-controls polish items for the interactive shell.

**Arrow-navigable execution confirmation**
- The `Yes, allow? [Y/n]` gate is now a stacked, arrow-selectable choice:
  `[a] Yes` / `[b] No`. Up/Down move the selection (with wrap), Enter confirms;
  `a`/`b` and `y`/`n` are one-key shortcuts. No typed answer required.
- The free-text composer box is hidden while the choice is active — a
  `ConditionalContainer` collapses the composer frame and footer whenever
  structured input owns the keyboard, so the box cannot compete with the choice.
- The selected row is marked with the accent arrow; the block has blank-row
  spacing around its header, options, and nav hint for a clear hierarchy.
- New `surfaces/interactive_shell/ui/hooks/` package holds the self-contained
  render + key-binding units (`confirmation_choice.py`, `plan_expand.py`).

**Collapsible plan overlay**
- A plan longer than a few steps collapses to a window around the current step
  with `… N earlier` / `… N more · Ctrl+P for full plan` markers; short plans
  and the expanded view show every step. Done / current / pending keep their
  glyphs. Ctrl+P toggles the full list (gated on a plan being present, so it
  keeps its default behavior otherwise).

PR #5862, opened 2026-08-28
fix(shell): unhang subprocess confirmation and stop risk-grading non-shell commands

YauhenBichel:
Two bug fixes for the arrow-navigable execution confirmation (#5861).

**Subprocess confirmation dead-end.** During a subprocess / exclusive-stdin turn
(the demo runner, code-implement) the live prompt application is suspended, so a
confirmation that parked on it for an arrow-nav answer hung — and the cooked
terminal echoed the arrow keys as `^[[A`, so it could not be closed. When the
prompt app is not running, `_confirm_via_prompt` now falls back to
`_confirm_via_readline`: it prints the option rows and reads a single line, where
a row tag (`a`/`b`/`c`), a digit, or a `y`/`n` answer resolves the choice, and
Ctrl-C / EOF cancels. Normal turns keep the bordered arrow-nav.

**Non-shell commands mislabeled.** The per-command risk grade and the
"always allow reversible commands" row only make sense for real shell commands,
but they were applied to every gated action — a slash command (`/auto low`) or a
tool showed "· medium risk · Changes local state" and offered to "always allow
reversible commands". Risk classification is now gated on `tool_type == "shell"`;
other actions render a plain "· needs confirmation" header with the policy reason
and a Yes/No choice, and "always allow" cannot raise the auto level for them.

### DIFF
--- a/core/agent_harness/prompts/opensre_system_prompt.md
+++ b/core/agent_harness/prompts/opensre_system_prompt.md
@@ -207,6 +207,8 @@ Your final message should read naturally, like an update from a concise teammate
 
 You can skip heavy formatting for single, simple actions or confirmations. In these cases, respond in plain sentences with any relevant next step or quick option. Reserve multi-section structured responses for results that need grouping or explanation.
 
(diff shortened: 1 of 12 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5871, opened 2026-08-30
fix(session): preserve writes after a torn JSONL tail

### LINKED ISSUE #5750: [BUG] A torn final line swallows the next session turn

ump45nose:
Fixes #5750

- Separate a torn final JSONL record from the next append so a valid post-crash turn remains readable.
- Parse store and repository session files as strict UTF-8 one line at a time, allowing a malformed line to be skipped without accepting replacement-corrupted data or blocking later records.
- Keep production session restore, recent-session summaries, and investigation history readable after a torn UTF-8 line.
- Promote the existing strict xfail into a passing regression and cover partial JSON, a crash in the middle of a UTF-8 code point, and production restoration.

Before the fix:

```text
FAILED test_soak_write_after_a_torn_tail_is_not_swallowed
AssertionError: the turn written after the torn tail is unreadable
UnicodeDecodeError: invalid continuation byte during JsonlSessionRepo.load_session
```

After the fix:

```text
test_soak_write_after_a_torn_tail_is_not_swallowed[partial-json] PASSED
test_soak_write_after_a_torn_tail_is_not_swallowed[partial-utf8-code-point] PASSED
tests/interactive_shell/sessions/test_store.py: 65 passed
tests/core/: 1971 passed, 16 skipped, 1 xfailed
```

Validation:

- `make lint`
- `make format-check`
- `make typecheck`
- `uv run pytest tests/interactive_shell/sessions/test_store.py -q`
- `uv run python -m pytest tests/core/ -q`

---

The store already holds its per-session file lock while resolving the parent and appending a record. Inside that lock, the change checks the final byte of a non-empty file. If it is not a newline, the writer emits one separator before the new JSON record. This preserves the append-only file, leaves the one crash-damaged record for the decode-failure path to skip, and prevents it from absorbing the next valid turn.

A crash can also stop in the middle of a multibyte UTF-8 code point. I considered decoding with replacement, but that could turn corrupted bytes into apparently valid record content. Instead, both the write-side store and the read-only repository split files into byte lines, decode each line strictly as UTF-8, and reject invalid UTF-8 or JSON only for that line. The same behavior now applies to production restoration, recent-session summaries, and investigation lookup/history.

---

---

Allow edits from maintainers is enabled.

### DIFF
--- a/core/agent_harness/session/persistence/jsonl_repo.py
+++ b/core/agent_harness/session/persistence/jsonl_repo.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import contextlib
 import json
+from collections.abc import Sequence
 from pathlib import Path
 from typing import Any
 
@@ -106,7 +107,7 @@ class JsonlSessionRepo:
     def _collect_investigation_records(
         path: Path,
         *,
-        lines: list[str] | None = None,
+        lines: Sequence[str | bytes] | None = None,
     ) -> list[dict[str, Any]]:
         with contextlib.suppress(Exception):
             loaded = _load_v2_lines(lines) if lines is not None else _load_v2_file(path)
@@ -153,7 +154,7 @@ class JsonlSessionRepo:
         results: list[dict[str, Any]] = []
         for path in sorted(root.glob("*.jsonl"), key=_mtime, reverse=True):
             with contextlib.suppress(Exception):
-                lines = path.read_text(encoding="utf-8").splitlines()
+                lines = path.read_bytes().splitlines()
                 results.extend(self._collect_investigation_records(path, lines=lines))
             if len(results) >= n * 3:
                 break
@@ -170,7 +171,7 @@ class JsonlSessionRepo:
         count = 0
         for path in root.glob("*.jsonl"):
             with contextlib.suppress(Exception):
-                lines = path.read_text(encoding="utf-8").splitlines()
+                lines = path.read_bytes().splitlines()
                 for rec in JsonlSessionRepo._collect_investigation_records(path, lines=lines):
                     inv_id = str(rec.get("investigation_id") or "").lower()
                     if not inv_id.startswith(normalized):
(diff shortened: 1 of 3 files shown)


ump45nose:
@greptile review

PR #5874, opened 2026-08-30
feat(shell): Live plan overlay, per-step work log, and turn-continuation for investigations

YauhenBichel:
## What

Makes the interactive-shell plan and investigation UX live and continuous:
the pinned checklist advances as work runs, each step accrues a work log that
prints as a completion breakdown, reviewed goals keep the turn going until
they're met, and several rendering rough edges are fixed.

## Why

During an investigation, `investigation_start` blocks the action agent from
calling `update_plan`, so the pinned "Plan · n/m" overlay froze at step 1 and
the turn often went idle with pending steps left unaddressed. This drives the
plan from the host side, records what each step did, and holds a goal-reviewed
turn open until the goal is satisfied.

## Changes

**Plan overlay (host-driven)**
- `investigation_progress.py`: map pipeline stream nodes to plan phases
  (ordinal, never regressing, publish → last step) and advance the pinned plan
  as stages fire.
- `console_state.py` pins the active session; `ProgressTracker` / `StreamRenderer`
  advance that plan per node.
- `display.ensure_active_step` normalizes the plan so exactly one step is
  in_progress when work remains; `update_plan_policy` refreshes the prompt so
  the overlay repaints immediately.
- Surfaces import these through `core.agent_harness.spi.task_plan`, not internal
  leaves (keeps the harness API border intact).

**Per-step work log + breakdown**
- `work_log.py`: host-owned, capped, deduped work lines recorded under the
  in_progress step (the `update_plan` schema stays status-only). Attributed from
  the tool tracker and the action render observer.
- On completion the REPL prints a "Plan complete" checklist with each step's  work lines.

**Goal-reviewed turn continuation**
- `react_loop` nudges and continues when a reviewed goal (`goal.verify`) is not
  yet met, bounded by the iteration cap with a last-lap accept. Descriptive
  goals without `verify` still end on the first no-tool reply.

**Rendering**
- "Invoking tools…" spinner phase renders in a distinct bold-brand accent
  while tools run.
- `print_repl_text` emits CRLF under `patch_stdout(raw=True)` so goal-progress
  and menu rows no longer staircase.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5876, opened 2026-08-30
fix(shell): Run opensre integrations read-only subcommands in the foreground

YauhenBichel:
## What
`opensre integrations list/show/verify/status` now run in the foreground
instead of backgrounding, and Grafana verify failures render compactly.

## Why
Backgrounding a read-only `integrations` call ended the action turn on
"started — task …" while `verify` was still running. The pinned plan then hung
on "Invoking tools · 1 task running" with nothing to observe, so the turn went
idle mid-plan. A connect-timeout also dumped the full `HTTPConnectionPool`
string into the verify table, overflowing the terminal.

## Changes
- `tools/interactive_shell/cli.py`: classify `integrations`
  `list`/`show`/`verify`/`status` (and bare `integrations`, which defaults to
  `list`) as read-only and run them foreground; `integrations setup` stays
  mutating/background. Updated the confirmation hint accordingly.
- `integrations/grafana/verifier.py`: on connect failure, report the endpoint
  and exception type, not the full pool/retry dump.
- `surfaces/shared/terminal/health/__init__.py`: fold the Detail column to the
  terminal width so long errors wrap instead of overflowing.

### DIFF
--- a/integrations/grafana/verifier.py
+++ b/integrations/grafana/verifier.py
@@ -36,9 +36,13 @@ def _discover_datasources(config: GrafanaIntegrationConfig) -> list[Any] | str:
         detail = f"Datasource discovery failed: HTTP {status} from {url}"
         return f"{detail}: {body}" if body else detail
     except requests.RequestException as exc:
-        return f"Datasource discovery failed: could not reach {config.endpoint} ({exc})."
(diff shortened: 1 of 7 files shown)


YauhenBichel:
@greptile review

PR #5879, opened 2026-08-31
fix(shell): Fix: an untracked tool_end clears only an id-less action shimmer, never a named one

YauhenBichel:
## What
Harden the in-flight action-shimmer stack so an id-less `tool_end` can't clear a
named, still-running action.

## Why
`clear_active_action("")` — the path taken when a `tool_end` event omits a tool
id — popped the oldest slot unconditionally (`del self._in_flight_actions[0]`).
If a named action (started with an id) was oldest, that untracked end wiped its
shimmer while it was still running, so the live "⟩ …" line vanished mid-command.

## How
`clear_active_action("")` now scans for and removes the first slot that *also*
has no id, leaving named actions intact. `None` (clear-all) and a specific id
are unchanged.


YauhenBichel:
@greptile review

PR #5880, opened 2026-08-31
feat(shell): emphasize key action words in the agent's live narration

YauhenBichel:
## What / Why
In-progress agent narration read as a wall of text. Instruct the model to bold
the one or two words that carry each action so a long response scans quickly
(e.g. "Now let me **trigger** the chaos experiment and **watch** the UI react").

## How
Fill the previously-empty "Responsiveness" section of the single system prompt
with a tight rule: bold the verb or its target only — never a whole phrase or
sentence, and never a file path or command (those use inline code). No code
change needed: the narration path (`message_update` → `_render_intermediate_message`
→ `render_markdown_block`) already renders `**bold**` via rich Markdown.

### DIFF
--- a/core/agent_harness/prompts/opensre_system_prompt.md
+++ b/core/agent_harness/prompts/opensre_system_prompt.md
@@ -35,6 +35,13 @@ Unless the user explicitly asks for a plan, asks a question about the code, is b
 
 ## Responsiveness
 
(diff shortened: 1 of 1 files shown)


YauhenBichel:
@greptile review

PR #5881, opened 2026-08-31
feat(shell): Action shimmer: shimmer-in → solid, and no duplicate command during a run

YauhenBichel:
## What / Why
Two fixes to the live running-action row so it matches the spec (and droid):

1. **Shimmer prior → solid in progress.** The action line used to glow the whole
   time. Now it shimmers in (a white glow rising over ~0.55s as the action is
   picked up) then holds a **solid** fill while it runs — "shimmer prior, solid
   in progress".
2. **No duplicate command.** For `shell_run` / `cli_exec` the shimmer repeated
   the command that the `$ <cmd>` line already shows, so a running command
   appeared twice. The shimmer now names the action only (`⟩ Execute`); generic
   tools (no `$ cmd`) keep the full `label · content`.

## How
- `SpinnerState.active_action_ansi`: rising-glow over `_SHIMMER_LEAD_SECONDS`,
  then `fade_fg_ansi(1.0)` solid.
- `ActionRenderObserver._set_active_action`: `_COMMAND_STREAMING_TOOLS` show the
  label only.

### DIFF
--- a/surfaces/interactive_shell/runtime/core/state.py
+++ b/surfaces/interactive_shell/runtime/core/state.py
@@ -259,10 +259,12 @@ class SpinnerState:
     # Verbs picked twice as often as the rest of their pool (default weight 1).
     _VERB_WEIGHTS = {"jacking in": 2, "crawling the datastream": 2}
 
-    # Theme-token shimmer for the running action line: a triangle wave over this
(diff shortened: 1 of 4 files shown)


YauhenBichel:
@greptile review

PR #5883, opened 2026-08-31
feat(shell): User-prompt accent bar + command output indented under its action

YauhenBichel:
## What
- User-prompt rows open with a left colour accent bar (`▌`) so a prompt is
  visually distinct from the agent response (which opens with `●`).
- Shell/CLI command output is indented under its `$ command` header with a `↳`
  marker (parent → child), matching droid's readable hierarchy.

## How
- `render_submitted_prompt`: prepend a `▌` bar (bold highlight) on every line of
  the submission.
- `print_command_output`: indent the result with `  ↳ ` when it fits the
  terminal width; wide output (tables) stays flush so it does not wrap.


YauhenBichel:
@greptile review

PR #5884, opened 2026-08-31
feat(shell): Syntax-highlight the command; remove the ⟩ action glyph

YauhenBichel:
## What
- The `$ command` line is syntax-highlighted (bash lexer) so its keywords,
  arguments, and operators are coloured, not a flat monochrome bold line.
- The live running-action line dropped its leading glyph: `⟩` read as `)` and
  the arrow alternative was worse, so it's now just an indented `Execute`.

## How
- `ReplSubprocessPresenter.print_bold_command`: render the command via
  `rich.syntax.Syntax(...).highlight(...)` under the bold `$ ` prefix.
- `SpinnerState`: replace `_ACTION_GLYPH` with a plain `_ACTION_INDENT`.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5887, opened 2026-08-31
feat(slack): map read-tool output into citeable report evidence

### LINKED ISSUE #5551: Add evidence mappers for slack
## Summary
can point to. The `slack` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `replay_slack_thread_locally`
- `slack_list_team_members`
- `slack_read_list`
- `slack_read_messages`
- `slack_search_messages`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="replay_slack_thread_locally"' integrations/`.

```python

def _map_replay_slack_thread_locally(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="replay_slack_thread_locally",
        )

@tool(name="replay_slack_thread_locally", ..., evidence_mapper=_map_replay_slack_thread_locally)
```

## How to check it works

   under `tests/integrations/slack/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['replay_slack_thread_locally', 'slack_list_team_members', 'slack_read_list', 'slack_read_messages', 'slack_search_messages']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'replay_slack_thread_locally', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `replay_slack_thread_locally`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k slack
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify slack
# edit /tmp/alert.json so the alert names a resource slack actually knows about
grep -E 'replay_slack_thread_locally|slack_list_team_members|slack_read_list|slack_read_messages|slack_search_messages' /tmp/rca.json
```

matching lines into the PR.

**No credentials for slack?** That is a normal outcome, not a reason to go

AniketR10:
Fixes #5551 

Each of the five Slack read tools (`replay_slack_thread_locally`, `slack_list_team_members`, `slack_read_list`, `slack_read_messages`, `slack_search_messages`) now carries an `evidence_mapper` that calls `record_evidence_entry`, so its output becomes a numbered, citeable report entry (e.g. `E1 — Slack Channel Messages — 3 messages from C0INCIDENT`) instead of being dropped; the search mapper also links the top match's permalink. Empty, unavailable, and failed payloads record nothing, so no "0 items" line costs context on every later turn. Removes those five tools from `evidence_mapper_baseline.txt` and adds a unit test per mapper under `tests/integrations/slack/`.

https://github.com/user-attachments/assets/a325ab07-b7ac-4ce5-a8be-377bc81b0388

---

These five tools are `BaseTool` classes, not `@tool` functions like the Grafana example, so each mapper is attached as a class attribute (`evidence_mapper = _map_slack_read_messages`). Each mapper sits in its own tool file.

- Every mapper uses the tool's own name as the `source`. Only the first entry per source is kept, so a shared `"slack"` key would have hidden four of the five.
- A mapper checks the result list itself, not the `status` field. One check handles empty reads, failures and unavailable tools.
- Entries are a short count plus context (channel, list title, thread ref) — about 100 characters. Search also adds the top match's link, with the query cut at 80 chars.
- No new evidence keys: the raw output is already stored, so adding more would save it twice.

---

---

### DIFF
--- a/integrations/slack/tools/slack_list_members_tool/tool.py
+++ b/integrations/slack/tools/slack_list_members_tool/tool.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
 from core.tool import BaseTool, SideEffectLevel
 from core.tool_framework import SUMMARIZE_OBSERVATION_TAG, tool
@@ -16,11 +17,28 @@ from integrations.slack.web_client import (
 )
 
 
+def _map_slack_list_team_members(
+    evidence: dict[str, Any], output: dict[str, Any], _tool_input: dict[str, Any]
+) -> None:
+    """Record the workspace roster as citeable evidence when members were read."""
+    members = output.get("members")
+    if not isinstance(members, list) or not members:
+        return
+    truncated = " (truncated)" if output.get("truncated") else ""
+    record_evidence_entry(
+        evidence,
+        source="slack_list_team_members",
+        label="Slack Workspace Roster",
+        summary=f"{len(members)} members{truncated}",
+    )
+
+
 class SlackListTeamMembersTool(BaseTool):
     """List the members of the Slack workspace the bot is installed in."""
 
     name = "slack_list_team_members"
     source = SOURCE
+    evidence_mapper = _map_slack_list_team_members
     description = (
         "List members of the Slack *workspace* the bot is installed in "
         "(id, username, real name, title, bot flag) via users.list. "
--- a/integrations/slack/tools/slack_read_list_tool/tool.py
+++ b/integrations/slack/tools/slack_read_list_tool/tool.py
@@ -6,6 +6,7 @@ import os
 import re
 from typing import Any
 
+from core.domain.types.evidence import record_evidence_entry
 from core.domain.types.tools import ToolSurface
 from core.tool import BaseTool, SideEffectLevel
 from core.tool_framework import tool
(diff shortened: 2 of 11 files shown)


muddlebee:
@AniketR10 demo and runnables missing from the issue

image

Davidson3556 (dismissed):
Aniket, the mappers themselves look right. I checked all five against the tools' real output envelopes and every key you read lines up, and the per-tool `source` keys were the correct call.

Two things before this can go in.

`slack_read_messages` and `slack_search_messages` cite a page-capped count as exact. History fetches one page of `limit` messages with no pagination (default 50), and search takes `count` (default 20):

```
'50 messages from C0INC'      # limit=50, one page, likely more
"20 matches for 'timeout'"    # count=20, same
```

Your other three mappers already handle this with `(truncated)` off the tool's own flag. These two have the ceiling sitting in `tool_input` (`limit` / `count`), so the same treatment works there.

Second, on muddlebee's label: the description needs the behaviour comparison from #5551, with the diff and the five answers, plus the output of checks 1-4. If you have no Slack workspace to run check 5 against, the issue asks you to say so in the PR and leave that one unticked rather than leave it out.

AniketR10:
@muddlebee @Davidson3556 working on the fixes

AniketR10:
@muddlebee @Davidson3556 fixed the search tree as truncated like other and added the full slack test, please review (my free model was rate limiting everytime so that is why it is showing 404 in the recording)

PR #5894, opened 2026-08-31
feat(shell): recess plain command output beneath the reply summary

YauhenBichel:
Render uncoloured shell stdout in SECONDARY so raw output reads as supporting detail under the bright reply, matching the intended focus hierarchy. Colour the command emits itself (errors, success) still wins over the base, so semantic output colours are preserved.

image


YauhenBichel:
@greptile review

PR #5895, opened 2026-08-31
feat(prompt): emphasize key facts in the final reply

YauhenBichel:
Extend the Responsiveness guidance from narration to the final answer: bold the load-bearing figures and entities so the result is scannable, and when a headline metric is misleading on its own (a percentage hiding a shared total), add one sentence naming the number that actually matters.

image


YauhenBichel:
@greptile review

PR #5897, opened 2026-08-31
feat(shell): fix reply table rendering and add focus-aware sound notifications

YauhenBichel:

Two shell-UX items surfaced by a live comparison with droid.

**1. Reply markdown tables render as tables, not raw pipes.**
A reply that led with a Markdown table rendered as one flattened line of raw
pipes instead of an aligned table: the inline `∴ ` response marker fused onto
the table's header row broke CommonMark block parsing, so the block was treated
as a paragraph and flattened. The marker now attaches inline only for prose;
when the reply leads with a block element (table, heading, list, code fence,
blockquote) it goes on its own line above it. Detection is a lightweight
first-line check, so the streaming path takes no extra Markdown parse.

**2. Opt-in, focus-aware sound notifications (`OPENSRE_SOUND=1`).**
A chime that fires only when the terminal is **not** the focused window — so it
alerts you when you've switched away, not while you're watching:
- Turn complete: chimes only for turns longer than ~8s (quick replies stay silent).
- Input needed: chimes when a clarification menu opens.

macOS focus is detected permission-free via `lsappinfo` (frontmost app vs. the
terminal we run in, from `TERM_PROGRAM`); off macOS, or when focus can't be
determined, it degrades to the duration gate. Playback uses macOS system sounds
via `afplay` with a terminal-bell fallback, never blocks the REPL or raises, and
is silent by default.

Table — before: `∴ | Rank | Folder | Size | |---:|---|---:| | 1 | ~/Library | 113G | ...`
(one flattened, unreadable line). After:

    ∴
     Rank  Folder     Size
     ─────────────────────
        1  ~/Library  113G
        2  ~/DevBox    47G

Sound — `OPENSRE_SOUND=1 uv run opensre`, then `run: sleep 12; echo done` and
switch to another window before it finishes → a soft completion chime; stay
focused on the terminal → silent.

### DIFF
--- a/config/constants/__init__.py
+++ b/config/constants/__init__.py
@@ -330,6 +330,7 @@ from config.constants.repl_autonomy import (
     format_auto_status_plain,
     parse_auto_level,
 )
+from config.constants.repl_sound import SOUND_MIN_TURN_SECONDS, SOUND_NOTIFICATIONS_ENV
 from config.constants.repl_theme import DEFAULT_THEME_NAME, THEME_NAMES, Theme
 from config.constants.reporting import SLACK_LINK_RE
 from config.constants.runtime_metadata import (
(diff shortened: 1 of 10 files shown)


YauhenBichel:
@greptile review

PR #5898, opened 2026-08-31
feat(jenkins): add evidence mappers for all 5 Jenkins investigation tools

### LINKED ISSUE #5544: Add evidence mappers for jenkins
## Summary
can point to. The `jenkins` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `get_jenkins_build_log`
- `get_jenkins_pipeline_stages`
- `list_jenkins_builds`
- `list_jenkins_jobs`
- `list_jenkins_running_builds`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="get_jenkins_build_log"' integrations/`.

```python

def _map_get_jenkins_build_log(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="get_jenkins_build_log",
        )

@tool(name="get_jenkins_build_log", ..., evidence_mapper=_map_get_jenkins_build_log)
```

## How to check it works

   under `tests/integrations/jenkins/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['get_jenkins_build_log', 'get_jenkins_pipeline_stages', 'list_jenkins_builds', 'list_jenkins_jobs', 'list_jenkins_running_builds']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'get_jenkins_build_log', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `get_jenkins_build_log`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k jenkins
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify jenkins
# edit /tmp/alert.json so the alert names a resource jenkins actually knows about
grep -E 'get_jenkins_build_log|get_jenkins_pipeline_stages|list_jenkins_builds|list_jenkins_jobs|list_jenkins_running_builds' /tmp/rca.json
```

matching lines into the PR.

**No credentials for jenkins?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5544 (part of epic #5519, tracking issue #1079)

Wires all 5 Jenkins investigation tools — `list_jenkins_builds`, `get_jenkins_build_log`, `get_jenkins_pipeline_stages`, `list_jenkins_jobs`, `list_jenkins_running_builds` — into `record_evidence_entry` so their output becomes citeable investigation-report evidence.

- `list_jenkins_builds`/`list_jenkins_jobs`/`list_jenkins_running_builds` use the "N+" convention: `list_jenkins_jobs`/`list_jenkins_running_builds` have an explicit `truncated` flag from the client (folder-depth/job-cap dropped items); `list_jenkins_builds` compares its `total` against the caller's requested `limit`, the client's own truncation point (the client slices `builds[:limit]`, so a `total` at that ceiling may understate the real count).
- `list_jenkins_builds` cites a `status` filter directly instead of a derived failed count when one is set — every returned build already matches the filter, so an unfiltered "0 failed" derived from a `status="SUCCESS"` query would misrepresent reality (same fix pattern as the airflow evidence-mapper PR earlier in this project).
- `get_jenkins_build_log` cites the console log length using the client's own tail-truncation flag.
- `get_jenkins_pipeline_stages` skips freestyle jobs (`is_pipeline=False`) — no stages there is an expected, unremarkable outcome, not a finding worth citing.
- Mappers live in a new `integrations/jenkins/tools/_evidence.py` sibling module rather than inline in the flat `tools/__init__.py` facade.

**All 5 tools carry their mapper (via the real registry):**
```
print({n: bool(g(n).evidence_mapper) for n in ['get_jenkins_build_log', 'get_jenkins_pipeline_stages', 'list_jenkins_builds', 'list_jenkins_jobs', 'list_jenkins_running_builds']})"
{'get_jenkins_build_log': True, 'get_jenkins_pipeline_stages': True, 'list_jenkins_builds': True, 'list_jenkins_jobs': True, 'list_jenkins_running_builds': True}
```

**Tests and quality gates:**
```
$ uv run pytest tests/tools/test_jenkins_tools.py tests/tools/investigation/stages/gather_evidence/ -q
49 passed

$ uv run pytest tests/ -k jenkins -q
97 passed

```

**No live Jenkins server available in this environment**, so leaving the existing `is_available`/`_resolve_client` wiring untouched — only the mappers and their tests were added.

I read the full `integrations/jenkins/client.py` before writing any mapper, because the issue's generic template doesn't tell you which tools carry a real explicit truncation signal versus which ones only echo `len(...)` after their own limit slice. `list_jenkins_jobs` and `list_jenkins_running_builds` both compute a genuine `truncated` boolean (folder-depth/job-count cap), so their mappers trust that signal directly. `list_jenkins_builds` has no such field — `total` is just `len(builds)` after `builds[:limit]` — so its mapper instead compares against the caller's requested `limit`, the same "N+" heuristic used elsewhere in this evidence-mapper backfill when no better server-side signal exists. `get_jenkins_pipeline_stages` skipping freestyle jobs follows the established "no noise on an expected empty result" convention already used by the merged argocd/groundcover evidence-mapper PRs in this same epic.

---

---

**Before** (main, no mapper):
```
BEFORE catalog_entries: None
```

**After** (this branch) — realistic samples for all 5 tools:
```json
[
  {
    "source": "list_jenkins_builds",
    "label": "Jenkins Builds",
    "summary": "2 build(s), 1 failed for 'deploy-prod'"
  },
  {
    "source": "get_jenkins_build_log",
    "label": "Jenkins Build Log",
    "summary": "900 char(s) of console log, for 'deploy-prod' #42"
  },
  {
    "source": "get_jenkins_pipeline_stages",
    "label": "Jenkins Pipeline Stages",
    "summary": "2 stage(s), 1 failed (build status: FAILED) for 'deploy-prod' #42"
  },
  {
    "source": "list_jenkins_jobs",
    "label": "Jenkins Jobs",
    "summary": "2 job(s), 1 failing"
  },
  {
    "source": "list_jenkins_running_builds",
    "label": "Jenkins Running Builds",
    "summary": "1 build(s) currently running"
  }
]
```

**1. What the reader gains.** The report can now cite which Jenkins builds failed, the console log length for a specific failed build, which pipeline stage failed, how many jobs are currently red, and whether anything is running right now — previously all 5 tools ran, cost tokens, and left nothing in the report.

**2. How much context it costs.** One short line per call; `get_jenkins_build_log` cites a character count rather than inlining the log text itself, so a multi-thousand-line console log never bloats the entry.

**3. What happens on an empty or error result.** Verified directly for all 5 tools plus an unavailable case:
```
list_jenkins_builds -> None
get_jenkins_build_log -> None
get_jenkins_pipeline_stages -> None
list_jenkins_jobs -> None
list_jenkins_running_builds -> None
list_jenkins_builds (unavailable) -> None
```
No entry for an empty list/log, a freestyle job with no stages, or an unavailable/error result.

**4. Whether anything is now recorded twice.** These are the only 5 tools in `integrations/jenkins/tools/`, and none had a mapper before this PR — no overlap.

**5. Whether an existing report changed shape.** This only adds entries under the 5 new Jenkins source keys; no other tool's merge path is touched.

## Live-report check (#5)
No live Jenkins server available in this environment, so I can't run `opensre investigate` against a real alert to confirm the entries land in an actual generated report. Per the issue's own note for that case, the checks above (registry wiring, `merge_tool_evidence` output, before/after comparison, full `-k jenkins` sweep) stand on their own — they exercise the real evidence path a live investigation would call.

### DIFF
--- a/integrations/jenkins/client.py
+++ b/integrations/jenkins/client.py
@@ -138,23 +138,38 @@ def _as_dict_list(value: object) -> list[dict[str, Any]]:
 def _nested_jobs_tree(leaf_fields: str, depth: int) -> str:
     """Build a Jenkins ``tree`` query that descends folders ``depth`` levels.
 
-    Each level requests ``leaf_fields`` plus a nested ``jobs[...]`` for the next
-    level, so folder-organized jobs are returned alongside top-level ones.
+    Each level requests ``leaf_fields`` plus a nested ``jobs[...]`` for the
+    next level, so folder-organized jobs are returned alongside top-level
+    ones. The deepest level also probes one level further with a bare
+    ``jobs[name]`` (name only, no other fields) -- without it, a folder
+    that exists exactly at the depth limit comes back with no ``jobs`` key
+    at all (Jenkins only returns what the tree asked for), indistinguishable
+    from a genuine leaf job. The probe lets ``_flatten_jobs`` tell the two
+    apart and report depth truncation instead of silently treating an
+    unexplored folder as a job with unknown status.
     """
-    tree = leaf_fields
+    tree = f"{leaf_fields},jobs[name]"
     for _ in range(max(0, depth - 1)):
         tree = f"{leaf_fields},jobs[{tree}]"
     return f"jobs[{tree}]"
 
 
-def _flatten_jobs(raw_jobs: list[dict[str, Any]], prefix: str = "") -> list[tuple[str, dict]]:
+def _flatten_jobs(
+    raw_jobs: list[dict[str, Any]], prefix: str = "", max_depth: int = 0
+) -> tuple[list[tuple[str, dict]], bool]:
     """Flatten a (possibly nested) Jenkins jobs tree into ``(full_path, job)`` pairs.
 
-    A node with a nested ``jobs`` list is a folder and is recursed into; its
-    children's names are prefixed with ``folder/`` so the path matches what
-    ``_job_api_path`` expects. Leaf jobs are returned with their full path.
+    Returns ``(flat, depth_truncated)``. A node with a nested ``jobs`` list is
+    a folder and is recursed into (its children's names are prefixed with
+    ``folder/``); leaf jobs (no ``jobs`` key) are appended with their full
+    path. At ``max_depth`` the tree query only probed one level further with
+    bare names (see ``_nested_jobs_tree``), so a folder still reporting a
+    non-empty ``jobs`` list here has children beyond the requested depth that
+    were never fully fetched -- it's dropped (it isn't a job) and
+    ``depth_truncated`` is set, rather than misreading it as a job.
     """
     flat: list[tuple[str, dict]] = []
+    depth_truncated = False
     for job in raw_jobs:
         name = str(job.get("name", "")).strip()
         if not name:
@@ -162,10 +177,18 @@ def _flatten_jobs(raw_jobs: list[dict[str, Any]], prefix: str = "") -> list[tupl
         full_path = f"{prefix}{name}"
         nested = job.get("jobs")
         if isinstance(nested, list):
-            flat.extend(_flatten_jobs(_as_dict_list(nested), prefix=f"{full_path}/"))
+            if max_depth <= 0:
+                if nested:
+                    depth_truncated = True
+                continue
+            child_flat, child_truncated = _flatten_jobs(
+                _as_dict_list(nested), prefix=f"{full_path}/", max_depth=max_depth - 1
+            )
+            flat.extend(child_flat)
+            depth_truncated = depth_truncated or child_truncated
         else:
             flat.append((full_path, job))
-    return flat
+    return flat, depth_truncated
 
 
 def _coerce_build_number(value: object) -> int | None:
@@ -336,7 +359,9 @@ class JenkinsClient:
             resp = self._get_client().get("/api/json", params={"tree": tree})
             resp.raise_for_status()
             data = _as_dict(resp.json())
-            flat = _flatten_jobs(_as_dict_list(data.get("jobs")))
+            flat, depth_truncated = _flatten_jobs(
+                _as_dict_list(data.get("jobs")), max_depth=_MAX_FOLDER_DEPTH - 1
+            )
             jobs = []
             for full_path, job in flat[:_MAX_JOBS]:
                 status, building = _status_from_color(job.get("color"))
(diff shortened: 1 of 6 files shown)


arnavpopley:
Both real, fixed properly:

**1. Build window ignored (my mapper's bug):** compared \`total\` against the raw requested \`limit\`, but the client never honors a limit above its own 50-build cap — a \`limit=1000\` hitting that real ceiling looked exact. Also, whenever \`status\` is set the client always scans only the 50 *most recent* builds regardless of \`limit\` with no signal for whether older history also matches — that result can never be proven exact, so it's now always qualified with "+", not just when the count happens to reach a threshold.

**2. Folder-depth truncation invisible (a latent bug in the client, not just the mapper):** \`list_jobs\`/\`list_running_builds\`'s \`truncated\` flag only caught the 200-job cap. A folder nested deeper than \`_MAX_FOLDER_DEPTH\` comes back from Jenkins with no \`jobs\` key at all (the tree query didn't ask that deep) — \`_flatten_jobs\` was silently misreading that as a genuine leaf job with unknown status, and the depth truncation itself was invisible to \`truncated\`. Added a shallow \`jobs[name]\` probe at the tree's deepest level so a truncated folder can be told apart from a real job — it's now dropped and flagged instead of polluting the job list.

New tests cover both: the 50-cap clamp, the always-qualified status-filter path, and a direct \`_flatten_jobs\` unit test suite plus an end-to-end \`list_jobs\` regression for the depth-truncation case.

@greptile review

arnavpopley:
Fixed the depth-truncation-discarding-empty-results finding.

Verified against `client.py`: a folder-depth-limited scan whose only matching jobs (or running builds) lie beyond `_MAX_FOLDER_DEPTH` legitimately returns `jobs=[]`/`running_builds=[]` with `truncated=True` still set — `_flatten_jobs` reports `depth_truncated` independently of whether anything was actually flattened. Both `map_list_jenkins_jobs` and `map_list_jenkins_running_builds` were early-returning on the empty list before ever checking `truncated`, so that incompleteness signal was silently dropped instead of being cited as "0+ job(s)".

Fix: only skip when the result is empty *and* untruncated (nothing exists, or a full scan genuinely found nothing) — an empty-but-truncated result now records `"0+ job(s), 0+ failing"` / `"0+ build(s) currently running"`, consistent with the existing "N+" convention.

Added a regression test for each mapper pinning the empty+truncated case, alongside the existing empty+untruncated skip test (made its `truncated: False` explicit rather than implicit-via-missing-key). Full jenkins suite (105 tests via `-k jenkins`) + lint/format/mypy pass.

arnavpopley:
@greptile review

PR #5899, opened 2026-08-31
feat(helm): add evidence mappers for all 5 Helm CLI investigation tools

### LINKED ISSUE #5543: Add evidence mappers for helm
## Summary
can point to. The `helm` tools listed below do return useful data, but that
evidence the agent can cite.

## Why this should be done

- `helm_get_release_manifest`
- `helm_get_release_values`
- `helm_list_releases`
- `helm_release_history`
- `helm_release_status`

in your pull request.

## Setup (one time)

## How to do it
`grep -rl 'name="helm_get_release_manifest"' integrations/`.

```python

def _map_helm_get_release_manifest(evidence, output, _input):
    if rows:
        record_evidence_entry(
            evidence,
            source="helm_get_release_manifest",
        )

@tool(name="helm_get_release_manifest", ..., evidence_mapper=_map_helm_get_release_manifest)
```

## How to check it works

   under `tests/integrations/helm/`.
   Expected: the new tests pass.

   ```
   print({n: bool(g(n).evidence_mapper) for n in ['helm_get_release_manifest', 'helm_get_release_values', 'helm_list_releases', 'helm_release_history', 'helm_release_status']})"
   ```

   ```
   ev={}; merge_tool_evidence(ev, 'helm_get_release_manifest', {'<field>': [1, 2]}, {}); \
   ```
   Expected: prints a list with an entry whose `source` is `helm_get_release_manifest`.

4. **Nothing else broke.**
   ```
   uv run pytest tests/ -k helm
   ```
   Expected: all pass.

```
```

```
```

  not before?
  are unaffected.

## Definition of done

## Out of scope

```bash
uv run opensre integrations verify helm
# edit /tmp/alert.json so the alert names a resource helm actually knows about
grep -E 'helm_get_release_manifest|helm_get_release_values|helm_list_releases|helm_release_history|helm_release_status' /tmp/rca.json
```

matching lines into the PR.

**No credentials for helm?** That is a normal outcome, not a reason to go

arnavpopley:
Closes #5543 (part of epic #5519, tracking issue #1079)

Wires all 5 Helm CLI investigation tools — `helm_list_releases`, `helm_release_status`, `helm_release_history`, `helm_get_release_values`, `helm_get_release_manifest` — into `record_evidence_entry` so their output becomes citeable investigation-report evidence.

- `helm_list_releases` uses the "N+" convention against the caller's requested `max_releases` — `helm list --max` silently caps the CLI's own output with no truncation signal echoed back to the tool layer.
- `helm_get_release_manifest` uses the client's own explicit `truncated` flag (a real signal, not inferred).
- `helm_get_release_values` cites **only a top-level key count**, never the keys or values themselves — the tool's own docstring explicitly warns values "may include secrets", so the evidence entry stays safe to surface in a report without having to decide per-key sensitivity.
- `helm_release_status` cites the deployment status from `status.info.status` (helm's own status JSON shape).
- `helm_release_history` cites revision count and how many are in a `failed` state.
- Mappers live in a new `integrations/helm/tools/_evidence.py` sibling module rather than inline in the `tools/__init__.py` facade.

**All 5 tools carry their mapper (via the real registry):**
```
print({n: bool(g(n).evidence_mapper) for n in ['helm_get_release_manifest', 'helm_get_release_values', 'helm_list_releases', 'helm_release_history', 'helm_release_status']})"
{'helm_get_release_manifest': True, 'helm_get_release_values': True, 'helm_list_releases': True, 'helm_release_history': True, 'helm_release_status': True}
```

**Tests and quality gates:**
```
$ uv run pytest tests/tools/test_helm_tools.py tests/tools/investigation/stages/gather_evidence/ -q
50 passed

$ uv run pytest tests/ -k helm -q
56 passed, 1 skipped

```

**No live Kubernetes cluster/Helm binary available in this environment**, so leaving the existing `is_available`/`helm_client_for_run` wiring untouched — only the mappers and their tests were added.

I read the full `integrations/helm/client.py` before writing any mapper, because it's the only source of truth for which fields carry a real truncation signal. `helm_get_release_manifest` already tracks `truncated` explicitly (the client caps the manifest text and sets the flag itself), so its mapper trusts that directly. `helm_list_releases` has no such field — the CLI's own `--max` flag silently caps `releases` — so its mapper instead compares the returned count against the caller's requested `max_releases`, the same "N+" heuristic used elsewhere in this evidence-mapper backfill when no better server-side signal exists. The `helm_get_release_values` design decision (key count, not key names or values) came directly from the tool's own docstring warning about secrets — citing anything more specific would risk leaking a secret's name or shape into a report the security team might not expect evidence entries to carry.

---

---

**Before** (main, no mapper):
```
BEFORE catalog_entries: None
```

**After** (this branch) — realistic samples for all 5 tools:
```json
[
  {
    "source": "helm_list_releases",
    "label": "Helm Releases",
    "summary": "2 release(s) in all namespaces"
  },
  {
    "source": "helm_release_status",
    "label": "Helm Release Status",
    "summary": "status: failed for 'checkout' (ns: prod)"
  },
  {
    "source": "helm_release_history",
    "label": "Helm Release History",
    "summary": "2 revision(s), 1 failed for 'checkout'"
  },
  {
    "source": "helm_get_release_values",
    "label": "Helm Release Values",
    "summary": "2 top-level value key(s) retrieved for 'checkout'"
  },
  {
    "source": "helm_get_release_manifest",
    "label": "Helm Release Manifest",
    "summary": "160 char(s) of rendered manifest for 'checkout'"
  }
]
```

**1. What the reader gains.** The report can now cite which release is failed/deployed, how many prior rollouts failed, how many value keys were retrieved for a release (without leaking secret contents), and the size of the rendered manifest — previously all 5 tools ran, cost tokens, and left nothing in the report.

**2. How much context it costs.** One short line per call. `helm_get_release_values` cites a count, never the values, so a release with secret-shaped values never leaks anything into the report. `helm_get_release_manifest` cites a character count, not the manifest text itself.

**3. What happens on an empty or error result.** Verified directly for all 5 tools plus an unavailable case:
```
helm_list_releases -> None
helm_release_status -> None
helm_release_history -> None
helm_get_release_values -> None
helm_get_release_manifest -> None
helm_list_releases (unavailable) -> None
```
No entry for an empty releases list, an empty status/history/values, an empty manifest, or an unavailable/error result.

**4. Whether anything is now recorded twice.** These are the only 5 tools in `integrations/helm/tools/`, and none had a mapper before this PR — no overlap.

**5. Whether an existing report changed shape.** This only adds entries under the 5 new Helm source keys; no other tool's merge path is touched.

## Live-report check (#5)
No live Kubernetes cluster or Helm binary available in this environment, so I can't run `opensre investigate` against a real alert to confirm the entries land in an actual generated report. Per the issue's own note for that case, the checks above (registry wiring, `merge_tool_evidence` output, before/after comparison, full `-k helm` sweep) stand on their own — they exercise the real evidence path a live investigation would call.

### DIFF
--- a/integrations/helm/tools/__init__.py
+++ b/integrations/helm/tools/__init__.py
@@ -8,6 +8,13 @@ from typing import Any
 
 from core.tool import BaseTool
 from integrations.helm.client_factory import helm_client_for_run
+from integrations.helm.tools._evidence import (
+    map_helm_get_release_manifest,
+    map_helm_get_release_values,
+    map_helm_list_releases,
+    map_helm_release_history,
+    map_helm_release_status,
+)
 from integrations.helm.unavailable import helm_base_unavailable
 
 
@@ -16,6 +23,7 @@ class HelmListReleasesTool(BaseTool):
 
     name = "helm_list_releases"
     source = "helm"
+    evidence_mapper = map_helm_list_releases
     description = (
         "List Helm releases (JSON metadata) using the local Helm CLI against the "
         "configured kubeconfig/context."
@@ -135,6 +143,7 @@ class HelmReleaseStatusTool(BaseTool):
 
     name = "helm_release_status"
     source = "helm"
+    evidence_mapper = map_helm_release_status
     description = "Fetch Helm release status (resources, hooks metadata, notes) as structured JSON."
     use_cases = [
         "Checking whether a Helm release is in failed/pending state",
@@ -224,6 +233,7 @@ class HelmReleaseHistoryTool(BaseTool):
 
     name = "helm_release_history"
     source = "helm"
+    evidence_mapper = map_helm_release_history
     description = "Fetch Helm revision history (status, chart version, description per revision)."
     use_cases = [
         "Seeing recent failed rollouts or rollbacks for a Helm release",
@@ -311,6 +321,7 @@ class HelmGetReleaseValuesTool(BaseTool):
 
     name = "helm_get_release_values"
     source = "helm"
+    evidence_mapper = map_helm_get_release_values
     description = "Fetch Helm values for a release as JSON. May include secrets — handle carefully."
     use_cases = [
         "Confirming image tags, replica counts, or feature flags shipped with a chart revision",
@@ -403,6 +414,7 @@ class HelmGetReleaseManifestTool(BaseTool):
 
     name = "helm_get_release_manifest"
     source = "helm"
+    evidence_mapper = map_helm_get_release_manifest
     description = (
         "Fetch the rendered Kubernetes manifest YAML for a Helm release (truncated if huge)."
     )
--- a/integrations/helm/tools/_evidence.py
+++ b/integrations/helm/tools/_evidence.py
@@ -1,0 +1,138 @@
+"""Evidence mappers for the Helm CLI investigation tools."""
+
+from __future__ import annotations
+
+from typing import Any
+
+from core.domain.types.evidence import record_evidence_entry
+
+_FAILED_REVISION_STATUS = "failed"
+
+
+def map_helm_list_releases(
+    evidence: dict[str, Any], output: dict[str, Any], tool_input: dict[str, Any]
+) -> None:
+    """Cite the release count and scope, qualifying page-capped totals.
+
+    ``helm list --max`` silently caps the CLI's own output with no
+    truncation signal echoed back, so a returned count at that ceiling may
+    understate the true number of releases -- use the "N+" convention
+    against the caller's requested ``max_releases``.
+    """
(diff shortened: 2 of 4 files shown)


arnavpopley:
Confirmed and fixed — \`list_releases\` clamps \`max_releases\` to \`MAX_MESSAGE_SIZE\` (4096) before ever calling the CLI, so a caller-requested value above that is never actually honored. Compared against \`min(max_releases, MAX_MESSAGE_SIZE)\` now instead of the raw request, with a regression test pinning a saturated 4096-release page against a much larger requested max.

@greptile review

PR #5914, opened 2026-09-01
refactor(analytics): break tracker/event_properties import cycle

YauhenBichel:
`investigation_tracker` and `event_properties` imported each other, which CodeQL flags as `py/cyclic-import` (alert #2263): the tracker owns the `track_investigation` lifecycle, while `event_properties` reaches back for the `InvestigationTracker` record and the loop-metric helpers when building payloads.

The shared symbols — the `InvestigationTracker` dataclass and the `_resolve_/_with_investigation_loop_metrics` helpers — move into a new leaf module `investigation_tracker_types`, which both sides (plus `capture` and the `investigate` CLI) import from instead of reaching into each other. The leaf imports only downward (`config.constants`, `investigation_loop`, `provider`, so no cycle can form. Pure move — no behaviour change.

A border test (`tests/analytics/test_investigation_tracker_types.py`) pins the acyclic direction so the alert can't regress.

### DIFF
--- a/infrastructure/analytics/capture.py
+++ b/infrastructure/analytics/capture.py
@@ -13,10 +13,12 @@ from infrastructure.analytics.event_properties import (
     _investigation_failed_properties,
     _investigation_outcome_properties,
     _onboard_completed_properties,
-    _with_investigation_loop_metrics,
 )
 from infrastructure.analytics.events import Event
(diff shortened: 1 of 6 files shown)


YauhenBichel:
@greptile review

PR #5920, opened 2026-09-01
feat(shell): forced sign-in screen + specific dev build version

YauhenBichel:
Continues the forced-sign-in redesign (banner shipped in #5915).

**Sign-in screen component** (`surfaces/interactive_shell/ui/sign_in.py`):
- `render_sign_in_screen` — launch banner + a bordered welcome box (blue
  "Welcome to OpenSRE CLI" title over the product description) + the sign-in
  prompt line. Copy from `config/constants/product.py`.
- `prompt_login_or_exit` — the `Login`/`Exit` menu (reuses `repl_choose_one`),
  returning a `SignInChoice` StrEnum.
- `run_sign_in_gate(console, is_signed_in, login)` — gates the REPL: proceeds
  when already signed in or on a non-interactive stdin, else loops the menu
  (Login retries on failure, Exit/Esc declines). Authentication is **injected**
  via the two callables, so the web-app device-auth flow plugs in without this
  screen depending on it.

**[2] Specific build version:** `get_opensre_version()` returns the full release
string on release builds (unchanged) and now expands a dev checkout to the same
shape from git — `0.1.Y.M.D+<branch>.<sha>` (e.g. `v0.1.2026.9.1+main.263d4f1`),
so `--version`, telemetry, and the banner all report a specific build. Detached
HEAD uses the sha alone (no fabricated branch); branch names are sanitised into
a valid PEP 440 local segment. Adds `read_git_head_branch` and refactors
`read_git_head_sha` to share one HEAD parse.

image

PR #5924, opened 2026-09-01
fix(version): reproducible dev build identity from commit date

YauhenBichel:
What: The development build version (0.1.Y.M.D+main.<sha>) took its date from datetime.now(UTC), so the same commit produced a different version / Sentry release / telemetry id on every calendar day.

Fix: _dev_build_version now reads the date from the HEAD commit's committer timestamp via a new read_git_head_commit_date() in build_info.py — a filesystem read of the zlib-deflated loose commit object, consistent with the module's existing no-subprocess design. Result is deterministic per commit. Fallbacks: base+main.<sha> when the commit object is packed/unreadable (still reproducible via sha), and bare base when git metadata is absent.

Tests: fixed-commit determinism, packed-object → sha-only fallback, and direct read_git_head_commit_date coverage (loose-object parse + packed → None).

Follow-up to the Greptile review on #5920.

### DIFF
--- a/config/runtime_metadata/build_info.py
+++ b/config/runtime_metadata/build_info.py
@@ -175,117 +175,24 @@ def _parse_committer_date(commit_object: bytes) -> datetime | None:
     return None
 
 
-_PACK_IDX_MAGIC = b"\xfftOc"
-_PACK_IDX_V2 = 2
-_PACK_OBJ_COMMIT = 1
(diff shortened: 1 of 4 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5927, opened 2026-09-01
fix(scheduler): make task-store writes crash-safe, preserve unreadable stores

### LINKED ISSUE #4927: [BUG] Scheduler task store writes non-atomically; a torn write silently deletes every scheduled task

arnavpopley:
Fixes #4927

`infrastructure/scheduling/scheduler/store.py`'s `_save_raw` wrote the task store with a plain `Path.write_text`, which truncates the destination file before rewriting it. If the process died during that window (SIGKILL, container eviction, power loss, ENOSPC), the JSON was left truncated. `_load_raw` then swallowed the resulting `JSONDecodeError` and returned `[]`, so the scheduler silently reported **no tasks** — and the next `add_task` wrote a fresh list over the damaged file, making the loss permanent. Every other local store in the repo already writes by atomic rename (`integrations/store.py::_atomic_write` is the closest analog); the scheduler task store was the one that didn't.

Two changes, both from the issue's own suggested fix:

1. **`_save_raw` writes atomically.** Through a temp file in the destination directory (`tempfile.mkstemp`), `flush` + `os.fsync`, then `os.replace` — atomic on POSIX and Windows. A crash at any point before the final `os.replace` leaves the *previous* file completely untouched (verified with a test that fails `os.replace` mid-write and asserts the store is byte-identical to before, with no temp-file debris left behind).
2. **A store that cannot be parsed is never overwritten by a subsequent write.** Split `_load_raw` into `_read_raw` (returns `(data, readable)`) and `_load_for_write` (moves an unreadable store aside via `mkstemp`-reserved unique naming — `scheduler_tasks.json.corrupt-<timestamp>-<random>` — before returning `[]`, so schedules stay recoverable from the quarantined file instead of vanishing). Only `add_task` uses `_load_for_write`, since it's the one call that unconditionally writes after loading; `remove_task`/`update_task` only write when they find their target, so an unreadable store they can't search is correctly left completely untouched rather than quarantined for no reason.

```
$ uv run python -c "
from pathlib import Path
import tempfile
from infrastructure.scheduling.scheduler.store import add_task, list_tasks
from infrastructure.scheduling.scheduler.types import Provider, ScheduledTask, TaskKind
store = Path(tempfile.mkdtemp()) / 'scheduler_tasks.json'
for hour in (7, 8, 9):
    add_task(ScheduledTask(name=f'digest-{hour}', kind=TaskKind.DAILY_SUMMARY, cron=f'0 {hour} * * *', timezone='UTC', provider=Provider.TELEGRAM, chat_id='-100', window_hours=24), store_path=store)
print('stored:', [t.name for t in list_tasks(store)])
full = store.read_text(encoding='utf-8')
store.write_text(full[: len(full) // 2], encoding='utf-8')
print('after torn write:', [t.name for t in list_tasks(store)])
add_task(ScheduledTask(name='digest-10', kind=TaskKind.DAILY_SUMMARY, cron='0 10 * * *', timezone='UTC', provider=Provider.TELEGRAM, chat_id='-100', window_hours=24), store_path=store)
print('after next add_task:', [t.name for t in list_tasks(store)])
"
stored: ['digest-7', 'digest-8', 'digest-9']
after torn write: []
after next add_task: ['digest-10']
```
The three original schedules are gone from disk, not merely unreadable.

After (this branch):
```
stored tasks: ['digest-7', 'digest-8', 'digest-9']
after partial write, list_tasks -> []
after next add_task -> ['digest-10']
quarantined files: ['/tmp/.../scheduler_tasks.json.corrupt-1788283816-cr7lylnd']
```
Same live-list behavior after the crash (the corrupt file genuinely can't be parsed, so `list_tasks` correctly can't recover it in place), but the original 3 tasks' data is now preserved byte-for-byte in the quarantined file rather than destroyed — recoverable by an operator, not gone.

```
$ uv run pytest tests/scheduler/test_store.py -q
27 passed in 1.59s

$ uv run pytest tests/scheduler/ -q
182 passed in 1.99s

$ uv run ruff check infrastructure/scheduling/scheduler/store.py tests/scheduler/test_store.py
All checks passed!

$ uv run ruff format --check infrastructure/scheduling/scheduler/store.py tests/scheduler/test_store.py
2 files already formatted

$ uv run mypy infrastructure/scheduling/scheduler/store.py tests/scheduler/test_store.py
```

---

I checked three prior closed PRs against this same issue before writing anything, to understand what had already been tried and why they didn't land. All three were closed the same day, in bulk, by a maintainer telling one contributor to "stop spamming with AI generated PRs" — a contributor-behavior issue, not a code-quality rejection. One of the three (the most refined iteration) had already reached Greptile 5/5 and included a real, reviewer-caught fix: an earlier version of the quarantine step built the aside filename from `time.time()` alone, which a reviewer pointed out would let two corruptions inside the same second collide on one filename and have the second `os.replace` erase the first casualty — exactly the class of loss this function exists to prevent. That got fixed with `tempfile.mkstemp`'s atomic-reservation guarantee instead of a bare timestamp string.

Given that design was already vetted to 5/5 and specifically addressed a subtle race a human reviewer had to point out once already, I implemented my own version of the same approach (read/write split, `mkstemp`-based unique quarantine naming, atomic write via temp-file+fsync+replace) rather than inventing a different one from scratch — but wrote it independently and verified every claim myself rather than copying: I ran the issue's exact repro against current `main` first to confirm the bug was still live (it was, unchanged), then re-derived the read/write-path distinction by tracing all four call sites (`list_tasks`, `add_task`, `remove_task`, `update_task`) — only `add_task` unconditionally calls `_save_raw` after loading, so only it needed the quarantine-before-write behavior; `remove_task`/`update_task` return early without writing when their target isn't found, and quarantining an unreadable store on a call that was never going to write to it would be a needless, surprising side effect.

Key files: `infrastructure/scheduling/scheduler/store.py` — `_read_raw` (the readability-reporting load), `_quarantine_unreadable` (the aside-move), `_load_for_write` (composes them for callers about to write), and the rewritten `_save_raw` (atomic write). `add_task`'s one changed line (`_load_raw` → `_load_for_write`) is the only call-site change; `list_tasks`, `remove_task`, and `update_task` are untouched.

Edge cases tested directly: a crash during `os.replace` itself (verifies the *previous* file survives byte-for-byte and no `.tmp` debris is left — not just "some file exists"); a store that's valid JSON but the wrong shape (`{"tasks": []}` instead of a list), which previously fell through to the same silent-empty path as broken JSON; `remove_task`/`update_task` against an unreadable store, confirming they leave it completely untouched (no quarantine, no write) since they have no target to act on; and two corruptions landing in the same wall-clock second, confirming both recovery copies survive independently rather than the second overwriting the first.

---

### DIFF
--- a/infrastructure/scheduling/scheduler/store.py
+++ b/infrastructure/scheduling/scheduler/store.py
@@ -57,6 +57,25 @@ def _load_raw(store_path: Path) -> list[dict[str, object]]:
     return _read_raw(store_path)[0]
 
 
+def _fsync_parent_dir(path: Path) -> None:
+    """Best-effort directory fsync so an ``os.replace`` survives a power loss on Unix.
+
+    ``os.replace`` is atomic, but the directory entry change it makes is only
+    guaranteed durable once the directory itself is fsynced -- without this, a
+    crash immediately after replace can leave the directory pointing at the
+    old inode again, even though the new file's own contents were fsynced.
+    Windows has no equivalent operation, so this is a no-op there.
+    """
+    if os.name == "nt":
+        return
+    with contextlib.suppress(OSError):
+        dir_fd = os.open(str(path.parent), os.O_RDONLY)
+        try:
+            os.fsync(dir_fd)
+        finally:
+            os.close(dir_fd)
+
+
 def _quarantine_unreadable(store_path: Path) -> None:
     """Move an unparseable store aside so a write cannot destroy it.
 
@@ -71,7 +90,16 @@ def _quarantine_unreadable(store_path: Path) -> None:
     )
     os.close(fd)
     aside = Path(aside_str)
-    os.replace(store_path, aside)
+    try:
+        os.replace(store_path, aside)
+    except OSError:
+        # mkstemp already created the (empty) aside file; a failed replace
+        # would otherwise leave it behind as a misleading empty "recovery"
+        # artifact that isn't actually the quarantined data.
+        with contextlib.suppress(OSError):
+            aside.unlink()
+        raise
+    _fsync_parent_dir(aside)
     logger.error(
         "Scheduler store at %s was unreadable and has been preserved at %s. "
         "Scheduled tasks in it are recoverable from that file.",
@@ -99,7 +127,10 @@ def _save_raw(store_path: Path, data: list[dict[str, object]]) -> None:
     window leaves a half-written file that no longer parses. Every other
     local store in the repo writes through a temp file in the destination
     directory, fsyncs, then ``os.replace``; see
-    ``integrations/store.py::_atomic_write``.
+    ``integrations/store.py::_atomic_write``. The directory fsync afterward
+    (``_fsync_parent_dir``) mirrors
+    ``infrastructure/analytics/provider.py::_write_text_atomic``, so the
+    rename itself survives a power loss, not just the file's own contents.
     """
     store_path.parent.mkdir(parents=True, exist_ok=True)
     serialized = json.dumps(data, indent=2, default=str) + "\n"
@@ -111,6 +142,7 @@ def _save_raw(store_path: Path, data: list[dict[str, object]]) -> None:
             handle.flush()
             os.fsync(handle.fileno())
         os.replace(tmp_path, store_path)
+        _fsync_parent_dir(store_path)
     except Exception:
         if tmp_path:
             with contextlib.suppress(OSError):
--- a/infrastructure/scheduling/scheduler/store.py
+++ b/infrastructure/scheduling/scheduler/store.py
@@ -2,8 +2,12 @@
 
 from __future__ import annotations
 
+import contextlib
 import json
 import logging
+import os
+import tempfile
+import time
 from collections.abc import Mapping
 from pathlib import Path
 from typing import Any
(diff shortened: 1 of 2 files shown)


arnavpopley:
@greptile review

arnavpopley:
Fixed both findings.

**Parent-directory fsync (real gap):** confirmed \`integrations/store.py::_atomic_write\` doesn't do this either, but \`infrastructure/analytics/provider.py::_write_text_atomic\` already has the exact precedent (\`_fsync_parent_dir\`, POSIX-only, no-op on Windows). Added an equivalent helper and call it after both \`_save_raw\`'s and \`_quarantine_unreadable\`'s \`os.replace\` — without it, a power loss (not a process kill — a page-cache-level guarantee, not a process-level one) immediately after replace could leave the directory entry pointing at the old inode even though the new file's own contents were fsynced.

**Failed-quarantine cleanup (secondary, non-blocking):** fixed too since it was cheap and directly related — \`mkstemp\` reserves the aside path by creating an empty file before \`os.replace\` runs; if replace then fails, that empty file was left behind looking like a (empty, wrong) recovery artifact. Now unlinked on failure, matching the existing \`_save_raw\` temp-file cleanup pattern in the same file.

Added a test asserting exactly 2 fsync calls per \`add_task\` (file contents + directory) and a test confirming a failed quarantine replace leaves no empty file and the original unreadable store untouched. Full suite (184 tests) + lint/format/mypy pass.

arnavpopley:
@greptile review

cerencamkiran (approved):

PR #5928, opened 2026-09-01
fix(grafana): key the client cache on credentials and TLS config, not just endpoint

### LINKED ISSUE #4192: [BUG] Grafana client cache reuses stale credentials and TLS configuration

arnavpopley:
Fixes #4192

`get_grafana_client_from_credentials` cached the built client under `creds_{account_id}_{endpoint}` only. A rotated service-account token, a changed Basic Auth pair, or a changed `verify_ssl`/`ca_bundle` for the same account+endpoint silently reused the previously cached client — carrying the old credentials — until process restart. In a long-running process, that means `/verify grafana` or any tool query keeps using stale/revoked credentials after a legitimate rotation.

Fix: fold a SHA-256 fingerprint of the credential/TLS tuple (`api_key`, `username`, `password`, `verify_ssl`, `ca_bundle`) into the cache key, so any change to those inputs builds a fresh client, while an identical config still reuses one. The fingerprint is hashed, never embedded in plaintext — the cache key itself can end up in logs or reprs, and the raw secret shouldn't. On a successful rebuild, the superseded credential version of the same `(account_id, endpoint)` is evicted from the cache so a long-running process doesn't accumulate one stale entry per rotation forever.

I checked for prior attempts before starting: four different contributors tried this issue between July and August, all closed. The first was closed by a maintainer citing "we are making some refactoring around integrations" — I checked whether that refactor landed (`git log` on `integrations/grafana/client.py` since that comment) and found only unrelated feature commits; the cache mechanism itself is unchanged from when the issue was filed, and three more contributors continued attempting this same fix afterward. The most refined of those four reached Greptile 5/5, with one real reviewer-caught refinement along the way: an earlier version derived the quarantine-style identity purely from account+endpoint text; a reviewer found that normalizing the endpoint the same way `GrafanaAccountConfig` does (`.strip().rstrip("/")`) was needed so equivalent-but-differently-typed endpoints don't get treated as separate cache identities. I incorporated that as part of the design rather than rediscovering it independently.

```
$ uv run python -c "
from integrations.grafana import client as grafana_client
import unittest.mock as mock
with mock.patch.object(grafana_client.GrafanaClient, 'discover_datasource_uids', lambda self: {}):
    first = grafana_client.get_grafana_client_from_credentials(endpoint='https://grafana.example.com', api_key='token-one', account_id='acct')
    second = grafana_client.get_grafana_client_from_credentials(endpoint='https://grafana.example.com', api_key='token-two', account_id='acct')
    print('same_object_after_credential_change =', first is second)
    print('second_call_uses_new_token =', second.read_token == 'token-two')
"
```

After (this branch):
```
same_object_after_credential_change = False
second_call_uses_new_token = True
cache size (should be 1, old evicted): 1
```

```
$ uv run pytest tests/integrations/grafana/test_client_cache_key.py -q
8 passed in 1.63s

$ uv run pytest tests/integrations/grafana/ -q
102 passed in 1.29s

$ uv run pytest -k grafana -q
327 passed, 16671 deselected in 10.72s

$ uv run ruff check integrations/grafana/client.py tests/integrations/grafana/test_client_cache_key.py
All checks passed!

$ uv run ruff format --check integrations/grafana/client.py tests/integrations/grafana/test_client_cache_key.py
2 files already formatted

$ uv run mypy integrations/grafana/client.py tests/integrations/grafana/test_client_cache_key.py
```

---

I read the full current `integrations/grafana/client.py` before writing anything, since it has real concurrency machinery around the cache (`_BuildGate`, `_grafana_client_lock`, an in-flight-build map for concurrent callers of the same key) that a naive fix could easily break. The fix only touches how the cache key is *computed* and adds one eviction step at the point the client is already inserted under the lock — it doesn't change the locking, gating, or retry logic at all, so the existing concurrency behavior (verified by the pre-existing `test_client_cache_race.py`, which still passes) is untouched.

I considered embedding the credentials directly in the key (e.g. `f"{account_id}_{endpoint}_{api_key}_{username}..."`) instead of hashing them, which would be simpler, but rejected it: cache keys are the kind of string that ends up in a log line, a `repr()`, or a debugger session during an investigation, and a raw API token sitting in the key defeats the purpose of the redaction work already done elsewhere in this codebase (e.g. `infrastructure/observability/trace/redaction.py`). Hashing gets the same uniqueness property (any credential change produces a different key) without ever writing the secret into a string that isn't already a designated "this holds a secret" field.

For eviction, I scope the cleanup to keys sharing the *same* `(account_id, endpoint)` identity prefix (up to and including the separator character), not a global sweep — a rotation for one account must never touch another account's cached client, which is exactly why I added `test_different_account_id_does_not_evict_the_other_accounts_client`. I traced through the string-matching manually to confirm `key.startswith(identity + "#")` can't false-positive across two different identities: since the separator is only appended once, right after the exact normalized endpoint, a different (shorter or unrelated) identity's prefix can never accidentally satisfy another's `startswith` check without also matching the separator at the exact same position.

Key files: `integrations/grafana/client.py` — `_cache_key_identity` (the account+endpoint identity, endpoint-normalized), `_cache_key` (identity + credential fingerprint), and the eviction block inside `_build_and_cache_client`'s existing locked section. `get_grafana_client` (the env-var entry point) needed no changes — it already delegates to `get_grafana_client_from_credentials`, so it picks up the fix for free.

Edge cases tested directly: an endpoint differing only by a trailing slash still sharing one cache entry (the normalization matters both for cache-hit behavior and for not treating `https://x.com` and `https://x.com/` as different identities that could each accumulate their own stale entries); Basic Auth and TLS config changes independently of `api_key`, since the bug report calls out all four as affected; and the eviction not crossing account boundaries.

---

### DIFF
--- a/integrations/grafana/client.py
+++ b/integrations/grafana/client.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import hashlib
 import logging
 import threading
 from dataclasses import dataclass, field
@@ -44,6 +45,55 @@ class _BuildGate:
 
 _grafana_build_inflight: dict[str, _BuildGate] = {}
 
+#: Hex chars of the credential/TLS fingerprint folded into the cache key.
+#: 16 hex chars = 64 bits of SHA-256 -- collision-safe for the handful of live
+#: (account, endpoint) configs a process holds, while keeping the key short.
+#: The raw secrets never appear in the key itself (it can leak into logs or
+#: reprs); only this hash does.
+_FINGERPRINT_HEX_LEN = 16
+#: Separates a cache key's ``(account, endpoint)`` identity from its trailing
+#: fingerprint. Two keys sharing the identity but differing in fingerprint are
+#: different credential versions of the same account+endpoint -- the older one
+#: is evicted on rotation.
+_FINGERPRINT_SEP = "#"
+
+
+def _cache_key_identity(*, account_id: str, endpoint: str) -> str:
+    """Identity shared by every credential version of one (account, endpoint).
+
+    Normalizes the endpoint the same way ``GrafanaAccountConfig`` does
+    (``strip().rstrip("/")``) so equivalent-but-differently-typed endpoints
+    map to the same identity.
+    """
+    normalized_endpoint = endpoint.strip().rstrip("/")
+    return f"creds_{account_id}_{normalized_endpoint}"
+
+
+def _cache_key(
+    *,
+    endpoint: str,
+    api_key: str,
+    account_id: str,
+    username: str,
+    password: str,
+    verify_ssl: bool,
+    ca_bundle: str,
+) -> str:
+    """Build a cache key that changes whenever auth or TLS config changes.
+
+    Keying only on (account_id, endpoint) let a rotated token, changed Basic
+    Auth, or a changed verify_ssl/ca_bundle silently reuse the previously
+    cached client until process restart. The credential tuple is hashed, never
+    embedded in plaintext, so a rotation yields a fresh key without leaking the
+    secret into the key string.
+    """
+    fingerprint = hashlib.sha256(
+        repr(
+            (api_key.strip(), username.strip(), password.strip(), verify_ssl, ca_bundle.strip())
+        ).encode()
+    ).hexdigest()[:_FINGERPRINT_HEX_LEN]
+    return f"{_cache_key_identity(account_id=account_id, endpoint=endpoint)}{_FINGERPRINT_SEP}{fingerprint}"
+
 
 class GrafanaClient(LokiMixin, TempoMixin, MimirMixin, GrafanaClientBase):
     """Unified client for querying Grafana Cloud Loki, Tempo, and Mimir."""
@@ -76,7 +126,15 @@ def get_grafana_client_from_credentials(
     ca_bundle: str = "",
 ) -> GrafanaClient:
     """Create a Grafana client from integration credentials."""
-    cache_key = f"creds_{account_id}_{endpoint}"
+    cache_key = _cache_key(
+        endpoint=endpoint,
+        api_key=api_key,
+        account_id=account_id,
+        username=username,
+        password=password,
+        verify_ssl=verify_ssl,
+        ca_bundle=ca_bundle,
+    )
     cached = _grafana_client_cache.get(cache_key)
     if cached is not None:
         return cached
@@ -198,5 +256,17 @@ def _build_and_cache_client(
         )
 
     with _grafana_client_lock:
+        # Evict superseded credential versions of the same (account, endpoint):
+        # on rotation a fresh fingerprint would otherwise leave the old
+        # token/password cached forever.
+        identity_prefix = (
+            f"{_cache_key_identity(account_id=account_id, endpoint=endpoint)}{_FINGERPRINT_SEP}"
+        )
+        for stale_key in [
+            key
+            for key in _grafana_client_cache
+            if key != cache_key and key.startswith(identity_prefix)
+        ]:
+            del _grafana_client_cache[stale_key]
         _grafana_client_cache[cache_key] = client
     return client
--- a/integrations/grafana/client.py
+++ b/integrations/grafana/client.py
@@ -41,9 +41,24 @@ class _BuildGate:
     done: threading.Event = field(default_factory=threading.Event)
     client: GrafanaClient | None = None
     error: BaseException | None = None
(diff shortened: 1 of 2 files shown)


arnavpopley:
@greptile review

arnavpopley:
Investigated the P1 finding and I don't think it holds — pushing back with evidence rather than complying, since the code disproves it.

**Stripping does match actual client behavior.** The finding says \`GrafanaAccountConfig\` uses \`read_token\`/\`username\`/\`password\`/\`ca_bundle\` verbatim, so fingerprinting the stripped values could collide two distinct credentials. But \`GrafanaAccountConfig\` extends \`StrictConfigModel\` (\`config/strict_config.py\`), which has a wildcard \`@field_validator("*", mode="before")\` that strips **every** string field on **every** subclass — not just the fields with their own explicit validator. Verified directly:

\`\`\`
>>> GrafanaAccountConfig(account_id='a', instance_url='https://x.com', read_token='token ', username=' alice', password='secret ').read_token
'token'
\`\`\`

So the built client's actual \`read_token\`/\`username\`/\`password\` are already normalized regardless of what's passed in — the client does **not** use these fields verbatim, contrary to the finding's premise. My fingerprint stripping matches that real behavior exactly: two inputs the client normalizes to the same value correctly share one cache entry (added \`test_whitespace_only_difference_shares_cache_entry\` to pin this), and two inputs that normalize differently still get different fingerprints and different clients (already covered by the existing rotation tests). Removing the \`.strip()\` calls, as the finding suggests, would make the fingerprint diverge from real behavior in the *other* direction — treating functionally-identical credentials as different, building redundant clients for no real difference. I clarified this in the docstring rather than changing the behavior.

**The concurrent-rotation ordering note is a real but genuinely non-blocking observation**, as Greptile itself labeled it, and I'm leaving it as-is: it requires an old-credential build and a new-credential build to be in flight concurrently *and* the old one to finish strictly after the new one — a narrow, adversarial ordering that isn't how a caller normally rotates (sequentially, not by spraying old and new credentials concurrently). Even in that window, it self-heals on the very next call using the current credentials, since that call misses the cache under its own fingerprint and rebuilds. Adding cross-key coordination for this would be a real complexity increase (the build-gate/lock machinery is currently scoped per cache-key, not per account+endpoint identity) for a transient, self-correcting edge case — not proportionate to this fix's scope.

Pushed a docstring update explaining the actual rationale (with the verified repro above) plus the new whitespace test. Full suite (9 new + 102 grafana + lint/format/mypy) passes.

arnavpopley:
@greptile review

arnavpopley:
Fixed the concurrent-rotation ordering race properly rather than dismissing it a second time — re-examined it and concluded it's real enough to be worth fixing at low cost, even though the impact is transient/self-healing rather than permanent data loss.

Reproduced it directly: two threads, one building with old credentials artificially slowed to finish second, one building with new credentials finishing first. Confirmed on the pre-fix code that the old build's completion evicted the new build's cache entry, exactly as the flagged sequence diagram describes.

Fix: each build leader gets a monotonic sequence number when it starts (assigned under the existing lock, so no new lock needed). A build only wins the cache slot — evicting older-fingerprint entries for its identity and inserting itself — if no newer-sequenced build for that same (account, endpoint) identity has already won. A build that loses this race still returns its own correctly-built client to its own caller; it simply isn't cached for reuse, so the next call for the current credentials finds nothing and rebuilds correctly (rather than the previous behavior, where the stale client would sit in the cache being silently reused).

Verified the fix directly resolves the exact race: with old-slow/new-fast threads, the cache now correctly ends up holding only the new client, and a subsequent call with the current credentials reuses it without an unnecessary rebuild. Added \`test_a_slow_old_build_finishing_after_a_new_one_does_not_win_the_cache\`, ran it 5x standalone to check for flakiness (real threads, no mocked timing beyond a sleep to force ordering) — stable every time.

Full suite (10 cache-key tests + 104 grafana + 329 broader \`-k grafana\`) + lint/format/mypy pass.

arnavpopley:
@greptile review

muddlebee:
This should be verified locally with grafana setup. Last we checked all was goood

PR #5929, opened 2026-09-01
fix(shell): advertise Ctrl+P for the plan overlay instead of Alt/Option+P

YauhenBichel:

The pinned task-plan overlay hinted "Alt/Option+P to view all", but macOS
terminals don't send the Option key as Meta by default, so pressing it did
nothing and users couldn't expand a collapsed plan.

Both keys were already bound to toggle the overlay. This change makes Ctrl+P
the advertised, primary key — it emits a control code on every terminal and
keyboard layout with no Meta-key configuration — and keeps Alt/Option+P bound
as a secondary for terminals set up to send Meta.

Changes:
- Hint text now reads "Ctrl+P to view all".
- `plan_expand.py`: Ctrl+P binding listed first; docstring and comments explain
  why it is the advertised key. Alt/Option+P (`escape`, `p`) still bound.
- `prompt_builder.py`: wiring comment updated.
- Tests updated to assert the new hint and ordering.

No behavior change beyond the advertised key: the toggle is still gated on a
plan with steps being on screen, and both keys fall through to their defaults
(Ctrl+P → history) when no plan is pinned.

### DIFF
--- a/surfaces/interactive_shell/runtime/core/prompt_builder.py
+++ b/surfaces/interactive_shell/runtime/core/prompt_builder.py
@@ -94,7 +94,7 @@ class PromptBuilder:
         # hook so a selection change repaints immediately.
         confirm_kb = install_confirmation_key_bindings(self.state, self._invalidate_prompt)
         install_session_key_bindings(self.pt_session, confirm_kb)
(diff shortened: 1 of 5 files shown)


YauhenBichel:
@greptile review

PR #5930, opened 2026-09-01
style(theme): soften text, base surface, and accents for less eye strain

YauhenBichel:
The shell's palette was high-glare over long sessions: a near-pure-black base
(`#0A0A0A`) under near-white body text (`#E0E0E0`) gives ~15:1 contrast, and the
per-theme accents were quite saturated. This softens all eight themes so the UI
is calmer to read, without dropping below the contrast floors the theme test
enforces.

Changes (all themes, in `infrastructure/terminal/theme.py`):
- Base surface `BG` `#0A0A0A` → `#15161A` (soft near-black instead of pure black).
- Body `TEXT` `#E0E0E0` → `#C8CCD4` (softer gray, less glare).
- `INPUT_SURFACE` `#141414` → `#1C1D22`, kept just above the base so the prompt
  reads as a subtle raised surface.
- `HIGHLIGHT` and `BRAND` accents desaturated ~35% per theme (hue and brightness
  preserved). `mono` has no saturation so it is unchanged; `WARNING` and `ERROR`
  are left as-is so status colors keep their salience.


YauhenBichel:
@greptile review

PR #5931, opened 2026-09-01
feat(shell): collapse tracebacks, frame output, soften text further

YauhenBichel:
Three readability changes to the command-output path:
- Frame every result under the ↳ gutter (drop the longest-line fit gate
  that flushed a whole block when one line was wide).
- Collapse a Python traceback in output to "ExcType: message" + the
  innermost frame + an (N more frames hidden) note; non-tracebacks are
  untouched.
- Soften body TEXT #C8CCD4 → #B6BAC2 (~9.3:1) across the softened themes,
  matching a Droid-like comfort level; contrast stays above the theme
  test's floors.


YauhenBichel:
@greptile review

PR #5932, opened 2026-09-01
fix(shell): clear a stale pinned plan when a new turn starts idle, always cap command output to a short head in the transcript

YauhenBichel:
1. Interrupting a task left its Plan overlay pinned; the next typed prompt
still rendered the old plan because session.task_plan was never reset. On
a genuine typed turn with no dispatch running, clear it — a leftover plan
from a finished or interrupted task must not bleed into the new turn. A
still-running task keeps its plan.

2. A full command dump (JSON blobs, long listings) is noise once
the agent has read it — the reply summary carries the answer. print_command_output
now shows a 10-line head under the ↳ gutter and folds the rest as
"… +N more lines", always, matching Claude Code / Droid / Cursor.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5937, opened 2026-09-02
feat(shell): readable tool output — hide raw data dumps, one-line calls, plan overlay

YauhenBichel:
Make the interactive-shell turn read like Claude Code / Droid instead of a wall
of raw data. When the agent runs a tool (e.g. `gh api`), the transcript was
dumping the raw JSON response under the reply; this hides that noise and keeps
the turn to: the tool call, the `∴` prose summary, and (for plain-text tools)
the real output.

- **Hide raw data blobs from the console** (`action_driver.py`): a generic tool
  result that is a JSON/record blob — valid, truncated (capped `gh api`
  responses arrive as invalid JSON), or a mid-object fragment — is detected by
  shape plus `":` key density (`_is_data_blob`) and dropped from the transcript;
  the reply summary carries the answer. Genuine plain-text output (logs, `ls`,
  `git`) still shows. The generic display cap is tightened and shows a
  `… N more lines` peek.
- **Collapse echoed dumps in the reply** (`streaming/renderer.py`): if the model
  pastes a raw block into its prose, it folds per paragraph to a one-line marker
  and keeps the summary around it.
- **One-line tool call** (`action_rendering.py`): `⏺ GitHub CLI · gh api …`
  instead of a label line plus an indented block.
- **Plan overlay, not header dumps** (architecture_audit `SKILL.md`): the skill
  drives the pinned `update_plan` overlay rather than printing `### [n/9]`
  headers into the transcript.
- Adds `infrastructure/terminal/peek.py` (peek + expand-marker helpers).

### DIFF
--- a/core/agent_harness/prompts/skills/architecture_audit/SKILL.md
+++ b/core/agent_harness/prompts/skills/architecture_audit/SKILL.md
@@ -36,18 +36,16 @@ HARD RULES (violating any = failed turn):
   + save observations + final report.
 
 STEP LABELING RULES (UX) — this skill hides its shell output (quiet=true), so
-narrate the 9 steps of the compact sequence:
-- Before every step's tool calls, emit this exact header format as assistant
-  text in the SAME response as the tool calls, then one short status sentence:
-    ### [n/9] <step name>
-    <One-sentence status.>
-- Never start tool calls for a new step without its header.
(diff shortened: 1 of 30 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5938, opened 2026-09-02
refactor(turns): extract tool-output display text into its own module

YauhenBichel:
Moves the tool-result → transcript-text formatting out of the 1320-line
action_driver.py into a focused core/agent_harness/turns/display_text.py, so
the turn driver is about driving the turn and there is one obvious home for
"how a tool result becomes the line the user sees."

## Why

action_driver.py had accreted a cohesive cluster of display logic — hiding raw
data blobs, capping verbose output to a short head, peeling expand markers,
stripping echoed plan snapshots — interleaved with the tool-calling loop. It
was hard to find and mixed two concerns. The cluster is self-contained (its
only couplings, _content_to_text and _HOST_RENDERED_TOOL_NAMES, moved with it),
so it lifts out cleanly.

## Changes

- New module core/agent_harness/turns/display_text.py holding
  format_generic_tool_payload, cap_for_display, is_data_blob, looks_like_json,
  split_output_truncation_markers, strip_plan_snapshots,
  preferred_tool_response_text, the display-cap constants, and their private
  helpers. Public names lost the leading underscore now that they are a real
  module API; genuinely internal helpers keep it.
- action_driver.py imports from the new module and no longer imports peek
  directly (only the moved code used it).
- Three test modules import the display helpers from their new home.

### DIFF
--- a/core/agent_harness/turns/action_driver.py
+++ b/core/agent_harness/turns/action_driver.py
@@ -42,6 +42,15 @@ from core.agent_harness.session.pending_choice import parse_ask_user_answers
 from core.agent_harness.session.terminal_access import execute_cli_onboard_on_missing_key
 from core.agent_harness.session_goal.goal import strip_session_goal_progress_tags
 from core.agent_harness.turns.conversation_recording import record_conversation_turn
+from core.agent_harness.turns.display_text import (
+    cap_for_display,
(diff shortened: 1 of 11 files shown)


YauhenBichel:
@greptile review

PR #5939, opened 2026-09-02
refactor(text): share one data-blob detector across both hide policies

YauhenBichel:
The interactive shell decided "is this text a JSON/record dump to keep out of
the transcript?" in two hand-rolled places with different thresholds:

- core `is_data_blob` (hides a tool-result payload): opens with `{`/`[`, or
  ≥ 2 `":` keys, no size floor.
- streaming renderer `_looks_like_raw_dump` (collapses a dump the model pasted
  into its reply): ≥ 200 chars and (≥ 3 `":` keys, or structural ratio ≥ 0.15,
  or the "output truncated" marker).

Same question, two divergent answers — a payload with 2 keys and 150 chars was
hidden by one and shown by the other. That drift is what made the raw-JSON
leak whack-a-mole to fix.

One detection mechanism in a dependency-free leaf,
infrastructure/text/data_blob.py — `looks_like_data_blob(text, *, min_json_keys,
min_chars=0, honor_open_bracket=False, structural_ratio=None,
truncation_marker=None)`. Each caller keeps its own policy, now stated
explicitly on top of the shared code, so the two sites cannot silently drift
apart again. The reply side stays deliberately more conservative than the
tool-payload side (collapsing real prose is worse than showing a small blob) —
that difference is now intentional and tested, not accidental.

No behavior change: each branch of the two originals was mapped before the
rewrite.

### DIFF
--- a/core/agent_harness/turns/display_text.py
+++ b/core/agent_harness/turns/display_text.py
@@ -19,6 +19,7 @@ from infrastructure.terminal.peek import (
     cap_output_for_display,
     format_view_all_marker,
 )
+from infrastructure.text import looks_like_data_blob
 
 # Tools whose result the host already rendered to the console; their payload is
 # not re-shown in the transcript.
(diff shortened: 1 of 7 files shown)


YauhenBichel:
@greptile review

PR #5941, opened 2026-09-02
fix(yandex-cloud): carry the command hand-off into the database tool descriptions

nowhere-in-space:
Part of #4605

The Yandex Cloud SKILL.md says that when a finding calls for a change, the answer should end with the exact `yc ...` command an operator can paste. That slice attaches to the two generic readers, `find_yc_api` and `execute_yc_operation`, so an investigation that only touched the Managed Databases tools never sees it - and ends with "return the timeout to its previous value" instead of the line that does it.

For a read-only integration that is the whole deliverable. It cannot change anything, so the command it writes out is the only remediation it can produce.

This carries the one relevant sentence into the three database tool descriptions: `list_yc_db_clusters`, `get_yc_db_cluster`, `read_yc_db_logs`. A sentence rather than adding those tools to the SKILL frontmatter, which would attach the full 2400-character slice to each of them - three copies of a document to deliver one idea. A test keeps the constant short so nobody later replaces it with the slice after all.

https://github.com/user-attachments/assets/defcbbc0-a45e-4294-bbe1-b547530b1b13

The wider version of this change added a line to the shared investigation prompt, since the mitigation phase is where the final wording is decided and no integration has a hook into it. I dropped that on purpose. It would have changed how every investigation ends, for every integration, to fix something observed in one - and the wording I had drafted told the model to reach for "the CLI of whichever platform the evidence came from", which is wrong for the sources that have no CLI at all. A Sentry or Jira finding is fixed by a person, not an invocation.

So this stays inside the integration that has the problem.

It is worth being straight about what this does and does not fix, because I checked afterwards and it is less than I hoped. It closes the case where the instruction was absent: a database-only investigation now carries it, and the demo above shows that against `origin/main`. It does not change the answer in the run I have: with the sentence in place, and `execute_yc_operation` called alongside so the slice was in context too, the conclusion was prose again. The same question against the same cluster on a stronger model does produce the command. So the variable that moved the outcome was the model, not the delivery of the instruction.

I am proposing it anyway because absent guidance is still worth fixing on its own terms, and because the next lever is a different one: hand the command shape back as data in the tool result - the cluster id is already there - rather than as an instruction the model has to remember to follow. That is a larger change and belongs in its own PR, with this measurement behind it. Happy to hold this one if you would rather see them together.

Edge cases: the sentence says these tools only read, which keeps it from reading as permission to run anything - the same SKILL.md separately forbids shelling out to `yc`, and the integration holds a hard read-only guarantee. The constant is defined once per module with a comment explaining why it is a sentence and not the slice, and asserted under 200 characters so the reasoning survives the next person to touch it.

---

### DIFF
--- a/integrations/yandex_cloud/mdb_catalog.py
+++ b/integrations/yandex_cloud/mdb_catalog.py
@@ -14,6 +14,14 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import Final
 
+#: Carried in every tool description of this family rather than as the full
+#: SKILL.md slice: the slice is 2400 characters and would be duplicated into each
+#: description, while the one thing an investigation loses without it is this
+#: sentence. A read-only integration cannot change anything, so the command it
+#: writes out is the remediation - and an investigation that only touched these
+#: tools never saw the generic readers where the slice is attached.
+READ_ONLY_HANDOFF: Final = "These tools only read, so when the finding calls for a change, end with the exact `yc ...` command an operator can paste."
+
 
 @dataclass(frozen=True)
 class ManagedDatabase:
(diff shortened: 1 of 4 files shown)


nowhere-in-space:
@greptile review

nowhere-in-space:
@greptile review

davincios:
Closing as part of the open-source roadmap refresh: investigation, evidence-mapper, AI-SRE, and refactor-only work is no longer on the public roadmap, which now focuses on scheduled tasks, CI/CD skills, and parallelization. Thank you for the contribution — apologies that the direction change lands on in-flight work.

PR #5942, opened 2026-09-02
feat(shell): tighter prompt stack, consistent tool-output hiding

YauhenBichel:
Two interactive-shell readability improvements.

- Tighter prompt stack. The prompt region no longer reserves a shimmer row for   the running action. Tools already paint a ⏺ line into scrollback, so the   reserved row only left a sparse gap (Plan → Thinking → Auto with  none) and made the bordered input box jump up and down as the row filled and  cleared mid-turn. The stack now stays put for the whole turn, and the spinner  phase carries "Invoking tools…" on its own.

- Consistent tool-output hiding. The inline ⏺ result preview decided "is this a  JSON/record blob to hide?" with its own private copy of the check, a third  one beside the two the shared detector already replaced. It now uses the same
  is_data_blob, so a payload is hidden or shown the same way everywhere in the  shell — no more a blob hidden in one spot and shown in another.

- Live status is one row: `Thinking…` / `Invoking tools… · {tool}` with a light-wave shimmer; no reserved blank action row or composer height jump.
- Idle shows a clipped `Ready · / for commands…` hint; footer drops competing `TERMINAL ■` chrome; confirm block density matches the streaming stack.
- Ctrl+O cycles a bounded peek ring (inline expand vs pager), strips CSI before painting, and single-flights expands; GitHub repo + Grafana metrics/logs get short prose `summary` fields.
- Retires unused cyberpunk verbs / dead shimmer paint; shares the aggressive `is_data_blob` policy for transcript hide.

### DIFF
--- a/core/agent_harness/turns/action_dedup.py
+++ b/core/agent_harness/turns/action_dedup.py
@@ -1,0 +1,167 @@
+"""Suppress replayed action-tool calls within one turn.
+
+A guarded tool (``slash_invoke`` / ``shell_run`` / ``cli_exec``) that the model
+re-emits with identical arguments after it already succeeded this turn is
+blocked, so a stuttering model cannot run the same side-effecting command
+twice. Kept apart from the turn driver: this is a self-contained hook policy,
+not part of driving the loop.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Sequence
+from typing import Any
+
+from core.llm.types import ToolCall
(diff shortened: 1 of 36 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #5952, opened 2026-09-02
feat(shell): render one-shot answers, make denials actionable, align multi-select echo

YauhenBichel:
Three interactive-shell / CLI polish fixes so the surfaces users actually touch
read cleanly.

- One-shot `ask` renders its answer. On a terminal the answer is rendered as
  Markdown (bold, lists, fenced code) through the shared theme — the same prose
  quality as the interactive shell — instead of printing raw `**...**`. Piped or
  `--json` output stays plain, ANSI-free text so scripts still parse it.

- `ask` approval denials are actionable. A one-shot has no prompt to approve at,
  so the old bare `Approval denied for: shell_run` was a dead end. It now names
  the exact flags that unblock the run:

      Approval denied for: shell_run
      Re-run with --allowed-tool shell_run to authorize these,
      or --dangerously-bypass-approvals to allow all.

- Multi-select choice echo is aligned. A multi-select answer arrives as one
  option per line, but the persisted echo indented only the first line, leaving
  the rest flush-left and ragged. Every selected line is now indented under the
  ✓ heading so the set reads as one block.

### DIFF
--- a/surfaces/cli/ask/service.py
+++ b/surfaces/cli/ask/service.py
@@ -198,6 +198,21 @@ def _successful_turn(result: TurnResult) -> bool:
     )
 
 
+def _approval_denied_message(denied_tools: tuple[str, ...]) -> str:
+    """Name the denied tools and the exact flags that authorize them.
+
+    A one-shot ``ask`` has no prompt to approve at, so a bare "denied" is a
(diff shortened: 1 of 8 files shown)


YauhenBichel:
@greptile review

PR #5955, opened 2026-09-02
feat(shell): OpenSRE visual identity — purple default, fully themed chrome, fix parity spacing

YauhenBichel:
Makes the interactive shell more user friendly

## Changes

- Purple is the default theme (OpenSRE identity, not warm-gold copycat chrome).
  amber/orange stay available via /theme for side-by-side comparison.
- Every component now follows the active palette. Two colors were hardcoded and
  stayed orange under any theme:
  - the ∴ reply marker / user-row accent / spinner glyph now resolve from the
    theme HIGHLIGHT;
  - the ⏺ tool-call payload (e.g. `⏺ command · /exit`) used the WARNING amber;
    it now uses the neutral SECONDARY so it never reads as off-theme orange.
- Transcript rhythm matches Droid: a blank row between turns and a blank row
  between the user echo and its reply, applied in every render path (TTY
  streaming, non-TTY, deferred gather).
- Exit block stays quiet grey (resume hint + goodbye) — not accented.
- Banner: ring "loops" mark, full build version, "Welcome to OpenSRE CLI" title
  and one-line product description (matching the sign-in screen), centered hero.
- Assistant reply hangs in the ∴ gutter; warmer assistant voice.

### DIFF
--- a/config/constants/__init__.py
+++ b/config/constants/__init__.py
@@ -309,6 +309,7 @@ from config.constants.posthog_mcp import (
 )
 from config.constants.product import (
     FORCE_SIGN_IN_ENV,
+    PRODUCT_DISPLAY_NAME,
     PRODUCT_NAME,
     RELEASE_STAGE,
     RELEASE_STAGE_BANNER,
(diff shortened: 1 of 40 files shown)


YauhenBichel:
@greptile review

PR #5973, opened 2026-09-02
feat(shell): OpenSRE visual identity — purple default, fully themed chrome, user friendly spacing

YauhenBichel:
Makes the interactive shell read as OpenSRE's own product — as polished with its own identity rather than a look-alike — and fixes two transcript bugs that made ordinary turns render wrong.

## Changes

- Purple is the default theme (OpenSRE identity, not warm-gold copycat chrome);  amber/orange remain available via /theme for comparison.
- Every component now follows the active palette. Two colors were hardcoded and
  stayed orange under any theme:
  - the ∴ reply marker / user-row accent / spinner glyph resolve from HIGHLIGHT;
  - the ⏺ tool-call payload (e.g. `⏺ command · /exit`) used the WARNING amber
    and now uses the neutral SECONDARY, so it never reads as off-theme.
- Transcript rhythm: a blank row between turns and a blank row between the user echo and its reply, applied in every render path (TTY  streaming, non-TTY, deferred gather). Gap placement now lives inside  render_submitted_prompt so each path (plain turn, autosubmit, handoff answer)  places it correctly.

### DIFF
--- a/config/constants/product.py
+++ b/config/constants/product.py
@@ -18,7 +18,7 @@ PRODUCT_DISPLAY_NAME: Final[str] = "OpenSRE"
 #: Sign-in / welcome screen copy, shown when the shell requires login.
 WELCOME_TITLE: Final[str] = "Welcome to OpenSRE CLI"
 WELCOME_DESCRIPTION: Final[str] = (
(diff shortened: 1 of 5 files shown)


YauhenBichel:
@greptile review

PR #5991, opened 2026-09-03
test(sandbox): cover Windows SystemRoot forwarding

### LINKED ISSUE #4937: [BUG] Sandbox allow_network is broken on Windows: SystemRoot missing from the subprocess env breaks Winsock (WinError 10106)

xuhaonan013:
Fixes #4937

- Add a focused regression test for the Windows sandbox subprocess environment.
- Verify that `SYSTEMROOT` reaches the actual `subprocess.run(..., env=...)` call used by `run_python_sandbox`.
- Keep production behavior unchanged because current `main` already contains the allowlist fix from #5850.

```text
uv run pytest tests/sandbox/test_runner.py tests/tools/test_python_execution_tool.py::TestPythonExecutionToolRestrictions::test_network_access_can_be_enabled tests/tools/test_python_execution_runtime_inputs.py::test_socket_reachability_check_works_with_allow_network
============================= 30 passed in 28.04s =============================

uv run python -m ruff check bootstrap config core gateway integrations infrastructure surfaces tools tests
All checks passed!

uv run python -m ruff format --check bootstrap config core gateway integrations infrastructure surfaces tools tests
3105 files already formatted

uv run mypy bootstrap config core gateway integrations infrastructure surfaces tools
Success: no issues found in 1800 source files

uv run python -m build
Successfully built opensre-0.1.tar.gz and opensre-0.1-py3-none-any.whl
```

---

The Windows failure was caused by the sandbox's narrow environment omitting the Windows root variable. Current `main` already forwards `SYSTEMROOT`, so this PR avoids duplicating production code and instead protects that contract. The test sets a deterministic Windows-style value, replaces only the runner's subprocess boundary, executes `run_python_sandbox`, and asserts that the exact environment supplied to the child contains the value. This remains platform-independent while directly covering the regression point.

---

### DIFF
--- a/tests/sandbox/test_runner.py
+++ b/tests/sandbox/test_runner.py
@@ -13,6 +13,7 @@ from typing import Any
 import pytest
 
 from config.constants import OPENSRE_TMP_DIR, ensure_opensre_tmp_dir
+from infrastructure.safety.sandbox import runner as sandbox_runner
 from infrastructure.safety.sandbox.runner import (
     MAX_TIMEOUT,
     SandboxResult,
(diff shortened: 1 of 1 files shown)


xuhaonan013:
CI note: the failed `test (cli-live-agent)` job is unrelated to this sandbox test. On fork PRs the workflow sets `-m "not live_llm"` but still runs the live-only `tests/tools/selection` directory, producing `0 items` and pytest exit code 5. I filed #5994 and opened the focused workflow fix in #5995. The same job completes successfully on #5995, whose remaining applicable CI jobs are also green.

For this PR itself, all 27 applicable checks passed, the focused Windows tests pass locally, and Greptile reports 5/5 with no unresolved findings.

PR #5998, opened 2026-09-03
fix(install): preflight glibc compatibility on Linux before download

### LINKED ISSUE #5993: [linux/x86_64/binary] v0.1.2026.9.3 requires GLIBC_2.35 and fails on Ubuntu 20.04
## Summary

The `v0.1.2026.9.3` linux-x64 release archive downloads and verifies correctly, but the binary cannot start on Ubuntu 20.04 because its bundled Python requires `GLIBC_2.35`.

**Expected:** Either the published linux-x64 binary starts on a commonly supported x86_64 Linux distribution, or the release/install documentation and installer preflight clearly state that glibc 2.35+ (Ubuntu 22.04+) is required.

**Actual:** On Ubuntu 20.04.6 LTS with glibc 2.31, every entry point fails before the CLI starts.

## Steps to reproduce

1. Download `opensre_0.1.2026.9.3_linux-x64.tar.gz` and its `.sha256` file from release `v0.1.2026.9.3`.
2. Verify with `sha256sum -c` (passes).
3. Extract with `tar -xzf` (passes).
4. Run `./opensre/opensre --version`.

Yes, every time on this environment.

## Environment

- Ubuntu 20.04.6 LTS (WSL2)
- Linux x86_64, kernel `6.6.87.2-microsoft-standard-WSL2`
- glibc 2.31
- Release tag `v0.1.2026.9.3`, commit `34b77c8`

## Logs and error output

```text
opensre_0.1.2026.9.3_linux-x64.tar.gz: OK
[PYI-23:ERROR] Failed to load Python shared library '/tmp/opensre-5402-20260903/extract/opensre/_internal/libpython3.13.so.1.0': /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.35' not found (required by .../libpython3.13.so.1.0)
```

The release workflow intentionally builds on Ubuntu 22.04 and caps the maximum imported glibc symbol at `GLIBC_2.35`, but the user-facing Linux release requirements do not currently state that compatibility floor.

Found while validating #5402.

muddlebee:
Fixes #5993

The linux-x64/arm64 release binaries are built on `ubuntu-22.04` in `release.yml`, which itself already asserts (`Check Linux binary glibc compatibility`) that the binary must not require newer than `GLIBC_2.35`. That step confirms the binary's floor is exactly glibc 2.35 — but nothing told the *installer* or the *user* about it, so on an older host (e.g. Ubuntu 20.04, glibc 2.31) `install.sh` happily downloaded, extracted, and then failed with a cryptic PyInstaller loader error (`Failed to load Python shared library ... GLIBC_2.35' not found`) exactly as reported in the issue.

This PR adds a preflight check to `install.sh`:
- `detect_glibc_version()` reads the host's glibc version via `getconf GNU_LIBC_VERSION` (falling back to `ldd --version`); known musl systems fail before download, while genuinely inconclusive detection warns and continues.
- `check_linux_glibc_compatibility()` runs right after `detect_platform` in `main()`, before any network call, and `die()`s with an actionable message (detected version, required floor, and a source-install fallback) if the host's glibc is older than `LINUX_MIN_GLIBC="2.35"`.
- Documented the floor in `README.md` and `docs/environments/linux-local.mdx`, next to the existing install instructions.

Rebuilding the release binaries against an older glibc floor (e.g. a manylinux-style container) would be a much larger infra change with its own risk (recompiling every native dependency against an older toolchain); this PR takes the fix the issue itself proposed as sufficient — fail fast with a clear message instead of a cryptic post-download crash.

Simulated on this machine by shimming `getconf` to report an old glibc version (real repro requires an actual Ubuntu 20.04 host, which isn't available here):

```
$ bash install.sh --main
Error: The OpenSRE Linux binary requires glibc 2.35+ (Ubuntu 22.04+, Debian 12+, Fedora 36+, or similarly current distributions). This system reports glibc 2.31, so the prebuilt binary will not start here.
Run OpenSRE from source instead: clone https://github.com/Tracer-Cloud/opensre, install uv (https://docs.astral.sh/uv/getting-started/installation/), then run 'make install' followed by 'uv run opensre'.
```

Exits immediately (before any download), vs. today's flow which downloads ~100MB+ and fails deep inside the PyInstaller loader.

Added regression coverage for incompatible glibc, failed-`getconf` fallback, and musl rejection before network access. The focused installer suite passes on this glibc 2.39 host, confirming compatible installations remain unchanged.

`make lint`, `make format-check`, and `make typecheck` all pass.

---

The bug is a distribution-compatibility gap, not a logic bug: the release pipeline intentionally targets glibc 2.35 (documented by its own CI assertion), and nothing surfaced that requirement to users before they hit a crash. The fix is a shell preflight check placed as early as possible in `main()` (right after platform detection, before any network I/O), using `getconf`/`ldd` — both already broadly available on any glibc Linux host without new dependencies — and comparing against a named constant (`LINUX_MIN_GLIBC`) kept next to a comment pointing back at the CI check it must stay in sync with. It rejects known musl/Alpine hosts because the glibc-linked binary cannot start there, and warns but continues only when the libc cannot be identified. I considered instead rebuilding the release binaries in an older-glibc container (e.g. manylinux2014), but that's a much larger, riskier infra change (recompiling every native wheel dependency against an older toolchain) than what the issue itself calls sufficient — clearly documenting/enforcing the existing floor.

---

---

https://claude.ai/code/session_016xigdmudH12x1BhZtfXJ4o

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -916,9 +916,8 @@ detect_glibc_version() {
   # getconf reports "glibc X.Y" on glibc systems and is part of the C library
   # itself, so prefer it; fall back to parsing `ldd --version` when getconf
   # is missing, or present but unable to report a version (e.g. a non-glibc
-  # getconf that doesn't know GNU_LIBC_VERSION). Empty output (not an error)
-  # means neither could determine a version, e.g. on musl/other libc systems
-  # where this check does not apply.
+  # getconf that doesn't know GNU_LIBC_VERSION). Empty output means no glibc
+  # version was found; the caller distinguishes known musl from an unknown libc.
   local version=""
 
   if command -v getconf >/dev/null 2>&1; then
@@ -935,6 +934,17 @@ detect_glibc_version() {
   fi
 }
 
+is_musl_linux() {
+  local ldd_output
+
+  command -v ldd >/dev/null 2>&1 || return 1
+  ldd_output="$(ldd --version 2>&1 || true)"
+  case "$ldd_output" in
+    *musl*|*MUSL*) return 0 ;;
+    *) return 1 ;;
+  esac
+}
+
 version_ge() {
   [ "$(printf '%s\n%s\n' "$1" "$2" | sort -V | tail -n 1)" = "$1" ]
 }
@@ -945,14 +955,23 @@ check_linux_glibc_compatibility() {
   [ "$platform" = "linux" ] || return 0
 
   glibc_version="$(detect_glibc_version)"
-  # Can't determine the version (non-glibc libc, or missing getconf/ldd) —
-  # don't block install on a check we can't perform.
-  [ -n "$glibc_version" ] || return 0
+  if [ -z "$glibc_version" ]; then
+    if is_musl_linux; then
+      die "$(cat <<EOF
+The OpenSRE Linux binary requires glibc ${LINUX_MIN_GLIBC}+, but this system uses musl libc, so the prebuilt binary will not start here.
+Use a glibc-based distribution or container, or run OpenSRE from source: clone https://github.com/${REPO}, install uv (https://docs.astral.sh/uv/getting-started/installation/), then run 'make install' followed by 'uv run opensre'.
+EOF
+)"
+    fi
+
+    warn "Could not determine this Linux system's libc version. The prebuilt binary requires glibc ${LINUX_MIN_GLIBC}+; continuing without a compatibility guarantee."
+    return 0
+  fi
 
   if ! version_ge "$glibc_version" "$LINUX_MIN_GLIBC"; then
     die "$(cat <<EOF
 The OpenSRE Linux binary requires glibc ${LINUX_MIN_GLIBC}+ (Ubuntu 22.04+, Debian 12+, Fedora 36+, or similarly current distributions). This system reports glibc ${glibc_version}, so the prebuilt binary will not start here.
-Run OpenSRE from source instead: clone https://github.com/${REPO}, install uv (https://docs.astral.sh/uv/getting-started/installation/), then run 'make install' followed by 'uv run opensre'.
+Upgrade to a supported distribution rather than replacing the system glibc manually, or run OpenSRE from source: clone https://github.com/${REPO}, install uv (https://docs.astral.sh/uv/getting-started/installation/), then run 'make install' followed by 'uv run opensre'.
 EOF
 )"
   fi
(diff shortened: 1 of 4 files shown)


muddlebee:
Fixed in 7fee22a8d: `detect_glibc_version()` now only returns early from the `getconf` branch when it actually produced a parseable version, so a present-but-uninformative `getconf` (e.g. non-glibc) falls through to `ldd --version` instead of silently skipping the check. Added a regression test (`test_install_sh_falls_back_to_ldd_when_getconf_cannot_report_version`) covering that fallback.

muddlebee:
@greptile review

install.sh
```diff
@@ -909,6 +912,52 @@ detect_platform() {
...
+
+check_linux_glibc_compatibility() {
+  local glibc_version
+
+  [ "$platform" = "linux" ] || return 0
+
+  glibc_version="$(detect_glibc_version)"
+  # Can't determine the version (non-glibc libc, or missing getconf/ldd) —
+  # don't block install on a check we can't perform.
+  [ -n "$glibc_version" ] || return 0
```
muddlebee: Known non-glibc Linux hosts are still allowed through even though this glibc-linked binary cannot start there. On Alpine/musl, both probes return no parsed version, so the installer proceeds to download and ends in the same opaque loader failure. Detect known unsupported libc implementations and fail with the source-install guidance; reserve fail-open for genuinely indeterminate hosts, and add a musl shim regression test.

muddlebee: Fixed in 8bffef46a: known musl hosts now fail before download with source/container guidance, unknown libc detection warns and continues, and a network-marker regression test covers the musl path.

install.sh
```diff
@@ -909,6 +912,45 @@ detect_platform() {
...
 
+detect_glibc_version() {
+  # getconf reports "glibc X.Y" on glibc systems and is part of the C library
+  # itself, so prefer it; fall back to parsing `ldd --version`. Both are
+  # empty (not an error) on musl/other libc systems, where this check does
+  # not apply.
+  if command -v getconf >/dev/null 2>&1; then
+    getconf GNU_LIBC_VERSION 2>/dev/null | sed -n 's/^glibc \([0-9][0-9.]*\)/\1/p'
+    return 0
+  fi
```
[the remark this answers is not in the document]
muddlebee: Fixed in 7fee22a8d: failed or unparseable getconf output now falls through to ldd, with regression coverage.

muddlebee:
@greptile review

muddlebee:
Superseded by #6284, which documents the Linux glibc compatibility requirement and source-install fallback without changing the installer.

PR #6000, opened 2026-09-03
feat(shell): group tool actions into full-width collapsible sections

YauhenBichel:
Implements "Group tool action rendering" (10.x): tool actions are now compact
and visually secondary to the reply, matching tool chrome instead
of a wall of long dotted lines.

## Changes

- Tool calls are buffered per turn (ActionLogEntry on TerminalSession) and
  flushed as an execution log above the reply — not streamed inline as bright
  primary steps.
- 2+ consecutive same-kind calls (all GitHub CLI, etc.) render as one
  full-width bordered box (title in the top border), one concise status line
  per call. A lone call renders as a single dim line — no one-row box.
- Concise labels only: the `key: value ·` argument strip is gone from the
  visible surface. Full call + result detail moves into the Ctrl+O panel.
- Ctrl+O expands the stashed detail; off a TTY the detail inlines so logs keep
  it. The box spans the full window width, aligned with the input composer.

Two fixes made while testing this path (independent, kept):
- Output folding no longer shreds a bordered table mid-box or over-truncates
  help; tables show whole (peek.py), inline budget raised 4→12 lines.
- The confirmation fallback (when a subprocess owns the TTY) reads one keypress
  in cbreak mode, so arrow keys no longer echo raw `^[[A` (confirm_keys.py).

image

image


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6004, opened 2026-09-03
feat(shell): quiet status bar and composer chrome

YauhenBichel:
## Changes

- Auto status: `auto_status_ansi` collapses its caption when the level is
  the High default (noise on every idle frame) and takes a `quiet` mode so
  it recedes to DIM while Thinking/Invoking owns the accent — one status
  row, not two competing headlines.
- Composer: removed the dedicated footer line under the box; the send hint
  now lives in the empty-box placeholder (`see what you can do · Enter
  send · ? help`). One stack, not a textbook under the box.
- Removed the ↗ answer hand-off marker (unused) and its dead export.
- Plan breakdown: work-note rows are emitted as raw truecolor DIM with
  markup/highlight off, so Rich's Style cache can't emit bright-black
  `[90m` from an earlier 16-color console on a truecolor turn.

### DIFF
--- a/config/constants/__init__.py
+++ b/config/constants/__init__.py
@@ -323,10 +323,12 @@ from config.constants.redis import (
 )
 from config.constants.repl_autonomy import (
     AUTO_LEVEL_ASK_TOOL_TYPES,
+    AUTO_LEVEL_BAR_CAPTIONS,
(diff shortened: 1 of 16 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6005, opened 2026-09-03
fix(shell): close the transcript holes left by inline menus

YauhenBichel:
Inline menus (Ask User, help, choice pickers) and the status region left
empty rows and doubled blanks in the transcript. This closes them so the
shell reads as one tight stack.

## Changes

- choice_menu: `erase_menu_lines` / `_erase_menu_block` gain a `delete`
  mode. On leaving a menu the rows are deleted with `CSI n M` (not just
  blanked with `ESC[J`), so a tall Ask User picker no longer leaves a hole
  between the reply and the ✓ recap, or above Thinking. Redraw still clears
  in place.
- `leave_inline_menu` parks at column zero without a trailing newline (the
  stream already prints one post-reply blank).
- ask_user / help_menu pass `delete=True` when erasing on close.
- terminal_ui: drop the doubled leading blank in the prompt region; keep
  the Auto permission line visible (DIM) while a turn is busy rather than
  replacing it.

### DIFF
--- a/config/constants/__init__.py
+++ b/config/constants/__init__.py
@@ -323,10 +323,12 @@ from config.constants.redis import (
 )
 from config.constants.repl_autonomy import (
     AUTO_LEVEL_ASK_TOOL_TYPES,
+    AUTO_LEVEL_BAR_CAPTIONS,
     AUTO_LEVEL_CAPTIONS,
     AUTO_LEVEL_TITLES,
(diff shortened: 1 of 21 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6012, opened 2026-09-04
perf(install): warm the frozen bundle at install time and bake the tool index

YauhenBichel:
## Problem

A fresh install's first `opensre` took 7–14s to reach the prompt (up to
18s on a cold machine). Every launch after that is ~1s. A 3s load-time
increase costs ~30% of users, so this is the single biggest drop-off in
the funnel.

## Root cause

The `curl` installer ships a PyInstaller onedir bundle: 781 MB, 301
native Mach-O libraries, all ad-hoc signed. macOS validates each image's
code signature on its first dlopen and caches the result. Two things
combine on the user's first launch:

1. The installer re-signs all 301 libs (to fix Invalid-Page SIGKILLs),
   which resets that cache — then verifies with `--version`, the fast path,
   which loads almost none of them. So the user's first real launch pays
   the whole validation bill. Measured on the bundle: re-sign then launch
   = 11.8s to first byte (`real 11.6s` vs `user 1.5s` — the process is
   waiting on the kernel, not computing); the run after = 1.0s.
2. Frozen builds imported every vendor tool module on the first turn
   (+1,881 modules: boto3, botocore, kubernetes, google) because the
   descriptor index is an AST scan of `.py` source the bundle does not
   carry, so it fell back to the full dynamic snapshot.

Ruled out along the way: `configure_process` (CLI profile = env only), the
sign-in gate, banner counters (filesystem only), version resolution (reads
git HEAD from files), integration warming (lazy), the tiktoken bootstrap
(only loads with litellm). Warm imports are 0.67s; dev `uv run opensre` is
3.4s to prompt.

## Fix

- **Install-time warm-up** (`install.sh`): after the re-sign, run the
  hidden `opensre _package-smoke` under the installer spinner. It imports
  the full tool registry, all verifiers and skills, so it dlopens
  essentially the whole library set and absorbs the validation while the
  user is already waiting on the installer. Quiet (JSON suppressed) and
  best-effort — `--version` already gates correctness, so a smoke failure
  warns rather than aborts.
- **Baked descriptor index**: `opensre.spec` runs `dump_descriptor_index()`
  against the checkout at build time and ships the result as
  `tools/descriptor_index.json`; `build_descriptor_index()` loads it when
  `sys.frozen`. `_load_surface_snapshot` now uses the index in frozen
  builds, keeping the surface-scoped load (54 action tools, 840 modules)
  instead of importing all 266 tools. The every-vendor fallback remains only
  for a bundle that shipped without the index.

## Results (rebuilt bundle, real PTY, macOS)

| Scenario | First byte → prompt |
|---|---|
| Before: first launch after install | 11.8–18.2s → 13–19.5s |
| After: first launch (installer ran the warm-up) | 1.56s → 2.8s |

### DIFF
--- a/.github/ci/test_scope_rules.py
+++ b/.github/ci/test_scope_rules.py
@@ -551,6 +551,7 @@ RULES: tuple[PathRule, ...] = (
     ),
     PathRule("infrastructure/safety/guardrails/", ("tests/infrastructure/safety/guardrails/",)),
     PathRule("infrastructure/safety/masking/", ("tests/masking/",)),
+    PathRule("opensre.spec", ("tests/packaging/",)),
     PathRule("infrastructure/deployment/packaging/", ("tests/packaging/",)),
     PathRule("infrastructure/safety/sandbox/", ("tests/sandbox/",)),
     PathRule(
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -384,10 +384,17 @@ jobs:
             echo "No onedir binary at ${APP_BIN}; skip re-sign"
             exit 0
           fi
+          # Nested libs are independent — parallelize with a small cap, then
+          # sign the main binary last (same order as install.sh).
+          JOBS="$(sysctl -n hw.ncpu 2>/dev/null || printf '4')"
+          if [ "$JOBS" -gt 4 ]; then
+            JOBS=4
+          fi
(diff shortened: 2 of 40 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6014, opened 2026-09-04
perf(launch): install the bundle by rename and keep the banner off the skill harness

YauhenBichel:
Two launch-path cuts, both measured on the same machine.

**Install by rename, not copy.** The installer verified the extracted bundle, copied it into `.opensre-app`, then ran the first-launch warm-up on the copy. macOS caches signature validation per file, so the copy paid the whole cold validation again. The extracted tree is now renamed to a staged directory under the install dir, where the re-sign, version check and warm-up run, and then renamed into place. A binary that fails verification is discarded and the existing install stays as it was (`stage_binary` / `verify_staged_binary` / `activate_staged_binary` / `discard_staged_binary`).

**Banner status without the skill harness.** The Skills and Integrations chips on the launch banner were computed through `list_action_skills`, which imports the action-skill harness and prompt_toolkit on the launch path. They now count bundled skill files and active store services from the filesystem, and `infrastructure.terminal`'s prompt exports are lazy, so the banner prints its two numbers without loading the prompt stack. This part was pushed to main by mistake as 635053593 and reverted in dea43449f; it lands here as intended.

## Measurement (same machine, same 17s download)

| | main | this PR |
| --- | --- | --- |
| Local install work after download | 31s | 21s |
| Total install | 47s | 38s |
| First `--version` after install | 0.23s | 0.23s |

### DIFF
--- a/infrastructure/terminal/__init__.py
+++ b/infrastructure/terminal/__init__.py
@@ -1,14 +1,12 @@
-"""Shared terminal presentation primitives (theme, prompts, error rendering)."""
+"""Shared terminal presentation primitives (theme, prompts, error rendering).
 
-from infrastructure.terminal.errors import render_error
-from infrastructure.terminal.prompt_support import (
-    CTRL_C_DOUBLE_PRESS_WINDOW_S,
-    handle_ctrl_c_press,
-    install_questionary_ctrl_c_double_exit,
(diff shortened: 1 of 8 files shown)


YauhenBichel:
@greptile review

PR #6052, opened 2026-09-07
fix(scheduler): preserve manual_loop prompts through creation confirmation

### LINKED ISSUE #6047: manual_loop schedules lose their prompt during creation
## Description

  `manual_loop` schedules do not preserve the prompt that should be executed on each run.

  The CLI creates a `ScheduledTask` without setting its prompt. Additionally, conversational schedule confirmation
  does not include the prompt when converting a pending schedule offer into a slash command.

  ## Relevant code

  - `surfaces/cli/commands/cron.py`
  - `infrastructure/scheduling/scheduler/tasks.py`
  - `core/agent_harness/session/pending_offer.py`

  `ScheduledTask` expects a prompt for manual-loop execution. When the prompt is missing, the task path returns a
  warning instead of executing the requested instruction.

  ## Reproduction

  Create a `manual_loop` schedule with a user instruction, either through the CLI or conversational scheduling
  flow.

  The schedule is created, but the instruction is not persisted with the scheduled task. When the scheduler runs,
  it has no prompt to execute.

  ## Expected behavior

  The prompt provided by the user should be:

  - Stored in the scheduled task.
  - Preserved through confirmation and serialization.
  - Available to the scheduler when the task executes.
  - Executed on every scheduled run.

  ## Actual behavior

  The schedule can be created without a usable prompt. When it runs, the scheduler cannot execute the intended
  instruction and may return a missing-prompt warning.

  ## Impact

  Users can create apparently valid `manual_loop` schedules that do not perform the requested work.

  ## Acceptance criteria

  - Add prompt support to `manual_loop` task creation.
  - Preserve the prompt in `PendingScheduleOffer`.
  - Include the prompt when serializing the offer into a slash command or scheduled task.
  - Add a regression test for CLI-created `manual_loop` schedules.
  - Add a regression test for conversational schedule confirmation.

Devesh36:
Fixes #6047

  ## Description

  `manual_loop` schedules were being created without the prompt they were supposed to execute. As a result, the scheduler accepted the task but
  later returned a missing-prompt warning instead of running the requested instruction.

  This change fixes prompt handling across both supported creation flows:

  - Adds a required `--prompt` option to `opensre cron add --kind manual_loop`.
  - Persists the prompt using the scheduler’s existing `loop_prompt` parameter.
  - Preserves the prompt in conversational schedule offers.
  - Includes the prompt when a confirmed offer is serialized into `/cron add`.
  - Rejects prompts for task kinds that do not support them.
  - Updates cron documentation and adds regression coverage.

  ## Testing

  - 62 focused tests passed.
  - `make lint`
  - `make format-check`
  - `make typecheck`
\

### DIFF
--- a/core/agent_harness/session/pending_offer.py
+++ b/core/agent_harness/session/pending_offer.py
@@ -61,6 +61,7 @@ class PendingScheduleOffer:
     timezone: str
     provider: str
     chat_id: str = ""
+    prompt: str = ""
     skill_name: str = ""
     skill_inputs: dict[str, str] = field(default_factory=dict)
 
@@ -77,6 +78,10 @@ class PendingScheduleOffer:
             "--provider",
             self.provider,
         ]
+        if self.kind == "manual_loop":
+            prompt = self.prompt.strip()
+            if prompt:
+                args.extend(["--prompt", prompt])
         if self.kind == "recurring_skill":
             skill = self.skill_name.strip()
             if skill:
--- a/docs/cron.mdx
+++ b/docs/cron.mdx
@@ -24,14 +24,17 @@ When you open the interactive shell (`opensre`) and no scheduled tasks exist yet
 ```bash
 # Add a free-form recurring ops digest to Telegram at 09:00 IST on weekdays
 opensre cron add --kind manual_loop --cron "0 9 * * 1-5" \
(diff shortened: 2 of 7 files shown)


muddlebee:
LGTM

PR #6058, opened 2026-09-07
fix(aws): do not abort an invocation listing on an unparseable START line

### LINKED ISSUE #6035: [BUG] get_recent_invocations raises IndexError on a Lambda START line with no request id
### Summary

`get_recent_invocations` raises `IndexError` when a CloudWatch event looks like a Lambda `START` line but carries no request id after the label.

**Expected:** an unparseable log line leaves `request_id` unset and the rest of the invocation summary is still returned. That is how the sibling parses in the same block already behave: `Duration` and `Memory Used` are both wrapped in `suppress(IndexError, ValueError)`.

**Actual:** the request-id parse is unguarded and raises. `get_recent_invocations` only catches `ClientError`, so the `IndexError` escapes the function and aborts the whole tool call rather than dropping one field.

The parse is:

```python
request_id = (
    message.split("RequestId: ")[1].split()[0] if "RequestId:" in message else None
)
```

The guard tests for `RequestId:` while the split uses `RequestId: ` with a trailing space, so the guard cannot prevent the failure it appears to be there for. Two inputs break it:

- `START RequestId:req9 Version: $LATEST` (no space after the colon) leaves `split()` returning one element, so `[1]` raises.
- `START RequestId: ` (label with no value) makes `[1]` an empty string, so `.split()[0]` raises.

This matters because `filter_log_events` returns every event in `/aws/lambda/<function>`, the function's own stdout included, so the parser is reading untrusted text rather than only AWS-generated preamble lines.

### Steps to reproduce

```python
from unittest.mock import MagicMock, patch
from integrations.aws.lambda_client import get_recent_invocations

client = MagicMock()
client.filter_log_events.return_value = {
    "events": [{"message": "START RequestId: ", "timestamp": 1700000000000}]
}
with (
    patch("integrations.aws.lambda_client._get_cloudwatch_logs_client", return_value=client),
    patch("integrations.aws.lambda_client.require_aws_credentials", return_value=None),
):
    get_recent_invocations("fn")
```

Yes

### How often does it occur?

Under specific conditions (a `START RequestId:` line with no value, or with no space after the colon)

### Operating system

Windows

### Logs and error output

```
  normal AWS line          -> ok, request_id='8f1d2c3b-0000'
  no space after colon     -> IndexError: list index out of range
  nothing after the label  -> IndexError: list index out of range
```

### Additional context

The `Duration` and `Memory Used` parses use the same `split(label)[1].split()[0]` shape and are already protected, so this is the one place in the block where a malformed line is fatal rather than skipped.

dchaudhari7177:
Fixes #6035

Wrapped the `START RequestId:` parse in `suppress(IndexError, ValueError)` with the field defaulting to `None`, matching the `Duration` and `Memory Used` parses a few lines below.

Before (on `main`) — both shapes abort the whole call:

```
'START RequestId:req9 Version: $LATEST\n'  -> IndexError: list index out of range
'START RequestId: \n'                      -> IndexError: list index out of range
'START RequestId: ok-1 Version: $LATEST\n' -> OK  request_id='ok-1'
```

After — the field is left unset and the rest of the summary survives:

```
no space after colon   success=True count=1 request_id=None logs_kept=1
label with no value    success=True count=1 request_id=None logs_kept=1
normal                 success=True count=1 request_id='ok-1' logs_kept=1
```

Test run:

```
$ pytest tests/integrations/aws/test_lambda_client.py tests/tools/test_lambda_errors_tool.py tests/tools/test_lambda_invocation_logs_tool.py -q
38 passed
```

With the new test in place and `lambda_client.py` reverted: `2 failed, 13 passed`.

---

**The problem.** The parse was

```python
request_id = (
    message.split("RequestId: ")[1].split()[0] if "RequestId:" in message else None
)
```

The guard tests for `RequestId:` but the split uses `RequestId: ` with a trailing space, so the two do not describe the same condition and the guard cannot prevent the failure it appears to exist for. Two shapes get past it: `START RequestId:req9 …` (no space, so `split` returns one element and `[1]` raises) and `START RequestId: ` (label with no value, so `[1]` is `""` and `"".split()[0]` raises).

`get_recent_invocations` catches only `ClientError`, so the `IndexError` propagates out and the caller loses every invocation, not just the one field. That is reachable from ordinary traffic rather than a malformed AWS response: `filter_log_events` returns every event in `/aws/lambda/<function>`, including whatever the function itself prints, so any application log line beginning `START RequestId:` reaches this parser.

1. *Fix the guard to `"RequestId: " in message`* — makes the two agree and stops the first shape, but `START RequestId: ` with nothing after it still passes the guard and still raises on `.split()[0]`. It fixes one of the two inputs and leaves a guard that looks total but is not.
2. *Parse with a regex, `RequestId:\s*(\S+)`* — this would also succeed on `RequestId:req9` and recover `req9`. Tempting, but it changes what the function extracts rather than only how it fails, and #6035 states the expected behaviour as leaving `request_id` unset. I left it out to keep the change to the defect; worth doing separately if you want the no-space form actually parsed.
3. *`suppress(IndexError, ValueError)`* — what I chose.

**Why this one.** The block already establishes the convention two parses down: `Duration` and `Memory Used` are both wrapped in exactly this, on the same reasoning, that one unreadable field should not cost the whole record. Following it makes the three parses uniform, covers both failing inputs and any third shape, and adds nothing new to the file — `suppress` is already imported.

**Key components.** Only the `request_id` assignment inside the `START RequestId:` branch changes: `request_id = None` up front, then the split inside `with suppress(...)`. No other control flow is touched, and a well-formed START line still yields the same id.

**Edge cases tested.** The new test is parametrized over both failing shapes and asserts the positive outcome, not just the absence of a raise: `success is True`, `request_id is None`, `start_time` still recorded, and the following stdout line still attached to the invocation's logs. `test_get_recent_invocations_success` continues to pin the well-formed path.

---

### DIFF
--- a/integrations/aws/lambda_client.py
+++ b/integrations/aws/lambda_client.py
@@ -204,9 +204,16 @@ def get_recent_invocations(
             if "START RequestId:" in message:
                 if current_invocation:
                     invocations.append(current_invocation)
-                request_id = (
-                    message.split("RequestId: ")[1].split()[0] if "RequestId:" in message else None
-                )
+                # The guard tested for "RequestId:" while the split used
+                # "RequestId: ", so it could not prevent the failure it looked
+                # like it was there for. filter_log_events returns the
+                # function's own stdout as well as AWS preamble, so this is
+                # parsing untrusted text; leave the field unset and keep the
+                # rest of the summary, as the Duration and Memory Used parses
+                # below already do.
+                request_id = None
+                with suppress(IndexError, ValueError):
+                    request_id = message.split("RequestId: ")[1].split()[0]
                 current_invocation = {
                     "request_id": request_id,
                     "start_time": timestamp,
--- a/tests/integrations/aws/test_lambda_client.py
+++ b/tests/integrations/aws/test_lambda_client.py
@@ -134,6 +134,40 @@ def test_get_recent_invocations_success(mock_logs_client) -> None:
     assert len(invocation["logs"]) == 4
 
 
+@pytest.mark.parametrize(
+    ("message", "label"),
+    [
+        ("START RequestId:req9 Version: $LATEST\n", "no space after the colon"),
+        ("START RequestId: \n", "label with no value"),
+    ],
+)
+def test_get_recent_invocations_tolerates_unparseable_request_id(
+    mock_logs_client, message, label
+) -> None:
+    """An unparseable START line leaves request_id unset, it does not abort the call.
+
+    filter_log_events returns everything in the log group, the function's own
+    stdout included, so the START parse reads untrusted text. The guard tested
+    for "RequestId:" while the split used "RequestId: ", so neither of these
+    two shapes was caught and the IndexError escaped get_recent_invocations.
+    """
+    mock_logs_client.filter_log_events.return_value = {
+        "events": [
+            {"timestamp": 1000, "message": message},
+            {"timestamp": 1100, "message": "hello world\n"},
+        ]
+    }
+
+    result = get_recent_invocations("test-func")
+
+    assert result["success"] is True, label
+    invocation = result["data"]["invocations"][0]
+    assert invocation["request_id"] is None
+    assert invocation["start_time"] == 1000
+    # the trailing stdout line still lands in this invocation
+    assert invocation["logs"] == [message, "hello world\n"]
+
+
 def test_get_recent_invocations_resource_not_found(mock_logs_client) -> None:
     error_response = {
         "Error": {"Code": "ResourceNotFoundException", "Message": "Log group not found"}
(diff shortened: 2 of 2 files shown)


dchaudhari7177:
Addressed the docstring rule in the latest commit.

AGENTS.md asks for concise, contract-focused docstrings, with bug history kept in commits or the PR description. The test docstring is now the single contract sentence, and the note about `filter_log_events` returning untrusted text moved to a comment on the fixture it actually explains — the reasoning about the mismatched guard lives in the commit message and the PR body instead.

`pytest tests/integrations/aws/test_lambda_client.py` — 15 passed; `ruff check` and `ruff format` clean at v0.15.12.

PR #6073, opened 2026-09-07
fix(shell): keep the prompt spinner alive for the demo's auto-submitted turn

YauhenBichel:
After picking a repository in the CI/CD analytics demo, the pinned layout
showed nothing from "Skill activated" until the analytics tool printed its
first line, often for a minute. An auto-submitted prompt is treated as a
`/goal` work turn, and the controller suspends the prompt bar until that
turn ends. In the pinned layout the spinner lives in that bar, so the
model's thinking phase had no visible progress at all.

- `TerminalSession.set_auto_prompt` queues text to be submitted as an
  ordinary typed turn. The prompt bar stays up, so the "Thinking…",
  "Executing…" and "Invoking tools…" phases keep animating, and the
  misleading "↗ /goal — work turn" label is no longer painted.
  `set_auto_command` is unchanged and still drives loop suggestions and
  interactive setup handoffs.
- `PromptBuilder.read_prompt_text` and the prompt refresh path honour the
  new flag; `Session.clear` and the harness terminal reset clear it.
- The demo picker uses `set_auto_prompt` for the analysis prompt and for
  a typed custom answer.

Verified with a scripted run of `/demo` in the pinned layout at Auto (Low):
between the repository pick and the tool's first output line the terminal
stream now carries 60 "Thinking…" frames plus the "Executing…" and
"Invoking tools…" phases, where before it carried none. Unit tests cover
the new session method and the picker's use of it.

### DIFF
--- a/core/agent_harness/session/terminal_access.py
+++ b/core/agent_harness/session/terminal_access.py
@@ -103,6 +103,8 @@ def clear_pending_autosubmit(session: Any) -> None:
         terminal.pending_prompt_default = None
     if hasattr(terminal, "pending_prompt_autosubmit"):
         terminal.pending_prompt_autosubmit = False
+    if hasattr(terminal, "pending_prompt_plain_turn"):
+        terminal.pending_prompt_plain_turn = False
 
 
 __all__ = [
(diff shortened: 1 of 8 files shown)


YauhenBichel:
@greptile review

PR #6075, opened 2026-09-07
fix(ci-analytics): read the whole window past GitHub's 1,000-run listing ceiling

YauhenBichel:
GitHub's workflow-run listing stops at 1,000 rows however far you page.
On an active repository the "last 30 days" report therefore covered only
the newest few days, with no coverage notice because the cap check was
set at 2,000. On Tracer-Cloud/opensre the demo showed 645 PR runs where
the window holds 5,572, and every KPI derived from it was an undercount.

- `collector.py`: each scope is fetched in time slices. A slice's first
  page carries `total_count`; a slice over the ceiling is split in half
  and refetched without paging the rest, slices run in parallel, and only
  a one-hour slice that still exceeds 1,000 runs is reported as a gap.
  Closed PRs are paged until their update time leaves the window instead
  of stopping at 500. Re-run history lookups check merged-PR re-runs
  first under a budget of 500 requests.
- `task_plan.py` / `rendering.py`: the post-turn plan breakdown is
  written as one buffered block with CRLF line endings through a new
  `print_repl_renderable`, so it stays left-aligned when the prompt has
  the terminal in raw mode.

### DIFF
--- a/integrations/github/tools/ci_analytics/collector.py
+++ b/integrations/github/tools/ci_analytics/collector.py
@@ -13,12 +13,17 @@ from integrations.github.client import GitHubApiError, GitHubRestClient
 from integrations.github.tools.ci_analytics.models import MergedPullRequest, WorkflowRun
 
 _PER_PAGE = 100
-_MAX_RUN_PAGES_PER_SCOPE = 20
-_MAX_PR_PAGES = 5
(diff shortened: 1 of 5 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6076, opened 2026-09-07
fix(ci-analytics): give re-run lookup priority only to PRs merged after the run

YauhenBichel:
Follow-up to #6075. Attempt-history lookups for passing re-runs are
checked in two phases so the shared budget is spent on merged-PR re-runs
first. The phase selector matched a re-run to a merged PR by head
repository and branch alone, without checking that the merge happened
after the run. On a reused branch name an unrelated re-run could take
priority budget away from a genuine merged-PR re-run, understating
CI-caused failures and blocked time under a tight budget.

- `metrics.py`: `on_critical_path` is now the public, documented rule: a
  PR number match when GitHub attached one, otherwise head repository and
  branch must match a PR merged later than the run was created.
- `collector.py`: the priority phase uses that same rule instead of its own
  branch-only lookup, so priority and the blocked-time metric can never
  disagree.
- Test: a re-run on a reused branch merged earlier in the window does not
  take the single available lookup; the genuine merged-PR re-run does.

### DIFF
--- a/integrations/github/tools/ci_analytics/collector.py
+++ b/integrations/github/tools/ci_analytics/collector.py
@@ -10,6 +10,7 @@ from typing import Any
 from urllib.parse import quote
 
 from integrations.github.client import GitHubApiError, GitHubRestClient
+from integrations.github.tools.ci_analytics.metrics import on_critical_path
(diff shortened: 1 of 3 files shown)


YauhenBichel:
@greptile review

PR #6080, opened 2026-09-07
fix(onboarding): render the demo's discovery steps in the transcript idioms

YauhenBichel:
The scan lead, chart, token hint, and "Analyzing…" line were printed as plain console output while everything the agent says goes through the reply gutter, so the demo read as a different program from the shell. The lead and the analysis line are now agent notes, the chart and the token hint hang in the same gutter, both pickers carry the Ask User header the agent's menus use, and the analytics tool indents its progress lines and painted report to the same body column.


YauhenBichel:
@greptile review

PR #6083, opened 2026-09-07
fix(scheduler): renew leases during long-running executions

### LINKED ISSUE #6082: [BUG][P1] Expired scheduler leases can duplicate external deliveries
## Problem

  Scheduled executions use a fixed 30-minute claim lease without heartbeats. A
  slow LLM call, network request, or delivery can outlive the lease.

  The recovery sweep then claims the same fire time and executes it again while the
  original worker may still be running. Completion fencing prevents stale database
  updates, but it cannot prevent duplicate Slack, Telegram, Discord, or other
  external messages.

  ## Affected code

  - `infrastructure/scheduling/scheduler/storage/run_store.py`
  - `infrastructure/scheduling/scheduler/executor.py`

  ## Expected behavior

  A running task must not be reclaimed while its worker is still alive.

  ## Suggested fix

  Add periodic lease heartbeats or extend the lease based on the task timeout.
  External delivery should also use an idempotency key based on task ID and fire
  time where supported.

  ## Acceptance criteria

  - Long-running tasks can renew their lease.
  - Recovery does not start a second execution while the original is healthy.
  - A worker that genuinely dies is still recoverable.
  - A regression test proves that a slow delivery produces only one external send.

Devesh36:
Fixes #6082

  Long-running scheduled tasks could outlive their fixed 30-
  minute claim lease. The recovery sweep could then reclaim
  the task while the original worker was still delivering,
  causing duplicate external messages.

  This change:

  - Adds owner-token-fenced claim renewal.
  - Runs a heartbeat during message building and delivery.
  - Preserves recovery for genuinely crashed workers.
  - Adds regression coverage for slow deliveries and stale
  claim owners.

  The scheduler now renews an active claim approximately every
  third of the lease duration. Renewal is guarded by the task
  ID, fire time, attempt number, owner token, and running
  status, so a reclaimed worker cannot extend a newer attempt.

  The heartbeat stops when execution completes, fails, is
  interrupted, or loses ownership. If the worker genuinely
  dies, the heartbeat stops with it and the existing recovery
  sweep can reclaim the expired task.

### DIFF
--- a/infrastructure/scheduling/scheduler/executor.py
+++ b/infrastructure/scheduling/scheduler/executor.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 import threading
 from collections.abc import Callable
+from datetime import UTC, datetime
 
 from infrastructure.scheduling.scheduler.delivery_bundle import resolve_delivery_adapter
 from infrastructure.scheduling.scheduler.delivery_plan import (
@@ -40,6 +41,7 @@ class _ClaimHeartbeat:
 
     def __init__(self, claim: ExecutionClaim) -> None:
         self._claim = claim
+        self._lease_expires_at = claim.lease_expires_at
         self._stop = threading.Event()
         self._lost = threading.Event()
         self._thread = threading.Thread(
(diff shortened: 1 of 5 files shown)


Devesh36:
@greptile review

Devesh36:
@greptile review

Devesh36:
@greptile review

PR #6084, opened 2026-09-07
feat(onboarding): CI reliability agent demo schedules a local loop and runs it once

YauhenBichel:
Onboarding: "Set up an agent that improves CI/CD reliability
over time". Picking it in `/demo` scans the machine, asks which repository
to watch and when to run (weekdays 08:00 local by default), creates a
local prompt loop that re-runs the CI/CD reliability analytics for the
last 7 days, runs the first pass, shows the report as it will land in
`/loops messages`, and offers the Slack demo next. Every step before the
loop is deterministic; no model turn is needed to configure it.

Delivery is pinned to the shell inbox, so the loop cannot post to Slack or
Telegram even when those are configured.

## Changes

- `loop.py`: loop prompt, creation with reuse of an existing loop for the
  same repository, local timezone, schedule card, report completeness check
- `schedule_ci_reliability_loop` action tool; the analytics skill's step 4
  calls it for "Set up an agent…" instead of the CI-fix onboarding
- `run_loop_now` in the shell runtime, shared with `/loops run`
- `analyze_github_ci_reliability` paints only on a real terminal; a
  headless run gets the full markdown back (loop messages were one line)
- Docs section "The CI reliability agent"

### DIFF
--- a/core/agent_harness/prompts/skills/cicd_analytics_demo/SKILL.md
+++ b/core/agent_harness/prompts/skills/cicd_analytics_demo/SKILL.md
@@ -25,6 +25,7 @@ WHEN TO USE:
 USE THESE TOOLS:
 - `scan_local_git_workspace`
 - `analyze_github_ci_reliability`
+- `schedule_ci_reliability_loop`
 - `ask_user_choice`
 
 DO NOT USE THIS SKILL FOR:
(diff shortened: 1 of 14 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6088, opened 2026-09-07
feat(scheduler): deterministic CI reliability loop, background service, raw snapshots

YauhenBichel:
Hardens the CI reliability agent demo after comparing it with what Droid
and Claude Code produce for the same request: both generate a launchd
service and a deterministic script. This PR closes those two gaps and
keeps opensre's advantages (60-second setup, baseline-adjusted KPI,
shell-only delivery).

- Deterministic tick: a manual loop may name a report builder; the runner
  uses it instead of a model turn. The CI reliability loop is built this
  way, so a delivery can no longer be a model refusal.
- Raw data: each delivery names a JSON snapshot of its figures under
  `~/.opensre/ci_reliability_reports/`.
- Background service: `/loops service install` registers a per-user
  launchd or systemd unit running `opensre cron start --service`; the demo
  offers it as the first next step when none is installed.
- The demo option now reads "Set up an agent that reports CI/CD
  reliability every weekday".

### DIFF
--- a/core/agent_harness/prompts/skills/cicd_analytics_demo/SKILL.md
+++ b/core/agent_harness/prompts/skills/cicd_analytics_demo/SKILL.md
@@ -81,14 +81,15 @@ headers [3/4] and [4/4] only.
 4) Offer what to do next.
    Call `ask_user_choice` with title `What would you like to do next?` and
    these exact options:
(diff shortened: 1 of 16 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6089, opened 2026-09-07
refactor(scheduler): centralize lease renewal and ownership fencing

### LINKED ISSUE #6087: refactor(scheduler): centralize lease renewal and ownership fencing
## Problem

Scheduler executions use durable SQLite leases so expired work can be recovered, but lease liveness is currently coupled directly to individual executions. The fix proposed in #6083 adds one heartbeat thread per execution and leaves renewal-error deadlines and recovery coordination split across `executor.py`, `runner.py`, and `run_store.py`.

That fixes the common slow-execution case, but it makes ownership harder to reason about and scales renewal work with concurrent executions.

## Proposed design

Introduce a process-level scheduler claim lease renewer:

- Maintain a registry of claims that are actively executing in this process.
- Use one dedicated renewal loop per process, independent of APScheduler's executor pool.
- Renew active claims in a single batched SQLite transaction.
- Track the last confirmed lease deadline for each claim.
- Treat a fenced renewal result as immediate ownership loss.
- Retry transient storage errors only while the last confirmed lease remains valid.
- Expose a claim-scoped ownership guard that the executor checks before every external delivery.
- Keep expired-claim recovery durable; crashed processes lose their in-memory registry and their claims become reclaimable after the last confirmed lease expires.

This issue is a focused replacement for the per-execution heartbeat design in #6083. Separating message generation from delivery through a durable outbox remains a follow-up architectural improvement.

## Correctness boundary

Fencing and lease ownership can prevent a stale attempt from *starting* another delivery. They cannot guarantee exactly-once delivery when a provider accepts a message and the process crashes before recording success. Provider-supported deterministic idempotency keys or a durable delivery outbox are required for that boundary.

## Acceptance criteria

- [ ] Concurrent executions share one renewal loop rather than starting one thread each.
- [ ] Active claims are renewed in a batch before their leases expire.
- [ ] A definitive fenced renewal marks only that claim as lost.
- [ ] A transient renewal error is retried while the last confirmed lease is valid.
- [ ] Persistent renewal errors mark the claim lost at its confirmed deadline.
- [ ] Lost claims cannot begin message delivery or later fan-out targets.
- [ ] A genuinely expired claim remains recoverable through the normal fenced executor.
- [ ] Registration, completion, failure, and shutdown do not leak active claims or renewal threads.
- [ ] Tests cover slow generation/delivery, concurrent claims, stale owners, transient and persistent storage failures, recovery, and delivery fencing.

## Related

- #6082
- #6083

muddlebee:
Fixes #6087

Related to #6082. This is the process-level batched lease-renewal alternative to #6083.

- add one process-level `ClaimLeaseRenewer` that registers every locally active claim and renews due claims in a shared batch
- track each claim's last storage-confirmed lease deadline using a monotonic clock
- sample monotonic time before UTC so a pause between clock readings cannot extend ownership
- retry transient renewal errors only inside that confirmed safety window and fail closed at expiry
- fence renewal and completion by task, fire time, attempt, owner token, running status, and an unexpired lease
- check ownership before delivery, before every fan-out retry, and before persisting delivery results
- stop the renewal loop automatically after the final active claim exits

This keeps the existing durable recovery sweep for genuinely abandoned work while removing one heartbeat thread per execution.

Local scheduler validation:

```text
$ uv run python -m pytest tests/scheduler -n 4 -q
270 passed

$ make test-scope
89 passed

$ uv run python -m pytest tests/quality/test_no_umbrella_names.py -q
38 passed

$ make lint
All checks passed!

$ make format-check
3115 files already formatted

$ make typecheck
Success: no issues found in 1804 source files
```

The regression suite covers concurrent claims sharing a renewal batch, slow execution surviving its original lease, fenced stale owners, transient and persistent renewal failures, recovery after expiry, fan-out retry fencing, renewal-loop cleanup, and a deterministic clock-sampling pause. The new clock-pause regression fails on the previous head and passes with the conservative deadline conversion.

---

The existing lease remains the durable crash-recovery authority. A local renewer owns only process liveness: executions register their claims, one daemon renewal loop batches active claims, and each registration receives a lightweight ownership guard.

A successful renewal moves the local monotonic deadline to the new storage-confirmed expiry. A definitive fenced result invalidates that claim immediately. Storage exceptions schedule another attempt but never extend the local deadline, so persistent failures eventually stop new external sends.

Alternatives considered:

- per-execution heartbeat threads: simpler locally, but thread and renewal work scale with concurrent executions
- renewal from the recovery sweep: cannot distinguish a slow live owner from a crashed owner
- a fixed longer lease: delays recovery and still fails for sufficiently slow executions
- a durable delivery outbox: valuable follow-up for provider-side send/ack ambiguity, but intentionally outside this ownership-focused change

Key components:

- `ClaimLeaseRenewer`: shared claim registry, renewal scheduling, batching, and lifecycle
- `ClaimOwnership`: claim-scoped validity check used by execution and delivery
- `renew_claims`: one immediate SQLite transaction containing fenced renewals
- `complete_run`: now also rejects completion after lease expiry

---

---

### DIFF
--- a/infrastructure/scheduling/scheduler/claim_lease.py
+++ b/infrastructure/scheduling/scheduler/claim_lease.py
@@ -47,7 +47,7 @@ class ClaimOwnership:
             return not self._state.lost and self._monotonic() < self._state.confirmed_until
 
 
-class ClaimLeaseManager:
+class ClaimLeaseRenewer:
     """Renew every active process claim from one shared background loop."""
 
     def __init__(
@@ -189,7 +189,7 @@ class ClaimLeaseManager:
                 state.renew_at = min(state.confirmed_until, retry_at)
 
 
-default_claim_lease_manager = ClaimLeaseManager()
+default_claim_lease_renewer = ClaimLeaseRenewer()
 
 
-__all__ = ["ClaimLeaseManager", "ClaimOwnership", "default_claim_lease_manager"]
+__all__ = ["ClaimLeaseRenewer", "ClaimOwnership", "default_claim_lease_renewer"]
--- a/infrastructure/scheduling/scheduler/claim_lease.py
+++ b/infrastructure/scheduling/scheduler/claim_lease.py
@@ -1,0 +1,195 @@
+"""Process-level lease renewal for active scheduler executions."""
+
+from __future__ import annotations
+
+import logging
+import threading
+import time
+from collections.abc import Callable, Collection, Iterator, Mapping
+from contextlib import contextmanager
+from dataclasses import dataclass
+from datetime import UTC, datetime
+
+from infrastructure.scheduling.scheduler.storage import (
+    ExecutionClaim,
+    claim_renewal_interval_seconds,
+    renew_claims,
+)
+
+logger = logging.getLogger(__name__)
+
+RenewClaims = Callable[[Collection[ExecutionClaim]], Mapping[ExecutionClaim, datetime]]
+
+
+@dataclass(slots=True)
+class _LeaseState:
+    confirmed_until: float
+    renew_at: float
+    lost: bool = False
+
+
+class ClaimOwnership:
+    """Expose whether one execution still owns its confirmed lease."""
+
+    def __init__(
+        self,
+        state: _LeaseState,
+        condition: threading.Condition,
+        monotonic: Callable[[], float],
+    ) -> None:
+        self._state = state
+        self._condition = condition
+        self._monotonic = monotonic
+
+    def valid(self) -> bool:
+        """Return whether storage still confirms an unexpired fenced claim."""
+        with self._condition:
+            return not self._state.lost and self._monotonic() < self._state.confirmed_until
+
+
+class ClaimLeaseManager:
+    """Renew every active process claim from one shared background loop."""
+
+    def __init__(
+        self,
+        *,
+        renew: RenewClaims = renew_claims,
+        renewal_interval_seconds: float | None = None,
+        monotonic: Callable[[], float] = time.monotonic,
+        utc_now: Callable[[], datetime] | None = None,
+    ) -> None:
+        self._renew = renew
+        self._renewal_interval_seconds = (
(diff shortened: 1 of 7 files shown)


muddlebee:
@greptile review

infrastructure/scheduling/scheduler/claim_lease.py
```diff
@@ -0,0 +1,195 @@
...
+            yield ownership
+        finally:
+            with self._condition:
+                state.lost = True
+                self._active.pop(claim, None)
+                self._condition.notify_all()
+
+    def _deadline_for(self, lease_expires_at: datetime) -> float:
+        remaining = max(0.0, (lease_expires_at - self._utc_now()).total_seconds())
+        return self._monotonic() + remaining
```
muddlebee: [P2] Sample monotonic time before UTC when converting lease expiry. A pause between the current UTC read and monotonic read is added to the local safety window. A deterministic reproduction with a lease expiring at t=10 and a pause advancing time to t=20 computes a deadline of t=30 and reports valid ownership after durable expiry. That permits a stale external send even though completion is fenced. Capture monotonic first so sampling delays only shorten the window, and add a deterministic regression test.

muddlebee (commented):
Reviewed head 1dac2cf. The shared renewal design remains preferable to #6083 for separation of responsibilities, batching, and direct expiry checks. One reproducible deadline-conversion issue needs correction; scheduler validation at this head passed all 269 tests.

infrastructure/scheduling/scheduler/claim_lease.py
muddlebee: Fixed in ff0072812: capture monotonic time before UTC so clock-sampling pauses cannot inflate the confirmed deadline. The deterministic ownership regression failed on the previous code and passes with the fix. Validation: lint, format-check, typecheck, and all 270 scheduler tests passed.

muddlebee:
@greptile review

PR #6024, opened 2026-09-05
feat(scheduler): bound scheduled run concurrency

cerencamkiran:
Fixes #5963

#### Describe the changes you have made in this PR -

- Merge current `main`, retaining the exact fire-time callback fixes from #6093 and the shared lease renewer / ownership fencing from #6089.
- Configure the existing scheduler executor with a default of two workers, overridable through `OPENSRE_SCHEDULER_MAX_CONCURRENT_RUNS`; reject invalid limits by falling back to the default.
- Persist admitted ticks before worker submission and remove the 60-second misfire deadline for queued work.
- Use one atomic `try_claim` path for new, pending, and expired attempts. Keep pending rows as durable, unowned work; the existing recovery sweep resumes them after restart through the same bounded worker pool.
- Prevent concurrent live claims for different ticks of the same task, including recovery and hosts sharing the run database. Filter busy and ineligible tasks before the recovery scan limit.
- Preserve scoped delivery and the process-wide agent turn gate. Remove the PR's duplicate per-run heartbeat and separate `queued_claim` / `try_start_run` path.
- Document overflow, same-task tick skipping, restart behavior, and the shared capacity limit.

The independent runbook skill metadata fix landed separately in #6134. This branch includes it through the main merge, and the metadata file is not part of this PR diff.

### Demo/Screenshot for feature changes and bug fixes -

Real APScheduler callbacks and SQLite history are exercised with barriers at limits 1 and 2: available workers enter together, overflow remains `pending`, and admitted five-minute-old ticks finish after workers are released. A fresh scheduler recovers a persisted pending tick with its original fire time. Competing claimants cannot execute different ticks of one task together.

```text
make lint                                      passed
make format-check                              passed
make typecheck                                 passed (1833 source files)
uv run python -m pytest tests/scheduler \
  gateway/tests/runtime/test_concurrency_gate.py \
  gateway/tests/runtime/test_scheduler_runners.py \
  tests/infrastructure/test_turn_capacity.py \
  tests/core/agent_harness/prompts/test_skill_metadata.py \
  tests/core/agent_harness/prompts/skills/test_runbook_investigation.py
                                               317 passed
```

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [x] I have reviewed every single line of the AI-generated code
[...]

PR #6093, opened 2026-09-08
fix(scheduler): pass fire times directly to scheduled callbacks

### LINKED ISSUE #6092: [BUG][P1] Scheduler callback can run before fire-time metadata is recorded
## Problem

  The scheduler stores fire times from `EVENT_JOB_SUBMITTED` in a shared map.
  However, APScheduler submits the worker callback before dispatching the
  submission event.

  The callback can therefore run before its fire time exists, fall back to the
  current wall-clock time, and leave stale metadata that may be consumed by the
  next tick. This can corrupt the task-run deduplication key.

  ## Affected code

  - `_on_job_submitted()`
  - `_scheduled_job()`

  ## Expected behavior

  Every callback must receive the exact `scheduled_run_time` that APScheduler
  assigned to that execution.

  ## Suggested fix

  Pass the scheduled fire time directly through the job callback or use a wrapper
  that captures it before submission. Do not use a shared listener map as the
  source of truth.

  ## Acceptance criteria

  - Callback execution does not depend on listener timing.
  - Misfires and coalesced runs retain the correct scheduled time.
  - A concurrent regression test proves that no callback falls back to wall-clock
    time.

Devesh36:
Fixes #6092

  ## Summary

  - Remove the `EVENT_JOB_SUBMITTED` fire-time mailbox.
  - Pass each APScheduler `scheduled_run_time` directly through the scheduler executor.
  - Eliminate the wall-clock fallback in scheduled callbacks.
  - Preserve the correct fire time for misfired and coalesced runs.
  - Add deterministic regression coverage for worker-before-listener execution.

  This complements the concurrency work in #6024. Its bounded executor can use
  `ScheduledThreadPoolExecutor(max_workers=limit)`.

  ## Testing

  - `uv run pytest tests/scheduler/test_runner.py tests/scheduler/test_executor.py tests/scheduler/test_apscheduler_executor.py -q`
    - 58 passed
  - `make lint`
  - `make format-check`
  - `make typecheck`

### DIFF
--- a/infrastructure/scheduling/scheduler/apscheduler_executor.py
+++ b/infrastructure/scheduling/scheduler/apscheduler_executor.py
@@ -1,0 +1,88 @@
+"""APScheduler executor that passes each scheduled fire time to the job."""
+
+from __future__ import annotations
+
+from concurrent.futures import Future
+from dataclasses import dataclass
+from datetime import datetime
+from typing import Any, cast
+
+from apscheduler.executors.base import run_job
+from apscheduler.executors.pool import ThreadPoolExecutor
+
+
+@dataclass(frozen=True)
+class _ScheduledJobInvocation:
+    """Expose one APScheduler run time through the job callable."""
+
+    job: Any
+    scheduled_run_time: datetime
+
+    @property
+    def id(self) -> str:
+        return cast(str, self.job.id)
+
+    @property
+    def func(self) -> Any:
+        def invoke(*args: Any, **kwargs: Any) -> Any:
+            return self.job.func(
+                *args,
+                scheduled_run_time=self.scheduled_run_time,
+                **kwargs,
+            )
+
+        return invoke
+
+    @property
+    def args(self) -> tuple[Any, ...]:
+        return cast(tuple[Any, ...], self.job.args)
+
+    @property
+    def kwargs(self) -> dict[str, Any]:
+        return cast(dict[str, Any], self.job.kwargs)
+
+    @property
+    def misfire_grace_time(self) -> int | None:
+        return cast(int | None, self.job.misfire_grace_time)
+
+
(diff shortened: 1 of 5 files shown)


Devesh36:
@greptile review

muddlebee:
**[P1] Required timestamp argument prevents scheduled-job registration**

The timestamp-race fix is justified, but the current head (`dffb006`) introduces a startup blocker: `_scheduled_job()` now requires the keyword-only `scheduled_run_time`, while `_register_jobs()` calls `scheduler.add_job()` without supplying it. APScheduler validates the callback arguments during registration, before `ScheduledThreadPoolExecutor` has an opportunity to inject the timestamp.

I reproduced this using the PR's `_register_jobs()` with a real APScheduler 3.11.3 `BackgroundScheduler` and one enabled task (task-store writes and agent execution isolated):

```text
ValueError: The following keyword-only arguments have not been supplied in kwargs: scheduled_run_time
```

Both blocking and background scheduler entrypoints use this registration path, so neither can register an enabled task. The current fake-scheduler tests bypass APScheduler's argument validation, which explains why CI can remain green.

Please make the callback signature registration-compatible while still explicitly rejecting a missing timestamp at execution time—do not restore the wall-clock fallback. Add a regression test that registers and executes a task through a real scheduler and verifies that the callback receives the intended fire time.

This is separate from the worker-pool/backpressure issue in #6094; that redesign need not be included here.

Devesh36:
@greptile review

Devesh36:
@muddlebee Fixed in commit ee4416b2c. `_scheduled_job` now accepts an omitted timestamp for APScheduler registration but raises immediately if execution occurs without one; it never falls back to wall-clock time. Added a real `BackgroundScheduler` regression test covering registration and execution with the intended fire time.

PR #6098, opened 2026-09-08
fix(onboarding): deterministic CI/CD loop suggestion, bottom-up blocked time, single answer line

YauhenBichel:
Addresses the review of the onboarding demos: the demo ran `opensre
health`, seemed to ask the same question twice, and the blocked-time KPI
was a single total rather than a bottom-up calculation.

- **No more `opensre health`.** The first-launch loop picker's CI/CD row
  auto-submitted a prompt into `github-ci-health`, a skill that only works
  under the scheduled runner; with nothing legitimate to call, the model ran
  `opensre health` and printed 61 missing integrations on a fresh machine.
  The row now runs the deterministic CI reliability agent demo (scan,
  repository, schedule, first pass) with no model turn. The `cli_exec` tool
  no longer advertises `health` as its preferred payload, and the skill
  states how to find the repository interactively.
- **Bottom-up blocked time.** The report shows "How it adds up, worst
  first": the five most blocked merged PRs with when each should have been
  green, when it actually was, and how long it waited, then the remainder
  and the sum with the typical wait. `PullRequestDelay` carries the expected
  and actual green times.
- **One answer line.** After a menu the shell printed the question again
  under an "Ask User" header, which read as asking twice. It is now a single
  dim question with the answer after it.
- The analytics skill asks each question once, continues with the answer,
  and never asks what a request means.

### DIFF
--- a/core/agent_harness/prompts/skills/cicd_analytics_demo/SKILL.md
+++ b/core/agent_harness/prompts/skills/cicd_analytics_demo/SKILL.md
@@ -50,6 +50,10 @@ HARD RULES:
   not fall back to a different data source.
 - Decision points use `ask_user_choice` with the exact option texts below.
   End the turn after calling it; the answer arrives as the next user message.
+- Ask each question once. When the answer arrives, continue with the next
+  step immediately: do not reload this skill, do not restate the options, and
+  never ask what the answer or the request "means". A repository name in the
+  request or in the answer is the repository; go straight to step 3.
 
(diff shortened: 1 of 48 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6113, opened 2026-09-08
fix(shell): make session goals legible: credit plan steps, pause on stall with options

YauhenBichel:
A typed request with two parts ("report failing checks, then set up a
recurring check") runs as a session goal. In practice the goal's checklist
never got a tick while the per-turn plan showed every step done, so the
loop repeated the same work for six turns, reprinted the full goal block
each time, and ended "budget exhausted". Along the way a slash command
typed into an open menu was taken as the answer.

- **Checklist credit from the plan.** Completed `update_plan` steps mark
  matching checklist items done (normalized text match), so a goal whose
  steps the plan completed closes instead of repeating them.
- **Stall pause with options.** Two turns without a tick pause the goal and
  open a menu: keep going one more turn (`/goal resume`), stop here
  (`/goal clear`), or type guidance. Menu options can now carry a slash
  command the shell runs directly (`PendingUserChoice.commands`).
- **One-line repaint.** An unchanged active goal prints one status line;
  the full block returns when a tick, a pause, or the end changes it.
- **Menus and commands.** A slash command typed into the custom row runs as
  a command rather than being handed to the model as the chosen option.

### DIFF
--- a/core/agent_harness/session/pending_choice.py
+++ b/core/agent_harness/session/pending_choice.py
@@ -14,7 +14,7 @@ and no "Reply with 1, 2, or 3" free-text parsing.
 from __future__ import annotations
 
 import re
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 _ANSWER_HEADER = re.compile(r"^(\d+)\.\s+(.+)$")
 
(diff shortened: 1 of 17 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6114, opened 2026-09-08
feat(goal): judge completion with a cheap model, tick the checklist by tool, stop on stalls

YauhenBichel:
Session goals (`/goal`) are now judged by a separate cheap model instead of a
rule table, and progress is tracked through a tool instead of prose tags.

- A classification-tier judge reads the condition, checklist and reply after
  every turn and returns met / not yet / impossible with a reason. "Met" still
  needs a successful tool call this turn or stored findings. The reason shows on
  the status line and steers the next turn. New `impossible` terminal state.
- Checklist items are ticked with `session_goal_complete`; a validator on the
  same model can refuse a tick and says why. A met verdict ticks every item.
  Numbered steps in the condition become the checklist; a plain condition gets
  none.
- Two turns with no tick and no successful tool call pause the goal and offer
  the options menu.
- One status line per turn with elapsed time and tokens; the condition prints
  once per goal; verdict reasons appear in the transcript.
- `github_cli` allows `gh run list|view` and `gh workflow list|view`; only
  `run rerun|cancel|delete|watch|download` and `workflow run|enable|disable`
  stay denied. `list_github_actions_workflow_runs` accepts `head_sha`.
- `/goal set` on Slack and Telegram starts the first turn in the same loop.
- The judge client is injected through `HeadlessAgent`; the retired
  `session_goal:done=` grammar and `confirm.py` are removed.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6116, opened 2026-09-08
feat(shell): offer a way out of the credit wall: top-up page or provider switch, as a menu

YauhenBichel:
When the hosted LLM route reports exhausted credits, the shell now gives the
user something to press instead of a URL to copy:

- `opensre account usage` (new): opens the usage and top-up page in the
  browser and prints the URL. `--no-browser` prints only; `--dev` targets the
  local webapp. Exposed in the shell as `/account usage`.
- After the credit-exhausted turn error, the shell opens its arrow-key menu:
  "Open the usage and top-up page in the browser" runs `/account usage`,
  "Switch to another LLM provider" runs `/model`, Esc leaves it. Both options
  map to slash commands, so choosing one never fires a model call that would
  hit the same wall.
- The URL on the error line is underlined, so it reads as a link before hover.
- Chat transports (Slack, Telegram) keep the plain-text message.

## Why

A terminal hyperlink still needs Cmd+click, and Apple Terminal ignores
hyperlinks entirely, so the link alone left users stuck. Claude Code answers a
spend or usage wall with an inline selectable prompt; this brings the shell to
the same place using the existing choice menu.

image


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6011, opened 2026-09-04
Fix: ollama model detection and validation

SaptarshiSarkar12:
Fixes #5768

#### Describe the changes you have made in this PR -

- Discover model choices from the configured Ollama host via `GET /api/tags` instead of showing the static provider catalog.
- Reject unavailable Ollama models before changing `.env` or resetting runtime clients.
- Keep model switching for hosted, CLI, and ambient-auth providers unchanged; switching does not perform an implicit paid completion probe.
- Reuse the Ollama discovery and tag-matching helpers in onboarding validation and the local-model lifecycle.
- Preserve Rich escaping for model identifiers shown by switching commands.

### Demo/Screenshot for feature changes and bug fixes - 

```text
$ uv run python -m pytest tests/cli/ tests/interactive_shell/ \
    tests/shared/llm_setup/test_ollama.py \
    tests/shared/llm_setup/test_validation.py -q
2259 passed, 5 skipped
```

---

## Code Understanding and AI Usage

**Did you use AI assistance (ChatGPT, Claude, Copilot, etc.) to write any part of this code?**
- [ ] No, I wrote all the code myself
- [x] Yes, I used AI assistance (continue below)

**If you used AI assistance:**
- [x] I have reviewed every single line of the AI-generated code
- [x] I can explain the purpose and logic of each function/component I added
- [x] I have tested edge cases and understand how the code handles them
- [x] I have modified the AI output to follow this project's coding standards and conventions

**Explain your implementation approach:**

The bug is specific to Ollama's locally available model inventory, so the final implementation keeps provider-specific behavior in Ollama-owned modules. A shared helper performs one bounded `/api/tags` request, validates the response shape, normalizes tags, and returns a stable model tuple. The interactive picker consumes that tuple, while direct switching performs the same availability check immediately before persistence. Hosted-provider credential verification remains an explicit authentication concern rather than an implicit side effect of model switching.

---

## Checklist before requesting a review
- [x] I have added proper PR title and linked to the issue
- [x] I have performed a self-review of my code
- [x] **I can explain the purpose of every function, class, and logic block I added**
- [x] I understand why my changes work and have tested them thoroughly
- [x] I have considered potential edge cases and understand how the code handles them
- [x] If it is a core feature, I have added thorough tests
- [x] My code follows the project's style guidelines and conventions

---

Note: Please check **Allow edits from maintainers** if you would like us to assist in the PR.

PR #6117, opened 2026-09-08
fix(shell): open the shell on a configured provider without an account

### LINKED ISSUE #6112: [BUG] No way to use a local Ollama model in the shell: sign-in is required, and the hosted lock tells you to sign out
## Summary

There is no supported way to run a local Ollama model in the interactive shell. The shell requires an OpenSRE account, a signed-in shell is locked to the hosted OpenAI route, and the lock message tells you to sign out first. Signing out means the shell stops opening at all, so the instruction is a dead end.

Reported on Discord: "How can we configure local ollama models now? I checked Preferences section for models but there was no option to use local models."

Expected: someone with Ollama running can select a local model and use it in the shell.

Actual:

- `opensre` with no arguments hits the sign-in gate. The only two choices are "Sign in or create account" and "Exit and stay signed out" (`surfaces/interactive_shell/ui/sign_in.py:72-99`, wired at `surfaces/interactive_shell/main.py:178`).
- Once signed in, `/model` refuses every change: "LLM settings are managed by your OpenSRE account ... Run `opensre account logout` before configuring a different provider or model" (`surfaces/interactive_shell/command_registry/model/switching.py:16-31`).
- `opensre account logout` clears the account, so the next `opensre` lands back on the gate and the shell never opens.

Our own docs already describe the loop, at `docs/interactive-shell-commands.mdx:565`: "local provider/model changes are locked until sign-out; signing out closes the shell."

The only thing that skips the gate today is `OPENSRE_IS_TEST=1`, `PYTEST_CURRENT_TEST`, or a CI environment (`infrastructure/analytics/source.py:11`). That is not something we can tell a user to set.

## Steps to reproduce

1. `opensre onboard local_llm` and finish the Ollama setup. It succeeds.
2. Run `opensre`. Sign in when the gate asks.
3. Run `/model set ollama llama3.2`. It is refused, and you are told to log out.
4. Run `opensre account logout`.
5. Run `opensre` again. You are back at the sign-in gate with no way into the shell.

Yes.

## Scope

Only the interactive shell is affected. `opensre ask`, the gateway, and cron still fall back to the configured `LLM_PROVIDER` when no account route exists (`config/account.py:197`), so Ollama works there.

The web Preferences page showing only hosted models is expected, since the account route is a hosted OpenAI proxy. It is where people look first, though, so whatever we do here needs a line in the docs too.

## Additional context

The mandatory gate is deliberate (#5920), and this is not a request to remove it. What is missing is a supported path for people running their own model. Two options that keep the gate intact:

1. Let the gate pass when a validated local or self-configured provider is already set, so `opensre onboard local_llm` followed by `opensre` works.
2. Add an explicit "use my own model" choice to the sign-in screen, shown only when a provider is configured.

Happy to implement whichever direction the team prefers.

## Environment

macOS 15, `main` at 751e7da2e.

Davidson3556:
Fixes #6112

Running a local model was unreachable from the shell. The shell requires an account, a signed-in shell is pinned to the hosted OpenAI route, and the `/model` lock tells you to sign out. Signing out then closed the shell, so the instruction had nowhere to land.

Two changes close the loop, both conditioned on the user having already configured their own provider:

- The startup gate offers a third choice, "Continue with my own model", which enters the shell signed out.
- `/account logout` keeps the shell open instead of closing it.

`has_user_configured_llm_provider()` backs both. It requires `LLM_PROVIDER` to be set explicitly plus resolvable credentials. The built-in default provider does not count, so a fresh install has no provider, still sees the same two-way sign-in choice, and cannot slip past the gate only to fail on its first turn.

The `/model` lock message now names `opensre onboard local_llm` when there is no provider to fall back to, and the docs that said signing out closes the shell are updated.

**Worth a maintainer's opinion before merge:** anyone who already has an API key in their environment now has a way past the sign-in screen. The gate sits directly on hosted credits, so that tradeoff is a product call rather than mine. Happy to narrow the condition to local providers only, or to drop the startup half and keep just the logout fix, if you prefer.

Real binary driven under a pty with an isolated `OPENSRE_HOME`.

No provider configured, unchanged from today:

```text
Sign in or create an OpenSRE account to use the interactive shell.
  ❯ Sign in or create account
    Exit and stay signed out
  ↑↓ Navigate • Enter Select • Esc cancel
```

With `LLM_PROVIDER=ollama`:

```text
Sign in or create an OpenSRE account to use the interactive shell.
  ❯ Sign in or create account
    Continue with my own model
    Exit and stay signed out
  ↑↓ Navigate • Enter Select • Esc cancel
```

Checks:

```text
make lint          All checks passed!
make format-check  3119 files already formatted
make typecheck     Success: no issues found in 1807 source files
uv run pytest tests/interactive_shell tests/config tests/shared tests/cli/wizard tests/quality -q
                   2334 passed, 1 skipped
```

---

The bug is a collision between two workstreams rather than one broken function: the mandatory account gate (#5920) and the hosted-model lock added alongside credit exhaustion (#5958) each make sense alone, and together they leave no route to a local model. Saptarshi's Ollama discovery fix (#6011) merged the day before he reported this on Discord, so the picker works; you just cannot reach it.

I considered letting the gate pass silently whenever a provider is configured. I rejected that because it removes the sign-in screen for a whole class of users without them choosing it, and the forced screen is deliberate. Making it an explicit third row keeps the screen, keeps the exposure, and only opens the door for someone who already did the setup work.

The predicate deliberately reads `LLM_PROVIDER` rather than `get_configured_llm_provider()`, which falls back to a default provider when unset. Using the resolved value would show the extra row on a machine with nothing configured, which trades one dead end for a worse one.

I extended the check to `/account logout` after tracing what a signed-out shell touches. Without it the loop is still closed from inside the shell: `/model` sends you to log out, and logging out ejects you.

Tests cover the two failure modes that matter: the row appears and enters the shell when a provider exists, and it is absent when one does not. `has_user_configured_llm_provider` has one test pinning that the default provider does not count.

---

### DIFF
--- a/config/account.py
+++ b/config/account.py
@@ -200,6 +200,11 @@ def ignore_account_llm_route() -> None:
     os.environ[OPENSRE_IGNORE_ACCOUNT_ROUTE_ENV] = "1"
 
 
+def restore_account_llm_route() -> None:
+    """Re-enable the hosted route after a login replaces the rejected session."""
+    os.environ.pop(OPENSRE_IGNORE_ACCOUNT_ROUTE_ENV, None)
+
+
 def account_llm_route() -> AccountLLMRoute | None:
     """Return the hosted OpenAI route only when account metadata and token exist.
 
@@ -224,6 +229,7 @@ __all__ = [
     "AccountLLMRoute",
     "account_llm_route",
     "ignore_account_llm_route",
+    "restore_account_llm_route",
     "account_metadata_path",
     "delete_account_record",
     "delete_account_token",
--- a/config/account.py
+++ b/config/account.py
@@ -20,6 +20,7 @@ from config.constants.account import (
     OPENSRE_ACCOUNT_TOKEN_ENV,
     OPENSRE_APP_URL_DEFAULT,
     OPENSRE_APP_URL_ENV,
+    OPENSRE_IGNORE_ACCOUNT_ROUTE_ENV,
 )
 from config.constants.paths import host_home
 from config.secrets.store import (
@@ -194,8 +195,21 @@ def delete_account_token() -> None:
     delete_secret(OPENSRE_ACCOUNT_TOKEN_ENV)
 
 
+def ignore_account_llm_route() -> None:
+    """Drop the hosted route for this process; the user chose their own model."""
+    os.environ[OPENSRE_IGNORE_ACCOUNT_ROUTE_ENV] = "1"
+
+
 def account_llm_route() -> AccountLLMRoute | None:
-    """Return the hosted OpenAI route only when account metadata and token exist."""
+    """Return the hosted OpenAI route only when account metadata and token exist.
+
+    A session the webapp rejected or could not validate still leaves its record
+    and token on disk, so the route outlives the sign-in it came from. Callers
+    that let the user pick a local provider instead first call
+    :func:`ignore_account_llm_route`.
+    """
+    if os.getenv(OPENSRE_IGNORE_ACCOUNT_ROUTE_ENV, "").strip():
+        return None
     record = load_account_record()
     if record is None or record.llm_provider != "openai" or not resolve_account_token():
         return None
@@ -209,6 +223,7 @@ __all__ = [
     "AccountRecord",
     "AccountLLMRoute",
     "account_llm_route",
+    "ignore_account_llm_route",
     "account_metadata_path",
     "delete_account_record",
     "delete_account_token",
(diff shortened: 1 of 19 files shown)


Davidson3556:
@greptile review

Davidson3556:
@greptile review

Davidson3556:
@greptile review

muddlebee:
@Davidson3556 thank you but closing for now, till we have product direction clarity. rest I have added comments in the discord channel.

PR #6118, opened 2026-09-08
feat(goal): count tokens for real and pause the goal when a turn fails

YauhenBichel:
## What

- Token counts. `AgentLLMResponse` gains `input_tokens` / `output_tokens`,   filled by the Anthropic, Bedrock, OpenAI Responses, chat-completions and  OpenAI-compatible paths. The tool loop sums them into `AgentRunResult`, the
  action driver copies them onto `ToolCallingTurnResult`, and the orchestrator  records them on `session.tokens`. `/goal` status lines and `/cost` now show  real spend; a provider that reports no usage adds nothing rather than an  estimate.
- Failed turns pause the goal. When a goal turn raises, the loop paints the  goal paused with "the last turn failed; fix the cause, then /goal resume"  and re-raises, so the host still prints the turn error and the credit-wall  menu still opens. Docs updated.

## Why

Every goal line read `+0 tokens`, and a goal on a credit wall kept retrying into the same error until its budget was gone. Both came up in the live comparison against Claude Code, which reports spend per goal and clears the goal on billing failures.

### DIFF
--- a/core/agent/react_loop.py
+++ b/core/agent/react_loop.py
@@ -169,6 +169,9 @@ class ReactLoop[RuntimeToolT: RuntimeTool]:
         self._terminated_by_tool = False
         self._cancelled = False
         self._iterations_used = 0
+        # Provider-reported usage summed over every model call of this run.
+        self._input_tokens = 0
+        self._output_tokens = 0
         self._stop_reason = "iteration_cap"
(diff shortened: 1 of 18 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6119, opened 2026-09-08
feat(goal): real token counts, pause on failed turns, live progress during goal turns; skill metadata

YauhenBichel:
Goal loop
- Token counts are real. Agent responses carry the provider's input and
  output usage, the tool loop puts each call's counts on its
  `provider_request_end` event, and the action driver records them on
  `session.tokens` as they happen. `/goal` lines and `/cost` show real spend,
  a run that raises on a later call keeps the spend of earlier calls, and a
  provider that reports nothing adds no estimate.
- A goal turn that raises (credit wall, rejected key) pauses the goal with
  "the last turn failed; fix the cause, then /goal resume" and re-raises, so
  the host still prints the turn error and the next message does not resume
  into the same failure. Covers a goal attached by `session_goal_set` in the
  same turn; state is stored before the paint so a paint error cannot mask
  the turn error.
- A `/goal` work turn keeps the prompt open. Suspending it hid the spinner,
  the live tool name and the Auto line, so a goal ran with no progress
  signal. Only exclusive-stdin commands hold the next prompt now.

Shell
- Apple Terminal gets one-cell dot spinner frames; braille there bleeds into
  the next cell and leaves fragments at column 0 on repaint. Other terminals
  keep braille. `TERM_PROGRAM_ENV` and `APPLE_TERMINAL_PROGRAM` live in
  `config/constants/terminal_host.py`.

Skills
- Every SKILL.md carries the metadata block (owner, usecases, requires, type,
  version) in the format `github-ci-fix-onboarding` established; owner is
  `Tracer Team` everywhere, including the onboarding card that named a
  person. The template documents the block and a test fails any card
  without it.

### DIFF
--- a/config/constants/__init__.py
+++ b/config/constants/__init__.py
@@ -989,6 +989,12 @@ if TYPE_CHECKING:
     from config.constants.tenancy import (
         INTEGRATIONS_STORE_PATH_ENV as INTEGRATIONS_STORE_PATH_ENV,
     )
+    from config.constants.terminal_host import (
+        APPLE_TERMINAL_PROGRAM as APPLE_TERMINAL_PROGRAM,
+    )
+    from config.constants.terminal_host import (
+        TERM_PROGRAM_ENV as TERM_PROGRAM_ENV,
+    )
     from config.constants.tooling import (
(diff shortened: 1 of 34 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6121, opened 2026-09-08
fix(goal): judge rejects self-contradicting replies; single ending path in the loop

YauhenBichel:
## What
- Judge prompt: a reply that contradicts itself (summary count vs table,
  yes in a row vs no in text, total vs rows) is NOT_REACHED with the
  contradiction named.
- run_until: cancelled, budget exhausted, achieved, impossible, stall pause
  and failed turn all end through `_end`, which stores state, drops queued
  autosubmit, then paints inside a guard. Removed an unreachable
  empty-result fallback, a leftover no-op assignment, a one-line wrapper and
  a stray re-export; the stall menu mapping is immutable.

## Why
Live comparison tonight: on the five-PR task opensre reported "3 re-runs"
above a table showing one, and the judge accepted it. The loop had six
hand-written endings with slightly different ordering.


YauhenBichel:
@greptile review

PR #6123, opened 2026-09-08
fix(goal): judge catches contradictions and repeats, loop ends through one path; action-log box width

YauhenBichel:
Judge
- A reply that contradicts itself is not met: the judge first compares every
  count in the prose with the reply's own table or list and reports a
  mismatch as "Contradiction: a vs b".
- A condition that demands a statement the tool results contradict is
  `impossible`, with the conflicting requirement named, instead of a goal
  that runs to its budget.

Loop
- The same judge verdict twice, with no checklist item ticked in between,
  pauses the goal and opens the stall menu. Repeats are detected on a
  normalised key of the reason (the model rewords the same finding each
  turn); the last verdict is persisted with the goal.
- Every ending (cancelled, budget exhausted, achieved, impossible, stall
  pause, failed turn) goes through one helper that stores state, drops
  queued work, then paints inside a guard.
- Dead code removed: unreachable empty-result fallback, a no-op assignment,
  a one-line wrapper, a stray re-export. Menu mappings are immutable.

Shell
- Action-log box rows are sized to the width the buffered CRLF writer
  renders at; they were one cell wider, so every row wrapped and the box
  corners landed on their own lines.

## Why

Live comparison against Claude Code on the same goals: opensre accepted a
reply that said "3 re-runs" above a table showing one, and on a
self-defeating condition it spent four turns and 588k tokens repeating the
same verdict. Claude Code's evaluator blocked the reply and its worker
declared the condition unsatisfiable.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6133, opened 2026-09-09
fix(goal): host evidence accept, reviewer off, stale plans yield, headless stall stays active

YauhenBichel:
`/goal` used to wait for the model to say it was done, or for a second
model pass on the same turn. That was easy to get wrong: a reply could
close the goal after tools ran, even when the answer was still wrong.
Slack / Telegram could also pause into a menu the user cannot see, and a
plan left over from an earlier request pulled unrelated turns back into
it ("how are you doing?" answered with "I will continue with the CI/CD
flow").

This PR does five things:

1. The host closes `/goal` when a real tool succeeded and no checklist
   item is still open. The cheap judge, reading the tool evidence, can
   only block a contradiction or say the goal is impossible. Reply tags
   like `session_goal:done=` do not close the goal.
2. The extra same-model "are we done?" check on each action turn is
   **off** unless you set `OPENSRE_REACT_GOAL_LLM_REVIEW=1`.
3. If `/goal` stalls with no terminal (ask, Slack, Telegram), the goal
   stays **active**. The next user message continues. The shell still
   shows the keep-going / stop menu. `/goal pause` is unchanged.
4. A plan from an earlier turn yields to the latest message. The CURRENT
   PLAN block now says the user's newest message decides the turn; the
   "keep working the plan" gate and nudge apply only to a turn that
   called `update_plan`; a skill's own answer turn no longer carries the
   generic "then update_plan and execute" guidance; the pinned plan
   leaves the prompt once its "Plan complete" breakdown is printed.
5. The first `/goal` turn no longer pastes the condition twice, and the
   demo A skill keys its next step on which question was answered, so
   "Set up an agent..." schedules the loop instead of re-running the
   analysis.

`opensre ask` still only runs the `/goal` loop when a goal is attached.
A normal one-shot ask is still one turn.

### Root causes found with request dumps (9 Sep, E5–E8)

- The casual-question hijack came from three places pushing every turn
  back into an old plan: the CURRENT PLAN directives ("do not start
  another workload", "execute it now"), the Ask User answered guidance,
  and the reviewer's plan gate. All three are scoped now.
- The demo A replay came from the skill text itself: the step-4 answer
  plus the repository in the conversation matched "a repository in the
  answer means go to step 3". Dumping the exact request showed no other
  cause (no plan existed in that run).
- The `413 request_too_large` seen on the hosted route is not a context
  problem: the failing request had 12 messages and tool results under
  2.5 kB, the same shape passed at 167 kB, and a 900,000-character probe
  was accepted. The message is the proxy's own rule; raised with the
  platform owner.

### Demo / experiments (9 Sep 2026)

All live runs used a real CLI boot (270 tools, `shell_run` present).
HEAD `f6ef5e62c`.

**1. `opensre ask` — no tools**
- Prompt: reply `pong`, do not call tools.
- Result: `pong`, 0 tools, 3.2s.

**2. `opensre ask` — write a file**
- Prompt: write `ok` with `shell_run`.
- Result: 1/1 tool, file on disk contains `ok`, 4.0s.

**3. `/goal` attached — write a file**
- Condition: write `ok` with one shell command and stop.
- Result: **achieved** in 1 turn, file on disk, 5.9s.

**4. `/goal` attached — claim the file is written, do not use tools**
- Result: goal stayed **active**. 0 tools. File missing. Did not close.
  (It returned after 1 turn with a “same verdict twice” label. That
  label is wrong on the first turn; the next message would continue.)

**5. Same-model reviewer on vs off (scripted + live write)**
- Live write-then-stop: both arms wrote the file and stopped. Reviewer
  on added about 14 output tokens and 0.8s. Keep it **off**.
- Scripted: reviewer off accepts after real work with 0 extra model
  calls. Reviewer on can reject the same work and add 1 call.

**6. Host rules (no live model)**

| Case | Result |
| --- | --- |
| Tools succeeded, judge says “not yet” | achieved |
| No tools, judge says “reached” | still active |
| Listing only / failed tool / reply tags | still active |
| Judge says the reply contradicts itself | still active |
| Headless stall | still active, no `/choose` menu |

**7. Earlier live miss (8 Sep, before this work)**
Same five-PR “re-run to green” task: Claude `/goal` and Droid got the
right table. OpenSRE closed after 1 turn because tools ran, and the
answer was wrong. That is why “a tool ran” is no longer enough by
itself when the reply contradicts the evidence — and why we still
do **not** close on talk alone.

**8. Peers (same machine, 9 Sep)**
- Claude Code 2.1.263: `/goal <condition>` starts work and a small
  model checks after each turn.
- Cursor: user card; the same agent marks complete; user owns pause.
- Droid 0.213.0: no `/goal`; `--mission` is a different product.

We match the shared user flow (user sets it, work starts, host keeps
going, stall leaves the goal set). We do **not** copy Droid Missions
onto `/goal`.

Still open: default `/goal` stop is still 5 turns (Claude/Cursor keep
going until done). Evidence is “a tool succeeded”, not a re-check of
live state.

### DIFF
--- a/config/constants/__init__.py
+++ b/config/constants/__init__.py
@@ -467,6 +467,9 @@ if TYPE_CHECKING:
     from config.constants.llm import (
         OPENSRE_LLM_NATIVE_STRUCTURED_OUTPUT_ENV as OPENSRE_LLM_NATIVE_STRUCTURED_OUTPUT_ENV,
     )
+    from config.constants.llm import (
+        OPENSRE_REACT_GOAL_LLM_REVIEW_ENV as OPENSRE_REACT_GOAL_LLM_REVIEW_ENV,
+    )
     from config.constants.mariadb import (
         MARIADB_DATABASE_ENV as MARIADB_DATABASE_ENV,
     )
--- a/config/constants/exports.py
+++ b/config/constants/exports.py
@@ -188,6 +188,7 @@ EXPORTS: dict[str, str] = {
     "LLM_AUTH_METHOD_ENV": "llm",
     "LLM_PROVIDER_ENV": "llm",
     "OPENSRE_LLM_NATIVE_STRUCTURED_OUTPUT_ENV": "llm",
+    "OPENSRE_REACT_GOAL_LLM_REVIEW_ENV": "llm",
     # mariadb
     "MARIADB_DATABASE_ENV": "mariadb",
     "MARIADB_HOST_ENV": "mariadb",
--- a/config/constants/llm.py
+++ b/config/constants/llm.py
@@ -12,6 +12,7 @@ here, and ``core`` may import them downward too.
 
 from __future__ import annotations
 
+import os
 from typing import Final
 
 # --- Connection env-var names ------------------------------------------------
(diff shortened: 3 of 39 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6141, opened 2026-09-09
fix(goal): retry a too-large request once, break from the previous goal, judge must quote the data

YauhenBichel:
Follow-ups from the live walkthrough of #6133. Three defects
seen in the shell, each with its root cause pinned in the code:

1. **Proxy 413 on goal turns.** The hosted route rejected a 12-message,
   ~57k-token request as "The LLM request is too large" on two of three
   attempts of the five-PR task. That message is now classified as context
   overflow. The ReAct loop drops its budget to 60% of the rejected
   request's estimate for the rest of the run, trims low-value tool
   exchanges in place, and sends the request once more. A second
   rejection propagates as before, so there is no retry loop.
2. **A new goal continued the previous goal's work.** The first turn of a
   close-issue goal printed a workflow path because the earlier goal's
   "continue the active goal, take the next step" prompts were still in
   the recent conversation. A goal's first turn now states that earlier
   goals are finished and their steps must not be continued. Resumed
   goals do not get that line.
3. **The judge blocked on a claim nobody could check.** In the five-PR
   run the judge called a contradiction citing a failed CI run that the
   data never showed. The verdict schema gains `evidence_quote`. When tool
   evidence exists, an IMPOSSIBLE verdict or a Contradiction reason must
   quote a passage that appears in the observations or the reply;
   otherwise the goal stays active with "not yet — the judge could not
   point at the data: …" instead of ending or pausing. Without tool
   evidence an impossible verdict still stands (the "say X, call no tool"
   case ends in one turn as before).


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6143, opened 2026-09-09
fix(shell): startup demo menu offers a skip row and no free-text row

YauhenBichel:
The startup onboarding menu had no way to just open the shell: Esc printed
"Selection cancelled" and a typed answer such as "open interactive shell"
went to the model as a demo answer and did nothing useful. The
"Or type your own answer..." row also invited that.

- The menu ends with `Skip the demo and open the shell`. The shell handles
  that row itself: no model turn, the onboarding skill is left, and the
  prompt reads "Demo skipped — type a request, or /demo to come back to it."
- `ask_user_choice` accepts an optional `allow_custom` boolean. The master
  onboarding skill declares `allow_custom: false` in its `pre_execute`
  menu, so the picker shows only the five rows (A–E). Every other menu
  keeps the free-text row.
- The "↗ /goal — work turn" marker above a submitted prompt now prints
  only while a goal is active. It used to appear above `/demo` because the
  queued startup picker also sets the autosubmit flag.


YauhenBichel:
@greptile review

PR #6144, opened 2026-09-09
feat(goal): judge checks the reply against an independent reading of the observations

YauhenBichel:
Live run E10 closed a `/goal` on a table that said Yes for two PRs whose
runs all passed on attempt 1. The judge's quote rule proves a line exists
in the observations, not that it supports the claim, so a confident wrong
table still passed.

Now, before judging, a first classification call reads the tool
observations with the assistant reply withheld and answers the condition
from them alone (counts, names, a yes or no per item). The judge gets that
reading as a section of its input and must set NOT_REACHED with
`Contradiction: reply says X, observations read Y` when the reply's key
facts differ. Turns without tool observations skip the reading.

### Demo / experiments

- Live E14, same five-PR task, today's top five (truth: all No): correct
  table, judge first asked for run data on three PRs the reply had not
  covered, closed on tool evidence at turn 3. E10 on the previous build
  accepted two invented Yes rows.
- Unit: the reading call never sees the reply; the judge prompt carries the
  reading and a Contradiction verdict keeps the goal active; the scripted
  restore test accounts for the extra call per turn.
- Gates: ruff, format, mypy clean; 919 tests across core/agent_harness and
  core/agent.

Cost: one extra cheap call per turn with observations. Goal turns remain
expensive for another reason (81 tool schemas and a 46 kB system prompt on
every iteration); that is the next item.

### DIFF
--- a/core/agent_harness/session_goal/evaluate.py
+++ b/core/agent_harness/session_goal/evaluate.py
@@ -36,6 +36,7 @@ from core.agent_harness.session_goal.judge import (
     SessionGoalJudgeVerdict,
     invoke_session_goal_judge,
     judge_reason_is_contradiction,
+    read_observations,
 )
 from core.agent_harness.session_goal.plan_credit import credit_completed_plan_steps
 from core.agent_harness.session_goal.progress import is_session_goal_progress_text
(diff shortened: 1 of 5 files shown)


YauhenBichel:
@greptile review

PR #6147, opened 2026-09-09
fix(goal): one-pass run history, bounded judge evidence, cached tokens shown

YauhenBichel:
From the live comparison on main (goal-comparison-final-2026-09-09.md):
Claude Code and Droid answer the five-PR re-run question in one pass by
reading each commit's run attempts; opensre's model listed runs without
attempts and needed three tries, and the goal ran out of turns before the
judge accepted the correct table. Cost was also unreadable: the status line
showed raw input tokens with no hint that most were served from the
provider's prompt cache.

1. **Run history in one pass.** `list_github_actions_workflow_runs` says it
   returns `status`, `conclusion` and `run_attempt` per run and that
   `head_sha` gives one commit's history; `github_cli` points re-run
   questions to it because `gh run list` hides attempts. With `head_sha`
   the rows are compact (id, name, status, conclusion, run_number,
   run_attempt, event, timestamps, url). Live: the model went straight to
   that tool for all five commits and produced the correct all-No table on
   turn 1 (E16).
2. **Judge stays available.** E16 also showed the cost of raw run payloads:
   the review input passed its 64k-char cap and the judge went
   "unavailable" every turn, so the correct table was never accepted. Each
   tool result is now bounded at 12k chars with a marker for the dropped
   remainder; the reading and the judge see the head of every observation.
3. **Cached tokens shown.** The OpenAI SDK clients carry
   `prompt_tokens_details.cached_tokens` onto the response; the react loop
   puts it on the usage event; the session records it; the goal keeps a
   cached baseline at attach (persisted); status lines read
   `+342k tok (280k cached)`. A direct probe of the hosted route reports
   3,840 of 4,021 tokens cached on a repeated prefix, so goal turns are far
   cheaper than the raw count suggests.

### DIFF
--- a/core/agent/react_loop.py
+++ b/core/agent/react_loop.py
@@ -387,6 +387,7 @@ class ReactLoop[RuntimeToolT: RuntimeTool]:
             span_attrs["content_chars"] = len(response.content or "")
         input_tokens = int(getattr(response, "input_tokens", 0) or 0)
         output_tokens = int(getattr(response, "output_tokens", 0) or 0)
+        cache_read_tokens = int(getattr(response, "cache_read_tokens", 0) or 0)
         self._input_tokens += input_tokens
         self._output_tokens += output_tokens
         response = self._host._after_response(provider_request, response)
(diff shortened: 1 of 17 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6149, opened 2026-09-09
feat(github): per-workflow verdict lines in a commit's run history

YauhenBichel:
In the live comparison on main, gpt-5.4-mini invented a "re-run to green"
for #6143 in about one run in two (its CI attempt 2 was cancelled). The
judge caught it every time, but each invented Yes cost the goal its turn
budget. Claude Code and Droid never invented one in four runs because they
read the attempt and its conclusion and copy them.

`list_github_actions_workflow_runs` with `head_sha` now adds
`workflow_verdicts`, one line per workflow on that commit:
`latest_attempt`, `latest_conclusion`, `re_run` (attempt above 1) and
`re_run_to_green` (attempt above 1 and success). The tool description
says to copy those lines into the answer rather than infer from the rows.
A cancelled second attempt reads `re_run: true, re_run_to_green: false`,
which is the case the model kept getting wrong.

The listing carries each run's latest attempt only, so earlier attempts'
conclusions are not claimed; `history_fully_fetched` still says when the
commit's history may be incomplete.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6150, opened 2026-09-09
feat(github): copyable summary sentences in a commit's run history

YauhenBichel:
After #6149 the evidence for "was this commit re-run to green" is complete
in one call, yet in live runs E35 and E36 gpt-5.4-mini wrote "Yes, CI" for
#6143 in seven of eight turns while the same tool result carried
`re_run_to_green: false` (attempt 2 was cancelled). The judge blocked it
every time, so nothing false was accepted, but each goal spent its turns.
A boolean was not enough for the worker; a sentence it can copy is the
cheapest next step.

- Each workflow verdict gains `summary`, for example
  "CI: attempt 2 ended cancelled; re-run, but not re-run to green." or
  "CodeQL: attempt 1 success; never re-run."; a run still in progress reads
  "is in_progress".
- The commit gains `history_summary`, for example "No workflow on this
  commit was re-run to green; re-run without green: CI." When the history
  is incomplete the sentence says so, since the sentence is what gets
  copied.
- The tool description tells the model to copy those sentences.
- Judge: an empty cell such as "—" for the workflow on a "no" row is
  correct, not missing. In E35 the worker corrected to No and the judge
  blocked that correct table on the empty cell.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6155, opened 2026-09-09
feat(demo): benchmark comparison, progress while reading, key results at the end of the CI report

YauhenBichel:
- **[A] compare with well-known repositories.** The next-step menu after the
  analysis offers "Compare these numbers with well-known open-source
  repositories". The branch loads `references/benchmarks.md` (which now
  points at `metrics.md`, not a missing file) and runs one analysis per
  benchmark with the same window, the user's repository first. Asked this
  before the change, the model compared star counts.
- **Same-day snapshots.** A 30-day analysis of apache/airflow takes about
  170 s live. Every live analysis now writes its figures and rendered
  report under `~/.opensre/ci_reliability_reports/`; a call for the same
  repository and window within 24 hours answers from it in a second, paints
  the saved report and says "as of <time>". The loop tick shares the writer.
  Prewarm the user's repository and the two benchmarks before the demo.
- **(a) progress while reading.** The collector reports each stage as it
  finishes (default-branch runs, pull request runs, merged PRs with counts
  and elapsed seconds, then the attempt-history check); the shell paints
  them as dim lines. Live: lines at 2 s, 3 s, 17 s, 51 s, 87 s, 101 s on
  Tracer-Cloud/opensre instead of one line after 107 s.
- **(b) results not buried.** No method definitions, no Σ/÷ lines; the
  blocked-PR table shows PR, author, day CI failed, blocked working hours,
  wall clock, with labelled totals under it; the report closes with a
  "Key results" block (default-branch red share with a bar, mean time back
  to green, CI-caused failure share with a bar, slowest normal run,
  developer time blocked per week). Markdown gets the same block.
- **Skill scope.** A slash command typed while a skill's menu is waiting
  (`/auto high` between the repository menu and its answer) no longer ends
  the skill, which had silently dropped the next-step menu in a live run.

### DIFF
--- a/core/agent_harness/prompts/skills/onboarding_cicd_fix/a_local_analysis/SKILL.md
+++ b/core/agent_harness/prompts/skills/onboarding_cicd_fix/a_local_analysis/SKILL.md
@@ -48,6 +48,7 @@ after_tool:
     args:
       title: What would you like to do next?
       options:
+        - Compare these numbers with well-known open-source repositories
         - Set up an agent that improves CI/CD reliability over time
         - Connect OpenSRE to Slack and hand off DevOps chores for your team
         - Exit demo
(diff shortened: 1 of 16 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6157, opened 2026-09-09
fix(shell): readable reply tables, visible quiet commands, menu allow_custom

YauhenBichel:
- Markdown tables render through ReplyMarkdown: bold header, dim column and
  header rules, cells fold instead of ellipsis; code spans keep colour only.
- Quiet shell_run prints a dim one-line command entry and hides the output.
- Menu recap lists the offered rows on a dim line.
- ask_user_choice accepts allow_custom on the run path; the schema property
  crashed every call that used it and the same menu was asked in a loop.


YauhenBichel:
@greptile review

PR #6158, opened 2026-09-09
fix(shell): rerun failed commands before answering; full command in tool frame rows

YauhenBichel:
- System prompt shell rules: fix and rerun a failing command (uv run python
  has the project libraries); never present a regex or guess as a measured
  figure; quiet hides output only.
- Tool frame rows keep the full collapsed command and clip to the terminal
  width instead of a fixed 72 characters, so file names survive.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6159, opened 2026-09-09
fix(shell): keep a menu batch's open questions when some were just answered

YauhenBichel:
Follow-up to #6158, which made `ask_user_choice` refuse a question the user
had already answered in the same message (demo B asked "When should it run?"
four times). Greptile flagged a P1: when a batched request mixed one answered
question with unanswered ones, the whole batch was refused and the open
questions were lost, so the model had to rebuild a smaller batch on its own.

The guard now works per question:

- answered questions leave the batch, the rest are queued as before;
- when one question remains, it is asked as a plain single decision so the
  two-question minimum for batches does not reject it;
- the tool refuses only when every question was already answered, and the
  error lists each question with its answer so the model can continue.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6160, opened 2026-09-09
fix(shell): reply rows keep a right margin; CI report paints like a reply

YauhenBichel:
Right margin. At 210 columns the closing reply showed a blank line inside a
wrapped paragraph and a one-space continuation. Reproduced: the reply row was
padded to exactly the terminal width, and a full-width row followed by a
newline makes Terminal.app wrap before the newline. Measured against the
other shells at 120 columns, Cursor wraps prose 7 columns short, Droid 1,
Claude Code 0; opensre wrapped at 113 in the streaming path but at the full
width in the closing path. Reply, note and streaming rows now render two
columns short of the console width (`reply_width` in the streaming renderer),
pinned by a 210-column test.

One look for the CI report. The CI analytics tool painted its report with its
own Rich objects: dim labels, dim table header, no column rules, a formula
footer. That is why the demo report looked like a second application next to
the Ω reply, and why the theme change alone did not remove the grey. The
painter now prints the tool's markdown twin through `ReplyMarkdown`, which
moved from surfaces to `infrastructure/terminal/markdown.py` so integrations
can use it (one canonical path; the surfaces module is removed; `repl_table`
shares the box constant). Dead painter helpers are gone. Painted output now
has the reply's body colour, bold headings, accent table header, dim rules
and the same two-cell indent as the Ω gutter. Coverage notices are reworded
in plain language.

Greptile P1 on #6159 (benchmarks truncated analysis details): the compact
markdown switch is separate from `include_benchmarks`. Only the schedule
closing asks for compact; a non-terminal caller asking for benchmarks keeps
the full analysis (test added).

Theme. The purple palette set body text to a 71% grey, so every reply element
was a grey shade with bold as the only structure. Purple body text is now
#D0D0D0, the same as the blue theme and Droid's body, secondary #B4B4BC, and
markdown table headers use the bold accent.

### DIFF
--- a/core/agent_harness/ports.py
+++ b/core/agent_harness/ports.py
@@ -160,12 +160,17 @@ class ToolProvider(Protocol):
         confirm_fn: ConfirmFn | None,
         is_tty: bool | None,
         resolved_integrations: dict[str, Any] | None = None,
+        turn_user_message: str = "",
     ) -> list[Any]:
         """Return the agent tools available for this turn.
 
         When ``resolved_integrations`` is supplied it is the turn's single
         resolved-integration view (from ``TurnSnapshot``); the provider builds
         tools from it instead of resolving again, so tools and the prompt agree.
+
+        ``turn_user_message`` is what the user sent this turn; tools that must
+        not repeat work the message already settled (an answered menu) read it
+        from the scope.
         """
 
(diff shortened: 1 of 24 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6161, opened 2026-09-09
fix(shell): drop a painted tool's log row and its restated closing

YauhenBichel:
The tool row printed under its own output. Tool rows are buffered during a
turn and flushed once at the end so consecutive calls can group into a box.
The analytics tool paints its report while it runs, so the row
"analyze github ci reliability" always arrived after the tables it names.
Droid and Cursor put the label first.

The figures printed twice. On a terminal the tool returned its summary
sentence as the turn's reply, so "8151 runs in 30 days, 1174 of 5727 PR runs
failed, 333 CI-caused" appeared again under a report whose key results had
just shown those numbers. The skill already tells the model not to restate
figures; this duplicate came from the tool, not the model.

The result payload already carried `rendered_in_shell`, true only when the
tool painted to a terminal. Two consumers now honour it:

- the shell's observer withdraws the buffered row for that call, using a new
  `TerminalSession.drop_action_log`;
- `format_generic_tool_payload` treats such a result as already shown, so
  nothing re-prints it. This guard is needed on its own: without it the
  formatter falls through to the payload's `summary` field and prints the
  figures anyway.

The analytics tool's painted path stops returning a reply text, in both the
live and the saved-snapshot branch. The report is the output.

A name list was the alternative and was rejected: it would have put a GitHub
tool name inside both the shell surface and core, and it would have silently
stopped working if the tool were renamed. A tool now says for itself that it
painted.

### DIFF
--- a/core/agent_harness/turns/display_text.py
+++ b/core/agent_harness/turns/display_text.py
@@ -22,10 +22,20 @@ from infrastructure.terminal.peek import (
 from infrastructure.text import is_data_blob
 
 # Tools whose result the host already rendered to the console; their payload is
-# not re-shown in the transcript.
+# not re-shown in the transcript. A tool outside this set says so per call with
+# ``rendered_in_shell`` in its payload (it paints only on a terminal surface).
 _HOST_RENDERED_TOOL_NAMES: frozenset[str] = frozenset(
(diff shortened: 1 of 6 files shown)


YauhenBichel:
@greptile review

PR #6162, opened 2026-09-09
fix(demo): the CI comparison is always built; the model chooses brevity instead

YauhenBichel:
Demo A ran without the benchmark comparison. The cause was a contradiction
the model resolved the wrong way: the skill says to call the analysis with
`include_benchmarks=true`, while the tool's own schema described the flag as
"Default false". The model followed the schema and passed false, so the
comparison against apache/airflow and fastapi/fastapi was missing and the
long counts appendix printed instead.

The flag conflated two different decisions:

- whether to compare, which reads same-day snapshots only and never fetches,
  so it costs nothing and is not a judgement call;
- how long the report is, which is a real choice.

`include_benchmarks` is gone from the public schema, so the model can no
longer turn the comparison off, and the comparison always runs. The schema
offers `compact` instead, which drops the counts appendix (executions,
failure classification, blocked time, workflows). The analysis skill now
calls the tool with `compact=true`; the benchmarks reference no longer
mentions a flag.

The failure mode is now safe in both directions: if the model forgets
`compact`, the reader gets the comparison plus more detail, instead of
losing the section the demo exists to show.

Live demo A on this branch, warm snapshots. The comparison is present:

    Metric                 │ Tracer-Cloud/opensre │ apache/airflow │ fastapi/fastapi
    Red time on main       │                 5.1% │           3.4% │            0.0%

On main, the same run called the tool with `include_benchmarks: false` and
produced no comparison at all. On this branch the model still passed
`compact: false` against the skill, and the only consequence is the extra
appendix: the table is there either way.

### DIFF
--- a/core/agent_harness/prompts/skills/onboarding_cicd_fix/a_local_analysis/SKILL.md
+++ b/core/agent_harness/prompts/skills/onboarding_cicd_fix/a_local_analysis/SKILL.md
@@ -127,7 +127,7 @@ the answer.
 ### 3. Analyze CI/CD reliability
 
 Call
-`analyze_github_ci_reliability(owner="<owner>", repo="<repo>", include_benchmarks=true)`
+`analyze_github_ci_reliability(owner="<owner>", repo="<repo>", compact=true)`
 for the chosen repository. The tool paints the report (key results first)
 and the comparison table itself. Do not restate figures, do not output
 `headline`, and do not call the tool again for benchmarks. A peer without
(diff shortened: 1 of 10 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6163, opened 2026-09-09
fix(shell): a plan update no longer rescues a restated closing

YauhenBichel:
Follow-up to #6161 and #6162, found by a live sweep of the shell rather than
by a test.

#6161 drops a closing that restates a report the tool already painted, the
same way the shell has long dropped a closing written over a shell command's
output. The predicate required *every* tool result in the turn to be painted.
Real turns are not shaped that way: the analytics skill calls `update_plan`
alongside the analysis, and `update_plan` is not a painted result, so the
rule never fired on the path that matters.

Two live captures show the damage. One closing restated the report as six
bullets. Another added "I skipped the benchmark comparison as requested" with
the comparison table sitting directly above it. A turn that ran the analysis
alone suppressed correctly, which is why earlier spot checks passed.

`update_plan` and the menu tools carry no content of their own — the host
draws them — so they are now ignored when deciding, and at least one painted
result is required. The "is this already on screen" predicate moves from
private to shared in `display_text` so the driver and the transcript
formatter answer the question the same way instead of each keeping a copy.

Same prompt, before and after, in a real shell with a warm snapshot.

Before ("Analyze CI/CD reliability for Tracer-Cloud/opensre, but skip the
benchmark comparison"):

    ...comparison table...
    Ω Good news: Tracer-Cloud/opensre had 8,151 runs in the last 30 days.
       • PR health: 1,174 of 5,727 PR runs failed; 333 were CI-caused.
       • Main branch: red for 36.4h total, with 41 breakages.
       • I skipped the benchmark comparison as requested.

After:

    ...comparison table...
     • Coverage notice: attempt history was checked for the first 500 requests only...
     • fastapi/fastapi: default branch stayed green in this window.
    (next-step menu)

### DIFF
--- a/core/agent_harness/turns/action_driver.py
+++ b/core/agent_harness/turns/action_driver.py
@@ -51,6 +51,7 @@ from core.agent_harness.turns.conversation_recording import record_conversation_
 from core.agent_harness.turns.display_text import (
     cap_for_display,
     format_generic_tool_payload,
+    host_rendered,
     looks_like_json,
     preferred_tool_response_text,
     split_output_truncation_markers,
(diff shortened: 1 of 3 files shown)


YauhenBichel:
@greptile review

PR #6164, opened 2026-09-09
fix(prompt): a count taken from a truncated read is wrong, not approximate

YauhenBichel:
Found by a live sweep of the shell, not by a test.

Asked for a table of the workflows with their job counts, the shell reported
3 jobs for ci.yml and 4 for release.yml. Both have 7. The transcript shows
why: it ran `sed -n '1,260p' .github/workflows/ci.yml` against a file of 868
lines, and `sed -n '1,240p' release.yml` against 854 lines, then counted the
jobs it could see. The reply presented those numbers as measured.

This is not the failure #6158 already covers. That rule fires when a command
fails; here every command succeeded and the answer was still wrong.

This is not the failure #6158 already covers. That rule fires when a command
fails; here every command succeeded and the answer was still wrong.

The existing guidance pushes in this direction. "Do not use python scripts to
attempt to output larger chunks of a file" steers away from parsing, and
`sed` appears in the list of reads to parallelise, while nothing warns that a
range read drops everything past the cut. The shell section now says that
counting or measuring from a file means reading all of it, that a range read
makes a count wrong rather than approximate, and that the length should be
checked first or the whole file parsed.

This one is prompt guidance rather than a guard, deliberately. Code cannot
tell that a model truncated its own input, and inspecting commands for range
reads is exactly the fragile pattern matching the codebase forbids.

Ground truth, by parsing the files:

    ci.yml       868 lines, 7 jobs   (model read to line 260, reported 3)
    release.yml  854 lines, 7 jobs   (model read to line 240, reported 4)

The same prompt in a session where the model parsed with the project's Python
gave all nine counts correctly, so the numbers are reachable; the reading
method was the difference.

A live rerun of the failing prompt at the same terminal width is in progress
and is not included in this description yet.


YauhenBichel:
@greptile review

PR #6165, opened 2026-09-09
feat(tools): read_structured_file, so counts come from a parser

YauhenBichel:
A live sweep of the shell found it giving confidently wrong numbers about the
user's own repository. Asked how many jobs `.github/workflows/ci.yml` defines,
it answered 3; the file defines 7. It had matched a pattern against indented
keys, which also matches `steps`, `env` and `permissions`, after reading only
the first 220 lines of an 868-line file. Three of five counting prompts came
back wrong.

Two rules added the same day (#6164 and its follow-up) cover exactly these
mistakes and did not hold. The same question answered correctly in one turn
and incorrectly in the next, decided by phrasing. Prompt text was not the
lever.

`read_structured_file` parses YAML, JSON or TOML and reports what one key
holds: its kind, how many entries it has, and their names. Counting jobs is
then reading the length of a mapping — no indentation to miscount, no tail of
the file to miss. It is read-only and takes a dotted path, so
`project.dependencies` walks into a TOML table.

Two details the tests pin, both real:

- A workflow's `on:` key is read by YAML 1.1 as the boolean `true`, so looking
  it up by the word the file shows found nothing. Every trigger question would
  have failed. The lookup accepts YAML's boolean spellings.
- A key whose own name contains a dot (`"a.b"`) is matched before the dot is
  treated as a path separator, so real key names win over the syntax.

The system prompt gains one line pointing at parsing rather than pattern
matching, and saying not to fill one column of a table with a plausible value
while admitting another column is unknown — the shell did exactly that.

Live shell, two turns, 60 seconds, with the tool available:

    ▌ How many top-level jobs does .github/workflows/ci.yml define?
    ⏺ read structured file
    Ω 7 entries under `jobs` in .github/workflows/ci.yml: quality-static,
      quality-typecheck, test, coverage-report, session-store-locked,
      package-preflight, ci-gate

    ▌ What triggers .github/workflows/codeql.yml? List them.
    ⏺ read structured file
    Ω 3 entries under `on` in .github/workflows/codeql.yml: push, schedule,
      workflow_dispatch

The session record shows only this tool and the slash command: no shell
scripting, no pattern matching, no range reads. Before this change the same
first question answered 3.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6166, opened 2026-09-09
fix(tools): YAML anchors load, and file counts skip generated directories

YauhenBichel:
Two things, both found by running the shell.

**YAML anchors made a file unreadable** (Greptile P1 on #6165). To keep a workflow's `on:` key from being reported as `True`, #6165 gave the loader its own mapping constructor. That replaced the one PyYAML uses to expand `<<`, so any document with an anchor failed to load and the tool called the whole file invalid — including workflows that share defaults through an anchor, which is a common shape. The constructor now expands merges before reading keys. A test loads an anchored document and checks the merged key is part of the job.

**File counts counted caches.** Asked how many test files a package has, the shell answered 233, then 263 in an earlier run; there are 103. `find -type f` counts everything on disk, and a `test_*` glob also matches the compiled `.pyc` files under `__pycache__`. `count_files` counts files under a directory matching a glob and prunes generated directories (`__pycache__`, `.venv`, `venv`, `site-packages`, `node_modules`, `build`, `dist`, `.git`, and the mypy, pytest, ruff and tox caches). Symlinked directories are not followed, so a link back up the tree cannot make the count run away. The count always covers the whole tree; only the example names are capped, so it cannot repeat the truncation that caused the original wrong numbers.

The skip set follows the convention already used by the workspace scan rather than adding an ignore-file parser, since the library for that is only a transitive dependency here.

No prompt rule accompanies this. Three were tried tonight for this class of error and none could be shown to change behaviour in a live session, so the tool stands on its own

Live shell, three questions, one tool call each:

    ▌ How many test files are under tests/interactive_shell?
    ⏺ count files
    Ω 103 files matching test* under tests/interactive_shell, not counting
      generated directories

    ▌ How many Python files are in surfaces/interactive_shell?
    ⏺ count files
    Ω 139 files matching *.py under surfaces/interactive_shell, not counting
      generated directories

    ▌ How many workflow files are in .github/workflows?
    ⏺ count files
    Ω 9 files matching *.yml under .github/workflows, not counting generated
      directories

All three match the repository. The first answered 233 before this change.
Note the model passed `test*`, not `test_*.py`, and still got 103: the skip
set is doing the work, not the phrasing.

Anchors, before and after, on a workflow using `<<: *defaults`:

    before: "… is not valid yaml: ConstructorError"
    after:  jobs -> 2 ('build', 'test'); jobs.build -> ('runs-on', 'steps')

### DIFF
--- a/docs/counting-files.mdx
+++ b/docs/counting-files.mdx
@@ -5,13 +5,9 @@ description: "Ask how many files of a kind a directory holds and get the project
 
 ## Overview
 
-"How many test files are in this package" has an exact answer, and the obvious
-shell command gets it wrong. `find -type f` counts everything on disk, so
-compiled caches under `__pycache__` are counted alongside the modules you wrote
-— in this repository that turns 103 test files into 233.
-
-OpenSRE counts the files that belong to the project. Ask in the interactive
-shell (`opensre`):
+OpenSRE counts the files that belong to your project: the modules you wrote,
+not the compiled caches and vendored packages sitting beside them. Ask in the
+interactive shell (`opensre`):
 
 ```
 How many test files are under tests/interactive_shell?
--- a/docs/counting-files.mdx
+++ b/docs/counting-files.mdx
@@ -1,0 +1,49 @@
+---
+title: "Counting files"
+description: "Ask how many files of a kind a directory holds and get the project's files, not its caches"
+---
+
+## Overview
+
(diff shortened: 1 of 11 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6168, opened 2026-09-09
fix(shell): the onboarding question is asked once, unless you ask for it

YauhenBichel:
Found by running all four onboarding demos, which had never been exercised
end to end. A and B were fine. C looped.

Picking "Run CI/CD improvements with a managed service (coming soon)" gave
three demo menus for one choice. The branch is a placeholder that answers and
stops, so the session has nothing active; the next plain message ("hi") routed
back to the master onboarding skill, whose `pre_execute` opens the menu, and
the cycle repeated. A viewer sees the shell unable to leave a dead end.

Two paths brought the question back, and both are closed:
- **The skill's entry hook.** The model may no longer reopen a menu the session
  has already opened; the host still may.
- **The model asking directly.** The session record showed it calling
  `ask_user_choice` with the onboarding question itself, which the hook rule
  does not touch. The shell already refuses a question answered in the *same
  turn*; that now covers the whole session. Answering a menu records the
  question, normalized for spacing and case, and a later attempt is refused
  with the answer so the model continues instead of asking again.

`/demo` and startup still open the menu. They clear the session's record for
that skill's questions first — asking again is what the command is for. This
mattered: an earlier revision of this change broke `/demo` completely (it
queued nothing and the shell printed nothing), because the session-wide guard
refused the entry hook too. There is now a test for a host entry *after* an
answer, which is the shape that regressed.

The managed-service skill also names what works now — repository analysis or
the recurring agent — instead of ending on "coming soon" alone.

Same run, before and after, picking the managed-service option:

    before: menu → "coming soon" → menu → master skill → menu → "coming soon"
    after:  menu → "coming soon. For now, I can help you analyze a repository's
            CI/CD performance or set up a recurring reliability agent."

Demo menus in the transcript: 3 → 1.

All three entry paths checked directly:

    startup                       menu opens
    model re-entering the skill   no menu
    /demo                         menu opens

`/demo` verified in a real shell after the fix, reopening the menu as expected.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6169, opened 2026-09-10
fix(skills): the scheduled-agent demo asks one declared question

YauhenBichel:
Running all four onboarding demos on main showed demo B drifting from its own
script. The skill says to ask "When should it run?" with two options, weekdays
or daily. The live run asked:

    When should the recurring CI/CD reliability loop run for
    `deepseek-ai/deepseek-harness`?
    Weekdays at 08:00 · Weekdays at 09:00 · Daily at 08:00 · Use a custom time

The schedule that followed was correct, so nothing failed — but the demo shows
a different question each run, and a title built from the repository name is
unique every time, which defeats the guard added in #6168 that stops a question
being asked twice. That guard compares titles.

The cause is the shape of the skill: it describes its menus in prose and asks
the model to reproduce them exactly. Skill A does not — it declares both menus
in frontmatter and the host runs them, which is why the demo menu is identical
every time. Demo B now does the same:

- The repository question becomes an `after_tool` hook on
  `scan_local_git_workspace`, with `options_from: local_git_scan_repos` so the
  candidate list still comes from the scan, plus the example repository.
- The cadence question is removed. The loop schedules weekdays at 08:00, and
  the schedule card already states that and tells the user how to change it
  with `/loops`. One less step, and nothing left for the model to invent.

The body now tells the model not to call `ask_user_choice` for the repository
question and not to ask about cadence at all; the plan steps drop from four to
three to match.

Live run after the change, answering the startup menu with B:

    ↳ Which repository should the agent watch?  deepseek-ai/deepseek-harness (822 commits, CI configured)
      offered: Tracer-Cloud/opensre (1588 commits, CI configured) ·
      deepseek-ai/deepseek-harness (822 commits, CI configured) ·
      YauhenBichel/py-harness (496 commits, CI configured) ·
      Use the open-source example repository (Tracer-Cloud/opensre)
    ...
    Scheduled: CI reliability check · deepseek-ai/deepseek-harness
    Runs weekdays at 08:00 Europe/London; next run 2026-09-10T07:00:00+00:00.

Cadence menus in the transcript: 1 → 0. Model-authored menus: 2 → 0.

### DIFF
--- a/core/agent_harness/prompts/skills/onboarding_cicd_fix/b_local_scheduled_loops/SKILL.md
+++ b/core/agent_harness/prompts/skills/onboarding_cicd_fix/b_local_scheduled_loops/SKILL.md
@@ -11,7 +11,7 @@ getting_started: Set up an agent that improves CI/CD reliability over time
 demo_order: 2
 metadata:
   owner: Vincent
-  last_changed_by: Vincent
+  last_changed_by: Yauhen
   last_changed_at: 2026-09-09
   usecases:
     - First-experience demo: schedule a weekday CI/CD reliability agent for one repo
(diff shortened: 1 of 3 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6171, opened 2026-09-10
fix(scheduler): persist successful manual task runs

### LINKED ISSUE #6170: [BUG] Manual cron and loop runs do not update task completion state
## Summary

  Successful manual task executions do not update the scheduled task's persisted state.

  `run_task_now()` returns directly from `execute_task()` and does not call `record_task_success()`.

  ## Affected code

  - `infrastructure/scheduling/scheduler/storage/task_store.py`

  Normal scheduled execution calls `record_task_success()` after a successful task run, but manual execution through
  `run_task_now()` does not.

  ## Reproduction

  1. Create an enabled scheduled task.
  2. Run it manually:

  ```bash
  opensre cron run <task-id>
 ```

  or run a loop manually:

  /loops run <loop-id>

  3. Check the task state:

  opensre cron list

  or:

  /loops list

  ## Expected behavior

  After a successful manual run:

  - last_run should be updated.
  - The task should reflect the latest successful execution.
  - disable_after_success=true should disable the task after success.

  ## Actual behavior

  The manual command reports success and the execution is recorded in SQLite run history, but:

  - last_run remains stale or empty.
  - The task remains enabled even when disable_after_success=true.
  - /cron list and /loops list display outdated task state.

  ## Impact

  Manual recovery or first-run testing can produce misleading scheduler state and may cause one-shot tasks to run again
  unexpectedly.

Devesh36:
Fixes #6170

  Manual `cron run` and `/loops run` executions now persist task completion state after a successful delivery.

  - Call `record_task_success(task.id)` after `run_task_now()` succeeds.
  - This updates `last_run` and honors `disable_after_success=true`, matching scheduled and recovered task execution.
  - Add a regression test confirming successful ad-hoc runs invoke the task-state updater.

  
  run_task_now() already delegates execution and delivery to execute_task(), but it returned immediately after success. Scheduled and recovery paths additionally call
  record_task_success(), which owns task-level state updates such as last_run and disable_after_success.

  I kept the behavior aligned with those existing paths: after a successful manual execution, call record_task_success(task.id) and return the execution result unchanged.
  Failed or refused manual runs do not update completion state.

  The regression test mocks a successful execution and asserts that the task-state updater is called with the correct task ID.

### DIFF
--- a/infrastructure/scheduling/scheduler/runner.py
+++ b/infrastructure/scheduling/scheduler/runner.py
@@ -35,6 +35,7 @@ from infrastructure.scheduling.scheduler.storage import (
     complete_run,
     default_task_store_path,
     get_recoverable_runs,
+    get_runs,
     get_task,
     list_tasks,
     record_task_success,
@@ -498,7 +499,9 @@ def run_task_now(task_id: str, runners: SchedulerRunners, *, only_failed: bool =
     fire_time = datetime.now(UTC).strftime("%Y-%m-%dT%H:%M:%SZ")
     result = execute_task(task, fire_time, runners, target_filter=target_filter)
     if result:
-        record_task_success(task.id)
+        run = next((run for run in get_runs(task.id) if run.fire_time == fire_time), None)
+        if run is not None and all(outcome.ok for outcome in run.targets):
+            record_task_success(task.id)
     return result
 
 
--- a/infrastructure/scheduling/scheduler/runner.py
+++ b/infrastructure/scheduling/scheduler/runner.py
@@ -496,7 +496,10 @@ def run_task_now(task_id: str, runners: SchedulerRunners, *, only_failed: bool =
             return False
 
     fire_time = datetime.now(UTC).strftime("%Y-%m-%dT%H:%M:%SZ")
-    return execute_task(task, fire_time, runners, target_filter=target_filter)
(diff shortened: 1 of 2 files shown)


Devesh36:
@greptile review

muddlebee:
Good catch on persisting manual runs. One follow-up on completion semantics: this PR now correctly avoids calling `record_task_success()` when a manual fan-out has a partial delivery, but the scheduled (`_scheduled_job`) and recovery (`_recover_runs`) paths still call it whenever `execute_task()` returns `True`.

Since fan-out intentionally returns `True` for a partial delivery (the run is stored as `SUCCESS`, with failed target outcomes), a one-shot scheduled or recovered task can still update `last_run` and disable itself while some destinations missed the message.

Could we extract one shared “fully delivered” completion predicate/helper and use it in all three paths—manual, scheduled, and recovery? That keeps `last_run` / `disable_after_success` consistent regardless of how the task was started.

muddlebee:
Tracked the shared completion-semantics follow-up in #6175.

muddlebee:
@codex review

Devesh36:
@muddlebee I’m keeping this PR scoped to manual-run persistence. The shared completion semantics for scheduled and recovery runs are tracked separately in #6175 and will be handled in a follow up PR

PR #6177, opened 2026-09-10
fix(shell): the schedule card is a list, not a paragraph

YauhenBichel:
The scheduled-agent demo closes with a card that arrived as one dense
paragraph — the schedule, where reports land, and three /loops commands
run together. The lines were separate in source; markdown folds
consecutive lines into a paragraph.

- loop_card returns a LoopCard (headline + one fact per bullet) instead
  of a list of sentences. The action tool sends its markdown; the
  startup demo paints the same content through ReplyMarkdown, so both
  entry points show one card.
- Four bullets instead of five sentences, none wrapping at 120 columns.
- The next run prints in the schedule's timezone ("next Fri 11 Sep
  08:00") rather than a UTC ISO stamp contradicting the local time
  stated beside it.
- System prompt: three or more actionable things are one bullet each,
  never a run of sentences; a card a tool already rendered is repeated
  line for line, never re-flowed into prose.


YauhenBichel:
@greptile review

PR #6178, opened 2026-09-10
PR title: feat(ci-analytics): first run is the demo — cost first, no cache, shipped benchmarks

YauhenBichel:
- The report leads with one sentence naming the cost ("Waiting on CI  cost 213.7h of developer time in the last 30 days, about 1.6h per   developer a week"), then Key results. The duplicate developer-time  bullet is removed.
- analyze_github_ci_reliability always reads GitHub. The same-day  snapshot answer path is deleted; snapshots are still written for the  scheduled loop, which is the only reader left.
- Benchmark columns ship with the product (apache/airflow and  fastapi/fastapi, measured with this tool over 30 days on 9 Sep 2026)  instead of depending on peer snapshots that a first run does not  have. The table states the measurement day; the "no same-day  snapshot" skip text is gone.
- Menu titles drop "(Esc to skip)"; the menu footer already says it,  and the answered recap now shows the bare question.

### DIFF
--- a/core/agent_harness/prompts/skills/onboarding_cicd_fix/SKILL.md
+++ b/core/agent_harness/prompts/skills/onboarding_cicd_fix/SKILL.md
@@ -27,7 +27,7 @@ metadata:
 pre_execute:
   - tool: ask_user_choice
     args:
-      title: Which demo would you like me to run? (Esc to skip)
+      title: Which demo would you like me to run?
       note: >-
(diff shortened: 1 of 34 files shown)


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6181, opened 2026-09-10
fix(observability): OTLP parser tolerates null JSON fields

mdiniz97:
No linked issue — small robustness bug found while auditing JSON shape handling in the shared OTLP parser.

`parse_otlp_trace` and `extract_span_attributes` assumed OTLP/JSON collection fields were either present or absent, never `null`. A payload like `{"batches": null}`, `{"resource": null}` or `{"attributes": null}` reached a `for` loop and raised `TypeError: 'NoneType' object is not iterable`; a `null` attribute value raised `TypeError`, and a non-dict `resource`/`attributes` raised `AttributeError`.

`integrations/tempo/client.py` calls `parse_otlp_trace(payload or {})` without a catch, so a Tempo response with a null field failed the `tempo_get_trace` tool instead of returning spans. The Grafana Cloud Tempo mixin catches the exception but then silently reports zero spans.

The parser now normalizes every collection and mapping before iterating (`batches`, `resource`, `scopeSpans`, `spans`, `attributes`, `status`, attribute entries and their `value`), consistent with the non-dict entry guards already in the function.

Before (new tests against `main`):

```
tests/infrastructure/observability/test_otlp_trace.py::test_extract_span_attributes_tolerates_null_and_non_dict_entries
    for attr in span.get("attributes", []):
E   TypeError: 'NoneType' object is not iterable

tests/infrastructure/observability/test_otlp_trace.py::test_parse_otlp_trace_tolerates_null_collections
    for batch in trace_data.get("batches", []):
E   TypeError: 'NoneType' object is not iterable
```

After:

```
$ uv run python -m pytest tests/infrastructure/observability/ -q
156 passed

$ uv run python -m pytest tests/integrations/grafana/test_tempo.py tests/integrations/tempo -q
18 passed
```

`make lint`, `make format-check` and `make typecheck` are clean.

---

The root cause is that `dict.get(key, default)` only covers a missing key, not an explicit `null`, and OTLP/JSON can carry null collections. I chose one normalization per field before the loop rather than sprinkling `or []` inline: `isinstance(x, list)` for collections and `isinstance(x, dict)` for mappings, matching the `isinstance(batch, dict)` guards already used one level down. `extract_span_attributes` also accepts a non-dict span now because `parse_otlp_trace` can pass a `resource` that is present but not a mapping.

---

- [x] I have added a proper PR title and described the change (no tracking issue exists for this small fix)

### DIFF
--- a/infrastructure/observability/otlp_parser.py
+++ b/infrastructure/observability/otlp_parser.py
@@ -28,11 +28,20 @@ def extract_span_attributes(span: dict[str, Any]) -> dict[str, Any]:
     with an unsupported value kind are skipped.
     """
     attributes: dict[str, Any] = {}
-    for attr in span.get("attributes", []):
+    if not isinstance(span, dict):
+        return attributes
+    raw_attributes = span.get("attributes")
+    if not isinstance(raw_attributes, list):
+        return attributes
+    for attr in raw_attributes:
+        if not isinstance(attr, dict):
+            continue
         key = attr.get("key", "")
         if not key:
             continue
-        value = attr.get("value", {})
+        value = attr.get("value")
+        if not isinstance(value, dict):
+            continue
         for kind in _OTLP_SCALAR_KINDS:
             if kind in value:
                 attributes[key] = value[kind]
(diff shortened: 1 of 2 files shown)


mdiniz97:
@greptile review

PR #6189, opened 2026-09-11
fix(scheduler): add scheduled Buzz delivery

### LINKED ISSUE #6188: Buzz is advertised by cron but has no scheduler adapter
## Summary

  The generic cron CLI accepts `buzz` as a provider, but the scheduler has no Buzz delivery adapter.

  As a result, users can successfully create a Buzz schedule that always fails during execution with `Unsupported provider`.

  ## Reproduction

  ```bash
  opensre cron add \
    --kind manual_loop \
    --cron "0 9 * * *" \
    --prompt "Check open incidents" \
    --provider buzz \
    --chat-id <channel-id>
   ```
  The task is created successfully. When the scheduler fires, delivery fails because Provider.BUZZ is not present in the installed scheduled-delivery
  adapter bundle.

  ## Expected behavior

  Either:

  1. implement scheduled Buzz delivery, or
  2. remove buzz from generic cron provider choices and reject it clearly.

  The CLI must not advertise providers that the scheduler cannot execute.

  ## Affected files

  - surfaces/cli/commands/cron.py
  - bootstrap/adapters.py
  - infrastructure/scheduling/scheduler/delivery_bundle.py
  - infrastructure/scheduling/scheduler/types.py

  ## Acceptance criteria

  - Every provider shown by cron add --help has a working scheduler adapter.
  - Unsupported providers are rejected during task creation.
  - Add a provider-conformance test covering CLI choices and installed adapters.

Devesh36:
Fixes #6188

  Adds the missing scheduled-delivery adapter for Buzz.

  - Adds `BuzzScheduledDelivery`, using existing Buzz credential resolution and `post_buzz_message`.
  - Registers `Provider.BUZZ` in the scheduler delivery bundle.
  - Requires a Buzz private key and the task’s explicit channel ID before delivery.
  - Replaces the known Buzz mismatch test with a conformance check: every provider advertised by `opensre cron add` has an installed scheduler adapter.
  - Adds focused tests for successful Buzz delivery and missing credentials/channel handling.

  Buzz already had the scheduler-compatible pieces: credential resolution preserves the configured store/env/keyring precedence, and post_buzz_message owns client invocation and secret-safe error handling. This PR adds only the thin vendor-owned adapter that connects those pieces to the scheduler contract, then registers it at the composition root.

Removing Buzz from the generic cron choices was considered, but Buzz is already documented as an outbound messaging channel. Adding the adapter preserves that expected capability and prevents validly created schedules from failing later with Unsupported provider.

### DIFF
--- a/bootstrap/adapters.py
+++ b/bootstrap/adapters.py
@@ -65,6 +65,7 @@ def scheduled_delivery_adapters() -> ScheduledDeliveryAdapters:
         InteractiveShellScheduledDelivery,
     )
     from infrastructure.scheduling.scheduler.types import Provider
+    from integrations.buzz.scheduled_delivery import BuzzScheduledDelivery
     from integrations.discord.scheduled_delivery import DiscordScheduledDelivery
     from integrations.rocketchat.scheduled_delivery import RocketChatScheduledDelivery
     from integrations.slack.scheduled_delivery import SlackScheduledDelivery
@@ -77,6 +78,7 @@ def scheduled_delivery_adapters() -> ScheduledDeliveryAdapters:
             Provider.DISCORD: DiscordScheduledDelivery(),
             Provider.ROCKETCHAT: RocketChatScheduledDelivery(),
             Provider.INTERACTIVE_SHELL: InteractiveShellScheduledDelivery(),
+            Provider.BUZZ: BuzzScheduledDelivery(),
         }
     )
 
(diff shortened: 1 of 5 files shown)


Devesh36:
@greptile review

muddlebee (changes requested):
@Devesh36 you will need to setup Buzz for this and test it out locally

Devesh36:
> @Devesh36 you will need to setup Buzz for this and test it out locally

wip

PR #6194, opened 2026-09-11
fix(work-items): validate Slack reminder destinations

### LINKED ISSUE #6192: Work-item Slack reminders accept undeliverable targets
## Summary

  Work-item reminder validation accepts Slack targets with an empty `chat_id`, even when no Slack webhook is configured.

  This allows a reminder to be created successfully, but the scheduler later fails because there is no usable destination.

  An explicit empty `delivery_targets` entry also prevents Slack default-channel resolution.

  ## Reproduction

  Create a Slack reminder without a chat ID or webhook:

  ```bash
  opensre work add \
    --title "Check deployment" \
    --remind-at "2026-09-12T09:00" \
    --channel-provider slack
  ```
  The reminder is accepted. At execution time, Slack delivery fails because the target has no channel and no webhook destination.

  ## Expected behavior

  Slack should only be accepted without a chat_id when a webhook is actually configured.

  If a Slack default channel is configured, the scheduler should resolve it before creating an explicit delivery target.

  ## Affected files

  - tools/system/work_items/delivery.py
  - infrastructure/scheduling/scheduler/delivery_plan.py
  - infrastructure/scheduling/scheduler/delivery.py

  ## Acceptance criteria

  - Empty Slack targets are rejected unless a webhook is configured.
  - Configured Slack default channels are resolved correctly.
  - Reminder creation reports an actionable validation error.
  - Add tests for webhook-only, default-channel, and unconfigured Slack cases.

Devesh36:
Fixes #6192

  ## Summary

  Work-item reminders could be created with a Slack target that had no `chat_id` and no configured webhook. The scheduler then had no usable destination at delivery time.

  - Resolve empty Slack work-item targets to the configured Slack default channel before serializing `delivery_targets`.
  - Permit an empty Slack target only when a configured webhook supplies its channel-bound destination.
  - Return an actionable validation error when neither a default channel nor webhook is available.
  - Add focused coverage for default-channel, webhook-only, and unconfigured Slack setups.

  The work-item target resolver now applies the existing scheduler Slack default-channel policy before explicit targets are persisted. Validation then allows a remaining empty target only for a configured incoming webhook, whose destination is fixed by Slack. This keeps work-item reminder creation aligned with scheduler delivery behavior without changing explicit channel handling.

### DIFF
--- a/infrastructure/scheduling/scheduler/delivery_plan.py
+++ b/infrastructure/scheduling/scheduler/delivery_plan.py
@@ -116,6 +116,8 @@ def _explicit_targets(task: ScheduledTask) -> tuple[DeliveryTarget, ...]:
     if not isinstance(parsed, list):
         return ()
 
+    slack_creds = resolve_slack_credentials(task.params)
+    slack_webhook = str(slack_creds.get("webhook_url") or "")
     targets: list[DeliveryTarget] = []
     for entry in parsed:
         if not isinstance(entry, dict):
@@ -128,6 +130,8 @@ def _explicit_targets(task: ScheduledTask) -> tuple[DeliveryTarget, ...]:
         except ValueError:
             continue
         chat_id = str(entry.get("chat_id", "")).strip()
+        if provider is Provider.SLACK and not chat_id:
+            chat_id = resolve_slack_delivery_chat_id(task, webhook_url=slack_webhook)
         targets.append(
             DeliveryTarget(
                 provider,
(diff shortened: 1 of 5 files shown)


Devesh36:
@greptile review

muddlebee:
@Devesh36 i think we need to test this with slack? as this is a func change

Devesh36:
> @Devesh36 i think we need to test this with slack? as this is a func change

tested !

muddlebee:
Please add demo or screenshot

PR #6233, opened 2026-09-12
feat(tool): enforce one action per model response with ToolRole

davincios:
Fixes #

**One action per model response, enforced in the runtime.**

A demo run (`analyzing-github-ci-performance`) lost its CI/CD report because the model batched `update_plan` + `ask_user_choice` + `memory_remember` in one response: the menu tool terminated the turn, the plan guard credited step 4 from a skill reference read, and the report text never rendered. The root cause is that we let the model batch several tool calls per response and then executed them in parallel, so ordering, turn-ending and rendering all became racy.

This PR replaces the `parallel_safe` / `ToolParallelism` / `execution_mode` knobs with a single `ToolRole` on the tool contract and a batch validator in `core/tool/execution.py`:

- `ToolRole.ACTION` (default): does the step's work — exactly one per response.
- `ToolRole.BOOKKEEPING` (`update_plan`, `memory_remember`, `session_goal_complete`): may accompany the one action.
- `ToolRole.TURN_ENDING` (`ask_user_choice`): hands control to the user and must be the only call in its response.

A response that violates the rule executes **nothing** — every call gets the same structured error (`metadata.batch_rejected=True`) telling the model to re-issue one action — so a half-executed batch can never leave the transcript in an ambiguous state. Permitted batches run sequentially in provider order; the `ThreadPoolExecutor` path, `_requires_sequential_execution`, and `_TOOL_EXECUTOR_WORKERS` are deleted.

Providers are asked for one call up front so the rejection path is a fallback, not the norm:

- OpenAI native (Responses + Chat): `parallel_tool_calls=False` (compat providers still omit the param).
- Anthropic + Bedrock-Anthropic: `tool_choice={"type":"auto","disable_parallel_tool_use":True}`.
- LiteLLM: `parallel_tool_calls=False` with `drop_params=True` so providers that lack the param don't fail.

Prompts and cards were inverted: the system prompt no longer says "Parallelize tool calls"; `skills/AGENTS.md` documents the roles; `scheduling-github-ci-fixes`, `delivering-morning-briefings`, and `connecting-slack` no longer instruct batching (bookkeeping still rides with the next single action, never with a menu). The same-batch branch in `action_dedup` is gone because two guarded actions can no longer share a response.

A follow-up PR renders the text that accompanies `ask_user_choice` as the assistant reply and requires it to be non-empty.

Runtime rejecting the exact batch from the failed demo (offline harness):

```
DEBUG core.tool.execution: tool_batch rejected calls=['update_plan', 'ask_user_choice', 'memory_remember']
tool result (all three, is_error=True):
  Nothing ran: ask_user_choice hands the turn to the user and must be the only tool call
  in a response, but this response requested 3 (update_plan, ask_user_choice, memory_remember).
  Re-issue ask_user_choice alone, after any other work.
```

Focused tests run locally (all green, plus `make pre-push` 9/9):

```
tests/core/tool tests/core/tool_framework tests/core/runtime/llm/test_agent_llm_client.py
tests/core/runtime/llm/test_litellm_compat.py tests/core/agent/orchestration
core/agent_harness/prompts/skills tests/tools/test_agent_memory.py tests/shared/test_tool_api_border.py
```

Full suite: 15393 passed; the only failure (`test_loop_results[40]`) also fails on `origin/main` in this environment (terminal geometry) and is unrelated.

---

- Problem: multi-call responses executed in parallel made turn-ending, plan credit and reply rendering order-dependent; the report was lost.
- Alternatives considered: (a) keep parallel execution and teach the menu hook to render pending text — patches one symptom; (b) execute the first call and error the rest — leaves the model with a half-applied batch and a "why did my other calls fail" prompt; (c) prompt-only — models ignore it under load. Chose a hard runtime rule with a whole-batch rejection, plus provider flags so the rejection rarely fires.
- Key pieces: `ToolRole` (`core/domain/types/tools.py`, a leaf both `contracts.py` and `registry.py` import); `response_batch_violation` + sequential `execute_tool_calls` (`core/tool/execution.py`); role declarations on the four non-action tools; provider flags in `agent_clients.py` / `litellm/clients.py`; prompt/card rewrites; workflow tests now script "rejected batch → re-issued single calls".

---


davincios:
@greptile review

PR #6247, opened 2026-09-14
fix(plan): plans complete only on evidence, verify before closing, and resolve blocked steps with the user

YauhenBichel:
Three problems with the live plan: all steps could be marked
complete without proof that verification happened; the closing step was
accepted as verification whatever it said; multi-step work could run
with no plan at all. Reproduced before changing anything: on main, a
plan born complete was reset but the reset message told the model to
"run the work", so it ran the commands the user had forbidden and earned
the ticks; "list, count, summarize" completed with nothing checked.

What the host enforces now (core/agent_harness/task_plan/, turns/):

- A step marked verifies: true is the only one labelled (verify). It
  never completes without its own tool return, and a text-only closing
  step closes for free only after such a step completed. Without one the
  close is refused and the tool result says so.
- The second work tool of a turn is refused while no plan with open work
  is stored. Bookkeeping, the hand-off, slash commands and failed
  commands do not count.
- A step newly marked blocked is resolved with the user, not skipped:
  the turn cannot end until ask_user_choice is queued; the turn carrying
  the user's answer is exempt, so a step they leave blocked does end it.
  "Plan ended" is printed only after that choice.
- A deliverable step in progress completes on its reply while later
  steps remain (demo A's report step was reset every run), never as the
  closing step.
- A slash command is evidence for the step it serves but not work for
  the plan rule, so skill B's /cron steps still complete.
- The reset message offers blocked for work that must not happen; a
  requested plan is written even when its marks are declined.

The rules live in one place each (completion.py, required.py,
conclusion.py, write_result.py) with the hooks in turns/plan_hooks.py;
prompt and skill cards state them once. Skill A is unchanged.

image

image


YauhenBichel:
@greptile review

PR #6249, opened 2026-09-14
fix(plan): the demo pick does not end on a skill load

YauhenBichel:
One demo A run in nine today stalled: the model loaded the chosen skill and ended the turn with "Next, I'll scan local repositories". No plan existed yet, so none of the plan gates applied and the goal reviewer accepted it. On stage that is a dead shell after the first menu.

The onboarding menu's answer turn now has one more host gate: if the turn loaded the chosen demo skill and did nothing else (no plan write, no step, no menu queued), its conclusion is rejected once with a nudge to write the plan and run the first step. Scoped to the hand-off from the onboarding master, so a question about a skill or a hand-off between two workflow skills is never pushed into running anything. Rules live in task_plan/conclusion.py and turns/goal_review.py; skill A is unchanged.

What in these five screens comes from the recent changes:

Screen 1–2: after "Skill activated" the model wrote the plan and ran the scan in the same turn, then the repository menu opened. That is the turn that stalled once today; here it went through on its own.
Screen 5, Plan complete · 6/6: every ✓ has its evidence under it — scan local git workspace, analyze github ci reliability …, Skill reference references/benchmarks.md. A reference read counting as a step's work, and no step bouncing from ✓ back to ○, are both from this week's rules. Before the deliverable fix this same run ended 5/6 · 1 blocked.
"Show CI performance report" ✓ with nothing under it: correct — its work is the reply itself (the deliverable rule), and it is not the closing step, so it did not need a verify step.
"Offer scheduling, Slack setup, or finish" ✓: closed on your Finish answer — an Ask User answer is evidence for the step that asked.
No (verify) anywhere, and no "unverified" complaint: skill A declares no verification step and the closing step had evidence, so the verification rule had nothing to say. That is by design; skill A is frozen.

Screenshot 2026-09-14 at 3 22 55 PM
Screenshot 2026-09-14 at 3 27 43 PM
Screenshot 2026-09-14 at 3 27 07 PM
Screenshot 2026-09-14 at 3 26 51 PM
Screenshot 2026-09-14 at 3 23 14 PM


YauhenBichel:
@greptile review

Issue #6252, opened 2026-09-14
37 tools skip the description contract under a closed ticket (#5498)


Davidson3556:
## Summary

37 tools skip `test_description_contract` under #5498, which is closed, so nothing tracks the gap and those tools are never checked.

## Expected vs actual behavior

**Expected:** a quarantine in `_DESCRIPTION_CONTRACT_ALLOWLIST` points at an open ticket, and entries leave the list as tools are fixed.

**Actual:** #5498 closed with all 37 entries still in the list. Every one fails the same rule, `use_cases is empty despite siblings in the same source`, and is reported as a skip rather than a failure:

```
240 passed, 37 skipped
```

`use_cases` is part of what the agent reads when choosing between tools from the same source, so these are the tools it has the least to go on for. The list includes `shell_run`, `slash_invoke`, `memory_recall` and every `work_task_*` tool.

| Source | Tools |
| --- | --- |
| interactive_shell | 8 |
| knowledge | 6 |
| dagster, mariadb, mongodb, mongodb_atlas | 5 each |
| railway | 2 |
| slack | 1 |

None of the entries are stale; each still violates the rule.

## Steps to reproduce

```bash
uv run python -m pytest tests/tools/test_description_contract.py -q -rs
```

## Can you reproduce it consistently?

Yes.

## Additional context

Plan: add `use_cases` per source in small PRs and remove each tool from the allowlist as it is fixed, so its test runs again. The allowlist should end empty, or keep only entries with an open ticket.

PR #6253, opened 2026-09-14
test(tools): give the knowledge tools use_cases and lift their quarantine

Davidson3556:
Part of #6252

Batch 1 of #6252: the `knowledge` tools.

These tools skipped `test_description_contract` under #5498, which closed with them still in `_DESCRIPTION_CONTRACT_ALLOWLIST`, so nothing checked them. Each had empty `use_cases` while sharing a source with tools that set them, which is the rule they broke:

- `memory_recall`
- `memory_forget`
- `work_task_list`
- `work_task_complete`
- `work_task_update`
- `work_task_prioritize`

Each gets `use_cases` written from its own description and implementation, worded to separate it from the other tools in the same source, since that is what the agent reads when choosing between them. `work_task_update` only claims what `main` supports today. They leave the allowlist, so their tests run again.

```
before: 240 passed, 37 skipped
after:  246 passed, 31 skipped
```

The fixed tools now report `PASSED` rather than `SKIPPED`. To check the un-quarantined tests are not vacuous, I removed one tool's `use_cases` again and confirmed its test fails instead of skipping, then restored it.

Checks: `make lint`, `make format-check`, `make typecheck`, and `uv run python -m pytest tests/tools/ tests/interactive_shell/` (5350 passed).

---

The quarantine had outlived its ticket, so the useful fix was the metadata rather than a new ticket number. I read each tool's description and handler before writing its entries, and phrased them as the situations that should lead the agent to that tool and not its siblings. Each entry is a single string literal on its own line, to stay clear of CodeQL's implicit-concatenation-in-list alert that `use_cases` metadata is prone to.

---


Davidson3556:
@codex review

muddlebee (changes requested):
@Davidson3556 you need to run these through live via LLM API to see if the changes are working directly, could you check??

Davidson3556:
image
image

Davidson3556:
Live LLM routing smoke test on commit `d2404367a` with OpenAI `gpt-5.4-mini`: all 6 prompts selected exactly the expected tool (`memory_recall`, `memory_forget`, `work_task_list`, `work_task_complete`, `work_task_update`, and `work_task_prioritize`), with key arguments validated. This was selection-only; no tools were executed.

Important scope note: generic tool `use_cases` are not currently included in the model-facing provider schemas. This confirms the current routing behavior from the existing descriptions and input schemas, but does not demonstrate that the newly added `use_cases` changed model behavior.

Davidson3556:
@muddlebee Your request for a live LLM check exposed the real issue: `use_cases` is not included in any current provider tool schema, so the earlier 6/6 run was only testing the unchanged descriptions. I traced the generic consumer to the investigation planner removed in #5981.

I've corrected the PR at `a645a6d`: it now removes the stale sibling/`use_cases` contract and clears the 37 quarantined tools, instead of adding metadata that cannot affect routing. The focused contract suite is 283/283 with no skips, the local pre-push gate is 9/9, and CI is green on this head.

Could you take another look?

@greptile review

PR #6254, opened 2026-09-14
test(tools): give the interactive-shell action tools use_cases and lift their quarantine

Davidson3556:
Part of #6252

Batch 2 of #6252: the `interactive_shell` tools.

These tools skipped `test_description_contract` under #5498, which closed with them still in `_DESCRIPTION_CONTRACT_ALLOWLIST`, so nothing checked them. Each had empty `use_cases` while sharing a source with tools that set them, which is the rule they broke:

- `cli_exec`
- `code_implement`
- `fix_sentry_issue_start`
- `llm_set_provider`
- `shell_run`
- `skill_view`
- `slash_invoke`
- `task_cancel`

Each gets `use_cases` written from its own description and implementation, worded to separate it from the other tools in the same source, since that is what the agent reads when choosing between them. These descriptions carry deliberate limits (`cli_exec` is not a discovery tool, `slash_invoke` is for explicit slash commands only), so every entry stays inside the scope its description already states. I checked where registered-tool `use_cases` are read: nothing renders them into the action agent's prompt today (only the slash catalog renders its own), so this restores the check without changing tool selection. They leave the allowlist, so their tests run again.

```
before: 240 passed, 37 skipped
after:  248 passed, 29 skipped
```

The fixed tools now report `PASSED` rather than `SKIPPED`. To check the un-quarantined tests are not vacuous, I removed one tool's `use_cases` again and confirmed its test fails instead of skipping, then restored it.

Checks: `make lint`, `make format-check`, `make typecheck`, and `uv run python -m pytest tests/tools/ tests/interactive_shell/ tests/core/agent_harness/` (5999 passed).

---

The quarantine had outlived its ticket, so the useful fix was the metadata rather than a new ticket number. I read each tool's description and handler before writing its entries, and phrased them as the situations that should lead the agent to that tool and not its siblings. Each entry is a single string literal on its own line, to stay clear of CodeQL's implicit-concatenation-in-list alert that `use_cases` metadata is prone to.

---


Davidson3556:
@codex review

PR #6251, opened 2026-09-14
feat(shell): resolve merge conflicts with a reviewed, approved merge that finishes as OpenSRE Agent

YauhenBichel:
Fixes the root cause of "OpenSRE cannot resolve merge conflicts": the shell handed
conflicts to code_implement, a background Claude Code launch whose only result was
{"ok": true} and whose prompt forbids commits, so files were edited but the merge was
never verified, staged or committed.

New tool: resolve_merge_conflicts (tools/cross_vendor)
- Prints every conflict hunk side by side (ours | theirs) before anything runs.
- Asks per file through the shell's own menu: keep ours, take theirs, combine with the
  coding agent, or free text. With several files it asks once for all of them first.
  Decisions stated in words go in as `decisions`; menu answers are read back by title.
- Applies keep/take with git; only "combine" files go to the coding agent (via the
  neutral coding-agent seam). Lockfiles are regenerated, never hand-merged.
- Verifies no marker or unmerged path remains, shows the result table, then commits
  the merge as OpenSRE Agent and pushes the branch after the /auto approval
  (High proceeds, Med/Low/Off ask, unattended proceeds), and waits for the pull
  request checks (up to 15 minutes), reporting passed/failed/not watched.
- ESC stops it before the commit and before the push; the merge is never aborted.
- While a merge is in progress, shell_run refuses git commit/push/checkout/reset/
  merge/rebase/stash/cherry-pick/revert and points to the tool.

Shared code: integrations/git/conflict_resolution.py (snapshot, unresolved check,
conclude_merge, per-file resolution description) now also backs the CI fix's
base-branch merge; integrations/github/pull_request_checks.py reuses the CI-fix
check waiter. code_implement reports whether its background task started.

Verified live on the five real conflicting PRs (6225, 6183, 6173, 6151, 6006) in
local clones, and end to end with the menu, approval, commit and push against a
local origin.

image

image

image

image

https://github.com/Tracer-Cloud/opensre/pull/6251

image

PR #6258, opened 2026-09-14
feat(merge): show the coding agent's steps while it resolves conflicts

YauhenBichel:
Follow-up to #6251. In a live comparison against Claude Code and Cursor on the
same conflict, OpenSRE was the only tool that finished the merge (commit as
OpenSRE Agent, push, check wait), but it showed a bare spinner for the three to
ten minutes the coding agent worked, while the other two stream every read,
edit and test run. This closes that gap.

What changes
- integrations/llm_cli/agent_exec.py: the process poller hands every stdout
  line to an optional callback as it arrives (drain thread), nothing else
  changes for callers without one.
- integrations/coding_agent: run_coding_task and every backend accept
  on_progress. The Claude Code backend switches to `--output-format
  stream-json --verbose` only when a listener is present; a small reader
  (claude_stream.py) turns tool-use events into one short line each
  ("Reading migrations.py", "Editing migrations.py", "Running: uv run pytest
  tests/scheduler", "Searching _schema_is_current"), collapses repeats, and
  takes the agent's final account from the result event. Pi, Codex and Cursor
  accept the argument and behave as before.
- tools/cross_vendor/resolve_merge_conflicts: the merge tool prints those
  lines dimmed under the running tool, between the "(resolving)" table and
  the merged-result table, so the user sees what the agent is doing and ESC
  has something to react to.

Verified live on a fresh clone of PR 6173: the steps appeared as they
happened, the merge was committed as OpenSRE Agent and pushed, 325 scheduler
tests pass on the result.


YauhenBichel:
@greptile review

PR #6281, opened 2026-09-16
fix(scheduler): reject unsupported platforms before filesystem writes

### LINKED ISSUE #6235: [BUG] Unsupported scheduler service install writes to the filesystem before returning
## Problem

  When background scheduler installation is requested on an unsupported platform, OpenSRE should immediately return a clear unsupported-platform message.

  Currently, it tries to create the scheduler log directory first.

  ## Current behavior

  `install_background_service(system="Windows")` attempts to create:

  ```text
  ~/.opensre/logs
  ```

  before checking whether the platform is supported.

  On a read-only home directory, the command fails with PermissionError instead of returning the expected unsupported-platform response.

  ## Expected behavior

  Unsupported platforms should:

  - Return the unsupported status immediately.
  - Avoid creating directories or files.
  - Avoid modifying the filesystem.
  - Show the existing fallback guidance.

  ## Acceptance criteria

  - Check platform support before creating the log directory.
  - Keep Windows unsupported.
  - Add a regression test confirming no filesystem writes occur on unsupported platforms.
  - Preserve existing macOS and Linux behavior.

Devesh36:
Fixes #6235

  Check platform support before resolving the scheduler command or creating the scheduler log directory.

  This ensures unsupported platforms such as Windows return the existing fallback response without modifying the filesystem or raising `PermissionError` when the OpenSRE home
  directory is read-only.

  Added a regression assertion confirming that no OpenSRE directory is created on unsupported platforms.

  The installer now returns the unsupported-platform state before performing any command resolution or filesystem operations. Supported macOS and Linux installation behavior remains
  unchanged. The Windows regression test redirects the configured OpenSRE home to a temporary path and verifies that it remains untouched.


muddlebee:
@Devesh36 closing this, there could be more linux based variatns for which this will faill..

PR #6286, opened 2026-09-16
prompt: add numbered-list formatting rule to the system prompt

davincios:
Fixes #

Adds a **Numbered list** rule to the system prompt's plain-text formatting section (`core/agent_harness/prompts/opensre_system_prompt.md`): use `1.`, `2.`, `3.` markers when order matters and `-` bullets when it does not, one step per line, never inlined into a sentence as `1) …, 2) …, 3) …`. Includes a bad/good example pair.

The system prompt is human-owned; this diff was written by hand by the author and committed verbatim. No other files change.

Prompt-only change; the diff is the demo:

```
+**Numbered list**
+- Use `1.`, `2.`, `3.` markers (number, period, space) when order matters — setup steps, a runbook, "do this, then that". Use `-` bullets when it does not.
+- One step per line, always. Never inline steps into a sentence as `1) …, 2) …, 3) …` or `first…, then…, finally…` — that renders as one paragraph and the user cannot follow it.
```

`make pre-push` 9/9 locally (lint, format, types, strict imports, registries, global contracts, `tests/core/`).

---

The prompt already required one bullet per actionable item but said nothing about ordered steps, so the model inlined them as `1) …, 2) …` inside a paragraph. This adds the explicit rule and a concrete bad/good pair next to the existing bullet guidance. An agent was used only to commit and open the PR; the prompt text itself is human-authored per `AGENTS.md`.

---


davincios:
@greptile review

muddlebee:
@greptile review

muddlebee:
@codex review

muddlebee:
@davincios IMO we should keep either one of them?? either **bullets** or **numbers** ?

image

davincios:
@muddlebee so I added it in there because numbered lists were showing up as inline like this 1), 2), 3) 

instead of:
1) 
2) 
3) 

It seems to have solved the issue. 

Or do I not understand your question correctly?

muddlebee:
> @muddlebee so I added it in there because numbered lists were showing up as inline like this 1), 2), 3) 
> 
> instead of:
> 1) 
> 2) 
> 3) 
> 
> It seems to have solved the issue. 
> 
> Or do I not understand your question correctly? 

Makes sense. Thank you.

PR #6288, opened 2026-09-16
fix(scheduler): preserve overdue executions during reconciliation

### LINKED ISSUE #6287: [BUG] Scheduler reload can discard an overdue execution of an unchanged task
### Problem

  `resync_scheduler_jobs()` re-registers every enabled task, including tasks whose configuration has not changed.

  `_register_jobs()` calls `scheduler.add_job(..., replace_existing=True)` without preserving the existing job's `next_run_time`. APScheduler calculates a new next execution from
  the current time.

  If reconciliation happens after a tick becomes due but before dispatch, replacing the job discards that overdue tick.

  The discarded execution never reaches `_queue_scheduled_run()`, so no pending run is persisted and the recovery sweep cannot recover it.

  ### Reproduction

  Using the real APScheduler `BackgroundScheduler`:

  1. Start the scheduler paused to hold dispatch.
  2. Register an enabled hourly task with `next_run_time` five seconds in the past.
  3. Call `resync_scheduler_jobs()` without changing the task.
  4. Inspect the job's `next_run_time` and durable run history.

  Observed:
  - The overdue execution moves to the next hourly tick.
  - No durable run exists for the discarded tick.

  Pausing dispatch makes the reload-versus-dispatch race reproducible without relying on timing.

  ### Expected behavior

  Reloading an unchanged task preserves its pending execution. An overdue tick should execute or remain durably recoverable.

  ### Suggested fix

  Only replace jobs whose scheduling configuration changed. Preserve existing fire times for unchanged jobs.

  ### Acceptance criteria

  - Reconciliation preserves an unchanged job's overdue execution.
  - Adding or editing another task does not discard that execution.
  - A regression test verifies execution or durable admission after reconciliation.

  ### Affected code

  `infrastructure/scheduling/scheduler/runner.py`
  - `_register_jobs()`
  - `resync_scheduler_jobs()`

Devesh36:
Fixes #6287 
Scheduler reconciliation could silently discard an overdue execution of an unchanged task. Every reload re-registered all enabled jobs with replace_existing=True, causing APScheduler to calculate their next execution from the current time. If reconciliation occurred after a tick became due but before dispatch, that tick moved to the next scheduled time.

The lost tick never reached the executor’s admission hook, so no pending run was persisted. The recovery sweep therefore had no durable record to recover. Adding or editing an unrelated task could trigger the same loss.

This change compares each existing job’s trigger and timezone with the current task schedule. When they match, reconciliation updates callback arguments and job metadata in place, preserving the job’s pending fire time. New tasks and tasks with changed schedules are registered through the existing replacement path. Disabled and removed tasks continue to be removed.

Regression tests use the real APScheduler BackgroundScheduler, paused to hold an overdue tick before dispatch. They verify that reconciliation preserves the tick, that it executes after resume, and that its original fire time appears in durable run history. Coverage includes:

  - Reloading without changing any task.
  - Adding another task.
  - Editing another task’s schedule.
  - Updating the overdue task’s name and prompt.
  - Applying cron and timezone changes to the edited task.


Devesh36:
@codex review

muddlebee:
@Devesh36 closing this as I dont see the practical utility or much delta benefit of the change.

feel free to re-open if still valid..

PR #6182, opened 2026-09-10
fix(install): authenticate GitHub API calls to avoid canary 403s (#6128)

mdiniz97:
Fixes #6128

#### Describe the changes you have made in this PR -

The installers query `api.github.com` unauthenticated. GitHub-hosted runners share egress IPs, so the unauthenticated limit (60 requests/hour/IP) is often already exhausted and the release metadata call fails with `403 Forbidden`. That is the canary failure: the latest run failed on macOS, an earlier one on Windows, and each retry rotates which platform loses the race.

Changes:

- `install.sh` / `install.ps1` now send `Authorization: Bearer $GITHUB_TOKEN`   (`GH_TOKEN` as fallback) on the GitHub API metadata request when either variable is set. Without a token, behavior is unchanged.
- The token is only attached to the API request. Release asset downloads keep the  unauthenticated header set because assets redirect to signed URLs on another  host, which must never receive the token.
- `.github/workflows/installer-canary.yml` passes its `contents: read`  `GITHUB_TOKEN` to the installer test step.
- Tests: `tests/cli/test_install_sh_github_auth.py` runs the real `download_text`  with a stubbed `curl` and pins the header, the `GH_TOKEN` fallback, and the  no-token path. `tests/cli/test_install_ps1_github_auth.py` pins the PowerShell  split (API headers authenticated, download headers not) and runs it under  PowerShell when available.

The Cloudflare install proxy serves `main` with `cache-control: no-store` (`infrastructure/deployment/cloudflare_install_proxy/src/index.mjs`), so the CDN path picks this up on the first canary run after merge.

### Demo/Screenshot for feature changes and bug fixes -

Before (canary run, macOS leg):

```
E   AssertionError: curl: (56) The requested URL returned error: 403
E     Error: Failed to query release metadata from GitHub.
```

After:

```
$ uv run python -m pytest tests/cli/test_install_sh_github_auth.py tests/cli/test_install_ps1_github_auth.py -q
4 passed, 1 skipped (PowerShell not installed locally)

$ uv run python -m pytest tests/cli/test_install_matrix.py tests/cli/test_install_sh_path.py \
    tests/cli/test_install_sh_resolution.py tests/cli/test_install_ps1_progress.py -q
71 passed, 5 skipped
```

`bash -n install.sh`, YAML parse, `make lint`, `make format-check` and
`make typecheck` are clean. The PowerShell runtime test runs on the Windows CI job.

---

## Code Understanding and AI Usage

[...]

PR #6295, opened 2026-09-17
fix(installer): authenticate GitHub API metadata calls when a token is set

### LINKED ISSUE #6291: Installer canary failing
The public install paths (curl/PowerShell against install.opensre.com, GitHub release resolution, or Homebrew) failed a post-publish canary run.

Run: https://github.com/Tracer-Cloud/opensre/actions/runs/35169947313
Trigger: repository_dispatch
Pinned tag: v0.1.2026.9.17

Check the failed matrix leg(s) above and the uploaded `installer-canary-*` diagnostics artifact for the failing platform.

nehraa:
Fixes #6291.

The macOS-arm64 canary leg on `v0.1.2026.9.17` hit 403s from api.github.com; Linux + Windows passed. Root cause: `install.sh` and `install.ps1` hit GitHub's API unauthenticated, and the macOS-arm64 runner IP pool exhausts the 60 req/hr anonymous limit before Linux/Windows do.

Precedence: `OPENSRE_GITHUB_TOKEN` > `GITHUB_TOKEN` > `GH_TOKEN`. Only the install scripts read it; release asset downloads stay anonymous so they keep working behind proxies that strip `Authorization`. The token reaches curl through stdin (`--config -`) and PowerShell through its header dict — never through process arguments, so other users on a shared host cannot read it out of `ps`.

```
$ make lint && make format-check && make typecheck   # all green
$ uv run pytest tests/cli/test_install_sh_github_auth.py tests/cli/test_install_ps1_github_auth.py -v
6 passed, 6 skipped (no pwsh on this runner; CI will run them on Windows + macOS)
```

Each test runs the real `install.sh` / `install.ps1` source against a fake `curl` that records both argv and stdin, so a future bug in the lookup or in the auth plumbing cannot hide behind a mocked test. This addresses the Greptile P1s that closed the prior attempts (#6182, #6246): argv leakage and a mocked-curl bypass of the lookup precedence.


nehraa:
Closed in favor of #6296 (branch fork-propagation issue on the original head ref).

PR #6305, opened 2026-09-18
fix(shell): braille "Thinking…" spinner everywhere, and no false "Could not load skill" error

YauhenBichel:
## Why
Two things made a routine shell turn look broken.
- In Apple Terminal the loader next to "Thinking…" used a dot and circle
  frame set, and its large circle read as an oversized loader.
- Asking for PR status printed "Error · Could not load skill ·
  tracking-github-work-status". That name is guidance attached to tool
  descriptions, not an action skill. The embedded block said "Use this
  skill", the model called skill_view, the lookup failed, and the user saw an
  error for something the agent recovered from a second later.

## What
- The spinner is braille in every terminal; the dot frame set is removed.
- Embedded tool guidance is tagged <tool_guidance> and says it is already
  loaded. The wording is length-neutral, since the block is capped at 2400
  characters.
- skill_view answers a tool-guidance name as a normal result that lists the
  tools to call. Unknown names still fail.
- The shell prints nothing for that case.


YauhenBichel:
@greptile review

YauhenBichel:
@greptile review

PR #6309, opened 2026-09-18
feat(hosted-gateway): client and check_hosted_gateway tool for the or…

YauhenBichel:
First step of the remote managed service: a client for the OpenSRE app's hosted-gateway routes and a read-only tool that reports whether the signed-in organization's Fargate gateway exists and is running. The app maps the account token to the organization; the tool takes no identifier. The token is sent only over https (http to loopback for development), redirects are not followed, and errors carry stable codes only.


YauhenBichel:
@greptile review

Issue #5963, opened 2026-09-02
Define bounded concurrency for scheduled skill runs


davincios:
## Problem

Scheduled-run concurrency is implicit. `infrastructure/scheduling/scheduler/runner.py` (`_register_jobs()`) sets no APScheduler executor or max-instances contract; gateway-hosted runs pass through `SchedulerRunners.gated()` and `queued_turn_slot()`, which blocks after work is claimed; and `TurnConcurrencyGate` defaults to one permit for `SMALL`, serializing every scheduled run with interactive chat. No test proves two distinct tasks overlap, excess tasks queue, or the same task cannot overlap itself.

## Value

Several reports due at 09:00 complete predictably, without unbounded fan-out and without bypassing gateway capacity controls.

## Acceptance criteria

- An explicit scheduled-run concurrency limit exists and is documented.
- Distinct tasks overlap up to that limit; the same task never overlaps itself.
- Excess runs remain queued and are observable in run history.
- Scheduled runs continue to respect the process-wide turn gate.
- Deterministic barrier-based tests cover limits 1, 2, and overflow.

Related: #5476 (capacity slot return) is a shared prerequisite; #5492 covers gateway turn concurrency generally.

PR #6325, opened 2026-09-19
test(turn-capacity): prove a slot returns on every abnormal turn exit

### LINKED ISSUE #5476: Prove a capacity slot always comes back
## Problem

`infrastructure/turn_host/concurrency.py` defaults to `SizeProfile.SMALL: 1` — a
single concurrent turn per process. Nothing asserts that a slot is returned when
a turn dies abnormally.

## Why it matters

At limit 1, **one leaked slot bricks the task**. Every subsequent message gets
`AT_CAPACITY_MESSAGE` forever, with no crash, no restart, and a healthy-looking
process. The failure is indistinguishable from "busy" until someone notices the
task has been busy for an hour.

The paths that can leak are the ones that skip a normal return: cancellation, an
exception inside `handle`, a timeout, and shutdown mid-turn.

## Steps

1. Read `infrastructure/process/turn_capacity/slots.py` — `turn_slot` (drop) and
   `queued_turn_slot` (wait) are the two policies, and both are context managers,
   so the question is whether every acquire is inside one.
2. Add a test per abnormal exit: cancel mid-`handle`, raise inside the turn body,
   and time out. After each, assert the gate reports zero held slots.
3. Add an accounting test: run N overlapping turns through the gate and assert
   `acquired == released` and that the in-flight count never exceeded the limit.
   Instrument the semaphore or wrap `TurnConcurrencyGate` in the test rather than
   sleeping and hoping.
4. If a leak is found, fix it by moving the acquire inside the context manager —
   do not add a compensating release in an error handler.

## Watch out

`tests/infrastructure/test_turn_capacity.py` already contains a `_fail_the_turn()`
helper that exists because CodeQL does not model `NoReturn` and reads code after
`pytest.fail` as reachable. Reuse it; do not write a bare `raise` inside a `with`
body and then silence the alert with a comment.

## Expected result

Every abnormal turn exit returns its slot. A test that fails if someone adds an
acquire path outside the context manager.

## Verification

```bash
uv run python -m pytest tests/infrastructure/test_turn_capacity.py \
  gateway/tests/runtime/test_concurrency_gate.py -q
```

Prove the accounting test is real — temporarily leak a permit (acquire without
the context manager) and confirm the test fails. Revert.

Overlap actually happens (the default of 1 serializes everything and proves
nothing):

```bash
OPENSRE_MAX_CONCURRENT_TURNS=4 uv run python -m pytest \
  tests/infrastructure/test_turn_capacity.py -q
```

## Checklist

- [ ] One test per abnormal exit: cancel, exception, timeout
- [ ] `acquired == released` over N overlapping turns; limit never exceeded
- [ ] Leak demonstrated and reverted; say so in the PR
- [ ] Any fix moves the acquire inside the context manager

## Out of scope

Choosing a production value for `OPENSRE_MAX_CONCURRENT_TURNS` — that needs RSS
measurement and is a separate decision. Per-session queuing.

## Claiming this issue

this issue.

rohitmane-dev:
Fixes #5476

Adds the missing proof that a turn capacity slot is always returned.

**No product code changed.** Auditing every acquire site showed both production callers — `TurnRunner.run` and `SchedulerRunners._gated_agent` — already run inside `turn_slot` / `queued_turn_slot`, so there was no leak to fix. The gap was coverage, which is what this PR closes.

`tests/infrastructure/test_turn_capacity.py` (4 → 10 tests):

- **Cancel** — a `BaseException` that `except Exception` would miss, so only a `finally` returns the slot.
- **Early return** — the real `TurnRunner.run` path when `host_cancel_requested` is set after the slot is taken.
- **Exception / timeout through `queued_turn_slot`** — the wait policy has its own `finally`; the existing tests only covered the drop policy's.
- **Accounting** — 12 turns over a limit of 4 against a real `TurnConcurrencyGate`, with a barrier *inside* the slot so the cap is measured rather than assumed: `acquired == released`, `in_flight == 0`, `max_in_flight == limit`. Threads are daemons so a leak fails the assertion instead of wedging the interpreter for the full 600s timeout.

`tests/quality/test_turn_gate_acquire_is_scoped.py` (new): an AST guard that fails any product module outside `turn_capacity/slots.py` calling `try_acquire` / `acquire` / `release` on a gate-named receiver. Keyed on the receiver rather than the method name, because `release` alone is far too common (locks, `platform.release()`) to flag safely.

Both new tests were proven to actually catch the bug, then reverted, as the issue asks:

- Removing the `finally` from `queued_turn_slot` failed 4 tests.
- A temporary hand-rolled acquire in a product module tripped the quality guard.

`git diff` confirms product code is untouched in this PR.

Reusing `_fail_the_turn()` and adding sibling `_cancel_the_turn()` / `_time_out_the_turn()` helpers keeps every abort behind a call, per the note in the issue about CodeQL not modelling `NoReturn`.

```console
$ make pre-push
PASS lint (0.6s)
PASS format (1.4s)
PASS integration-registry (6.3s)
PASS imports (9.6s)
PASS tool-registry (10.1s)
PASS types (5.2s)
PASS affected-tests (5.8s)
PASS tool-contracts (18.5s)
PASS global-contracts (33.3s)
Validation: 9/9 passed in 40.2s.

$ uv run python -m pytest tests/infrastructure/test_turn_capacity.py \
    gateway/tests/runtime/test_concurrency_gate.py -q
20 passed in 1.38s

$ OPENSRE_MAX_CONCURRENT_TURNS=4 uv run python -m pytest \
    tests/infrastructure/test_turn_capacity.py -q
10 passed in 0.78s
```

Leak demonstration, reverted before commit:

```console
# finally removed from queued_turn_slot
4 failed, 6 passed

# temporary hand-rolled acquire added to a product module
E  infrastructure/turn_host/_leak_probe_tmp.py:9: gate.try_acquire()
E  infrastructure/turn_host/_leak_probe_tmp.py:10: gate.release()
1 failed, 7 passed
```

Note for triage: #5963 listed this issue as "a shared prerequisite", but the PR that closed it (#6024) only touched the APScheduler executor, scheduler runner and run store — it never touched `infrastructure/process/turn_capacity/` or these tests, so the slot-return coverage was still missing.

---

The problem is that at the `SMALL` size profile the turn limit is one permit, so a single leaked slot makes the process answer `AT_CAPACITY_MESSAGE` forever with no crash and no restart — it looks like a permanently busy, healthy task. Nothing asserted that a permit comes back when a turn dies abnormally.

The first step was an audit rather than a fix: find every place a permit is taken. There are only two in product code, `TurnRunner.run` and `SchedulerRunners._gated_agent`, and both already sit inside the `turn_slot` / `queued_turn_slot` context managers, whose `finally` returns the permit. So the correct outcome here is tests plus a guard, not a product change — adding a compensating release in an error handler would have been the wrong fix, and the issue explicitly warns against it.

The alternative I considered for the accounting test was running N threads and sleeping to hope for overlap. That is exactly the "sleeping and hoping" the issue warns about and would be flaky under CI load, so instead a `threading.Barrier(limit)` *inside* the slot parks exactly `limit` turns together. That makes `max_in_flight == limit` a deterministic assertion rather than a probabilistic one, and it is why `turns` is a multiple of `limit` — every wave fills the barrier exactly.

One non-obvious detail: the accounting threads are daemons. Without that, a leaked permit does not fail the run, it wedges it — the leaked turns park forever on `acquire`, the interpreter waits on non-daemon threads, and CI reports a 600s timeout instead of the assertion. That was found by actually performing the leak demonstration rather than assuming the test was sound.

For the quality guard, matching on the method name alone was not viable: `release` appears all over the codebase on locks, semaphores and `platform.release()`. Keying on a receiver whose name contains "gate" keeps it precise, and the one legitimate acquire/release pair in `turn_capacity/slots.py` is allowlisted by path.

---


rohitmane-dev:
Thanks — all three findings were fair, and the second one was a genuine bug in the test rather than a style point. Fixed in 2000fd6db.

**1. Gate aliases bypass guard.** Correct: keying only on a receiver containing `"gate"` meant `capacity.try_acquire()` walked straight past. `try_acquire` belongs to the `TurnGate` protocol alone, so it is now an offense wherever it is called, regardless of the variable name. `acquire` / `release` stay receiver-keyed because they are shared with locks, semaphores and `platform.release()`, and matching those everywhere would flag unrelated code. Verified with a probe module using a deliberately non-gate name:

```
E  infrastructure/turn_host/_alias_probe_tmp.py:9: capacity.try_acquire()
1 failed, 7 passed
```

**2. Barrier failures are swallowed.** This one was a real defect. `threading.BrokenBarrierError` subclasses `RuntimeError`, so `except RuntimeError` in the worker caught the broken barrier as if it were the turn's expected failure — exactly the signal the test exists to surface. The expected failure now raises a dedicated `_TurnFailed`, so the barrier error propagates. Verified by sizing the inner barrier so it can never fill:

```
E  assert [BrokenBarrierError(), ...] == []
1 failed in 10.85s
```

**3. BaseException catch violates policy.** Agreed, and it is spelled out in `AGENTS.md`. Now catches `Exception`, with the `noqa: BLE001` removed rather than kept. `BrokenBarrierError` is an `Exception`, so finding 2 is still caught.

The one remaining `BaseException` subclass is the `_TurnCancelled` test double, which is deliberate: a cancellation that is not an `Exception` is precisely what proves only a `finally` returns the slot.

`make pre-push` still passes 9/9.

@greptile review

